PR #7, opened 2020-10-07
improve accuracy of jit type signature

joelberkeley-secondmind:
this PR improves the accuracy of the `jit` decorator's signature

### DIFF
--- a/trieste/utils/misc.py
+++ b/trieste/utils/misc.py
@@ -11,15 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Callable
+from typing import Callable, TypeVar
 
 import numpy as np
 import tensorflow as tf
 
 from ..type import TensorType
 
+C = TypeVar("C", bound=Callable)
+""" A type variable bound to `typing.Callable`. """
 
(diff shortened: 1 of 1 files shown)


awav (commented):
How is that different from what we have now? Why is that important?

joelberkeley-pio:
the type signature `def jit(apply: bool = True, **optimize_kwargs) -> Callable[[Callable], Callable]:` loses information about the function it decorates, because `Callable` [is equivalent to](https://docs.python.org/3/library/typing.html#typing.Callable) `Callable[..., Any]`. For example,
```
@jit()
def foo(x: int) -> str:
    ...
```
would be untyped

trieste/utils/misc.py
```diff
@@ -11,15 +11,18 @@
...
 import tensorflow as tf
 
 from ..type import TensorType
 
+C = TypeVar("C", bound=Callable)
+""" A type variable bound to `typing.Callable`. """
 
-def jit(apply: bool = True, **optimize_kwargs) -> Callable[[Callable], Callable]:
+
+def jit(apply: bool = True, **optimize_kwargs) -> Callable[[C], C]:
```
joelberkeley-secondmind, line 25: i'm leaving sorting out the argument types for another time

ChrisMorter (approved):
Looks good to me

PR #9, opened 2020-10-07
add expected constrained improvement acquisition function

joelberkeley-secondmind:


trieste/acquisition/function.py
```diff
@@ -298,3 +298,96 @@ def probability_of_feasibility(
...
+
+        if len(objective_dataset.query_points) == 0:
+            raise ValueError(
+                "Expected improvement is defined with respect to existing points in the objective"
+                " data, but the objective data is empty."
+            )
+
+        constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)
+        pof = constraint_fn(objective_dataset.query_points)
+        is_not_feasible = pof < self._min_feasibility_probability
```
joelberkeley-secondmind: **todo** confirm it's correct to use a hard cut-off and not integrate over best points weighted by pof

trieste/acquisition/function.py
```diff
@@ -298,3 +298,96 @@ def probability_of_feasibility(
...
+        :param models: The models over each dataset in ``datasets``.
+        :return: The expected constrained improvement acquisition function.
+        :raise KeyError: If `objective_tag` is not found in ``datasets`` and ``models``.
+        :raise ValueError: If the objective data is empty, or no feasible points are found in
+            ``datasets``.
+        """
+        objective_model = models[self._objective_tag]
+        objective_dataset = datasets[self._objective_tag]
+
+        if len(objective_dataset.query_points) == 0:
```
joelberkeley-secondmind: can do `len(dataset) == 0`

joelberkeley-secondmind:
this PR is waiting on feedback from victor (which he hasn't been asked for yet)

trieste/acquisition/function.py
vpicheny: yes, the hard cut-off is consistent with the existing papers. Your idea of integrating is intriguing, but I don't think it's feasible. Say at `objective_dataset.query_points` all `pof `is equal to 0.001 (everything you have evaluated so far is very probably infeasible). What value should you pick? In theory you should use `eta = +inf`so that EI is a constant and your acquisition reduces to `pof`. Here we would actually choose the maximum of the mean as eta for numerical stability.

trieste/acquisition/function.py
vpicheny: actually we need here `query points` to compute  `mean`

joelberkeley: len query_points is the same as len dataset

trieste/acquisition/function.py
```diff
@@ -405,3 +405,93 @@ def probability_of_feasibility(
...
+                "Expected improvement is defined with respect to existing points in the objective"
+                " data, but the objective data is empty."
+            )
+
+        constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)
+        pof = constraint_fn(objective_dataset.query_points)
+        is_not_feasible = pof < self._min_feasibility_probability
+
+        if tf.reduce_all(is_not_feasible):
+            return constraint_fn
```
joelberkeley, line 486: @vpicheny there we go, very literal

vpicheny: shoudn't it be `constraint_fn(at)`?

trieste/acquisition/function.py
```diff
@@ -405,3 +405,93 @@ def probability_of_feasibility(
...
+
+        mean, _ = objective_model.predict(objective_dataset.query_points)
+        range_of_mean = tf.reduce_max(mean) - tf.reduce_min(mean)
+        penalization = self._PENALIZATION_FACTOR * range_of_mean * tf.cast(
+            is_not_feasible, mean.dtype
+        )
+        eta = tf.reduce_min(mean + penalization, axis=0)
```
vpicheny: this is unnecessarily complex if we return `constraint_fn` when no feasible point has been identified. We don't need `range_of_mean`, and PENALIZATION_FACTOR can be set to `+inf`.

trieste/acquisition/function.py
joelberkeley: it could be `lamdba at: constraint_fn(at)`, but that's the same thing

trieste/acquisition/function.py
joelberkeley: if we're going to do that, we can just mask out any unfeasible points, I think

joelberkeley: like
```python
        is_feasible = pof >= self._min_feasibility_probability

        if not tf.reduce_any(is_feasible):
            return constraint_fn

        mean, _ = objective_model.predict(objective_dataset.query_points)
        eta = tf.reduce_min(tf.boolean_mask(mean, is_feasible), axis=0)
```

vpicheny: yes, that sounds perfect

vpicheny (approved):

PR #11, opened 2020-10-09
prepare for PyPI release v0.2.0

joelberkeley-pio:


docs/conf.py
```diff
@@ -49,7 +49,7 @@
 author = 'The Trieste Contributors'
 
 # The full version, including alpha/beta/rc tags
-release = 'v0.1.0'
+release = 'v0.2.0'
```
joelberkeley-pio, line 52: we're uploading to the already existent trieste pypi package. That has versions 0.1.0 and 0.1.1, so bumping to 0.2.0 clearly delineates this from the old trieste project. We may choose to delete the 0.1.0 and 0.1.1 versions

setup.py
```diff
@@ -19,8 +19,8 @@
 
 setup(
     name="trieste",
-    version="0.1.0",
-    author="The Trieste authors",
+    version="0.2.0",
+    author="The Trieste contributors",
```
joelberkeley-pio, line 23: authors -> contributors to align with license text

setup.py
```diff
@@ -19,8 +19,8 @@
 
 setup(
     name="trieste",
-    version="0.1.0",
-    author="The Trieste authors",
+    version="0.2.0",
```
joelberkeley-pio, line 22: we're uploading to the already existent trieste pypi package. That has versions 0.1.0 and 0.1.1, so bumping to 0.2.0 clearly delineates this from the previous trieste project. We may choose to delete the 0.1.0 and 0.1.1 versions

docs/conf.py
joelberkeley-pio, line 52: we should move to a VERSION file so we don't forget to update both versions, but I'm not going to do that now

vpicheny (approved):

docs/conf.py
```diff
@@ -49,7 +49,7 @@
 author = 'The Trieste Contributors'
 
 # The full version, including alpha/beta/rc tags
-release = 'v0.1.0'
+release = '0.2.0'
```
joelberkeley-pio, line 52: i think the `v` was a mistake before

PR #14, opened 2020-10-12
use more mock data in model tests

joelberkeley:
not having enough data meant there were no inducing points for `SGPR`, which lead my GPU to crash on tests. This fixes that


ChrisMorter (approved):
LGTM

PR #15, opened 2020-10-12
make optimizer attributes private

joelberkeley:
I don't believe it's appropriate for `observer` and `search_space` to be public members of the `BayesianOptimizer`, so I've created this PR to change that


ChrisMorter (approved):
LGTM

PR #17, opened 2020-10-12
use result type in optimizer

joelberkeley:
The return type of `BayesianOptimizer.optimize` is a combination of `LoggingState` and `OptimizationResult`. There are a few problem points in these:

* there is duplication between the two: the `datasets` and `models` fields
* `OptimizationResult` should have an `acquisition_state` field if users are going to restart the optimization process from a previous run. Note this would become another duplicate field
* error handling in `OptimizationResult` is not safe. It's very easy to forget to check for the presence of an `error` in `OptimizationResult`, and start unknowingly working with data from a failed run. If the process failed while updating either the model or data, this can mean e.g. returning a model that's not been updated with the data, or a corrupted model.

To resolve all of these problems, I propose

* a single type that contains all the things of interest in each step. I've called this a `Record`
* since the history of `Record`s from successful steps is _always_ returned whether the process failed or not, but the final `Record` is only returned on success, I've separated the two in the `OptimizationResult`
* if the process failed, the final `Record` isn't created or returned, so there's no chance of receiving corrupted state. On failure, we return only the error. There are several ways we could do this. I've adapted a method used in Rust/Scala/Haskell/Swift: creating a result type which is a container for either a value or an error, along with helper methods for handling the contents. Note it's not as powerful as it is in those other languages cos we don't have pattern matching

### DIFF
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -49,8 +49,6 @@ initial_data = observer(initial_query_points)
 #
 # The Bayesian optimization procedure estimates the next best points to query by using a probabilistic model of the objective. We'll use Gaussian process regression for this, provided by GPflow. The model will need to be trained on each step as more points are evaluated, so we'll package it with GPflow's Scipy optimizer.
 #
-# Note we could leave it to the optimizer to build this model, but we'll want to inspect it later so we'll build it ourselves.
-#
 # Just like the data output by the observer, the optimization process assumes multiple models, so we'll need to label the model in the same way.
 
 # %%
@@ -60,13 +58,13 @@ def build_model(data):
     gpr = gpflow.models.GPR(data.astuple(), kernel, noise_variance=1e-5)
     gpflow.set_trainable(gpr.likelihood, False)
 
-    return {OBJECTIVE: trieste.models.create_model({
+    return {OBJECTIVE: {
         "model": gpr,
         "optimizer": gpflow.optimizers.Scipy(),
         "optimizer_args": {
             "minimize_args": {"options": dict(maxiter=100)},
         },
-    })}
+    }}
 
 model = build_model(initial_data[OBJECTIVE])
 
@@ -79,20 +77,13 @@ model = build_model(initial_data[OBJECTIVE])
 #
 # We'll run the optimizer for fifteen steps.
 #
-# The `optimize` method returns several things (see the `optimize` documentation for details), but we're only interested in the data, which captures the points where the objective was queried and the resulting values. Note that the optimizer updates the model in place.
-#
-# However, since the optimization loop catches errors so as not to lose progress, we must check if any errors occurred so we know the data is valid. We'll do that crudely here by re-raising any such errors. You may wish instead to use the history to restore the process from an earlier point.
+# The optimization loop catches errors so as not to lose progress, which means the optimization loop might not complete and the data from the last step may not exist. Here we'll handle this crudely by asking for the data regardless, using `.try_get_final_data()`, which will re-raise the error if one did occur. For a review of how to handle errors systematically, there is a [dedicated tutorial](recovering_from_errors.ipynb). Finally, like the observer, the optimizer outputs labelled datasets, so we'll get the (only) dataset here by indexing with tag `OBJECTIVE`.
 
 # %%
-from trieste.bayesian_optimizer import OptimizationResult
-
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
-result: OptimizationResult = bo.optimize(15, initial_data, model)
-
-if result.error is not None: raise result.error
-
-dataset = result.datasets[OBJECTIVE]
+result = bo.optimize(15, initial_data, model)
+dataset = result.try_get_final_data()[OBJECTIVE]
 
 # %% [markdown]
 # ## Explore the results
@@ -150,12 +141,14 @@ plot_regret(observations, ax[0], num_init=num_initial_points, idx_best=arg_min_i
 plot_bo_points(query_points, ax[1], num_init=num_initial_points, idx_best=arg_min_idx)
 
 # %% [markdown]
-# We can visualise the model over the objective function by plotting the mean and 95% confidence intervals of its predictive distribution.
+# We can visualise the model over the objective function by plotting the mean and 95% confidence intervals of its predictive distribution. Like with the data before, we can get the model with `.try_get_final_models()` and indexing with `OBJECTIVE`.
 
 # %%
 from util.plotting_plotly import plot_gp_plotly
 
-fig = plot_gp_plotly(model[OBJECTIVE].model, mins, maxs, grid_density=30)
+fig = plot_gp_plotly(
+    result.try_get_final_models()[OBJECTIVE].model, mins, maxs, grid_density=30
+)
 
 fig = add_bo_points_plotly(
     x=query_points[:, 0],
@@ -171,14 +164,14 @@ fig = add_bo_points_plotly(
 fig.show()
 
 # %% [markdown]
-# We can also inspect the model hyperparameters, and use the history to see how the length scales evolved over iterations
+# We can also inspect the model hyperparameters, and use the history to see how the length scales evolved over iterations. Note the history is saved at the *start* of each step, and as such never includes the final result, so we'll add that ourselves.
 
 # %%
-gpflow.utilities.print_summary(model[OBJECTIVE].model)
+gpflow.utilities.print_summary(result.try_get_final_models()[OBJECTIVE].model)
 
 ls_list = [
     step.models[OBJECTIVE].model.kernel.lengthscales.numpy()  # type: ignore
-    for step in result.history
+    for step in result.history + [result.final_result.unwrap()]
 ]
 
 ls = np.array(ls_list)
@@ -191,11 +184,8 @@ plt.plot(ls[:, 1])
 # If we need more iterations for better convergence, we can run the optimizer again using the data produced from the last run, as well as the model. We'll visualise the final data.
 
 # %%
-result = bo.optimize(5, result.datasets, model)
-
-if result.error is not None: raise result.error
-
-dataset = result.datasets[OBJECTIVE]
+result = bo.optimize(5, result.try_get_final_data(), result.try_get_final_models())
+dataset = result.try_get_final_data()[OBJECTIVE]
 
 arg_min_idx = tf.squeeze(tf.argmin(dataset.observations, axis=0))
 fig, ax = plot_function_2d(branin, mins, maxs, grid_density=40, contour=True)
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -77,13 +77,13 @@ model = build_model(initial_data[OBJECTIVE])
 #
 # We'll run the optimizer for fifteen steps.
 #
-# The optimization loop catches errors so as not to lose progress, which means the optimization loop might not complete and the data from the last step may not exist. Here we'll handle this crudely by asking for the data regardless, using `.try_get_final_data()`, which will re-raise the error if one did occur. For a review of how to handle errors systematically, there is a [dedicated tutorial](recovering_from_errors.ipynb). Finally, like the observer, the optimizer outputs labelled datasets, so we'll get the (only) dataset here by indexing with tag `OBJECTIVE`.
+# The optimization loop catches errors so as not to lose progress, which means the optimization loop might not complete and the data from the last step may not exist. Here we'll handle this crudely by asking for the data regardless, using `.try_get_final_datasets()`, which will re-raise the error if one did occur. For a review of how to handle errors systematically, there is a [dedicated tutorial](recovering_from_errors.ipynb). Finally, like the observer, the optimizer outputs labelled datasets, so we'll get the (only) dataset here by indexing with tag `OBJECTIVE`.
 
 # %%
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
 result = bo.optimize(15, initial_data, model)
-dataset = result.try_get_final_data()[OBJECTIVE]
+dataset = result.try_get_final_datasets()[OBJECTIVE]
 
 # %% [markdown]
 # ## Explore the results
@@ -184,8 +184,8 @@ plt.plot(ls[:, 1])
 # If we need more iterations for better convergence, we can run the optimizer again using the data produced from the last run, as well as the model. We'll visualise the final data.
 
 # %%
-result = bo.optimize(5, result.try_get_final_data(), result.try_get_final_models())
-dataset = result.try_get_final_data()[OBJECTIVE]
+result = bo.optimize(5, result.try_get_final_datasets(), result.try_get_final_models())
+dataset = result.try_get_final_datasets()[OBJECTIVE]
 
 arg_min_idx = tf.squeeze(tf.argmin(dataset.observations, axis=0))
 fig, ax = plot_function_2d(branin, mins, maxs, grid_density=40, contour=True)
--- a/docs/notebooks/inequality_constraints.pct.py
+++ b/docs/notebooks/inequality_constraints.pct.py
@@ -128,7 +128,7 @@ bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
 data = bo.optimize(
     num_steps, initial_data, models, rule, track_state=False
-).try_get_final_data()
+).try_get_final_datasets()
 
 # %% [markdown]
 # To conclude, we visualise the resulting data. Orange dots show the new points queried during optimization. Notice the concentration of these points in regions near the local minima.
--- a/docs/notebooks/inequality_constraints.pct.py
+++ b/docs/notebooks/inequality_constraints.pct.py
@@ -126,15 +126,15 @@ rule = trieste.acquisition.rule.EfficientGlobalOptimization(eci)
 num_steps = 20
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
-result = bo.optimize(num_steps, initial_data, models, acquisition_rule=rule)
-
-if result.error is not None: raise result.error
+data = bo.optimize(
+    num_steps, initial_data, models, rule, track_state=False
+).try_get_final_data()
 
 # %% [markdown]
 # To conclude, we visualise the resulting data. Orange dots show the new points queried during optimization. Notice the concentration of these points in regions near the local minima.
 
 # %%
-constraint_data = result.datasets[CONSTRAINT]
+constraint_data = data[CONSTRAINT]
 new_query_points = constraint_data.query_points[-num_steps:]
 new_observations = constraint_data.observations[-num_steps:]
 new_data = (new_query_points, new_observations)
--- a/docs/notebooks/recovering_from_errors.pct.py
+++ b/docs/notebooks/recovering_from_errors.pct.py
@@ -67,7 +67,7 @@ acquisition_rule = trieste.acquisition.rule.TrustRegion()
 #
 # In this tutorial we'll try to complete fifteen optimization loops, which, with the broken observer, may take more than one attempt. The optimizer returns an `OptimizationResult`, which is simply a container for both:
 #
-#   * the final result, which uses a `Result` type (not to be confused with `OptimizationResult`), to safely encapsulates the final data, models and acquisition state if the process completed successfully, or an error if one occurred
+#   * the final result, which uses a `Result` type (not to be confused with `OptimizationResult`) to safely encapsulate the final data, models and acquisition state if the process completed successfully, or an error if one occurred
 #   * the history of the successful optimization steps.
 #
 # We can access these with the `astuple` method.
--- a/docs/notebooks/recovering_from_errors.pct.py
+++ b/docs/notebooks/recovering_from_errors.pct.py
@@ -1,0 +1,133 @@
+# %% [markdown]
+# # Recovering from errors during optimization
+
+# %%
+import numpy as np
+import tensorflow as tf
+import random
+
+np.random.seed(1793)
+tf.random.set_seed(1793)
+random.seed(3)
+
+# %% [markdown]
+# Sometimes the Bayesian optimization process encounters an error from which we can recover, without the need to restart the run from the beginning. In this tutorial, we'll simulate such an error and show how to recover from it.
+#
+# We'll use a similar setup to the [EI notebook](expected_improvement.ipynb), but use an observer that intermittently breaks when evaluated, and needs manual attention to get running again. We can simulate fixing the observer with its `manual_fix` method.
+
+# %%
+import trieste
+from trieste.acquisition.rule import OBJECTIVE
+from trieste.utils.objectives import branin
+
+
+class FaultyBranin:
+    def __init__(self):
+        self._is_broken = False
+
+    def manual_fix(self):
+        self._is_broken = False
+
+    def __call__(self, x):
+        if random.random() < 0.05:
+            self._is_broken = True
+
+        if self._is_broken:
+            raise Exception("Observer is broken")
+
+        return {OBJECTIVE: trieste.data.Dataset(x, branin(x))}
+
+
+observer = FaultyBranin()
+
+# %% [markdown]
+# ## Set up the problem
+# We'll use the same set up as before, except for the acquisition rule, where we'll use `TrustRegion`, which (with non-trivial state) will better illustrate how to recover.
+
+# %%
+import gpflow
+
+search_space = trieste.space.Box(
+    tf.cast([0.0, 0.0], tf.float64), tf.cast([1.0, 1.0], tf.float64)
+)
+initial_data = observer(search_space.sample(5))
+
+variance = tf.math.reduce_variance(initial_data[OBJECTIVE].observations)
+kernel = gpflow.kernels.Matern52(variance, [0.2, 0.2]) + gpflow.kernels.White(1e-12)
+gpr = gpflow.models.GPR(
+    initial_data[OBJECTIVE].astuple(), kernel, noise_variance=1e-5
+)
+gpflow.set_trainable(gpr.likelihood, False)
+models = {OBJECTIVE: trieste.models.GaussianProcessRegression(gpr)}
+
+acquisition_rule = trieste.acquisition.rule.TrustRegion()
+
+# %% [markdown]
+# ## Run the optimization loop
+#
+# In this tutorial we'll try to complete fifteen optimization loops, which, with the broken observer, may take more than one attempt. The optimizer returns an `OptimizationResult`, which is simply a container for both:
+#
+#   * the final result, which uses a `Result` type (not to be confused with `OptimizationResult`), to safely encapsulates the final data, models and acquisition state if the process completed successfully, or an error if one occurred
+#   * the history of the successful optimization steps.
+#
+# We can access these with the `astuple` method.
+
+# %%
+bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+
+result, history = bo.optimize(15, initial_data, models, acquisition_rule).astuple()
+
+# %% [markdown]
+# We can see from the logs that the optimization loop failed, and this can be sufficient to know what to do next if we're working in a notebook. However, sometimes our setup means we don't have access to the logs. We'll pretend from here that's the case.
+
+# %% [markdown]
+# ## Handling success
+#
+# We don't know if the optimization completed successfully or not, so we'll only try to access and plot the data if it was successful. We can find out if this was the case with `result`'s `is_ok` attribute. If it was successful, we know there is data in the `result`, which we can `unwrap` and view.
+
+# %%
+if result.is_ok:
+    data = result.unwrap().datasets[OBJECTIVE]
+    print("best observation: ", tf.reduce_min(data.observations))
+
+# %% [markdown]
+# ## Handling failure
+#
+# If on the other hand, the optimization didn't complete successfully, we can fix our observer, and try again. We can try again by using the data, model and acquisition state from the last successful step, which is the last element of the `history`.
+#
+# Note we can view any `Result` by printing it. We'll do that here to see what exception was caught.
+
+# %%
+if result.is_err:
+    print("result: ", result)
+
+    observer.manual_fix()
+
+    result, new_history = bo.optimize(
+        15 - len(history),
+        history[-1].datasets,
+        history[-1].models,
+        acquisition_rule,
+        history[-1].acquisition_state
+    ).astuple()
+
+    history.extend(new_history)
+
+# %% [markdown]
+# We can repeat this until we've spent our optimization budget, using a loop if appropriate. But here, we'll just plot the data if it exists, safely by using `result`'s `is_ok` attribute.
+
+# %%
+from util.plotting import plot_bo_points, plot_function_2d
+
+if result.is_ok:
+    data = result.unwrap().datasets[OBJECTIVE]
+    arg_min_idx = tf.squeeze(tf.argmin(data.observations, axis=0))
+    _, ax = plot_function_2d(
+        branin, search_space.lower, search_space.upper, 30, contour=True
+    )
+    plot_bo_points(data.query_points.numpy(), ax[0, 0], 5, arg_min_idx)
+
+# %% [markdown]
+# ## LICENSE
+#
+# [Apache License 2.0](https://github.com/secondmind-labs/trieste/blob/develop/LICENSE)
--- a/docs/tutorials.rst
+++ b/docs/tutorials.rst
@@ -36,6 +36,7 @@ The following tutorials (or sections thereof) explain how to use and extend spec
 * :doc:`How do I set up a basic Bayesian optimization routine?<notebooks/expected_improvement>`
 * :ref:`How do I make a custom acquisition function?<notebooks/failure_ego:Create a custom acquisition function>`
 * :ref:`How do I customize the model optimization routine?<notebooks/failure_ego:Create a custom optimize method>`
+* :doc:`How do I recover a failed optimization loop?<notebooks/recovering_from_errors>`
 
 Run the tutorials interactively
 -------------------------------
--- a/trieste/bayesian_optimizer.py
+++ b/trieste/bayesian_optimizer.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 import copy
 import traceback
 from dataclasses import dataclass
-from typing import Generic, List, Mapping, Optional, TypeVar, cast
+from typing import Generic, List, Mapping, Optional, Tuple, TypeVar, cast, overload
 
 import gpflow
 import tensorflow as tf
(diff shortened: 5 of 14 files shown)


st--:
This looks like a really neat way of handling the potentially-errored results, and it makes the notebooks much cleaner! One question  that isn't quite clear to me from the documentation is, if the optimization succeeds, whether bo.optimize().result is the same as bo.optimize().history[-1] (or perhaps not `is`-the same, but `==`-equal)? If you could clarify that (in the docs) that'd be great!

joelberkeley:
thanks @st-- . yes I've been deliberating a while about whether `.history` should include `.result.unwrap()`. I think not because:
- if it's in both, it will be less clear what the intended way of accessing it is. Also there's no guarantee the `.history[-1]` is the last step (unless you `if len(history) == num_steps` which is longer, more fragile, and less clear than `if result.is_ok`), so assuming it's so may lead to confusion
- we shouldn't add to the history at both the beginning and end, else if you concatenate the history of consecutive runs, you duplicate entries. But it's not possible to recover the one at the beginning from the `.result`

It does have the unfortunate consequence that if you want to analyse the whole run, you have to `history + [result.unwrap()]`. I don't know how to handle that. Maybe a method `def complete_history(self) = self.history + [self.result.unwrap()]`, not sure

I'm also writing a notebook to demonstrate a failed optimization run and how to recover from an earlier point

trieste/data.py
```diff
@@ -52,6 +53,12 @@ def __post_init__(self) -> None:
                 f" {self.query_points.shape}, {self.observations.shape}."
             )
 
+    def astuple(self) -> Tuple[tf.Tensor, tf.Tensor]:
+        """
+        :return: The :attr:`query_points` and :attr:`observations` as a 2-tuple.
+        """
+        return self.query_points, self.observations
```
joelberkeley-secondmind, line 60: move to its own PR

joelberkeley, line 60: move to a separate PR

trieste/observer.py
```diff
@@ -13,14 +13,14 @@
...
-from typing import Callable, Dict
+from typing import Callable, Mapping
 
 import tensorflow as tf
 
 from .data import Dataset
 
 
-Observer = Callable[[tf.Tensor], Dict[str, Dataset]]
+Observer = Callable[[tf.Tensor], Mapping[str, Dataset]]
```
joelberkeley, line 23: move to separate PR

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,150 @@
...
+
+    result, new_history = bo.optimize(
+        15 - len(history),
+        history[-1].datasets,
+        history[-1].models,
+        trust_region,
+        history[-1].acquisition_state
+    ).astuple()
+
+    history.extend(new_history)
```
joelberkeley: there's an obvious method here
```python
class BayesianOptimizer:
    ...
    def recover(
        self, total_steps: int, history: List[Record[S]], acquisition_rule[S, SP]
    ) -> OptimizationResult[S]:
       result, new_history = self.optimize(
            total_steps - len(history),
            history[-1].datasets,
            history[-1].models,
            acquisition_rule,
            history[-1].acquisition_state
       )
       return OptimizationResult(result, history + new_history)
```
but I'll leave that for another PR

trieste/utils/misc.py
```diff
@@ -55,3 +57,112 @@ def to_numpy(t: TensorType) -> np.ndarray:
         return t.numpy()
 
     return t
+
+
+T_co = TypeVar("T_co", covariant=True)
+""" An unbounded covariant type variable. """
+
+
+class Result(Generic[T_co], ABC):
```
joelberkeley, line 66: @st-- 's qu: how does this compare to the `result` package ?

trieste/bayesian_optimizer.py
```diff
@@ -39,47 +40,65 @@
...
+        **Note:** In contrast to the standard library function :func:`dataclasses.astuple`, this
+        method does *not* deepcopy instance attributes.
 
-    error: Optional[Exception]
-    """ The exception that occurred, if any. """
+        :return: The :attr:`final_result` and :attr:`history` as a 2-tuple.
+        """
+        return self.final_result, self.history
+
+    def try_final_data(self) -> Mapping[str, Dataset]:
```
ChrisMorter: `try_get_final_data` might make it clearer what this does, even if it is slightly longer? I was a bit confused the first time I read this, I initially thought it was something to do with getting model predictions for the most recent data.

trieste/bayesian_optimizer.py
```diff
@@ -39,47 +40,65 @@
...
+        """
+        Convenience method to attempt to get the final data.
+
+        :return: The final data, if the optimization completed successfully.
+        :raise Exception: If an exception occurred during optimization.
+        """
+        return self.final_result.unwrap().datasets
+
+    def try_final_models(self) -> Mapping[str, TrainableProbabilisticModel]:
```
ChrisMorter: Similar to above, I think  this could do with `get` in the same.

ChrisMorter (approved):
Looks good, just a couple of comments regarding naming.

trieste/bayesian_optimizer.py
joelberkeley: hmmm yes i see what you mean

trieste/utils/misc.py
joelberkeley: i think [returns](https://github.com/dry-python/returns) is the best candidate for this, but I'd like to hold off adding that dependency until we need more functionality from it. For now, we don't need much.

docs/notebooks/expected_improvement.pct.py
```diff
@@ -81,18 +79,13 @@ def build_model(data):
 #
 # The `optimize` method returns several things (see the `optimize` documentation for details), but we're only interested in the data, which captures the points where the objective was queried and the resulting values. Note that the optimizer updates the model in place.
 #
-# However, since the optimization loop catches errors so as not to lose progress, we must check if any errors occurred so we know the data is valid. We'll do that crudely here by re-raising any such errors. You may wish instead to use the history to restore the process from an earlier point.
+# However, the optimization loop catches errors so as not to lose progress, which means the optimization loop might not complete and the data from the last step may not exist. Here we'll handle this crudely by attempting to force the data, which will raise any errors that did occur. For a review of how to handle errors systematically, there is a [dedicated tutorial](recovering_from_errors.ipynb).
```
st--: What does "to force the data" mean? not clear to me

docs/tutorials.rst
```diff
@@ -36,6 +36,7 @@ The following tutorials (or sections thereof) explain how to use and extend spec
 * :doc:`How do I set up a basic Bayesian optimization routine?<notebooks/expected_improvement>`
 * :ref:`How do I make a custom acquisition function?<notebooks/failure_ego:Create a custom acquisition function>`
 * :ref:`How do I customize the model optimization routine?<notebooks/failure_ego:Create a custom optimize method>`
+* :doc:`How do I recover a failed optimization loop?<notebooks/recovering_from_errors>`
```
st--, line 39: what's the difference between `:ref:` and `:doc:`?

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,144 @@
...
+# ## Set up the problem
+# We'll use the same set up as before, except for the acquisition rule, where we'll use `TrustRegion`, which will better illustrate how to recover.
+
+# %%
+import gpflow
+
+search_space = trieste.space.Box(
+    tf.cast([0.0, 0.0], tf.float64), tf.cast([1.0, 1.0], tf.float64)
+)
+initial_data = observer(search_space.sample(5))
```
st--, line 53: This call won't fail even though you get five samples because it's guaranteed not to fail on the first call. But is that something that'd be worth mentioning in a short comment?

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,144 @@
...
+
+# %% [markdown]
+# ## Run the optimization loop
+#
+# The optimizer returns an `OptimizationResult`. We can use that object's `astuple` method to get a 2-tuple of the final result and the history of the successful optimization steps.
+
+# %%
+bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+
+result, history = bo.optimize(15, initial_data, models, acquisition_rule).astuple()
```
st--, line 78: Mention in a comment what the "15" is? or define a constant for it and pass it in by name?

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,144 @@
...
+
+# %% [markdown]
+# We can see from the logs that the optimization loop failed, and this can be sufficient to know what to do next if we're working in a notebook. However, sometimes our setup means we don't have access to the logs. We'll pretend from here that's the case.
```
st--, line 81: How can the user see that from the logs ?

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,144 @@
...
+
+# %% [markdown]
+# ## Handling success
+#
+# We don't know if the optimization completed successfully or not, so we'll only try to access and plot the data if it was successful. We can find out if this was the case with the `is_ok` attribute on the `Result`. If it was successful, we know there is data in the `result`, which we can `unwrap` and view.
```
st--: It's a bit confusing that first you only talk about `OptimizationResult`, and then that _contains_ the `Result` that you now talk about here (which from reading the code I understand to be the first element of .astuple()). What might be a way of explaining that difference more clearly?

docs/notebooks/recovering_from_errors.pct.py
st--, line 78: What is the `bo_result` (for lack of a better name) good for other than splitting it into `result` and `history` using astuple()?

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,144 @@
...
+        history[-1].datasets,
+        history[-1].models,
+        acquisition_rule,
+        history[-1].acquisition_state
+    ).astuple()
+
+    history.extend(new_history)
+
+# %% [markdown]
+# We can repeat the above until the optimization process is complete, or we can put the whole process in a loop
```
st--:
```suggestion
# We can repeat the above until the optimization process is complete, or we can put the whole process in a loop.
```

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,144 @@
...
+
+# %%
+while result.is_err:
```
st--: This requires a first manual call to `bo.optimize()`: maybe worth adding a comment on that?

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,144 @@
...
+        history[-1].datasets,
+        history[-1].models,
+        acquisition_rule,
+        history[-1].acquisition_state
+    ).astuple()
+
+    history.extend(new_history)
+
+# %% [markdown]
+# If we get to this point, we know the optimization process must have completed, because `is_ok` must be `True`, so we can safely plot the data
```
st--:
```suggestion
# If we get to this point, we know the optimization process must have completed, because `is_ok` must be `True`, so we can safely plot the data.
```
Also, what if you don't get to this point: could you end up in an infinite loop here?

trieste/utils/__init__.py
```diff
@@ -13,4 +13,4 @@
 # limitations under the License.
 """ This package contains library utilities. """
 from . import objectives
-from .misc import jit, shapes_equal, to_numpy
+from .misc import Err, Ok, Result, T_co, jit, shapes_equal, to_numpy
```
st--, line 16: Is `T_co` actually needed outside of the `misc` package? seems a slightly weird export to me.

trieste/utils/misc.py
```diff
@@ -55,3 +57,114 @@ def to_numpy(t: TensorType) -> np.ndarray:
...
+    """ Wraps the result of a successful evaluation. """
+
+    def __init__(self, value: T_co):
+        """
+        :param value: The result of a successful evaluation.
+        """
+        self._value = value
+
+    def __repr__(self) -> str:
+        """"""
```
st--, line 132:
```suggestion
```
?
(Or
```suggestion
        ""
```
at least looks more like 'empty string' to me)

trieste/utils/misc.py
```diff
@@ -55,3 +57,114 @@ def to_numpy(t: TensorType) -> np.ndarray:
...
+        return f"Ok({self._value!r})"
+
+    @property
+    def is_ok(self) -> bool:
+        """ True always. """
```
st--:
```suggestion
        """ Always `True`; this object contains a value. """
```

trieste/utils/misc.py
```diff
@@ -55,3 +57,114 @@ def to_numpy(t: TensorType) -> np.ndarray:
...
+    """ Wraps the exception that occurred during a failed evaluation. """
+
+    def __init__(self, exc: Exception):
+        """
+        :param exc: The exception that occurred.
+        """
+        self._exc = exc
+
+    def __repr__(self) -> str:
+        """"""
```
st--, line 158:
```suggestion
```
or "" as before

trieste/utils/misc.py
```diff
@@ -55,3 +57,114 @@ def to_numpy(t: TensorType) -> np.ndarray:
...
+        return f"Err({self._exc!r})"
+
+    @property
+    def is_ok(self) -> bool:
+        """ False always. """
```
st--:
```suggestion
        """ Always `False`; this object contains an error. """
```

st-- (commented):
I've not reviewed trieste.bayesian_optimizer yet (will try and do so shortly).

docs/notebooks/expected_improvement.pct.py
joelberkeley: will clarify

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: will do

joelberkeley: I've done it so that users can access the data using the workflow that suits them. Originally it was a plain tuple, but I wanted to let people use `.result` or `.history` for clarity, or if they prefer chaining calls. Then victor said it's a little convoluted to get the data as `.result.unwrap().datasets`, so I added the `try_final_data` etc methods

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: It will appear when run as a notebook. I've written this primarily for notebook users/people reading the built docs

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: yeah we could. I'll mention that we're confident we'll finish in this case case cos there's only a 30% chance of failure

docs/tutorials.rst
joelberkeley: `:ref:` points to a section in a file, `:doc:` points to a whole file

trieste/utils/misc.py
joelberkeley: i added this because the docs inherit `object.__repr__`'s docstring, which doesn't add anything particularly useful and appears all over the docs. This overrides that (so that `__repr__` doesn't appear in the docs). As for `""""""`, PEP-257 says docstrings should always use triple-double quotes, and I expect some linters will flag up anything else

trieste/utils/__init__.py
joelberkeley: ah, yes, it's needed for the docs. We don't put every module in the docs (to keep both import paths and the docs simple). The `misc` module is one of those, so we need to import everything that needs to appear in the API

docs/notebooks/recovering_from_errors.pct.py
st--: is there an issue to keep track of this?

trieste/bayesian_optimizer.py
```diff
@@ -39,47 +40,65 @@
...
+    The final result of the optimization process. This contains either a :class:`Record` or an
+    exception.
     """
 
-    models: Mapping[str, TrainableProbabilisticModel]
-    """
-    The models over the :attr:`datasets` (unless :attr:`error` is populated, in which case this is
-    the models from the point at which the process was interrupted). """
+    history: List[Record[S]]
+    r""" The history of the :class:`Record`\ s from each step of the optimization process. """
```
st--: Could you clarify whether history[-1] == final_result?

docs/notebooks/expected_improvement.pct.py
```diff
@@ -81,18 +79,13 @@ def build_model(data):
...
-
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
-result: OptimizationResult = bo.optimize(15, initial_data, model)
-
-if result.error is not None: raise result.error
-
-dataset = result.datasets[OBJECTIVE]
+result = bo.optimize(15, initial_data, model)
+dataset = result.try_final_data()[OBJECTIVE]
```
st--: I'd find it helpful to get at least a brief comment on what the `[OBJECTIVE]` does (can be a link to another place in the docs), as well as what `try_final_data` does (get you the data, or raise exception) - this is mostly explained in the markdown cell above, but connecting the textual description to which line in the code does that would be helpful (i.e., bo.optimize() will never throw an exception [perhaps with the exception of actual bugs in the code], but try_final_data() might - this is not clear from this notebook)

docs/notebooks/expected_improvement.pct.py
```diff
@@ -155,7 +148,7 @@ def build_model(data):
 # %%
 from util.plotting_plotly import plot_gp_plotly
 
-fig = plot_gp_plotly(model[OBJECTIVE].model, mins, maxs, grid_density=30)
+fig = plot_gp_plotly(result.try_final_models()[OBJECTIVE].model, mins, maxs, grid_density=30)
```
st--: same as above - brief comment on what `try_final_models` does?

docs/notebooks/expected_improvement.pct.py
```diff
@@ -174,11 +167,11 @@ def build_model(data):
 # We can also inspect the model hyperparameters, and use the history to see how the length scales evolved over iterations
 
 # %%
-gpflow.utilities.print_summary(model[OBJECTIVE].model)
+gpflow.utilities.print_summary(result.try_final_models()[OBJECTIVE].model)
 
 ls_list = [
     step.models[OBJECTIVE].model.kernel.lengthscales.numpy()  # type: ignore
-    for step in result.history
+    for step in result.history + [result.final_result.unwrap()]
```
st--, line 174: this relates to my comment on the OptimizationResult class - would be nice to explicitly clarify that result.history[-1] != result.final_result (if it does exist)

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: actually, I'm not sure this loop adds much, and won't work in many cases, so I'll remove it

docs/notebooks/recovering_from_errors.pct.py
st--: ok, up to you whether you think that's obvious enough by itself or whether it might be helpful to clarify that "logs" refers to "what bo.optimize() prints to stdout [rather than some logfile that I have to dig up somewhere]"?

trieste/bayesian_optimizer.py
```diff
@@ -198,24 +253,16 @@ def optimize(
 
             except Exception as error:
                 tf.print(
```
st--, line 259: Unrelated to this PR, but I was surprised by this combination of tf.print() (whose only purpose, really, is for running inside tf.function()) and Exception handling (which I _believe_ would not play well together with tf.function() - can you confirm that?)... Is this code run both using and not using tf.function wrapper?

trieste/utils/misc.py
st--: Ah, I didn't know the docs finally handle inheriting docstrings (I remember having found the lack of this quite frustrating in the past-- I even wrote a metaclass that copies docstrings from the superclass if not overwritten...). In that case, it might be helpful for `is_ok` to NOT overwrite the docstring, and get the generic one from the parent class?

trieste/bayesian_optimizer.py
joelberkeley: very good spot. hmmm ... i don't expect we can do without exception handling, so perhaps `tf.print` is useless

docs/notebooks/expected_improvement.pct.py
joelberkeley: i've updated the text. Is it clear now?

docs/notebooks/expected_improvement.pct.py
joelberkeley: done

trieste/utils/misc.py
joelberkeley: this is using sphinx-autoapi which is a different engine to what you're using

joelberkeley: i'd prefer to be specific in the subclasses since we can

joelberkeley: note the users are more likely to read the docs for `Result` than its implementations (don't know what that implies)

trieste/bayesian_optimizer.py
joelberkeley: will do

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: done

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: instead, i'll make it so that it could fail, but the seed is such that it doesn't

joelberkeley: (I think it's better if it can always fail)

docs/notebooks/expected_improvement.pct.py
joelberkeley: done

trieste/bayesian_optimizer.py
joelberkeley: added note to https://github.com/secondmind-labs/trieste/issues/90#issuecomment-763660627

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: removed the loop instead

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: #124

trieste/utils/misc.py
joelberkeley: The docs are pretty compact already (i.e. it's clear this contains a value in the built docs). I don't think this extra bit is needed

I've formatted the `True` and `False` with backticks

trieste/bayesian_optimizer.py
joelberkeley: done

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: i've extended the markdown section

docs/notebooks/expected_improvement.pct.py
```diff
@@ -150,12 +141,14 @@ def build_model(data):
...
 # %% [markdown]
-# We can visualise the model over the objective function by plotting the mean and 95% confidence intervals of its predictive distribution.
+# We can visualise the model over the objective function by plotting the mean and 95% confidence intervals of its predictive distribution. Like with the data before, we can get the model with `.try_get_final_models()` and indexing with `OBJECTIVE`.
 
 # %%
 from util.plotting_plotly import plot_gp_plotly
 
-fig = plot_gp_plotly(model[OBJECTIVE].model, mins, maxs, grid_density=30)
+fig = plot_gp_plotly(
+    result.try_get_final_models()[OBJECTIVE].model, mins, maxs, grid_density=30
```
st--, line 150: Unrelated to this PR but it's a bit weird that you have to do `model[...].model` ... is that something you'd want to clear up? (Note that this is different from the convention for getting the dataset, which is just `datasets[...]`.) In that case we could turn this into an issue to keep track of it.

docs/notebooks/expected_improvement.pct.py
```diff
@@ -171,14 +164,14 @@ def build_model(data):
 fig.show()
 
 # %% [markdown]
-# We can also inspect the model hyperparameters, and use the history to see how the length scales evolved over iterations
+# We can also inspect the model hyperparameters, and use the history to see how the length scales evolved over iterations. Note the history is saved at the *start* of each step, and as such never includes the final result, so we'll add that ourselves.
```
st--, line 167: :+1:

docs/notebooks/expected_improvement.pct.py
```diff
@@ -79,20 +77,13 @@ def build_model(data):
...
-
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
-result: OptimizationResult = bo.optimize(15, initial_data, model)
-
-if result.error is not None: raise result.error
-
-dataset = result.datasets[OBJECTIVE]
+result = bo.optimize(15, initial_data, model)
+dataset = result.try_get_final_data()[OBJECTIVE]
```
st--: Should this be `try_get_final_datasets` to be consistent e.g. with _get_models ? (`datasets` also makes it a bit clearer that it's several and you want to index to get a specific one, IMO)

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -126,15 +126,15 @@ def create_bo_model(data):
...
+data = bo.optimize(
+    num_steps, initial_data, models, rule, track_state=False
+).try_get_final_data()
 
 # %% [markdown]
 # To conclude, we visualise the resulting data. Orange dots show the new points queried during optimization. Notice the concentration of these points in regions near the local minima.
 
 # %%
-constraint_data = result.datasets[CONSTRAINT]
+constraint_data = data[CONSTRAINT]
```
st--, line 137:
```suggestion
constraint_data = datasets[CONSTRAINT]
```
as per my comment on expected_improvement notebook ("data" vs "datasets")

docs/notebooks/recovering_from_errors.pct.py
st--: In that case it might be useful to clarify that (e.g. point out in practice someone would have to also develop a fault-tolerant way of collecting the initial dataset)

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,133 @@
...
+models = {OBJECTIVE: trieste.models.GaussianProcessRegression(gpr)}
+
+acquisition_rule = trieste.acquisition.rule.TrustRegion()
+
+# %% [markdown]
+# ## Run the optimization loop
+#
+# In this tutorial we'll try to complete fifteen optimization loops, which, with the broken observer, may take more than one attempt. The optimizer returns an `OptimizationResult`, which is simply a container for both:
+#
+#   * the final result, which uses a `Result` type (not to be confused with `OptimizationResult`), to safely encapsulates the final data, models and acquisition state if the process completed successfully, or an error if one occurred
```
st--:
```suggestion
#   * the final result, which uses a `Result` type (not to be confused with `OptimizationResult`), to safely encapsulate the final data, models and acquisition state if the process completed successfully, or an error if one occurred;
```

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -0,0 +1,133 @@
...
+#   * the history of the successful optimization steps.
+#
+# We can access these with the `astuple` method.
```
st--, line 73: As a user I'd find it valuable to have in this notebook at least a mention of/pointer to the other methods on the OptimizationResult object, such as the try_get_models etc.

trieste/bayesian_optimizer.py
```diff
@@ -39,47 +40,69 @@
...
-    """ The acquisition state after this optimization step. """
+    """ The acquisition state. """
 
 
+# this should be a generic NamedTuple, but mypy doesn't support them
+#  https://github.com/python/mypy/issues/685
 @dataclass(frozen=True)
 class OptimizationResult(Generic[S]):
-    """ Container for the final result of the optimization process. """
+    """ The final result, and the historical data of the optimization process. """
```
st--, line 60:
```suggestion
    """ Container for both the final result and the historical data of the optimization process. """
```

trieste/bayesian_optimizer.py
```diff
@@ -39,47 +40,69 @@
...
+    The final result of the optimization process. This contains either a :class:`Record` or an
+    exception.
     """
 
-    models: Mapping[str, TrainableProbabilisticModel]
+    history: List[Record[S]]
+    r"""
+    The history of the :class:`Record`\ s from each step of the optimization process. These
+    :class:`Record`\ s are created at the *start* of each loop, and as such will never include the
+    :attr:`final_result`.
```
st--, line 72: :+1:

trieste/bayesian_optimizer.py
```diff
@@ -39,47 +40,69 @@
...
+    def astuple(self) -> Tuple[Result[Record[S]], List[Record[S]]]:
+        """
+        **Note:** In contrast to the standard library function :func:`dataclasses.astuple`, this
+        method does *not* deepcopy instance attributes.
+
+        :return: The :attr:`final_result` and :attr:`history` as a 2-tuple.
+        """
+        return self.final_result, self.history
+
+    def try_get_final_data(self) -> Mapping[str, Dataset]:
```
st--:
```suggestion
    def try_get_final_datasets(self) -> Mapping[str, Dataset]:
```
for consistency throughout the code

trieste/bayesian_optimizer.py
```diff
@@ -39,47 +40,69 @@
...
+        """
+        Convenience method to attempt to get the final data.
```
st--, line 86:
```suggestion
        Convenience method to attempt to get the final datasets.
```

trieste/bayesian_optimizer.py
```diff
@@ -39,47 +40,69 @@
...
+
+        :return: The final data, if the optimization completed successfully.
```
st--, line 88:
```suggestion
        :return: The final datasets, if the optimization completed successfully.
```

trieste/bayesian_optimizer.py
st--: Or, you could change the implementation of this to assert that Dataset only has a single key:value pair and return its value (and perhaps like wise for try_get_final_model?)

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -126,15 +126,15 @@ def create_bo_model(data):
 num_steps = 20
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
-result = bo.optimize(num_steps, initial_data, models, acquisition_rule=rule)
-
-if result.error is not None: raise result.error
+data = bo.optimize(
+    num_steps, initial_data, models, rule, track_state=False
+).try_get_final_data()
```
st--:
```suggestion
datasets = bo.optimize(
    num_steps, initial_data, models, rule, track_state=False
).try_get_final_datasets()
```

trieste/bayesian_optimizer.py
```diff
@@ -115,48 +161,57 @@ def optimize(
...
+            that fails to complete). The history will never include the final optimization result.
         :raise ValueError: If any of the following are true:
+
+            - ``num_steps`` is negative.
             - the keys in ``datasets`` and ``model_specs`` do not match
             - ``datasets`` or ``model_specs`` are empty
             - the default `acquisition_rule` is used and the tags are not `OBJECTIVE`.
         """
+        if num_steps < 0:
+            raise ValueError(f"num_steps must be at least 0, got {num_steps}")
```
st--, line 213: What does `num_steps == 0` do!? Is that equivalent to "do nothing"? Should that error as well? I guess it does some model construction. What's the use case for that? :thinking:

trieste/bayesian_optimizer.py
```diff
@@ -176,12 +231,16 @@ def optimize(
...
+        history: List[Record[S]] = []
 
         for step in range(num_steps):
+            if track_state:
+                history.append(Record(datasets, models, acquisition_state))
+
             try:
                 if track_state:
-                    _save_to_history(history, datasets, models, acquisition_state)
+                    models = {tag: gpflow.utilities.deepcopy(m) for tag, m in models.items()}
```
st--, line 242: You can replace this with just
```suggestion
                    models = {tag: copy.deepcopy(m) for tag, m in models.items()}
```
if you're willing to constraining to tensorflow_probability >=0.12 (i.e. tensorflow >=2.4).

trieste/utils/misc.py
```diff
@@ -55,3 +57,114 @@ def to_numpy(t: TensorType) -> np.ndarray:
...
+        """
+        self._value = value
+
+    def __repr__(self) -> str:
+        """"""
+        return f"Ok({self._value!r})"
+
+    @property
+    def is_ok(self) -> bool:
+        """ `True` always. """
```
st--, line 137: I still find this docstring more confusing than helpful, especially as `is_err` won't say "`False` always." but have the parent class's docstring... but this is getting very nitpicky - won't block the PR on this. :wink:

st-- (approved):
Looks very good. One thing I'd really like to see changed before merging though is renaming `try_get_final_data` to `try_get_final_datasets` to make it consistent with how it actually behaves (as well as with `try_get_final_models`).

docs/notebooks/expected_improvement.pct.py
joelberkeley: yeah it's weird. It's actually a hack which I have been working to fix - I just change the function to use the `ProbabilisticModel` API rather than the underling gpflow model API. It comes under #33

docs/notebooks/expected_improvement.pct.py
joelberkeley: done. I actually plan to rename `datasets` everywhere to `data`. Until we get there, `try_get_final_datasets` aligns with the `datasets` attribute

trieste/bayesian_optimizer.py
joelberkeley: done

trieste/bayesian_optimizer.py
joelberkeley: it is, I included it because an unsupervised system (e.g. production) may dynamically populate `num_steps`, and there's no reason it should fail for zero steps. It does build the models, it's true, but that's pretty cheap (it's just instantiating some classes), so I didn't bother doing anything about that

trieste/bayesian_optimizer.py
joelberkeley: i have another PR for this #97

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: that's more in scope of the intro notebooks. This isn't so much a notebook about the `OptimizationResult`, as how to recover the optimization process

PR #18, opened 2020-10-13
Added support for a Max-value Entropy Search Acquisition Function

henrymoss:
Hi. 

Whilst getting my head round the package (great work by the way!) I coded up a version of Max-value Entropy search (https://arxiv.org/abs/1703.01968) suitable for Trieste.

I have also included a notebook detailing its use (similar to your Thompson sampling one) and added an integration test.

Let me know if you have an issues.


joelberkeley-secondmind:
Hi Henry, thanks for the pull request! And congratulations for being our first contributor :D 

We'll have a look through it and get back to you this week.

trieste/acquisition/rule.py
```diff
@@ -117,6 +117,50 @@ def acquire(
         return point, None
 
 
+class MaxValueEntropySearch(AcquisitionRule[None, SearchSpace]):
```
vpicheny: do we have a need for a new rule? it seems to me that MES should be an acquisition function only, that uses EfficientGlobalOptimization

trieste/acquisition/function.py
```diff
@@ -130,6 +133,120 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        # find quantiles Pr(y*<y1)=r1 and Pr(y*<y2)=r2
+
+        right = tf.squeeze(fmean[tf.squeeze(idx)])
+        left = right
+        probf = lambda x: tf.math.exp(tf.math.reduce_sum(tfp.distributions.Normal(tf.cast(0,tf.float64),tf.cast(1,tf.float64)).log_cdf(-(x - fmean) / fsd), axis=0))
+        
+
+        # get range for binary search
+        i = 0
+        while probf(left) < 0.75:
```
vpicheny: do we have a risk of running an infinite loop here?

trieste/acquisition/function.py
```diff
@@ -130,6 +133,120 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+
+
+        return lambda at: self._acquisition_function(model, samples, at)
+
+    @staticmethod
+    def _acquisition_function(model: ModelInterface, samples: tf.Tensor, at: QueryPoints) -> tf.Tensor:
+        return mes(model, samples, at)
+
+
+def mes(model: ModelInterface, samples: tf.Tensor, at: QueryPoints) -> tf.Tensor:
```
vpicheny: replace with an explicit name, eg maxvalue_entropy_change (or ..._reduction, or any better name)?

trieste/acquisition/rule.py
henrymoss: The issue is that MES requires some extra arguments that govern the performance of the approximations its uses. A single EfficientGlobalOptimization rule would make more sense, but will need to be able to take generic extra arguments when initializing.

trieste/acquisition/function.py
henrymoss: I have used this exact bit of code many times before and it will always terminate. It keeps adding larger and larger elements to "left" until it hits the threshold.

trieste/acquisition/rule.py
joelberkeley-secondmind: it appears the need to add arguments is arising from requiring the rule to have MES as a default argument. I would say the correct flow would be to pass `MES` to `EfficientGlobalOptimization`:
```python
trieste.acquisition.rule.EfficientGlobalOptimization(
    trieste.acquisition.MES(search_space, num_samples=10, grid_size=5000)
)
```
Does that make sense?

trieste/acquisition/function.py
joelberkeley-secondmind: agreed. I think `MES -> MaxValueEntropySearch` and `mes` -> `max_value_entropy_search` would be good. What do you reckon?

docs/notebooks/max_value_entropy_search.pct.py
```diff
@@ -0,0 +1,159 @@
+# %% [markdown]
```
joelberkeley-secondmind: A notebook could be a nice way to demonstrate the new acquisition rule, but it could be in a different PR. We're aiming to keep PRs minimal. @vpicheny also suggested we could compare EI and max value entropy search in a single notebook, perhaps the existing intro notebook

trieste/acquisition/rule.py
henrymoss: I had missed that this was what the EfficientGlobalOptimization class was for! I've got rid of the extra MES rule now.

docs/notebooks/max_value_entropy_search.pct.py
henrymoss: I've made the changes and removed the notebook. I'll commit it in another PR. 

It might be nice to have it as part of the intro notebook as you suggest. Perhaps with a figure plotting the different acquisition function surfaces according to the GP once fit on the initial points?

trieste/acquisition/rule.py
joelberkeley-secondmind: great. That you'd missed it suggests we need to document it more loudly or clearly

docs/notebooks/max_value_entropy_search.pct.py
vpicheny: that's exactly what I had in mind

henrymoss: Nice. I will try that in another PR.

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        Need to sample possible min-values from our posterior. To do this we implement a Gumbel sampler.
+        we approximate Pr(y*^hat<y) by Gumbel(alpha,beta) then sample from Gumbel.
+
+        :param dataset: Unused.
+        :param model: The model over the specified ``dataset``.
+        :return: The MES function.
+        """
+
+        # generate random grid to fit Gumbel distribution
+        # grid consists of already queried locations and grid_size other random locations
```
joelberkeley-secondmind: comments tend to drift out of line with the code, so I'd prefer it if we left them out

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        query_points = self._search_space.sample(self._grid_size)
+        query_points = tf.concat([dataset.query_points,query_points],0)
+        fmean, fvar = model.predict(query_points)
+        fsd = tf.math.sqrt(fvar)
+        idx = tf.math.argmax(-fmean[:tf.shape(dataset.query_points)[0]])
```
joelberkeley-secondmind: i'd expect `fmean` to have the same length as `query_points`, and `argmax ~ - argmin`, so could this be
```suggestion
        idx = tf.argmin(fmean[:-1])
```

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+    :param model: The model of the objective function.
+    :param samples: Samples from p_min
+    :param at: The points for which to calculate the expected improvement.
+    :return: The expected improvement at ``at``.
+    """
+    fmean, fvar = model.predict(at)
+    fsd = tf.math.sqrt(fvar)
+    
+    # clip below to improve numerical stability
+    fsd = tf.clip_by_value(fsd, 1.0e-8, tf.float64.max)
```
joelberkeley-secondmind: on the other hand, i think these comments are useful. Can you put them on the same line so as to prevent drift? Like
```suggestion
    fsd = tf.clip_by_value(fsd, 1.0e-8, tf.float64.max)  # clip below to improve numerical stability
```

joelberkeley-secondmind: (that is, if they fit)

trieste/acquisition/function.py
joelberkeley-secondmind: also why are we not including the last row?

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        query_points = tf.concat([dataset.query_points,query_points],0)
+        fmean, fvar = model.predict(query_points)
+        fsd = tf.math.sqrt(fvar)
+        idx = tf.math.argmax(-fmean[:tf.shape(dataset.query_points)[0]])
+
+        # Fit Gumbel distriubtion
+        # scaling so that gumbel scale is proportional to IQ range of cdf Pr(y*<z)
+        # find quantiles Pr(y*<y1)=r1 and Pr(y*<y2)=r2
+        right = tf.squeeze(fmean[tf.squeeze(idx)])
+        left = right
```
joelberkeley-secondmind: this could be `left = right = tf.squeeze(fmean[tf.squeeze(idx)])` since `tf.Tensor`s are immutable

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        probf = lambda x: tf.math.exp(tf.math.reduce_sum(tfp.distributions.Normal(tf.cast(0,tf.float64),tf.cast(1,tf.float64)).log_cdf(-(x - fmean) / fsd), axis=0))
```
joelberkeley-secondmind: this line is quite hard to read. Can we use a function e.g.
```suggestion
        def probf(x: tf.Tensor) -> tf.Tensor :
            unit_normal = tfp.distributions.Normal(tf.cast(0, tf.float64), tf.cast(1, tf.float64))
            log_cdf = unit_normal.log_cdf(- (x - fmean) / fsd)
            return tf.exp(tf.reduce_sum(log_cdf), axis=0))
```

trieste/acquisition/function.py
joelberkeley-secondmind: alternatively, since we recalculate the posterior best later,
```suggestion
        posterior_best = tf.squeeze(fmean[tf.squeeze(idx)])
        left = right = posterior_best
```
then reuse `posterior_best` later

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+            right = -2. ** i * tf.math.reduce_min(fmean - 5. * fsd) + (1. + 2. ** i) * tf.squeeze(fmean[tf.squeeze(idx)])
+            i += 1
+
+        # Binary search for 3 percentiles
+        q1, med, q2 = map(lambda val: bisect(lambda x: probf(x) - val, left, right, maxiter=10000, xtol=0.00001),
+                            [0.25, 0.5, 0.75])
+
+        # solve for gumbel params
+        beta = (q1 - q2) / (tf.math.log(tf.math.log(4. / 3.)) - tf.math.log(tf.math.log(4.)))
+        alpha = med + beta * tf.math.log(tf.math.log(2.))
```
joelberkeley-secondmind: is this algorithm 1 in the paper? How come you used `alpha` and `beta` when the paper uses `a` and `b` and uses `\alpha` for the result of the acquisition function?

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+            left = 2. ** i * tf.math.reduce_min(fmean - 5. * fsd) + (1. - 2. ** i) * right
+            i += 1
+        i = 0
+        while probf(right) > 0.25:
+            right = -2. ** i * tf.math.reduce_min(fmean - 5. * fsd) + (1. + 2. ** i) * tf.squeeze(fmean[tf.squeeze(idx)])
+            i += 1
+
+        # Binary search for 3 percentiles
+        q1, med, q2 = map(lambda val: bisect(lambda x: probf(x) - val, left, right, maxiter=10000, xtol=0.00001),
+                            [0.25, 0.5, 0.75])
```
joelberkeley-secondmind: this is a bit hard to read. How about using a function
```suggestion
        def search(val: float) -> float:  # or some more informative name
            return bisect(lambda x: probf(x) - val, left, right, maxiter=10000, xtol=0.00001)

        q1, med, q2 = map(search, [0.25, 0.5, 0.75])
```

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
 
+class MaxValueEntropySearch(SingleModelAcquisitionBuilder):
+    """
+    Builder for the max-value entropy search acqusiiton function (for function minimisation)
+    """
+
+    def __init__(self,search_space: SearchSpace, num_samples: int = 10, grid_size: int = 5000):
+        """
+        :param num_samples: integer determining how many samples to draw of the minimum 
+            (does not need to be large).
```
joelberkeley-secondmind: can you elaborate here? what does large mean?

joelberkeley-secondmind: also no need to mention it's an integer

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        :param grid_size: number of random locations in grid used to fit the gumbel distribution 
+            and approximately generate the samples of the minimum (recommend scaling with problem dimension).
```
joelberkeley-secondmind: scaling linearly with problem dimension? and to clarify, is the problem dimension the query point dimension?

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        """
+        self._num_samples = num_samples
+        self._grid_size = grid_size
```
joelberkeley-secondmind: What are the restrictions on the size of these numbers? Do they need to be `> 0`, `>= 0`? Can we add argument validation, sth like here https://github.com/secondmind-labs/trieste/blob/develop/trieste/acquisition/function.py#L199

trieste/acquisition/function.py
```diff
@@ -130,6 +133,117 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
     return (eta - mean) * normal.cdf(eta) + variance * normal.prob(eta)
 
 
+class MaxValueEntropySearch(SingleModelAcquisitionBuilder):
+    """
+    Builder for the max-value entropy search acqusiiton function (for function minimisation)
```
joelberkeley-secondmind:
```suggestion
    Builder for the max-value entropy search acquisition function (for function minimisation)
```

trieste/acquisition/function.py
henrymoss: This is just to find a good initialisation for the bisection search. We get the best location from the already queried points (not the whole sampled grid).

fmean has tf.shape(dataset.query_points)[0] + tf.shape(self_.grid_size)[0] points, so fmean[:-1] is not what we want here.

trieste/acquisition/function.py
henrymoss: I'll change that. No harm in being consistent with the paper!

trieste/acquisition/function.py
henrymoss: The dimensions of the search space

trieste/acquisition/function.py
henrymoss: I have changed this bit now (following the MES implementation in BOTorch).

My approach here turns out to be overkill.

henrymoss:
I've adjusted the binary search bit. Its definitely nicer now, so thanks for your help!

trieste/acquisition/function.py
```diff
@@ -130,6 +133,110 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+
+        if num_samples <= 0:
+            raise ValueError(
+                f"num_samples must be positive, got {num_samples}"
+            )
+        self._num_samples = num_samples
+
+        if grid_size <= 0:
+            raise ValueError(
+                f"num_samples must be positive, got {grid_size}"
```
joelberkeley-secondmind:
```suggestion
                f"grid_size must be positive, got {grid_size}"
```

trieste/acquisition/function.py
```diff
@@ -130,6 +133,110 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+
+    def prepare_acquisition_function(
+        self, dataset: Dataset, model: ModelInterface
+    ) -> AcquisitionFunction:
+        """
+        Need to sample possible min-values from our posterior. 
+        To do this we implement a Gumbel sampler.
+        We approximate Pr(y*^hat<y) by Gumbel(a,b) then sample from Gumbel.
+
+        :param dataset: Unused.
```
joelberkeley-secondmind: the dataset is used below (which has me notice the docstrings for EI need correcting)
```suggestion
        :param datasets: The data from the observer.
```

trieste/acquisition/function.py
```diff
@@ -130,6 +133,110 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+          title={Max-value entropy search for efficient Bayesian optimization},
+          author={Wang, Zi and Jegelka, Stefanie},
+          journal={arXiv preprint arXiv:1703.01968},
+          year={2017}
+        }
+
+    :param model: The model of the objective function.
+    :param samples: Samples from p_min
+    :param at: The points for which to calculate the expected improvement.
+    :return: The expected improvement at ``at``.
```
joelberkeley-secondmind: is "expected improvement" right here?

trieste/acquisition/function.py
```diff
@@ -130,6 +133,110 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        return lambda at: self._acquisition_function(model, gumbel_samples, at)
+
+    @staticmethod
+    def _acquisition_function(model: ModelInterface, samples: tf.Tensor, at: QueryPoints) -> tf.Tensor:
+        return max_value_entropy_search(model, samples, at)
+
+
+def max_value_entropy_search(model: ModelInterface, samples: tf.Tensor, at: QueryPoints) -> tf.Tensor:
+    r"""
+    Computes the information gain, i.e the change in entropy of p_min if we would evaluate x.
```
joelberkeley-secondmind: what is `p_min`?

trieste/acquisition/function.py
```diff
@@ -130,6 +133,110 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+        :param model: The model over the specified ``dataset``.
+        :return: The MES function.
+        """
+        query_points = self._search_space.sample(self._grid_size)
+        query_points = tf.concat([dataset.query_points,query_points],0)
+        fmean, fvar = model.predict(query_points)
+        fsd = tf.math.sqrt(fvar)
+
+        def probf(x: tf.Tensor) -> tf.Tensor :  # Build empirical CDF
+            unit_normal = tfp.distributions.Normal(tf.cast(0, tf.float64), tf.cast(1, tf.float64))
```
joelberkeley-secondmind: is hard-coding `tf.float64` going to prohibit users who are using `tf.float32` workflows? i guess we could use the user's y dtype with `dataset.observations.dtype`

tests/integration/test_bayesian_optimization.py
```diff
@@ -39,11 +40,19 @@
...
+                                        search_space = Box(
+                                            tf.constant([0.0, 0.0], tf.float64), 
+                                            tf.constant([1.0, 1.0], tf.float64)
+                                        ),
+                                        num_samples = 1, 
+                                        grid_size = 100).using(OBJECTIVE)
+                                    )),
+        (22, TrustRegion()),
+        (17, ThompsonSampling(500, 3)),
+    ])
```
joelberkeley-secondmind: we'll need unit tests for MES to get sufficient coverage, and integration tests take a long time to run so I recommend we don't have any for MES. I appreciate unit tests will be a lot harder to write than this. I can't help with that this week as I'm going on holiday, so I propose we merge this with the integration test only for now, and we can change it to unit tests when I'm back. It's up to you whether you want to be involved in that. Does that work for you?

trieste/acquisition/rule.py
```diff
@@ -27,7 +27,7 @@
 from ..models import ModelInterface
 from ..space import SearchSpace, Box
 from ..type import QueryPoints
-from .function import AcquisitionFunctionBuilder, ExpectedImprovement
+from .function import AcquisitionFunctionBuilder, ExpectedImprovement, MaxValueEntropySearch
```
joelberkeley-secondmind: doesn't look like this is needed any more

trieste/acquisition/function.py
```diff
@@ -130,6 +133,110 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
 
 
+class MaxValueEntropySearch(SingleModelAcquisitionBuilder):
+    """
+    Builder for the max-value entropy search acquisition function (for function minimisation)
+    """
+
+    def __init__(self,search_space: SearchSpace, num_samples: int = 10, grid_size: int = 5000):
+        """
+        :param search_space: The search space over which the acquisiiton function is defined.
```
joelberkeley-secondmind: if we use MES with the `TrustRegion` acquisition rule, the acquisition function is optimized over the trust region search space, which may be different from the global search space, so might the following be clearer? 
```suggestion
        :param search_space: The global search space over which the Bayesian optimization problem is defined.
```

trieste/acquisition/function.py
```diff
@@ -130,6 +133,110 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+    
+    normal = tfp.distributions.Normal(tf.cast(0,tf.float64), tf.cast(1,tf.float64))
+    gamma = (samples - fmean) / fsd
+
+    minus_cdf = 1 - normal.cdf(gamma)
+    minus_cdf = tf.clip_by_value(minus_cdf, 1.0e-8, 1) # clip below to improve numerical stability
+    f_acqu_x = -gamma * normal.prob(gamma) / (2 * minus_cdf) - tf.math.log(minus_cdf)
+    f_acqu_x = tf.math.reduce_mean(f_acqu_x, axis=1)
+        
+    return tf.reshape(f_acqu_x,[-1, 1])
```
joelberkeley-secondmind:
```suggestion
    return tf.math.reduce_mean(f_acqu_x, axis=1, keepdims=True)
```
to allow for batch dimensions?

trieste/acquisition/function.py
joelberkeley-secondmind: yes, you're right about the shape, i missed that.

trieste/acquisition/function.py
joelberkeley-secondmind: was the first approach one of your own?

trieste/acquisition/function.py
henrymoss: At the moment the max_value_entropy_search function doesn't take dataset as an input?

tests/integration/test_bayesian_optimization.py
henrymoss: Yeah, I'll make the small changes you suggested today. Then we can talk about unit tests next week. I'm keen to hear your ideas on appropriate tests.

trieste/acquisition/function.py
henrymoss: Good spot!

henrymoss: I'm keen to do a batch version of this based on some of my PhD work that I am typing up. I'll do this as another PR later. I think it will require a bit more work to support batch optimisation.

tests/integration/test_bayesian_optimization.py
henrymoss: I'll remove MES from the integration tests.

tests/integration/test_bayesian_optimization.py
```diff
@@ -39,11 +40,12 @@
...
-@pytest.mark.parametrize('num_steps, acquisition_rule', [
-    (12, EfficientGlobalOptimization()),
-    (22, TrustRegion()),
-    (17, ThompsonSampling(500, 3)),
-])
+@pytest.mark.parametrize('num_steps, acquisition_rule', 
+    [(12, EfficientGlobalOptimization()),
+        (22, TrustRegion()),
+        (17, ThompsonSampling(500, 3)),
+    ])
```
joelberkeley-secondmind: do you mind reverting the formatting here, and removing the import?

trieste/acquisition/function.py
joelberkeley-secondmind: that's true. we could use `fmean.dtype` there? or in both cases?

joelberkeley-secondmind: was the reason for `tf.float64` for dtype compatibility or specifically to have high precision?

henrymoss:
Thanks so much for all the help. My next PR will be much faster!

trieste/acquisition/function.py
```diff
@@ -130,6 +133,110 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
+
+        def binary_search(val: float) -> float:  # Fit Gumbel quantiles
+            return bisect(lambda x: probf(x) - val, left, right, maxiter=10000, xtol=0.00001)
+        
+        q1, med, q2 = map(binary_search, [0.25, 0.5, 0.75])
+
+        b = (q1 - q2) / (tf.math.log(tf.math.log(4. / 3.)) - tf.math.log(tf.math.log(4.)))
+        a = med + b * tf.math.log(tf.math.log(2.))
+
+        uniform_samples = tf.random.uniform([self._num_samples],dtype=tf.dtypes.float64)
```
joelberkeley-secondmind: just one missed
```suggestion
        uniform_samples = tf.random.uniform([self._num_samples],dtype=fmean.dtype)
```
and we're good to merge

joelberkeley-secondmind (approved):

PR #22, opened 2020-10-17
DO NOT MERGE remove master branch

joelberkeley:
**DO NOT MERGE** without asking. There are external changes that need to happen at the same time as this being merged.

the master branch is superfluous. This PR does the code prep needed to remove it. Instead of building the docs on pushes to master (which, incidentally, is fragile to people pushing to master outside of releases), this PR proposes we build the docs when a semver-compliant tag is pushed. If a tag is pushed by mistake, thus building the docs for the wrong point in the code, we can delete and redeploy the latest correct tag (taking care with associated github releases)

We'll also need to
  - delete the master branch
  - check the github releases are still in place
  - remove the master branch protection rules
  - probably mention in slack #trieste so that people aren't confused

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -14,7 +14,7 @@
 
 on:
   push:
-    branches: master
+    tags: v[0-9]+.[0-9]+.[0-9]+*
 
 jobs:
   docs:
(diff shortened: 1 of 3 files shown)


ChrisMorter (approved):
LGTM

PR #23, opened 2020-10-17
define a static probabilistic model interface

### LINKED ISSUE #35: Add static model interface for inference only
I would like a static (untrainable) model interface for use with the acquisition functionality, so that I can decouple the model training from the acquisition process.

This also improves reasoning power in the acquisition functionality because you know the model won't be modified

joelberkeley:


trieste/models/model_interfaces.py
```diff
@@ -24,24 +24,14 @@
...
-
-    @abstractmethod
-    def update(self, dataset: Dataset) -> None:
-        """
-        Update the model given the specified ``dataset``. Does not train the model.
-
-        :param dataset: The data with which to update the model.
-        """
-        raise NotImplementedError
+class ProbabilisticModel(ABC):
```
joelberkeley-secondmind, line 26: this class is the main change: an untrainable probabilistic model, which is a superclass of the old `ModelInterface` (now called `TrainableProbabilisticModel`)

trieste/models/model_interfaces.py
```diff
@@ -161,7 +164,7 @@ def loss_fn() -> tf.Tensor:
         self._optimize_fn = optimization_fn
 
 
-class GPflowPredictor(ModelInterface, ABC):
+class GPflowPredictor(ProbabilisticModel, ABC):
     """ A trainable wrapper for a GPflow Gaussian process model. """
```
ChrisMorter: Is comment is now wrong, since it's not trainable?

trieste/models/config.py
```diff
@@ -74,19 +74,19 @@ def create_model_interface(self) -> ModelInterface:
         raise NotImplementedError(f"Not supported type {type(self.model)}")
 
 
-ModelSpec = Union[Dict[str, Any], ModelConfig, ModelInterface]
+ModelSpec = Union[Dict[str, Any], ModelConfig, TrainableProbabilisticModel]
 """ Type alias for any type that can be used to fully specify a model. """
 
 
-def create_model_interface(config: ModelSpec) -> ModelInterface:
+def create_model(config: ModelSpec) -> TrainableProbabilisticModel:
```
ChrisMorter, line 81: I realise that this isn't new, but this variable name is a bit confusing since ModelSpec can be a TrainableProbabilisticModel, which isn't really a config at all.

ChrisMorter (approved):
Looks good, just a couple of minor comments

trieste/models/model_interfaces.py
joelberkeley-secondmind: yes, nice catch

trieste/models/config.py
joelberkeley-secondmind: yeah this is a design issue that i want to fix. Really we shouldn't accept a `TrainableProbabilisticModel` as an argument to `create_model`, so I think `config` is OK here

joelberkeley-secondmind: there's another PR in the works that may address this so I'll leave it for now

PR #24, opened 2020-10-18
rename variables `dataset(s)` to `data`

joelberkeley:


joelberkeley-secondmind:
@vpicheny does this make sense to you?

PR #25, opened 2020-10-19
Added MES into introduction notebook

henrymoss:
Hi 

I have added in a demonstration of MES into the introduction notebook.

Let me know if this is what you had in mind.

Henry

### DIFF
--- a/docs/notebooks/introduction.pct.py
+++ b/docs/notebooks/introduction.pct.py
@@ -131,6 +131,8 @@ if result.error is not None: raise result.error
 
 dataset = result.datasets[OBJECTIVE]
 
+ei_optimization_progress = np.minimum.accumulate(dataset.observations)[5:]
+
 # %% [markdown]
 # ## Explore the results
 # We can now get the best point found by the optimizer. Note this isn't necessarily the point that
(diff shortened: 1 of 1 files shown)


docs/notebooks/introduction.pct.py
```diff
@@ -243,3 +245,79 @@
...
+fig.suptitle("Expected Improvement Acquisition Function")
+
+# %% [markdown]
+# To compare the performance of the optimization achieved by these two different acquisition functions, we re-run the above BO loop using MES. 
+
+# %% [markdown]
+# We re-initialize the model and define a new acquisiton rule.
+
+# %%
+gpr = gpflow.models.GPR(astuple(initial_data[OBJECTIVE]), kernel, noise_variance=1e-5)
```
joelberkeley: this will use the kernel from before, which has been trained. Do we not need a new kernel? A function would be good for that
```python
def build_model(data):
    kernel = gpflow.kernels.Matern52(variance=variance, lengthscales=0.2 * np.ones(2,))
    gpr = gpflow.models.GPR(astuple(data), kernel, noise_variance=1e-5)
    set_trainable(gpr.likelihood, False)
    
    return trieste.models.create_model_interface({
        "model": gpr,
        "optimizer": gpflow.optimizers.Scipy(),
        "optimizer_args": {"options": dict(maxiter=100)},
    })
```

docs/notebooks/introduction.pct.py
```diff
@@ -243,3 +245,79 @@
...
+
+model = {OBJECTIVE: trieste.models.create_model_interface(
+    {
+        "model": gpr,
+        "optimizer": gpflow.optimizers.Scipy(),
+        "optimizer_args": {"options": dict(maxiter=100)},
+    }
+)}
+
+builder = trieste.acquisition.MaxValueEntropySearch(search_space,num_samples = 5, grid_size = 5000).using(OBJECTIVE)
```
joelberkeley: the acquisition function builders are immutable, so you can reuse use the one you defined earlier

docs/notebooks/introduction.pct.py
```diff
@@ -243,3 +245,79 @@
...
+# # Try an alternative acquisition function
+
+# %% [markdown]
+# By default, Trieste uses Expected Improvement (EI) as its acqusition function when performing Bayesian optimization. However, many alternative acqusition functions have been developed. One such alternative is Max-value Entropy Search (MES), which approximates the distribution of current estimate of the global minimum and tries to decrease its entropy with each optimization step.
+
+# %% [markdown]
+# We plot these two acquisition functions across our search space. Areas with high acquisition function scores (i.e bright regions) are those rated as promising locations for the next evaluation of our objective function. We see that EI wishes to continue exploring the search space, whereas MES wants to focus resources on evaluating a specific region.
+
+# %%
+mes_acq_function = trieste.acquisition.MaxValueEntropySearch(search_space,num_samples = 5, grid_size = 5000).using(OBJECTIVE).prepare_acquisition_function(initial_data, model)
```
joelberkeley:
```suggestion
mes_acq_function = trieste.acquisition.MaxValueEntropySearch(
    search_space, num_samples=5, grid_size=5000
).using(OBJECTIVE).prepare_acquisition_function(initial_data, model)
```
for formatting

docs/notebooks/introduction.pct.py
```diff
@@ -243,3 +245,81 @@
     num_init=len(dataset.query_points),
     idx_best=arg_min_idx,
 )
+
+# %% [markdown]
+# # Try an alternative acquisition function
```
joelberkeley, line 252:
```suggestion
# ## Try an alternative acquisition function
```
else it will render as a title in the docs

docs/notebooks/introduction.pct.py
```diff
@@ -243,3 +245,81 @@
...
+
+# %% [markdown]
+# By default, Trieste uses Expected Improvement (EI) as its acqusition function when performing Bayesian optimization. However, many alternative acqusition functions have been developed. One such alternative is Max-value Entropy Search (MES), which approximates the distribution of current estimate of the global minimum and tries to decrease its entropy with each optimization step.
```
joelberkeley, line 255:
```suggestion
# By default, Trieste uses Expected Improvement (EI) as its acqusition function when performing Bayesian optimization. However, many alternative acqusition functions have been developed. One such alternative is Max-value Entropy Search (MES), which approximates the distribution of the current estimate of the global minimum and tries to decrease its entropy with each optimization step.
```

joelberkeley, line 255: i didn't realise you can put this all on one line, and it appears nicely in jupyter and the docs. that's much easier than manually splitting the lines. I'll do that in future

docs/notebooks/introduction.pct.py
```diff
@@ -243,3 +245,81 @@
...
+
+# %% [markdown]
+# We plot these two acquisition functions across our search space. Areas with high acquisition function scores (i.e bright regions) are those rated as promising locations for the next evaluation of our objective function. We see that EI wishes to continue exploring the search space, whereas MES wants to focus resources on evaluating a specific region.
```
joelberkeley, line 258:
```suggestion
# We plot these two acquisition functions across our search space. Areas with high acquisition function scores (i.e. bright regions) are those rated as promising locations for the next evaluation of our objective function. We see that EI wishes to continue exploring the search space, whereas MES wants to focus resources on evaluating a specific region.
```

docs/notebooks/introduction.pct.py
```diff
@@ -243,3 +245,81 @@
...
+
+# %%
+mes_acq_function = trieste.acquisition.MaxValueEntropySearch(
+    search_space, num_samples=5, grid_size=5000
+).using(OBJECTIVE).prepare_acquisition_function(initial_data, model)
+ei_acq_function = trieste.acquisition.ExpectedImprovement().using(OBJECTIVE).prepare_acquisition_function(initial_data, model)
```
joelberkeley: can you split this line, else it runs over the edge of the docs page?

joelberkeley:
this is a really nice addition to the intro notebook, thanks. The graphs are particularly clear.

What's your opinion: currently we don't demonstrate how to pass EI to the optimizer, but mention it's the default. Do you think this notebook would be clearer if we manually specified the EI acquisition function for the first BO loop, and mentioned as an aside that it's actually the default and you don't need to specify it? (btw I don't think this is something to do in this PR)

docs/notebooks/introduction.pct.py
```diff
@@ -131,6 +131,8 @@
 
 dataset = result.datasets[OBJECTIVE]
 
+ei_optimization_progress = np.minimum.accumulate(dataset.observations)[5:]
```
joelberkeley: would this be clearer further down, closer to where we use it? you may need to rename `dataset` further down if you do move it

henrymoss:
I've made all your recommended changes. 

I think it is worth manually specifying the EI acquisition function. Remember, that I was initially confused about if EGO was just EI or a wrapper for general BO acquisition functions.

docs/notebooks/introduction.pct.py
```diff
@@ -191,7 +195,7 @@
 # intervals of its predictive distribution.
 
 # %%
-fig = plot_gp_plotly(gpr, mins, maxs, grid_density=30)
+fig = plot_gp_plotly(model['OBJECTIVE'].model, mins, maxs, grid_density=30)
```
joelberkeley:
```suggestion
fig = plot_gp_plotly(model[OBJECTIVE].model, mins, maxs, grid_density=30)
```
someone could rename `OBJECTIVE`, or change the string it points to, and miss this

docs/notebooks/introduction.pct.py
```diff
@@ -211,7 +215,7 @@
 # evolved over iterations
 
 # %%
-print_summary(gpr)
+print_summary(model['OBJECTIVE'].model)
```
joelberkeley:
```suggestion
print_summary(model[OBJECTIVE].model)
```

joelberkeley (approved):

PR #31, opened 2020-11-04
add tests for expected improvement

joelberkeley:


tests/unit/acquisition/test_function.py
```diff
@@ -85,14 +92,41 @@ def test_expected_improvement_builder_builds_expected_improvement(
...
-        n = tfp.distributions.Normal(0, 1)
-        return - x * n.cdf(-x) + n.prob(-x)
+@random_seed()
+@pytest.mark.parametrize('best, rtol, atol', [
+    (tf.constant([50.0]), 0.01, 1e-3),
+    (BRANIN_GLOBAL_MINIMUM[None], 0.01, 1e-3),
+    (BRANIN_GLOBAL_MINIMUM[None] * 1.01, 0.01, 1e-3)
+])
+# todo this fails for variance scale 100. is that ok?
+@pytest.mark.parametrize('variance_scale', tf.constant([0.1, 1.0, 10.0], dtype=tf.float64))
```
joelberkeley: i need to make the rtol 0.03 for the last two tests if I add variance scale 100. Is that a problem?

tests/unit/acquisition/test_function.py
```diff
@@ -85,14 +90,37 @@ def test_expected_improvement_builder_builds_expected_improvement(
...
+])
+@pytest.mark.parametrize('variance_scale', tf.constant([0.1, 1.0, 10.0], dtype=tf.float64))
+def test_expected_improvement(
+    variance_scale: tf.Tensor, best: tf.Tensor, rtol: float, atol: float
+) -> None:
+    best = tf.cast(best, dtype=tf.float64)
+
+    x_range = tf.linspace(0.0, 1.0, 11)
+    x_range = tf.cast(x_range, dtype=tf.float64)
+    xs = tf.reshape(tf.stack(tf.meshgrid(x_range, x_range, indexing='ij'), axis=-1), (-1, 2))
```
joelberkeley, line 111: make a separate function to do this?

vpicheny: you have a create grid function in plotting utilities that does just this

tests/unit/acquisition/test_function.py
```diff
@@ -85,14 +90,37 @@ def test_expected_improvement_builder_builds_expected_improvement(
...
+
+    class _Model(GaussianMarginal):
+        kernel = tfp.math.psd_kernels.MaternFiveHalves(variance_scale, length_scale=0.25).apply
+
+        def predict(self, query_points: TensorType) -> Tuple[TensorType, TensorType]:
+            return branin(query_points), self.kernel(query_points, query_points)[:, None]
+
+    num_samples_per_point = 1_000_000
```
vpicheny: this seems very very high.1_000 should be sufficient (depends on the tolerance of course)

vpicheny: doesn't hurt if the test is fast though

tests/util/model.py
```diff
@@ -33,13 +33,13 @@ def optimize(self) -> None:
         pass
 
 
-class StaticWithUnitVariance(StaticModelInterface, ABC):
+class GaussianMarginal(StaticModelInterface, ABC):
     def sample(self, query_points: QueryPoints, num_samples: int) -> ObserverEvaluations:
         mean, var = self.predict(query_points)
         return tfp.distributions.Normal(mean, var).sample(num_samples)
```
vpicheny, line 39: why do we use tfp instead of just predict_f_samples?

trieste/acquisition/function.py
```diff
@@ -124,10 +124,13 @@ def expected_improvement(model: ModelInterface, eta: tf.Tensor, at: QueryPoints)
...
 
     :param model: The model of the objective function.
-    :param eta: The "best" observation.
+    :param eta: The "best" observation, with shape [1].
     :param at: The points for which to calculate the expected improvement.
     :return: The expected improvement at ``at``.
+    :raise ValueError (or InvalidArgumentError): If eta has the wrong shape.
     """
+    tf.debugging.assert_shapes([(eta, [1])], "eta has an invalid shape")
+
```
vpicheny: can we instead check that eta is either shape [1] or `at.shape[-1]`? I actually have a case with Manchester for that

tests/unit/acquisition/test_function.py
vpicheny: I don't understand what variance scale 100 means

tests/unit/acquisition/test_function.py
```diff
@@ -85,14 +90,37 @@ def test_expected_improvement_builder_builds_expected_improvement(
...
 
-def test_expected_improvement() -> None:
-    def _ei(x: tf.Tensor) -> tf.Tensor:
-        n = tfp.distributions.Normal(0, 1)
-        return - x * n.cdf(-x) + n.prob(-x)
+@random_seed()
+@pytest.mark.parametrize('best, rtol, atol', [
+    (tf.constant([50.0]), 0.01, 1e-3),
+    (BRANIN_GLOBAL_MINIMUM[None], 0.01, 1e-3),
+    (BRANIN_GLOBAL_MINIMUM[None] * 1.01, 0.01, 1e-3)
```
joelberkeley: if we use the same tolerances for all of them, maybe don't parametrize over them?

joelberkeley: is 1e-3 a bit high? depends on the problem scale?

tests/unit/acquisition/test_function.py
joelberkeley: it's the `amplitude` argument to the Matern52 kernel in the model, so determines the size of the model variance

tests/unit/acquisition/test_function.py
joelberkeley: yeah. I was thinking of porting that over. An alternative is to sample from a `Box`.

joelberkeley: but I'd have to write tests for the former

tests/unit/acquisition/test_function.py
joelberkeley: 1000 wasn't passing. is that a problem?

the test is getting fast enough at 1,000,000 unless you run it in pycharm (which for some reason is much slower)

vpicheny: Maybe the problem comes from the relative tolerance, the relative error can be quite large for values very close to zero. I would be in favour of only using the absolute tolerance.

joelberkeley: it's the high variance tests that need lots of samples

joelberkeley: haha, the 0.1 variance test works with 4 samples

trieste/acquisition/function.py
joelberkeley: this shape could actually be `['_', 1]`, but i thought i'd just keep things unambiguous

tests/unit/acquisition/test_function.py
```diff
@@ -85,14 +90,39 @@ def test_expected_improvement_builder_builds_expected_improvement(
...
-        n = tfp.distributions.Normal(0, 1)
-        return - x * n.cdf(-x) + n.prob(-x)
+@random_seed()
+@pytest.mark.parametrize('best, rtol, atol', [
+    (tf.constant([50.0]), 0.01, 1e-3),
+    (BRANIN_GLOBAL_MINIMUM[None], 0.02, 1e-3),
+    (BRANIN_GLOBAL_MINIMUM[None] * 1.01, 0.02, 1e-3)
+])
+@pytest.mark.parametrize('variance_scale, num_samples_per_point', [
+    (0.1, 1000), (1.0, 50_000), (10.0, 1_000_000), (100.0, 1_000_000)
```
ChrisMorter: More of an observation than a specific suggestion:

Some of these parameters only really serve to make sure that the approximation method used by the test is sufficiently accurate to be within the tolerance of the true value. So they're not really specifying the test case, they're provided some extra implementation detail needed to make the test work. 

The test may be be more readable if this wasn't the case, but I can't think of a way to do this which wouldn't increase the runtime a lot, or introduce another (probably worse) form of complexity.

ChrisMorter (approved):
Looks good to me

tests/unit/acquisition/test_function.py
vpicheny: I'd choose: amplitude=100, rtol=0, atol=1e-2. atol can be increased to 1e-1 if the test doesn't pass

tests/unit/acquisition/test_function.py
vpicheny: try 1_000 with 100 variance, increase to 10_000 if needed

tests/unit/acquisition/test_function.py
vpicheny: see proposition above

vpicheny: see above

tests/unit/acquisition/test_function.py
```diff
@@ -85,14 +90,39 @@ def test_expected_improvement_builder_builds_expected_improvement(
...
-def test_expected_improvement() -> None:
-    def _ei(x: tf.Tensor) -> tf.Tensor:
-        n = tfp.distributions.Normal(0, 1)
-        return - x * n.cdf(-x) + n.prob(-x)
+@random_seed()
+@pytest.mark.parametrize('best, rtol, atol', [
+    (tf.constant([50.0]), 0.01, 1e-3),
+    (BRANIN_GLOBAL_MINIMUM[None], 0.02, 1e-3),
+    (BRANIN_GLOBAL_MINIMUM[None] * 1.01, 0.02, 1e-3)
+])
```
joelberkeley, line 98: move `atol` to `variance_scale` section

tests/unit/acquisition/test_function.py
joelberkeley-secondmind: yes i know what you mean. i use `parametrize` for all sorts of things, even parametrizing over which function to test, so I think this is OK, though perhaps not common

tests/unit/acquisition/test_function.py
joelberkeley: it fails with 1e-2 and 1e-1 if rtol=0

tests/unit/acquisition/test_function.py
joelberkeley: i can get it down to 100_000. The values you ask for generally don't pass. Maybe worth investigating if this is a problem in the test/implementation

tests/unit/acquisition/test_function.py
joelberkeley: created issue #42

vpicheny (approved):

PR #43, opened 2020-11-09
Reintroduce MES

henrymoss:
Fixed Gumbel sampler subroutine for max-value entropy search.
 
We now fit the Gumbel (minima) distribution instead of the Gumbel (maxima) distribution, which is what we require for function minimisation.

I have also added some comments explaining where the equations come from to avoid future confusions on this issue.


docs/notebooks/introduction.pct.py
```diff
@@ -270,16 +270,16 @@ def build_model(data):
...
-result_mes: OptimizationResult = bo.optimize(15, initial_data, model, acq_rule)
+result_mes: OptimizationResult = bo.optimize(20, initial_data, model, acq_rule)
 
 if result_mes.error is not None: raise result_mes.error
 
 dataset_mes = result_mes.datasets[OBJECTIVE]
-mes_optimization_progress = np.minimum.accumulate(dataset_mes.observations)[5:]
-ei_optimization_progress = np.minimum.accumulate(dataset.observations)[5:]
+mes_optimization_progress = np.minimum.accumulate(dataset_mes.observations)
+ei_optimization_progress = np.minimum.accumulate(dataset.observations)
```
joelberkeley: does this mean we're also plotting the first 5?

trieste/acquisition/function.py
```diff
@@ -163,9 +163,20 @@ def prepare_acquisition_function(
         self, dataset: Dataset, model: ModelInterface
     ) -> AcquisitionFunction:
         """
-        Need to sample possible min-values from our posterior. 
+        Need to sample possible min-values y* from our posterior. 
         To do this we implement a Gumbel sampler.
-        We approximate Pr(y*^hat<y) by Gumbel(a,b) then sample from Gumbel.
+        We approximate Pr(y*<y) by Gumbel(a,b) then sample from Gumbel.
+
+        The Gumbel distribution for minimas has a cumulative density function
```
```suggestion
        The Gumbel distribution for minima has a cumulative density function
```

trieste/acquisition/function.py
```diff
@@ -163,9 +163,20 @@ def prepare_acquisition_function(
...
+        of f(y)= 1 - exp(-1 * exp((y - a) / b)), i.e. the q^th quantile is given by
```
```suggestion
        of :math:`f(y) = 1 - exp(-1 * exp((y - a) / b))`, i.e. the q\ :sup:`th` quantile is given by
```
and could you add the `:math:` context elsewhere please?

joelberkeley: also, was the `-1 * exp(...)` intentional over `- exp(...)`?

trieste/acquisition/function.py
```diff
@@ -176,24 +187,27 @@ def prepare_acquisition_function(
         fmean, fvar = model.predict(query_points)
         fsd = tf.math.sqrt(fvar)
 
-        def probf(x: tf.Tensor) -> tf.Tensor :  # Build empirical CDF
+        def probf(y: tf.Tensor) -> tf.Tensor :  # Build empirical CDF for Pr(y*^hat<y)
             unit_normal = tfp.distributions.Normal(tf.cast(0, fmean.dtype), tf.cast(1, fmean.dtype))
-            log_cdf = unit_normal.log_cdf(- (x - fmean) / fsd)
-            return tf.exp(tf.reduce_sum(log_cdf, axis=0))
+            log_cdf = unit_normal.log_cdf(- (y - fmean) / fsd)
```
joelberkeley: we appear to be mixing f and y. Is that right?

trieste/acquisition/function.py
joelberkeley: i'm curious about docs for research code. In other contexts, I wouldn't normally include the explanation of the implementation, as typically the only thing the user needs to know is what values they get back for any given input, but here, part of the importance of the function is how it works, so that researchers can compare algorithms

docs/notebooks/introduction.pct.py
henrymoss: I decided it was helpful to see the starting point of each algorithm, i.e what happens before we switch to BO. It helps get an idea about useful BO is. What do you think?

trieste/acquisition/function.py
henrymoss: I feel like it is useful? Let me know if you want me to get rid of it.

joelberkeley: no no need to, it was more of a musing than a suggestion

joelberkeley:
see https://www.itl.nist.gov/div898/handbook/eda/section3/eda366g.htm for the min-value gumbel form

trieste/acquisition/function.py
joelberkeley: well, at least in name

trieste/acquisition/function.py
```diff
@@ -176,24 +187,27 @@ def prepare_acquisition_function(
...
+            return 1 - tf.exp(tf.reduce_sum(log_cdf, axis=0))
```
joelberkeley, line 193: curious ... any reason you didn't use
```
tfp.distributions.Normal(fmean, fsd).cdf(y)
``

Comment (joelberkeley) [inline trieste/acquisition/function.py:218]: i'm finding it difficult to wrap my head around the signs here. Could you explain them?

**Comment by joelberkeley** [comment]
I've just run the integration test with these changes and it appears the changes actually make BO take longer by one step at 26 vs 25 steps (at least for that setup). Why do you think this is, and do you think it's a concern?

**Comment by joelberkeley** [comment]
We also spoke about renaming it to `MinValueEntropySearch`. Do you want to do that here?

File: trieste/acquisition/function.py (hunk 3 above)
Comment (joelberkeley) [reply trieste/acquisition/function.py]: @awav why resolved?

File: trieste/acquisition/function.py [hunk 6]
```diff
@@ -176,24 +187,27 @@ def prepare_acquisition_function(
         fmean, fvar = model.predict(query_points)
         fsd = tf.math.sqrt(fvar)
 
-        def probf(x: tf.Tensor) -> tf.Tensor :  # Build empirical CDF
+        def probf(y: tf.Tensor) -> tf.Tensor :  # Build empirical CDF for Pr(y*^hat<y)
```
awav: @henrymoss 

it would be better to move some docs into this function explaining what it is doing (comment will not be a part of documentation). Also `Pr(y*^hat<y)` looks quite cumbersome, I'd go with ":math:``Pr(y' < y)``", where y' is a y*^hat.

awav (approved):
LGTM. Please, address other comments before merging. Thanks!

trieste/acquisition/function.py
```diff
@@ -137,9 +137,9 @@ def expected_improvement(model: ProbabilisticModel, eta: tf.Tensor, at: QueryPoi
     return (eta - mean) * normal.cdf(eta) + variance * normal.prob(eta)
 
 
-class MaxValueEntropySearch(SingleModelAcquisitionBuilder):
+class MinValueEntropySearch(SingleModelAcquisitionBuilder):
     """
-    Builder for the max-value entropy search acquisition function (for function minimisation)
+    Builder for the min-value entropy search acquisition function (for function minimisation)
```
joelberkeley-secondmind: the only thing remaining afaict is that it would be good to mention this is Max-value Entropy Search modified for objective minimization (unless calling it min-value entropy search is standard in the literature)

henrymoss: Yeah that a good idea. I've changed it.

joelberkeley (approved):

joelberkeley-secondmind:
FYI we removed MES for the 0.3.0 release since it was untested and had this bug in it. We fully intend to reintroduce it

henrymoss:
MES now works. I have added some very simple unit tests. I will build a notebook in a separate PR

trieste/acquisition/function.py
```diff
@@ -157,6 +158,133 @@ def expected_improvement(model: ProbabilisticModel, eta: TensorType, at: TensorT
...
+
+        if num_samples <= 0:
+            raise ValueError(f"num_samples must be positive, got {num_samples}")
+        self._num_samples = num_samples
+
+        if grid_size <= 0:
+            raise ValueError(f"grid_size must be positive, got {grid_size}")
+        self._grid_size = grid_size
+
+        self.gumbel_samples = None
```
joelberkeley: is this for testing or do users need access to it?

henrymoss: Was primarily just for testing.

But it could be handy to have access to this in the future. Potentially as part of a stopping criterion for the BO loop.

joelberkeley: ok. I reckon that as a feature on the public API, this would be scope creep, so can we leave it for future work?

henrymoss: Sure. Although this will mean I need to remove some of the tests

trieste/acquisition/function.py
```diff
@@ -157,6 +158,133 @@ def expected_improvement(model: ProbabilisticModel, eta: TensorType, at: TensorT
...
+    See the following for details:
+
+    ::
+
+        @article{wang2017max,
+          title={Max-value entropy search for efficient Bayesian optimization},
+          author={Wang, Zi and Jegelka, Stefanie},
+          journal={arXiv preprint arXiv:1703.01968},
+          year={2017}
+        }
```
joelberkeley: since we're here, can we convert this to the sphinxcontrib-bibtex format? That would just mean moving the reference to docs/refs.bib and using
```
See :cite:`wang2017max` for details.
```
here

tests/unit/acquisition/test_function.py
```diff
@@ -171,6 +173,37 @@ def test_expected_improvement(
...
+    data = Dataset(tf.zeros([0, 1]), tf.ones([0, 1]))
+    search_space = Box([0, 0], [1, 1])
+    with pytest.raises(ValueError):
+        MinValueEntropySearch(search_space, num_samples=-5).prepare_acquisition_function(
+            data, QuadraticMeanAndRBFKernel()
+        )
+    with pytest.raises(ValueError):
+        MinValueEntropySearch(search_space, grid_size=-5).prepare_acquisition_function(
+            data, QuadraticMeanAndRBFKernel()
+        )
```
joelberkeley: do you need the calls to `prepare_acquisition_function` here?

trieste/acquisition/function.py
joelberkeley: that's ok by me. Do you not reckon?

henrymoss: Okay doke. Will remove

tests/unit/acquisition/test_function.py
henrymoss: Don't think so. I will check and remove.

henrymoss: Do you have any ideas for more tests? It seems quite sparse

joelberkeley: Your idea about testing the maximum is the same as LCB for specific beta would be really good at least until we have a test for the full function form. You could do that with a really fine grid search (no need for a fancy optimizer)

You could also add more shape checking for `min_value_entropy_search` (e.g. check `at`, predictive mean and `samples` are compatible) and test that if you felt like it

tests/unit/acquisition/test_function.py
```diff
@@ -171,6 +173,37 @@ def test_expected_improvement(
     npt.assert_allclose(ei, ei_approx, rtol=rtol, atol=atol)
 
 
+def test_min_value_entropy_search_builder_raises_for_empty_data() -> None:
+    data = Dataset(tf.zeros([0, 1]), tf.ones([0, 1]))
+    search_space = Box([0, 0], [1, 1])
+    with pytest.raises(ValueError):
+        MinValueEntropySearch(search_space).prepare_acquisition_function(
+            data, QuadraticMeanAndRBFKernel()
+        )
```
joelberkeley: if you do it like this, you can be sure the `ValueError` is raised by the call to `prepare_acquisition_function` not by the call to `__init__`
```suggestion
    builder = MinValueEntropySearch(search_space)
    with pytest.raises(ValueError):
        builder.prepare_acquisition_function(
            data, QuadraticMeanAndRBFKernel()
        )
```

tests/unit/acquisition/test_function.py
henrymoss: To compare with LCB i need a way to access the Gumbel samples. So I guess that is out for now.  Will do some extra shape checking though.

henrymoss:
I've made the suggested changes and added some more tests. I have a PI vs MES test now.

henrymoss:
Still can't get the notebook tests to pass. Is this an issue with my code?

tests/unit/acquisition/test_function.py
```diff
@@ -171,6 +174,63 @@ def test_expected_improvement(
...
+
+
+def test_min_value_entropy_search_raises_for_empty_gumbel_sample() -> None:
+    with pytest.raises(ValueError):
+        min_value_entropy_search(QuadraticMeanAndRBFKernel(), tf.constant([]), tf.constant([[]]))
+
+
+def test_min_value_entropy_search_raises_for_gumbel_sample_with_additional_dimensions() -> None:
+    with pytest.raises(ValueError):
+        min_value_entropy_search(QuadraticMeanAndRBFKernel(), tf.constant([[]]), tf.constant([[]]))
```
joelberkeley: this is a perfect opportunity for `parametrize`
```python
@pytest.mark.parametrize("samples", [tf.constant([]), tf.constant([[]])])
def test_min_value_entropy_search_raises_for_gumbel_samples_with_invalid_shape(samples: TensorType) -> None:
    with pytest.raises(ValueError):
        min_value_entropy_search(QuadraticMeanAndRBFKernel(), samples, tf.constant([[]]))
```

joelberkeley:
> Still can't get the notebook tests to pass. Is this an issue with my code?

No. There's a cholesky error in the batch section of inequality_constraints.pct.py. This sometimes happens and without any obvious reason. You can fix it by adding some `gpflow.kernels.White(1e-12)` to the model kernel

trieste/acquisition/function.py
```diff
@@ -157,6 +158,129 @@ def expected_improvement(model: ProbabilisticModel, eta: TensorType, at: TensorT
...
+    :param model: The model of the objective function.
+    :param samples: Samples from p_min
+    :param at: The points for which to calculate the expected improvement.
+    :return: The entropy reduction provided by an evaluation of ``at``.
+    """
+
+    if len(samples) == 0:
+        raise ValueError("Need at least one max-value sample.")
+    if len(samples.shape) != 1:
+        raise ValueError("Max-value samples must be 1-dimensional.")
```
joelberkeley: we usually use tf.debugging for shape checks
```suggestion
    tf.debugging.assert_rank(samples, 1)
    tf.debugging.assert_positive(len(samples))
```
and `len(samples)` will `TypeError` for scalar samples, so perhaps put that check second?

joelberkeley (approved):

henrymoss:
Cheers Joel. I have made all the changes except the fixing of that notebook. I had a quick look but couldn't easily access the kernel of the model.

tests/unit/acquisition/test_function.py
```diff
@@ -171,6 +174,61 @@ def test_expected_improvement(
...
+    xs = tf.reshape(tf.stack(tf.meshgrid(x_range, x_range, indexing="ij"), axis=-1), (-1, 2))
+
+    gumbel_sample = tf.constant([1.0], dtype=tf.float64)
+    mes_evals = min_value_entropy_search(model, gumbel_sample, xs)
+
+    mean, variance = model.predict(xs)
+    gamma = (tf.cast(gumbel_sample, dtype=mean.dtype) - mean) / tf.sqrt(variance)
+    norm = tfp.distributions.Normal(tf.cast(0, dtype=mean.dtype), tf.cast(1, dtype=mean.dtype))
+    pi_evals = norm.cdf(gamma)
+    npt.assert_array_equal(tf.argmax(mes_evals), tf.argmax(pi_evals))
```
joelberkeley, line 229: why do you think this passes exactly? I would have expected it to pass approximately

PR #46, opened 2020-11-10
VGP update

vpicheny:
this PR replaces the current VGP update that re-initializes q_mu and q_sqrt to 0 and 1 resp. to values based on the current model posterior

### DIFF
--- a/docs/notebooks/failure_ego.pct.py
+++ b/docs/notebooks/failure_ego.pct.py
@@ -33,7 +33,6 @@ from util.plotting_plotly import (
 )
 from util.plotting import plot_gp_2d, plot_function_2d, plot_bo_points
 
-
 # %%
 gpflow.config.set_default_jitter(1e-5)
 gpflow.config.set_default_float(np.float64)
--- a/docs/notebooks/failure_ego.pct.py
+++ b/docs/notebooks/failure_ego.pct.py
@@ -33,6 +33,7 @@ from util.plotting_plotly import (
 )
 from util.plotting import plot_gp_2d, plot_function_2d, plot_bo_points
 
+
 # %%
 gpflow.config.set_default_jitter(1e-5)
 gpflow.config.set_default_float(np.float64)
--- a/tests/unit/models/test_model_interfaces.py
+++ b/tests/unit/models/test_model_interfaces.py
@@ -180,6 +180,27 @@ def test_vgp_update_updates_num_data() -> None:
 
 
 @random_seed(1357)
+def test_vgp_update_updates_vi() -> None:
+    x_observed = tf.constant(np.arange(100).reshape((-1, 1)), dtype=gpflow.default_float())
+    y_observed = _3x_plus_gaussian_noise(x_observed)
+    model = VariationalGaussianProcess(_vgp(x_observed, y_observed))
+
+    gpflow.optimizers.Scipy().minimize(
+        model.loss, model.trainable_variables,
+    )
+    old_q_mu = model.q_mu.value()
+    old_q_sqrt = model.q_sqrt.value()
+
+    model.update(Dataset(x_observed, y_observed))
+
+    new_q_mu = model.q_mu.value()
+    new_q_sqrt = model.q_sqrt.value()
+
+    npt.assert_allclose(old_q_mu, new_q_mu)
+    npt.assert_allclose(old_q_sqrt, new_q_sqrt)
+
+
+@random_seed(1357)
 def test_gaussian_process_regression_default_optimize(gpr_interface_factory) -> None:
     model = gpr_interface_factory(*_mock_data())
     loss = model.loss()
--- a/tests/unit/models/test_model_interfaces.py
+++ b/tests/unit/models/test_model_interfaces.py
@@ -190,8 +190,8 @@ def test_vgp_update_updates_vi() -> None:
     )
     old_q_mu = model.model.q_mu.numpy()
     old_q_sqrt = model.model.q_sqrt.numpy()
-
-    model.update(Dataset(x_observed, y_observed))
+    data = Dataset(x_observed, y_observed)
+    model.update(data)
 
     new_q_mu = model.model.q_mu.numpy()
     new_q_sqrt = model.model.q_sqrt.numpy()
--- a/tests/unit/models/test_model_interfaces.py
+++ b/tests/unit/models/test_model_interfaces.py
@@ -124,6 +124,15 @@ def _vgp(x: tf.Tensor, y: tf.Tensor) -> VGP:
     return m
 
 
+def _vgp2(x: tf.Tensor, y: tf.Tensor) -> VGP:
+    likelihood = gpflow.likelihoods.Gaussian()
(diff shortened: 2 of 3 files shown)


trieste/models/model_interfaces.py
```diff
@@ -274,13 +274,18 @@ def update(self, dataset: Dataset):
         assert dataset.observations.shape[-1] == y.shape[-1]
         data = (dataset.query_points, dataset.observations)
         num_data = data[0].shape[0]
-        num_latent_gps = model.num_latent_gps
+
+        new_q_mu, new_q_var = self.model.predict_f(dataset.query_points, full_cov=True)
+
+        new_q_sqrt = tf.linalg.cholesky(
+            new_q_var
+            + gpflow.config.default_jitter() * (tf.eye(num_data, dtype=new_q_var.dtype)[None])
```
vdutor: FYI: gpflow has `gpflow.utilities.ops.eye(num, value=gpflow.config.default_jitter())` that you could use for this.

trieste/models/model_interfaces.py
```diff
@@ -274,13 +274,18 @@ def update(self, dataset: Dataset):
         assert dataset.observations.shape[-1] == y.shape[-1]
         data = (dataset.query_points, dataset.observations)
         num_data = data[0].shape[0]
-        num_latent_gps = model.num_latent_gps
+
+        new_q_mu, new_q_var = self.model.predict_f(dataset.query_points, full_cov=True)
```
vdutor: You probably want to be careful and assert that `self.model.q_sqrt.shape.ndims == 3` (i.e. `q_diag=False`) as GPflow also support only learning the marginals of `u` which results in a 2-dim `q_sqrt` (`self.model.q_sqrt.shape.ndims == 2`).

vdutor: Can you also assert that `white == False` for this model (not sure this applicable for VGP's though).

vdutor (approved):
LGTM - left you a few comments to make the code a bit less brittle.

trieste/models/model_interfaces.py
vpicheny: I think that VGP does not have a whiten option

vpicheny: sorry, actually VGP always uses white=True for conditionals

st--: When whitening (as @vpicheny pointed out hard-coded for VGP), predict_f and q_mu/q_sqrt live in different spaces.

u = f(X) still lives in the same space, but with whitening we have u = L v, where L = cholesky(Kuu), and q(v) = N(q_mu, q_sqrt q_sqrt^T) (p(v) = N(0, I)).

You get the correct q(u) = N(f_mean, f_cov) from f_mean, f_cov = predict_f(X, full_cov=True). They are related to q(v) by f_mean = L q_mu, f_cov = L q_sqrt q_sqrt^T L^T.

To update q(v), you need to backtransform: q_mu = L^{-1} f_mean and q_sqrt = cholesky(S) where S = L^{-1} f_cov L^{-T}.

So you need three extra triangular solves and an additional cholesky:
```python
f_mean, f_cov = model.predict_f(X_new, full_cov=True)
K = model.kernel(X_new)
L = tf.linalg.cholesky(K)
new_q_mu = tf.linalg.triangular_solve(L, f_mean)
Linv_f_cov = tf.linalg.triangular_solve(L, f_cov)
S = tf.linalg.triangular_solve(L, tf.linalg.matrix_transpose(Linv_f_cov))
new_q_sqrt = tf.linalg.cholesky(S)
```
(not tested, might have to account for some extra dimensions here and there)

Alternatively, you could just use an SVGP with Z=X and whiten=False.

vdutor: Yes, what Ti is saying is exactly right! I've written some code before to do this, so copy-pasting it here for your help:
```python
# predict with old model
 f_mu, f_cov = self.model.predict_f(X_new, full_cov=True) # [N, L], [L, N, N]

Knn = model.kernel(X_new, full_cov=True)  # [N, N]
Lnn = tf.linalg.cholesky(Knn +  tf.eye(num_data, dtype=Knn.dtype) * jitter)  # [N, N]
q_mu_v = tf.linalg.triangular_solve(Lnn, f_mu)  # [N, L]
tmp = tf.linalg.triangular_solve(Lnn[None], f_var)  # [L, N, N], L^{-1} f_cov
S_v = tf.linalg.triangular_solve(Lnn[None], tf.linalg.matrix_transpose(tmp))  # [L, N, N]
new_q_sqrt = tf.linalg.cholesky(S_v + tf.eye(num_data, dtype=Knn.dtype) * jitter)  # [L, N, N]
```
Equations:
```
u = L * v, with v = N(0, 1) and K(X, X) = L L^T
q_mu_v = L^{-1} q_mu
q_sqrt_v = L^{-1} q_cov L^{-T}, q_cov = q_sqrt @ q_sqrt
```

st--: Great, you've sorted the jitter/dimensions aspect of it (just need to replace `f_var` -> `f_cov`). :)

st--:
How about adding a test that, if the data points actually don't change, the computed new q_mu and q_sqrt are actually equivalent to the previous values ?

trieste/models/model_interfaces.py
vpicheny: updated with your suggested code - thx a bunch

docs/notebooks/failure_ego.pct.py
```diff
@@ -33,6 +33,7 @@
 )
 from util.plotting import plot_gp_2d, plot_function_2d, plot_bo_points
 
+
```
joelberkeley-secondmind: why victor?!

vpicheny: lines are added automatically when I run the notebook, this is indeed annoying

joelberkeley-secondmind: maybe we should just go with it

joelberkeley-secondmind: or are they going to keep adding them indefinitely?

tests/unit/models/test_model_interfaces.py
```diff
@@ -124,6 +124,15 @@ def _vgp(x: tf.Tensor, y: tf.Tensor) -> VGP:
     return m
 
 
+def _vgp2(x: tf.Tensor, y: tf.Tensor) -> VGP:
```
vpicheny: @joelberkeley-secondmind this new model is necessary to test the update because the one with linear kernel has a very bad conditioning.

tests/unit/models/test_model_interfaces.py
```diff
@@ -144,6 +153,10 @@ def _3x_plus_10(x: tf.Tensor) -> tf.Tensor:
     return 3.0 * x + 10
 
 
+def _sin_10_x(x: tf.Tensor) -> tf.Tensor:
```
vpicheny: @joelberkeley another test function that provides data easier to handle by GPs

joelberkeley-secondmind: you ain't fooling me: that's not sin(10x)

PR #47, opened 2020-11-11
Allow models to work with Scipy and TensorFlow optimizers

awav:


trieste/models/model_interfaces.py
```diff
@@ -15,18 +15,28 @@
...
 import numpy as np
 import tensorflow as tf
 
+from .optimizer import Optimizer
 from .. import utils
 from ..datasets import Dataset
 from ..type import ObserverEvaluations, QueryPoints, TensorType
 
 
+@dataclass
```
joelberkeley-secondmind: why `@dataclass`?

trieste/models/model_interfaces.py
```diff
@@ -15,18 +15,28 @@
...
 class ModelInterface(ABC):
     """ A trainable probabilistic model. """
 
+    @abstractmethod
+    def optimize(self, dataset: Dataset) -> None:
+        """
+        Optimize the model.
+        """
+        raise NotImplementedError
```
joelberkeley-secondmind: that `optimize` is called immediately after `update` and has the same signature is even more reason to merge the two. Users can always define separate methods in their own implementations, indeed we could do that for the GPflow wrappers below, but wouldn't have to

trieste/models/model_interfaces.py
```diff
@@ -60,110 +70,19 @@ def sample(self, query_points: QueryPoints, num_samples: int) -> ObserverEvaluat
...
+    """ A trainable wrapper for a GPflow Gaussian process model. """
 
-    @abstractmethod
-    def loss(self) -> tf.Tensor:
-        """ The training loss (to be minimized) on this model. """
-        raise NotImplementedError
+    def __init__(self, optimizer: Optional[Optimizer] = None):
+        if optimizer is None:
+            optimizer = Optimizer(gpflow.optimizers.Scipy())
+        self._optimizer = optimizer
```
joelberkeley-secondmind, line 80: can this be on a module-private subclass of `GPflowPredictor`? It makes it easier to define implementations of this interface

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+"""
+
+
+@dataclass
+class Optimizer:
+    optimizer: Union[gpflow.optimizers.Scipy, tf.optimizers.Optimizer]
+    minimize_args: Dict[str, Any] = field(default_factory=lambda: {})
+    compile: bool = False
+
+    def create_loss(self, model: tf.Module, dataset: Dataset) -> Callable:
```
joelberkeley-secondmind: can this be private?

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+        x = tf.convert_to_tensor(dataset.query_points)
+        y = tf.convert_to_tensor(dataset.observations)
+        data = (x, y)
+        return create_loss_function(model, data, self.compile)
+
+    def optimize(self, model: tf.Module, dataset: Dataset):
+        loss_fn = self.create_loss(model, dataset)
+        variables = model.trainable_variables
+        return self.optimizer.minimize(loss_fn, variables, **self.minimize_args)
```
joelberkeley-secondmind, line 46: do we need this return? it makes the return type a lot more complicated

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
+from typing import Any, Dict, Optional, Callable, Union, Iterable, Tuple
+from functools import singledispatch
+from dataclasses import dataclass, field
+import tensorflow as tf
+import gpflow
+
+from gpflow.models import ExternalDataTrainingLossMixin, InternalDataTrainingLossMixin
+from ..utils import jit
+from ..datasets import Dataset
+
```
joelberkeley-secondmind, line 11: there are a number of partial or missing type hints in this file. Do you mean to fill them in?

trieste/models/model_interfaces.py
```diff
@@ -175,22 +94,22 @@ def predict(self, query_points: QueryPoints) -> Tuple[ObserverEvaluations, Tenso
     def sample(self, query_points: QueryPoints, num_samples: int) -> ObserverEvaluations:
         return self.model.predict_f_samples(query_points, num_samples)
 
+    def optimize(self, dataset: Dataset):
+        self.optimizer.optimize(self.model, dataset)
```
joelberkeley-secondmind, line 96: since the optimizer doesn't access anything private on the model, we should be able to completely decouple the optimizers and models. What do you think about doing that?

joelberkeley-secondmind: this may be a separate option from merging the two methods

trieste/models/model_interfaces.py
awav: No reason, we can remove it.

trieste/models/model_interfaces.py
awav: discussed online

trieste/models/model_interfaces.py
awav: @joelberkeley ^^^

trieste/models/optimizer.py
awav: Callable -> Callable[[], SOMETHING]

trieste/models/optimizer.py
awav: Leave it public.

awav: @joelberkeley ^^^

trieste/models/optimizer.py
awav: yes, I will add a return type then

trieste/models/model_interfaces.py
joelberkeley: you could do this like
```
class _WithOptimizer(GPflowPredictor):
    def __init__(self, optimizer = None):
        self._optimizer = optimizer

    def optimize(self):
        self._optimizer.optimize(self.model)
```
then inherit from this class instead. Does that explain it?

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+
+DatasetBuilder = Callable[[Dataset, Optional[int]], Union[Iterable, Tuple]]
```
joelberkeley: I would expect a `DatasetBuilder` to build datasets. Was there something wrong with the name `Batcher`?

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+        elif self.dataset_builder is None:
+            d = tf.data.Dataset.from_tensor_slices((dataset.query_points, dataset.observations))
+            size = len(dataset)
+            data: Iterable = iter(
+                d.shuffle(size)
+                .batch(self.batch_size)
+                .prefetch(tf.data.experimental.AUTOTUNE)
+                .repeat()
+            )
+            return creator_fn(data)
```
joelberkeley: this looks like it can be written as a default `dataset_builder`, in which case the `batch_size` can be baked into the `dataset_builder` (i.e. as part of the function itself, rather than as an argument), and isn't needed on this class, which feels intuitive to me: a dataset builder builds minibatches from a Dataset. This process can be defined without the notion of a batch size

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+            return create_loss_function(model, data, self.compile)
+
+        if self.dataset_builder is None and self.batch_size is None:
+            x = tf.convert_to_tensor(dataset.query_points)
+            y = tf.convert_to_tensor(dataset.observations)
+            return creator_fn((x, y))
+        elif self.dataset_builder is None:
+            d = tf.data.Dataset.from_tensor_slices((dataset.query_points, dataset.observations))
+            size = len(dataset)
+            data: Iterable = iter(
```
joelberkeley: what happens if you leave this type hint off?

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+
+@dataclass
+class Optimizer:
+    optimizer: Union[gpflow.optimizers.Scipy, tf.optimizers.Optimizer]
+    minimize_args: Dict[str, Any] = field(default_factory=lambda: {})
+    compile: bool = False
+
+    def create_loss(self, model: tf.Module, dataset: Dataset) -> Callable:
+        x = tf.convert_to_tensor(dataset.query_points)
+        y = tf.convert_to_tensor(dataset.observations)
```
joelberkeley, line 39: we aren't using `tf.convert_to_tensor` anywhere else, so do we need it here?

joelberkeley: btw you can also `from dataclasses import astuple; astuple(dataset)` to unpack a dataset

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+        return create_loss_function(model, data, self.compile)
+
+    def optimize(self, model: tf.Module, dataset: Dataset):
+        loss_fn = self.create_loss(model, dataset)
+        variables = model.trainable_variables
+        return self.optimizer.minimize(loss_fn, variables, **self.minimize_args)
+
+
+@dataclass
+class TFOptimizer(Optimizer):
```
joelberkeley, line 50: it looks like this class is more about minibatching than tf per se. Is that right? If so, how about `MiniBatchOptimizer`?

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+    max_iter: int = 100
+    batch_size: Optional[int] = None
+    dataset_builder: Optional[DatasetBuilder] = None
+
+    def __post_init__(self):
+        if self.max_iter is None:
+            raise ValueError("Max iterations expected to be set")
```
joelberkeley: `max_iter` can only be `int`

trieste/models/optimizer.py
```diff
@@ -0,0 +1,118 @@
...
+    def create_loss(self, model: tf.Module, dataset: Dataset) -> Callable:
+        def creator_fn(data: Union[Tuple, Iterable]):
+            return create_loss_function(model, data, self.compile)
+
+        if self.dataset_builder is None and self.batch_size is None:
+            x = tf.convert_to_tensor(dataset.query_points)
+            y = tf.convert_to_tensor(dataset.observations)
+            return creator_fn((x, y))
+        elif self.dataset_builder is None:
+            d = tf.data.Dataset.from_tensor_slices((dataset.query_points, dataset.observations))
```
joelberkeley, line 64: can we have a more descriptive name? Looks like `d` is a pretty specific thing

trieste/models/optimizer.py
joelberkeley, line 11: we'll need docstrings for everything public eventually

trieste/models/model_interfaces.py
joelberkeley: with regards to testing, I think it's OK to change lots of tests if the design asks for it

awav: I think the question is more about delaying the delivery of this PR or simply move it to another ticket.

trieste/models/optimizer.py
awav: It is not necessary a batcher, it can be a data transformer or any other form of mapping function from a dataset to dataset.

trieste/models/optimizer.py
awav: nothing

trieste/models/optimizer.py
awav: I know about `astuple` and convert_to_tensor here is for safety reasons if query_points or observations are numpys.

trieste/models/optimizer.py
awav: dt? db? df? :)
Yes, you are right, but I'm okay with that name. And the reason for that - the scope of the variable is really short. It is two lines of the `if-else` statement.

trieste/models/optimizer.py
awav: The `batch_size` is the only variable that changes here, the user is not obliged to write his own builder all the time. It is enough to set `batch_size` to an appropriate value.

trieste/models/optimizer.py
awav: TF name distinguishes the class of optimizers which are available, not the batching mechanism. Strong _no_ to `MiniBatchOptimizer` :)

trieste/models/model_interfaces.py
joelberkeley: sorry i've conflated conversations here. my suggestion was for something else. I'm ok to leave the optimizers as part of the models. I'll put the other suggestion where I meant to put it

trieste/models/optimizer.py
joelberkeley: can we do that then?

trieste/models/optimizer.py
joelberkeley: we'll need tests for all public functions

trieste/models/optimizer.py
joelberkeley: ok, but `DatasetBuilder` gives the wrong impression, as it doesn't make `Dataset`s

trieste/models/model_interfaces.py
joelberkeley, line 78: this is where I meant the bit about
```
class _WithOptimizer(GPflowPredictor):
    def __init__(self, optimizer = None):
        self._optimizer = optimizer

    def optimize(self):
        self._optimizer.optimize(self.model)
```

trieste/models/optimizer.py
awav: DatasetTransformer

trieste/models/optimizer.py
awav:
```suggestion
```

trieste/models/optimizer.py
awav:
```suggestion
            data = iter(
```

trieste/models/model_interfaces.py
awav:
```suggestion
```

trieste/models/optimizer.py
```diff
@@ -0,0 +1,133 @@
...
+) -> Optimizer:
+    return Optimizer(optimizer, **optimizer_args)
+
+
+Data = Union[Tuple[tf.Tensor, tf.Tensor], Iterable[Tuple[tf.Tensor, tf.Tensor]]]
+
+
+@singledispatch
+def create_loss_function(model, dataset: Data, compile: Optional[bool] = False) -> LossClosure:
+    pass
```
st--: would it be helpful to explicitly raise a NotImplementedError or TypeError or so in case someone passes a model that happens to not inherit from either Internal- or ExternalData...Mixin?

trieste/models/model_interfaces.py
```diff
@@ -26,6 +28,22 @@
...
+
+    @abstractmethod
+    def update(self, dataset: Dataset) -> None:
+        """
+        Update the model given the specified ``dataset``. Does not train the model.
+
+        :param dataset: The data with which to update the model.
+        """
+        raise NotImplementedError
+
```
joelberkeley-secondmind: I don't know if you meant to do this, but this undoes the previous model PR, specifically we want there to be an interface for prediction only

trieste/models/optimizer.py
```diff
@@ -0,0 +1,134 @@
...
+Type alias for a function that maps a dataset from a :class:`~optiflow.datasets.Dataset`
+and generates an iterator over the transformation.
+"""
+
+LossClosure = Callable[[], tf.Tensor]
+"""
+Type alias for a loss closure that is used generally for optimizing.
+"""
+
+OptimizeResult = Union[scipy.optimize.OptimizeResult, None]
```
joelberkeley, line 24: is this name going to be confusing, as we already have an `OptimizationResult` for the main optimizer?

awav: Should we rename the main optimizer result?

joelberkeley: what would you suggest to? I think it's currently quite a good name. Indeed, I was hoping to rename it to `OptimizeResult` cos it's shorter. You've only used this alias once. Is it indispensable?

joelberkeley-secondmind:
what's the status of this PR?

joelberkeley:
what's the status of this PR?

docs/notebooks/failure_ego.pct.py
```diff
@@ -190,7 +201,8 @@ def prepare_acquisition_function(self, dataset, model):
 
 result = bo.optimize(20, initial_data, models, acquisition_rule=rule)
 
-if result.error is not None: raise result.error
+if result.error is not None:
+    raise result.error
```
awav:
```suggestion
if result.error is not None: raise result.error
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -175,6 +185,7 @@ class ProbabilityOfValidity(trieste.acquisition.SingleModelAcquisitionBuilder):
     def prepare_acquisition_function(self, dataset, model):
         return lambda at: trieste.acquisition.lower_confidence_bound(model, 0.0, at)
 
+
```
awav:
```suggestion
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -152,6 +159,7 @@ def optimize(self):
             natgrad_opt.minimize(self.model.training_loss, var_list=variational_params)
             adam_opt.minimize(self.model.training_loss, var_list=self.model.trainable_variables)
 
+
```
awav:
```suggestion
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -118,14 +119,20 @@ def observer(x):
...
     set_trainable(gpr.likelihood, False)
     return gpr
 
 
 def create_classification_model(data):
-    kernel = gpflow.kernels.SquaredExponential(variance=100., lengthscales=0.2 * np.ones(2, ))
+    kernel = gpflow.kernels.SquaredExponential(
+        variance=100.0,
+        lengthscales=0.2 * np.ones(2),
+    )
```
awav:
```suggestion
    kernel = gpflow.kernels.SquaredExponential(variance=100.0, lengthscales=0.2 * np.ones(2))
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -118,14 +119,20 @@ def observer(x):
 # %%
 def create_regression_model(data):
     variance = tf.math.reduce_variance(data.observations)
-    kernel = gpflow.kernels.Matern52(variance=variance, lengthscales=0.2 * np.ones(2, ))
+    kernel = gpflow.kernels.Matern52(
+        variance=variance,
+        lengthscales=0.2 * np.ones(2),
+    )
```
awav:
```suggestion
    kernel = gpflow.kernels.Matern52(variance=variance, lengthscales=0.2 * np.ones(2))
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -94,14 +93,16 @@ def masked_branin(x):
...
 def observer(x):
     y = masked_branin(x)
     mask = np.isfinite(y).reshape(-1)
     return {
         OBJECTIVE: trieste.data.Dataset(x[mask], y[mask]),
-        FAILURE: trieste.data.Dataset(x, tf.cast(np.isfinite(y), tf.float64))
+        FAILURE: trieste.data.Dataset(x, tf.cast(np.isfinite(y), tf.float64)),
     }
 
+
```
awav:
```suggestion
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -94,14 +93,16 @@ def masked_branin(x):
 OBJECTIVE = "OBJECTIVE"
 FAILURE = "FAILURE"
 
+
```
awav:
```suggestion
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -94,14 +93,16 @@ def masked_branin(x):
...
 def observer(x):
     y = masked_branin(x)
     mask = np.isfinite(y).reshape(-1)
     return {
         OBJECTIVE: trieste.data.Dataset(x[mask], y[mask]),
-        FAILURE: trieste.data.Dataset(x, tf.cast(np.isfinite(y), tf.float64))
+        FAILURE: trieste.data.Dataset(x, tf.cast(np.isfinite(y), tf.float64)),
```
awav:
```suggestion
        FAILURE: trieste.data.Dataset(x, tf.cast(np.isfinite(y), tf.float64))
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -52,11 +50,12 @@
 
 # %%
 def masked_branin(x):
-    mask_nan = np.sqrt((x[:, 0] - 0.5) ** 2 + (x[:, 1] - .4) ** 2) < 0.3
+    mask_nan = np.sqrt((x[:, 0] - 0.5) ** 2 + (x[:, 1] - 0.4) ** 2) < 0.3
     y = np.array(objectives.branin(x))
     y[mask_nan] = np.nan
     return tf.convert_to_tensor(y.reshape(-1, 1), x.dtype)
 
+
```
awav:
```suggestion
```

docs/notebooks/failure_ego.pct.py
```diff
@@ -52,11 +50,12 @@
 
 # %%
 def masked_branin(x):
-    mask_nan = np.sqrt((x[:, 0] - 0.5) ** 2 + (x[:, 1] - .4) ** 2) < 0.3
+    mask_nan = np.sqrt((x[:, 0] - 0.5) ** 2 + (x[:, 1] - 0.4) ** 2) < 0.3
```
awav:
```suggestion
    mask_nan = np.sqrt((x[:, 0] - 0.5) ** 2 + (x[:, 1] - .4) ** 2) < 0.3
```

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -71,12 +72,14 @@ def constraint(input_data):
 OBJECTIVE = "OBJECTIVE"
 CONSTRAINT = "CONSTRAINT"
 
+
```
awav:
```suggestion
```

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -88,10 +91,7 @@ def observer(query_points):
 
 # %%
 util.plot_init_query_points(
-    search_space,
-    Sim,
-    astuple(initial_data[OBJECTIVE]),
-    astuple(initial_data[CONSTRAINT])
+    search_space, Sim, astuple(initial_data[OBJECTIVE]), astuple(initial_data[CONSTRAINT])
```
awav:
```suggestion
    search_space,
    Sim,
    astuple(initial_data[OBJECTIVE]),
    astuple(initial_data[CONSTRAINT]),
```

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -141,23 +144,18 @@ def create_bo_model(data):
 
 result = bo.optimize(num_steps, initial_data, models, acquisition_rule=rule)
 
-if result.error is not None: raise result.error
+if result.error is not None:
+    raise result.error
```
awav:
```suggestion
if result.error is not None: raise result.error
```

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -148,16 +153,10 @@ def create_bo_model(data):
 
 # %%
 constraint_data = result.datasets[CONSTRAINT]
-new_data = (
-    constraint_data.query_points[-num_steps:], constraint_data.observations[-num_steps:]
-)
+new_data = (constraint_data.query_points[-num_steps:], constraint_data.observations[-num_steps:])
```
joelberkeley: Is this another black format? I intentionally split lines like this to avoid a horizontal scroll bar on the docs site

docs/notebooks/failure_ego.pct.py
```diff
@@ -141,7 +139,7 @@ def create_classification_model(data):
 
 # %%
 class NatGradTrainedVGP(VariationalGaussianProcess):
-    def optimize(self):
+    def optimize(self, dataset: trieste.data.Dataset):
```
joelberkeley:
```suggestion
    def optimize(self, dataset):
```

docs/conf.py
```diff
@@ -33,18 +33,18 @@
...
+    "You may obtain a copy of the License at\n"
+    "\n"
+    "    http://www.apache.org/licenses/LICENSE-2.0\n"
+    "\n"
+    "Unless required by applicable law or agreed to in writing, software\n"
     'distributed under the License is distributed on an "AS IS" BASIS,\n'
-    'WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n'
-    'See the License for the specific language governing permissions and\n'
+    "WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n"
+    "See the License for the specific language governing permissions and\n"
```
joelberkeley, line 47: here's another black format

notebooks
```diff
@@ -1 +1 @@
-docs/notebooks/
\ No newline at end of file
+docs/notebooks
```
joelberkeley, line 1: is this needed?

docs/notebooks/inequality_constraints.pct.py
awav: Therefore, you have to use `,` after the last entry. `black` would ignore ignore it.
Initial thing was:

```
new_data = (
      constraint_data.query_points[-num_steps:], constraint_data.observations[-num_steps:]
)
```

Actually, I don't think that neither of them are good.

notebooks
awav: This is a symbolic link, and I didn't change it. `GitHub` shows it in a different way, don't know why.

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -148,16 +153,16 @@ def create_bo_model(data):
 
 # %%
 constraint_data = result.datasets[CONSTRAINT]
-new_data = (
-    constraint_data.query_points[-num_steps:], constraint_data.observations[-num_steps:]
-)
+new_query_points = constraint_data.query_points[-num_steps:]
+new_observations = constraint_data.observations[-num_steps:]
+new_data = (new_query_points, new_observations)
```
awav, line 158: Apologies, this is not a black change. But I couldn't leave as is after @joelberkeley remarks :)

docs/notebooks/introduction.pct.py
```diff
@@ -86,8 +86,11 @@ def build_model(data):
 
     return {OBJECTIVE: trieste.models.create_model({
         "model": gpr,
-        "optimizer": gpflow.optimizers.Scipy(),
-        "optimizer_args": {"options": dict(maxiter=100)},
+        "optimizer": {
```
vpicheny: why do we not specify scipy anymore?

tests/integration/test_bayesian_optimization.py
```diff
@@ -65,7 +66,12 @@ def build_model(data: Dataset) -> GaussianProcessRegression:
     model = build_model(initial_data[OBJECTIVE])
 
     res = BayesianOptimizer(observer, search_space).optimize(
-        num_steps, initial_data, {OBJECTIVE: model}, acquisition_rule
+        num_steps,
+        initial_data,
+        {OBJECTIVE: model},
+        acquisition_rule,
+        # TODO(awav): see HashableWeakRef error in GPflow
```
vpicheny: remove todo

awav, line 73:
```suggestion
```

trieste/models/optimizer.py
vpicheny: `OptimizeResult` sounds like an action. `OptimisationResult` or `OptimizerResult` make more sense

vpicheny (approved):

Issue #41, opened 2020-11-06
MC-based batch-EI


vpicheny:
A Monte-Carlo implementation of EI that works for a single query point or a batch of query points. The returned EI is always a scalar. Compared to analytical EI, the parameter eta is optional.

This acquisition fct uses the reparametrisation trick.

joelberkeley:
@vpicheny test

PR #53, opened 2020-11-12
Monte-Carlo acquisition function

vpicheny:
First implementation of a Monte-Carlo acquisition function class, with an application to MC expected improvement.

### DIFF
--- a/trieste/acquisition/function.py
+++ b/trieste/acquisition/function.py
@@ -61,7 +61,7 @@ class SingleModelAcquisitionBuilder(ABC):
 
         class _Anon(AcquisitionFunctionBuilder):
             def prepare_acquisition_function(
-                    self, datasets: Mapping[str, Dataset], models: Mapping[str, ModelInterface]
+                self, datasets: Mapping[str, Dataset], models: Mapping[str, ModelInterface]
             ) -> AcquisitionFunction:
                 return single_builder.prepare_acquisition_function(datasets[tag], models[tag])
 
@@ -72,7 +72,7 @@ class SingleModelAcquisitionBuilder(ABC):
 
     @abstractmethod
     def prepare_acquisition_function(
-            self, dataset: Dataset, model: ModelInterface
+        self, dataset: Dataset, model: ModelInterface
     ) -> AcquisitionFunction:
         """
         :param dataset: The data to use to build the acquisition function.
@@ -88,7 +88,7 @@ class ExpectedImprovement(SingleModelAcquisitionBuilder):
     """
 
     def prepare_acquisition_function(
-            self, dataset: Dataset, model: ModelInterface
+        self, dataset: Dataset, model: ModelInterface
     ) -> AcquisitionFunction:
         """
         :param dataset: The data from the observer.
--- a/trieste/acquisition/function.py
+++ b/trieste/acquisition/function.py
@@ -24,6 +24,9 @@ from ..space import SearchSpace
 
 from scipy.optimize import bisect
 
+from gpflow.config import default_float, default_jitter
+from gpflow.utilities.ops import leading_transpose
+
 AcquisitionFunction = Callable[[QueryPoints], tf.Tensor]
 """ Type alias for acquisition functions. """
 
@@ -138,7 +141,7 @@ class MaxValueEntropySearch(SingleModelAcquisitionBuilder):
     Builder for the max-value entropy search acquisition function (for function minimisation)
     """
 
-    def __init__(self,search_space: SearchSpace, num_samples: int = 10, grid_size: int = 5000):
+    def __init__(self, search_space: SearchSpace, num_samples: int = 10, grid_size: int = 5000):
         """
         :param search_space: The global search space over which the Bayesian optimisation problem is defined.
         :param num_samples: Number of sample draws of the minimal value.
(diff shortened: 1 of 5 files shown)


trieste/acquisition/function.py
```diff
@@ -405,3 +410,93 @@ def probability_of_feasibility(
...
+            mean_for_sample = tf.linalg.adjoint(mean)  # [..., L, N]
+            mean_shape = tf.shape(mean)
+            num_latent = mean_shape[-1]
+
+            jittermat = (
+                    tf.eye(num_latent, batch_shape=mean_shape[:-1], dtype=default_float()) * default_jitter()
+            )  # [..., N, D, D]
+
+            chol = tf.linalg.cholesky(cov + jittermat)  # [..., N, L, L]
+            samples = mean_for_sample[..., None] + tf.linalg.matmul(chol, self.eps)  # [..., N, L, S]
```
vpicheny, line 457: dimensions are probably incorrect after this point. I haven't tested it yet

trieste/acquisition/function.py
```diff
@@ -405,3 +410,93 @@ def probability_of_feasibility(
...
+    """
+    Builder for the expected improvement function where the "best" value is taken to be the minimum
+    of the posterior mean at observed points.
+    """
+
+    def __init__(self, eps_shape: [int]):
+
+        super().__init__()
+        assert len(eps_shape) == 3
+        self.eps = tf.random.normal(eps_shape, dtype=default_float())  # [S, B, L]
```
vpicheny, line 423: for simplicity now the user specifies everything. Actually he should only have to care about S; B is one unless you choose a batch-sequential strategy but this should be given by the rule, and L is the number of latent processes that can be inferred from the model.

trieste/acquisition/function.py
```diff
@@ -405,3 +408,93 @@ def probability_of_feasibility(
...
+
+    def predict_f_samples_with_reparametrisation_trick(self, model: ModelInterface, at: QueryPoints) -> tf.Tensor:
```
joelberkeley, line 425: can this be a module-level function?

joelberkeley: indeed, it being a method means that we need to have a thing called a `MonteCarloAcquisitionFunction`, which I'm not convinced we do

trieste/acquisition/function.py
```diff
@@ -405,3 +408,93 @@ def probability_of_feasibility(
...
+        """
+        # mean, cov = model.predict_f(at, full_cov=True, full_output_cov=True)
+
+        if len(at.shape) == 2:
+            mean, cov = model.predict(at)
+            # mean: [N, L]
+            # cov: [N, L]
+            samples = mean[None, :, None, :] + \
+                      tf.math.sqrt(cov)[None, :, None, :] * self.eps[:, None, ...]  # [S, N, B, L]
+        else:
```
joelberkeley, line 442: why the two branches?

trieste/acquisition/function.py
joelberkeley: do we support batch sequential? cos if not, i suggest we don't make it an option

trieste/acquisition/function.py
```diff
@@ -405,3 +408,93 @@ def probability_of_feasibility(
...
+    """
+    Builder for the expected improvement function where the "best" value is taken to be the minimum
+    of the posterior mean at observed points.
+    """
+
+    def __init__(self, eps_shape: [int]):
+
+        super().__init__()
+        assert len(eps_shape) == 3
+        self.eps = tf.random.normal(eps_shape, dtype=default_float())  # [S, B, L]
```
joelberkeley, line 423: this is Z right? this would be creating it once per problem, not once per step

trieste/acquisition/function.py
```diff
@@ -405,3 +408,93 @@ def probability_of_feasibility(
...
+        :return: The expected improvement function.
+        """
+        mean, _ = model.predict(dataset.query_points)
+        eta = tf.reduce_min(mean, axis=0)
+        return lambda at: self._acquisition_function(self, model, eta, at)
+
+    @staticmethod
+    def _acquisition_function(
+            self, model: ModelInterface, eta: tf.Tensor, at: QueryPoints
+    ) -> tf.Tensor:
```
joelberkeley, line 494: i'm starting to think this `_acquisition_function` method style is confusing. It's arbitrary how it's defined, it's _never_ needed, and it neither is or returns an `AcquisitionFunction`

trieste/models/model_interfaces.py
```diff
@@ -37,7 +37,7 @@ def update(self, dataset: Dataset) -> None:
         raise NotImplementedError
 
     @abstractmethod
-    def predict(self, query_points: QueryPoints) -> Tuple[ObserverEvaluations, TensorType]:
+    def predict(self, query_points: QueryPoints, full_cov: bool = False) -> Tuple[ObserverEvaluations, TensorType]:
```
joelberkeley, line 40: on reflection, changing the structure of function return values based on arguments isn't recommended. I'll continue to think about it, but for posterity, the other options are

- have a completely separate method on `ProbabilsiticModel`, `predict_full_cov`, or some other name
- put the above method on a subclass of `ProbabilsiticModel`, but this will require us to make the model type generic across the code base. There are other cases where that might would help, but sth about it feels wrong
- have `predict` return a Distribution type, which can have the cov functionality on that (my personal favourite)

trieste/acquisition/function.py
joelberkeley: it might be good to split these two branches into two functions if we're not going to treat single-point acquisitions as batched acquisitions with batch size 1

trieste/acquisition/function.py
joelberkeley: why should batch size be determined by the rule instead of the acquisition function?

trieste/acquisition/function.py
joelberkeley, line 442: on more reflection, I don't see why we need to do MC and batch in the same PR, so we'd benefit from splitting it up

trieste/acquisition/function.py
vpicheny: cf https://github.com/secondmind-labs/trieste/issues/41 with correct maths

trieste/acquisition/function.py
vpicheny: We will support batches (cf https://github.com/secondmind-labs/trieste/issues/41). 

I think that the rule is the proper place to define the batch size.

trieste/acquisition/function.py
vpicheny: this is very much copying the gpflow style for conditionals, etc.

trieste/acquisition/function.py
vpicheny: OK then this is not the right place, we need to renew thoses

trieste/acquisition/function.py
vpicheny: isn't it used when we plot the acq fct?

trieste/models/model_interfaces.py
vpicheny: I think for now that my favorite solution is predict_full_cov

trieste/acquisition/function.py
joelberkeley: i don't think it's ever used

trieste/models/model_interfaces.py
```diff
@@ -169,8 +169,8 @@ class GPflowPredictor(ModelInterface, ABC):
     def model(self) -> GPModel:
         """ The underlying GPflow model. """
 
-    def predict(self, query_points: QueryPoints) -> Tuple[ObserverEvaluations, TensorType]:
-        return self.model.predict_f(query_points)
+    def predict(self, query_points: QueryPoints, full_cov: bool = False) -> Tuple[ObserverEvaluations, TensorType]:
+        return self.model.predict_f(query_points, full_cov=full_cov, full_output_cov=False)
```
joelberkeley:
```suggestion
        return self.model.predict_f(query_points, full_cov=full_cov)
```
`full_output_cov` defaults to `False`

PR #59, opened 2020-11-16
add tests for search spaces

joelberkeley:


trieste/space.py
```diff
@@ -146,12 +147,13 @@ def __contains__(self, value: TensorType) -> Union[bool, tf.Tensor]:
         :param value: A point to check for membership of this :class:`SearchSpace`.
         :return: `True` if ``value`` is a member of this search space, else `False`. May return a
             scalar boolean `tf.Tensor` instead of the `bool` itself.
-        :raise ValueError: If ``value`` has a different shape from the search space.
+        :raise ValueError (or InvalidArgumentError): If ``value`` has a different dimensionality
+            from the search space.
         """
         if not shapes_equal(value, self._lower):
             raise ValueError(
```
ChrisMorter, line 154: Probably not something for this PR, but perhaps it would be worth introducing an error type specifically for shape errors?

ChrisMorter (approved):
Looks good

trieste/space.py
joelberkeley: yeah that would be nice. I did try to add that exact thing before, but it wasn't received well

joelberkeley: this wasn't why it wasn't received well, but it would be tricky: a lot of shape checking is done by `tf.debugging.assert_shapes`, and you can't `try: except:` in a `tf.function` so we can't change the error type

PR #61, opened 2020-11-16
Monte-Carlo based batch expected improvement.

### LINKED ISSUE #40: MC-based acquisition function supporting reparametrization trick
Class of acquisition function that relies on GP sampling. The sampling uses the reparametrization trick, so that:
```python
def sample_with_trick(at, model, Z)
    mu, sigma = model.predict(at)
    samples = mu + chol(sigma) * Z
    return samples
```
The acquisition function uses a set of Z values fixed over a single BO iteration but renewed for a new iteration.

vpicheny:
Monte-Carlo based batch expected improvement.

- creates a new acq rule to deal with batches. The rule takes a search space a batch size and creates and expanded search space and a wrapper around the acquisition function so that its input is a vector.
- tests the rule on a notebook


docs/notebooks/montecarlo.pct.py
```diff
@@ -0,0 +1,177 @@
...
+    gpr = gpflow.models.GPR(astuple(data), kernel, noise_variance=1e-5)
+    set_trainable(gpr.likelihood, False)
+
+    return {OBJECTIVE: trieste.models.create_model_interface({
+        "model": gpr,
+        "optimizer": gpflow.optimizers.Scipy(),
+        "optimizer_args": {"options": dict(maxiter=100)},
+    })}
+
+model = build_model(initial_data[OBJECTIVE])
```
joelberkeley, line 93: we shouldn't be repeating this (everything above) in every notebook. It makes it harder for users to find what they're looking for, and is a maintenance headache

trieste/acquisition/rule.py
```diff
@@ -290,7 +293,75 @@ def acquire(
...
+    def expand_search_space(self, search_space: SearchSpace):
+        assert isinstance(search_space, Box)
+
+        lower_bound, upper_bound = search_space.lower, search_space.upper
+        expanded_lower_bound = tf.tile(lower_bound, [self._num_query_points])
+        expanded_upper_bound = tf.tile(upper_bound, [self._num_query_points])
+        return Box(expanded_lower_bound, expanded_upper_bound)
+
+    def vectorize_batch_acquisition(self, acquisition_function, dim):
+        return lambda at: acquisition_function(tf.reshape(at, [-1, self._num_query_points, dim]))
```
joelberkeley: do users need these methods (can you give an example)? if not, can we make them private?

trieste/acquisition/rule.py
```diff
@@ -290,7 +293,75 @@ def acquire(
...
+        self._num_query_points = num_query_points
+
+        if builder is None:
+            num_samples = 100
+            builder = MonteCarloExpectedImprovement(eps_shape=[num_samples, num_query_points, 1]).using(OBJECTIVE)  # [S, B, L]
+
+        self._builder = builder
+
+    def expand_search_space(self, search_space: SearchSpace):
+        assert isinstance(search_space, Box)
```
joelberkeley: if this class only works with `Box`es, you can change the declaration to
```
class BatchAcquisitionRule(AcquisitionRule[None, Box]):
```
and the method to
```
    def acquire(
        self,
        search_space: Box,
        ...
    ) -> Tuple[QueryPoints, None]:
```
You can do this for any search space e.g. if it only works with a `DiscreteSearchSpace`

trieste/acquisition/rule.py
```diff
@@ -290,7 +293,75 @@ def acquire(
...
+
+    def acquire(
+        self,
+        search_space: SearchSpace,
+        datasets: Mapping[str, Dataset],
+        models: Mapping[str, ModelInterface],
+        state: None = None,
+    ) -> Tuple[QueryPoints, None]:
+        """
+        Return the query point that optimizes the acquisition function produced by `builder` (see
```
joelberkeley, line 348: is this more than one query point now?

trieste/acquisition/rule.py
```diff
@@ -27,16 +27,19 @@
...
+from ..space import Box
 
 S = TypeVar("S")
 """ Unbound type variable. """
 
 SP = TypeVar("SP", bound=SearchSpace, contravariant=True)
 """ Contravariant type variable bound to :class:`SearchSpace`. """
 
+AcquisitionFunction = Callable[[QueryPoints], tf.Tensor]
+""" Type alias for acquisition functions. """
```
joelberkeley: you can import type aliases

trieste/acquisition/function.py
```diff
@@ -405,3 +408,106 @@ def probability_of_feasibility(
...
+
+            mean, cov = model.predict(at, full_cov=True)
+            # mean: [N, B, L]
+            # cov: [N, L, B, B]
+            mean_for_sample = tf.linalg.adjoint(mean)  # [N, L, B]
+            mean_shape = tf.shape(mean_for_sample)
+            B = mean_shape[-1]
+
+            jittermat = (
+                    tf.eye(B, batch_shape=mean_shape[:-1], dtype=default_float()) * default_jitter()
```
joelberkeley: can you use `cov.dtype` instead of `default_float()`? that would mean the code works regardless of concrete types

trieste/acquisition/function.py
```diff
@@ -405,3 +408,106 @@ def probability_of_feasibility(
...
+        if len(at.shape) == 2:
+            mean, cov = model.predict(at)
+            # mean: [N, L]
+            # cov: [N, L]
+            samples = mean[None, :, None, :] + \
+                      tf.math.sqrt(cov)[None, :, None, :] * self.eps[:, None, ...]  # [S, N, B, L]
+        else:
+            assert at.shape[1] == self.eps.shape[1]
+
+            mean, cov = model.predict(at, full_cov=True)
```
joelberkeley: can you remind me why you're using `full_cov=True` here but not when there's no batching?

trieste/acquisition/rule.py
```diff
@@ -290,7 +293,75 @@ def acquire(
...
         state_ = TrustRegion.State(acquisition_space, eps, y_min, is_global)
 
         return point, state_
+
+
+class BatchAcquisitionRule(AcquisitionRule[None, SearchSpace]):
+    """ Implements a acquisition rule for a batch of query points """
+
+    def __init__(self, num_query_points: int,
+                 builder: Optional[AcquisitionFunctionBuilder] = None):
```
joelberkeley: can you pass any acquisition function builder into this, and it will work?

trieste/acquisition/function.py
```diff
@@ -405,3 +408,106 @@ def probability_of_feasibility(
     mean, var = model.predict(at)
     distr = tfp.distributions.Normal(mean, tf.sqrt(var))
     return distr.cdf(tf.cast(threshold, at.dtype))
+
+
+class MonteCarloAcquisition(SingleModelAcquisitionBuilder):
```
joelberkeley: Can you pass a `MonteCarloAcquisition` to `EfficientGlobalOptimization` or `TrustRegion`?

trieste/acquisition/rule.py
joelberkeley: needs updating for multiple query points

trieste/acquisition/rule.py
```diff
@@ -290,7 +293,75 @@ def acquire(
...
+
+        self._builder = builder
+
+    def expand_search_space(self, search_space: SearchSpace):
+        assert isinstance(search_space, Box)
+
+        lower_bound, upper_bound = search_space.lower, search_space.upper
+        expanded_lower_bound = tf.tile(lower_bound, [self._num_query_points])
+        expanded_upper_bound = tf.tile(upper_bound, [self._num_query_points])
+        return Box(expanded_lower_bound, expanded_upper_bound)
```
joelberkeley: any reason we can't do this for discrete search spaces?

joelberkeley: e.g. can we tile all the points in a discrete search space and achieve the same?

trieste/acquisition/rule.py
```diff
@@ -290,7 +293,75 @@ def acquire(
...
         point = _optimizer.optimize(acquisition_space, acquisition_function)
         state_ = TrustRegion.State(acquisition_space, eps, y_min, is_global)
 
         return point, state_
+
+
+class BatchAcquisitionRule(AcquisitionRule[None, SearchSpace]):
+    """ Implements a acquisition rule for a batch of query points """
+
+    def __init__(self, num_query_points: int,
```
joelberkeley: you mentioned in [the other PR](https://github.com/secondmind-labs/trieste/pull/53/files#r522213555) that you think the batch size should be passed to the rule. Can you elaborate on why?

trieste/acquisition/rule.py
joelberkeley: having chatted on slack, the answer is no

trieste/acquisition/rule.py
vpicheny: I think that it can be useful for plotting, post-optim analysis, etc.

vpicheny: happy to make them private if you think it's important (but why is it important?)

trieste/acquisition/rule.py
vpicheny: sounds good. Ideally we'll be able to change it back to SearchSpace  in the future but this may be another ticket

docs/notebooks/montecarlo.pct.py
vpicheny: agreed - what's the solution then?

trieste/acquisition/function.py
vpicheny: we should definitely be able to do that

trieste/acquisition/rule.py
joelberkeley: public methods are features, so they can indicate scope creep

trieste/acquisition/function.py
vpicheny: When there is no batching you're evaluating the acquisition function for each query point independently. Hence, the covariance between query points is useless. When you have a batch, the covariance within the batch may be useful. But you're still evaluating batches independently so the covariance between batches is useless.

Actually, the current setting might be too batchEI-specific. You could want to evaluate a batch-acq_fct without needing the within-batch covariance.

trieste/acquisition/rule.py
joelberkeley: to think about it another way: there are several ways to make this functionality available, so we need to think about which is best. That may not take very long, but it is extra work that's not needed to complete this work. Similarly, public functionality needs tests - we'll need tests for every public method. That tends to take time, so we need to ask if this is the best use of our time

trieste/acquisition/function.py
joelberkeley: how much extra computation does it take to compute the full cov?

trieste/acquisition/rule.py
vpicheny: it seems more intuitive to me. You set the rule with the batch size: the rule is the evaluate the B set of points that optimize the acq function. 
The alternative, where the acq fct defines the batch size, seems backwards. You would then use the properties of the acqfct to define the optimizer for instance.

trieste/acquisition/rule.py
vpicheny: no reason why not. But I think that this `expand` should be a feature of the search space. Should we create a PR for that?

trieste/acquisition/function.py
vpicheny: Depends on the batch size but that could be really a lot. It is very important to do the minimum number of operations here.

trieste/acquisition/function.py
```diff
@@ -493,3 +494,94 @@ def prepare_acquisition_function(
...
+    """
+    Builder for the expected improvement function where the "best" value is taken to be the minimum
+    of the posterior mean at observed points.
+    """
+
+    def __init__(self, eps_shape: [int]):
+
+        super().__init__()
+        assert len(eps_shape) == 3
+        self.eps = tf.random.normal(eps_shape, dtype=default_float())  # [S, B, L]
```
vpicheny: @joelberkeley you mentioned in a previous comment that this would lead to a fixed sample throughout the BO process, which is not what we want. How should we do it then?

trieste/acquisition/function.py
```diff
@@ -493,3 +494,94 @@ def prepare_acquisition_function(
...
+class MonteCarloAcquisition(SingleModelAcquisitionBuilder):
+    """
+    Builder for the expected improvement function where the "best" value is taken to be the minimum
+    of the posterior mean at observed points.
+    """
+
+    def __init__(self, eps_shape: [int]):
+
+        super().__init__()
+        assert len(eps_shape) == 3
```
vpicheny: This is still WIP. The batch size B should be defined through the rule, not by hand as here, and the number of latent variables L is given by the models. Only the sample size S may be chosen by the user. 
Also, I am unsure how this should be extended to multiple models.

trieste/acquisition/rule.py
vpicheny: I don't know how!

trieste/models/model_interfaces.py
```diff
@@ -27,7 +27,7 @@ class ProbabilisticModel(ABC):
     """ A probabilistic model. """
 
     @abstractmethod
-    def predict(self, query_points: QueryPoints) -> Tuple[ObserverEvaluations, TensorType]:
+    def predict(self, query_points: QueryPoints, full_cov: bool = False) -> Tuple[ObserverEvaluations, TensorType]:
```
vpicheny, line 30: from previous PR: still under discussion if we should use a `full_cov` argument or a separated `predict_full_cov`

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -161,6 +169,127 @@ def create_bo_model(data):
...
+        objective_dataset = datasets[self._objective_tag]
+
+        if len(objective_dataset) == 0:
+            raise ValueError(
+                "Expected improvement is defined with respect to existing points in the objective"
+                " data, but the objective data is empty."
+            )
+
+        constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)
+        pof = constraint_fn(objective_dataset.query_points)
```
vpicheny, line 221: this dependency on the old probability of feasibility is an ugly hack, I was lazy to replace it with a MC one for consistency. Will do next.

docs/notebooks/montecarlo.pct.py
vpicheny: @joelberkeley do you have something in mind? the only alternative I see is to add all of this to the intro notebook but it's already quite long with the MES extension

trieste/acquisition/function.py
```diff
@@ -524,3 +527,194 @@ def prepare_acquisition_function(
...
         """
+
+
+class SingleModelBatchAcquisitionBuilder(BatchAcquisitionFunctionBuilder):
+    """
+    Convenience acquisition function builder for an acquisition function (or component of a
+    composite acquisition function) that requires only one model, dataset pair.
+    """
+
+    def using(self, tag: str) -> BatchAcquisitionFunctionBuilder:
```
vpicheny, line 538: this is a large chunk of repeated code but I don't know how to avoid it

trieste/acquisition/function.py
```diff
@@ -524,3 +527,194 @@ def prepare_acquisition_function(
...
+
+def predict_independent_f_samples_with_reparametrisation_trick(
+        model: ProbabilisticModel, at: QueryPoints, eps: tf.Tensor
+) -> tf.Tensor:
+    """
+    Returns independent samples according to the reparametrization trick.
+    The sample size S is determined by eps,
+    N is the number of query points and L is the number of latent processes.
+    :param model:
+    :param at: N query points [N, D]
```
vpicheny, line 576: I'm a bit unhappy with the rank 2 constraint, I'll probably remove it

trieste/acquisition/function.py
```diff
@@ -524,3 +527,194 @@ def prepare_acquisition_function(
...
+        model: ProbabilisticModel, at: QueryPoints, eps: tf.Tensor
+) -> tf.Tensor:
+    """
+    Returns batch-correlated samples according to the reparametrization trick.
+    The sample size S is determined by eps,
+    the batch size B must be compatible between eps and at,
+    N is the number of batches of query points
+    and L is the number of latent processes.
+    :param model:
+    :param at: N batches of query points [N, B, D]
```
vpicheny, line 595: same here, this should be able to broadcast

trieste/acquisition/function.py
joelberkeley-secondmind: i've finally has a chance to sit down and really think this through. Let's chat, cos it's complicated

trieste/acquisition/function.py
joelberkeley-secondmind: btw `tf.rank(x) == len(x.shape)`

trieste/acquisition/function.py
```diff
@@ -524,3 +527,194 @@ def prepare_acquisition_function(
...
+
+    def prepare_acquisition_function(
+            self, dataset: Dataset, model: GPflowPredictor
+    ) -> AcquisitionFunction:
+        """
+        :param dataset: The data from the observer.
+        :param model: The model over the specified ``dataset``.
+        :return: The expected improvement function.
+        """
+        eps_shape = [self.num_samples, model.model.num_latent_gps]
```
joelberkeley-secondmind, line 637: in general, `model.model` is not possible, because the acquisition functionality uses `ProbabilisticModel`s which don't have an underlying model. If we're working with vector targets, we can get the number of latent gps from e.g. the predictive mean: `mean.shape[-1]` unless we're working with higher-rank targets, which I've not heard any request to

trieste/acquisition/function.py
```diff
@@ -524,3 +527,194 @@ def prepare_acquisition_function(
...
+    )  # [N, L, B, B]
+
+    eps = tf.transpose(eps, [2, 1, 0])  # [S, B, L] -> [L, B, S]
+
+    chol = tf.linalg.cholesky(cov + jittermat)  # [N, L, B, B]
+    samples = mean_for_sample[..., None] + tf.linalg.matmul(chol, eps)  # [N, L, B, S]
+    return tf.transpose(samples, [3, 0, 2, 1])
+
+
+class MonteCarloExpectedImprovement(SingleModelAcquisitionBuilder):
```
joelberkeley-secondmind, line 620: can you clarify for me: is there actually such a thing as "Monte-Carlo Expected Improvement", or is there just "Expected Improvement" achieved via Monte Carlo sampling with the reparametrization trick? Cos if they're the same, why not merge them into a single acquisition and just use the analytic form where it's possible and the MC trick otherwise (which if i understand correctly is the batch case)? Well, I guess to keep batch and non-batch separate, we'll have to keep them separate, but in that case is it more: "ExpectedImprovement" vs "BatchExpectedImprovement"?

trieste/acquisition/function.py
joelberkeley-secondmind: ah, ok, I've scrolled down a bit to "BatchMonteCarloExpectedImprovement", which partially answers my question. So my next question is then: what does "MonteCarloExpectedImprovement" that "ExpectedImprovement" doesn't?

docs/notebooks/montecarlo.pct.py
joelberkeley-secondmind: most of the notebooks make basically the same gpflow GPR, so we can just make a function in the notebook utils that does this

joelberkeley-secondmind: as for the bit above that
```
# this first line is a symptom of the single/multi-model API conflict, which will hopefully go
observer = mk_observer(branin, OBJECTIVE)
# I plan to make Box accept lists, so the rest will be a one-liner 
lower_bound = tf.cast(mins, gpflow.default_float())
upper_bound = tf.cast(maxs, gpflow.default_float())
search_space = trieste.space.Box(lower_bound, upper_bound)
```

joelberkeley-secondmind: above that, we'll make plotting functions for observers, so that should be a one-liner too

joelberkeley-secondmind: the rest is text, and we shouldn't be duplicating that, and imports

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -161,6 +169,127 @@ def create_bo_model(data):
...
+    """
+    def __init__(
+        self,
+        objective_tag: str,
+        constraint_builder: AcquisitionFunctionBuilder,
+        num_samples,
+        num_query_points,
+        threshold,
+        min_feasibility_probability: Union[float, tf.Tensor] = 0.5,
+    ):
```
joelberkeley-secondmind, line 191: there's no need for `__init__` args for classes you're only going to use once

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -144,7 +152,7 @@ def create_bo_model(data):
 if result.error is not None: raise result.error
 
 # %% [markdown]
-# To conclude, we visualise the resulting data. Orange dots show the new points queried during optimization. Notice the concentration of these points in regions near the local minima.
+# Finally, we visualise the resulting data. Orange dots show the new points queried during optimization. Notice the concentration of these points in regions near the local minima.
```
joelberkeley-secondmind, line 155: At last!

trieste/acquisition/function.py
joelberkeley-secondmind: ah, ok, I've scrolled down a bit to "BatchMonteCarloExpectedImprovement", which partially answers my question. So my next question is then: what does "MonteCarloExpectedImprovement" that "ExpectedImprovement" doesn't?

and the same qu for `MonteCarloNegativeLowerConfidenceBound` vs `NegativeLowerConfidenceBound`

trieste/acquisition/function.py
```diff
@@ -524,3 +527,194 @@ def prepare_acquisition_function(
...
+) -> tf.Tensor:
+    """
+    Returns batch-correlated samples according to the reparametrization trick.
+    The sample size S is determined by eps,
+    the batch size B must be compatible between eps and at,
+    N is the number of batches of query points
+    and L is the number of latent processes.
+    :param model:
+    :param at: N batches of query points [N, B, D]
+    :param eps: a [S, B, L] tensor of N(0, 1) i.i.d. samples
```
joelberkeley-secondmind, line 596: you could pass the shape to this function and let the function make the N(0,1)

trieste/acquisition/function.py
```diff
@@ -524,3 +527,194 @@ def prepare_acquisition_function(
...
+
+    mean, cov = model.predict(at, full_cov=True)
+    # mean: [N, B, L]
+    # cov: [N, L, B, B]
+    mean_for_sample = tf.linalg.adjoint(mean)  # [N, L, B]
+    mean_shape = tf.shape(mean_for_sample)
+    B = mean_shape[-1]
+
+    jittermat = (
+            tf.eye(B, batch_shape=mean_shape[:-1], dtype=default_float()) * default_jitter()
```
joelberkeley-secondmind:
```suggestion
            tf.eye(B, batch_shape=mean_shape[:-1], dtype=cov.dtype) * default_jitter()
```
since we're adding it to `cov` anyway

joelberkeley-secondmind: i'm not sure about using gpflow config yet. global config's a bit of a statement* and isn't necessary here (we can make jitter a parameter with default to this function)

\* mark has suggested dropping jitter from the gpflow config for this reason

docs/notebooks/inequality_constraints.pct.py
joelberkeley: but anyway, why the change? finally sounds like we've been waiting the whole notebook for this

joelberkeley:
have we merged everything from here (except MC LCB which I recall you saying we won't merge)?

PR #68, opened 2020-11-23
Combining two box search spaces

### LINKED ISSUE #67: Combine Box search spaces
A method to combine an arbitrary number of box spaces into a single one. This is done simply by tiling lower and upper bounds.

This feature is useful for batch rules.

vpicheny:


trieste/space.py
```diff
@@ -171,3 +172,20 @@ def discretize(self, num_samples: int) -> DiscreteSearchSpace:
             this :class:`Box`.
         """
         return DiscreteSearchSpace(points=self.sample(num_samples))
+
+    def __mul__(self, other_box: Box):
```
joelberkeley:
```suggestion
    def __mul__(self, other: Box) -> Box:
```
is the usual naming, and with the return type

tests/unit/test_space.py
```diff
@@ -212,3 +212,12 @@ def test_box_discretize_returns_search_space_with_correct_number_of_points(
...
         dss.sample(num_samples + 1)
+
+
+def test_box_combined_with_itself_returns_new_box_with_bounds_twice_as_large(
+) -> None:
+    box = Box(tf.zeros((3,)), tf.ones((3,)))
+    new_box = box * box
+
+    assert len(new_box.lower) == 6
+    assert len(new_box.upper) == 6
```
joelberkeley, line 222: how about a test that checks the contents of the lower and upper bounds for two `Box`es
```
box1 = Box(tf.constant([0., 1.]), tf.constant([2., 3.]))
box2 = Box(tf.constant([4., 5., 6.]), tf.constant([7., 8., 9.]))

res = box1 * box2
npt.assert_allclose(res.lower, [0, 1, 4, 5, 6])
npt.assert_allclose(res.upper, [2, 3, 7, 8, 9])
```

joelberkeley: and also one with different dtypes, that it errors, with
```
box1 = Box(tf.constant([0., 1.]), tf.constant([2., 3.]))
box2 = Box(tf.constant([4., 5.], tf.float64), tf.constant([6., 7.], tf.float64))

with pytest.raises(TypeError):
    _ = box1 * box2
```

trieste/space.py
```diff
@@ -171,3 +172,20 @@ def discretize(self, num_samples: int) -> DiscreteSearchSpace:
...
+        Combines 2 Box spaces into a single one by concatenating their respective lower and upper bounds.
+        :param other_box: :class:`Box`.
+        :return: the new combined :class:`Box`.
+        :raise ValueError (or InvalidArgumentError): If ``other_box`` is not from :class:`Box`.
+        """
+        if self.lower.dtype is not other_box.lower.dtype:
+            raise TypeError(
+                f"Bounds of both boxes must have the same dtype, got {self.lower.dtype} and"
+                f" {other_box.lower.dtype}"
+            )
```
joelberkeley: this is nice, but in this case we need to `return NotImplemented` for invalid args (see [the docs](https://docs.python.org/3/reference/datamodel.html#object.__mul__))

joelberkeley: you'll need to update the docstring accordingly (though pretend it raises `TypeError` rather than returning `NotImplemented`, cos that's what the user will actually see).

trieste/space.py
```diff
@@ -171,3 +172,20 @@ def discretize(self, num_samples: int) -> DiscreteSearchSpace:
             this :class:`Box`.
         """
         return DiscreteSearchSpace(points=self.sample(num_samples))
+
+    def __mul__(self, other_box: Box):
+        """
+        Combines 2 Box spaces into a single one by concatenating their respective lower and upper bounds.
```
joelberkeley:
```suggestion
        Return the Cartesian product of the two :class:`Box`\ es (concatenating their respective lower and upper bounds).
```

joelberkeley: is that accurate?

trieste/space.py
vpicheny: is this what you had in mind?

trieste/space.py
```diff
@@ -171,3 +172,19 @@ def discretize(self, num_samples: int) -> DiscreteSearchSpace:
...
+        """
+        Return the Cartesian product of the two :class:`Box`\ es (concatenating their respective lower and upper bounds).
+        :param box: :class:`Box`.
+        :return: the new combined :class:`Box`, or NotImplemented if respective bounds have different dtypes.
+        """
+        if self.lower.dtype is not box.lower.dtype:
+            return NotImplemented(
+                f"Bounds of both boxes must have the same dtype, got {self.lower.dtype} and"
+                f" {box.lower.dtype}"
+            )
```
joelberkeley, line 186: haha, that's a funny bug. `NotImplemented` doesn't take arguments (it's a builtin constant not an exception type). So when you try to pass it arguments, python raises a `TypeError`, and your test passes

trieste/space.py
joelberkeley: you've gone with sth else, rather than my suggestion. Why so?

trieste/space.py
joelberkeley: hehe, almost, but this has a funny bug. `NotImplemented` doesn't take arguments (it's a builtin constant not an exception type). So when you try to pass it arguments, python raises a `TypeError`, and your test passes due to the wrong `TypeError`

tests/unit/test_space.py
```diff
@@ -212,3 +212,31 @@ def test_box_discretize_returns_search_space_with_correct_number_of_points(
 
     with pytest.raises(ValueError):
         dss.sample(num_samples + 1)
+
+
+def test_box_combined_with_itself_returns_new_box_with_bounds_twice_as_large(
+) -> None:
```
joelberkeley-secondmind, line 218: do these not fit on one line?

joelberkeley-secondmind: these fit on one line. Any reason they're on two?

trieste/space.py
```diff
@@ -171,3 +172,19 @@ def discretize(self, num_samples: int) -> DiscreteSearchSpace:
             this :class:`Box`.
         """
         return DiscreteSearchSpace(points=self.sample(num_samples))
+
+    def __mul__(self, box: Box):
+        """
+        Return the Cartesian product of the two :class:`Box`\ es (concatenating their respective lower and upper bounds).
+        :param box: :class:`Box`.
+        :return: the new combined :class:`Box`, or NotImplemented if respective bounds have different dtypes.
```
joelberkeley-secondmind:
```suggestion
        :return: the new combined :class:`Box`
        :raise TypeError: If the lhs and rhs :class:`Box` bounds have different types.
```
since the user will never actually see a `NotImplemented` (`NotImplemented` is python's under the hood way of deferring `a * b` to `b * a` if it's defined, which it isn't here)

joelberkeley-secondmind (approved):

PR #69, opened 2020-11-24
Acquisition rule for batches of points

vpicheny:
Creates a new acquisition rule class that returns a batch of query points instead of a single point. 
Requires an acquisition function that accepts batches of points as input, i.e. of shape [..., B, D] (B=batch size, D=search space dimension).
Comes with two tests: one checking that the rule returns a batch of the right size, the other that checks that the rule returns the batch that optimises a dummy batch acquisition function.

### DIFF
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -192,29 +192,29 @@ def test_trust_region_for_unsuccessful_local_to_global_trust_region_reduced() ->
 
 
 class _BatchModelMinusMeanMaximumSingleBuilder(BatchAcquisitionFunctionBuilder):
-    def using(self, tag: str) -> BatchAcquisitionFunctionBuilder:
-        """
-        :param tag: The tag for the model, dataset pair to use to build this acquisition function.
-        :return: An acquisition function builder that selects the model and dataset specified by
-            ``tag``, as defined in :meth:`prepare_acquisition_function`.
-        """
-        single_builder = self
-
-        class _Anon(BatchAcquisitionFunctionBuilder):
-            def prepare_acquisition_function(
-                self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
-            ) -> BatchAcquisitionFunction:
-                return single_builder.prepare_acquisition_function(datasets[tag], models[tag])
-
-            def __repr__(self) -> str:
-                return f"{single_builder!r} using tag {tag!r}"
-
-        return _Anon()
+    # def using(self, tag: str) -> BatchAcquisitionFunctionBuilder:
+    #     """
+    #     :param tag: The tag for the model, dataset pair to use to build this acquisition function.
+    #     :return: An acquisition function builder that selects the model and dataset specified by
+    #         ``tag``, as defined in :meth:`prepare_acquisition_function`.
+    #     """
+    #     single_builder = self
+    #
+    #     class _Anon(BatchAcquisitionFunctionBuilder):
+    #         def prepare_acquisition_function(
+    #             self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
+    #         ) -> BatchAcquisitionFunction:
+    #             return single_builder.prepare_acquisition_function(datasets[tag], models[tag])
+    #
+    #         def __repr__(self) -> str:
+    #             return f"{single_builder!r} using tag {tag!r}"
+    #
+    #     return _Anon()
 
     def prepare_acquisition_function(
-        self, dataset: Dataset, model: ProbabilisticModel
+        self, dataset: Dataset, model: Mapping[str, ProbabilisticModel]
     ) -> BatchAcquisitionFunction:
-        return lambda at: -tf.reduce_max(model.predict(at)[0], axis=-2)
+        return lambda at: -tf.reduce_max(model[OBJECTIVE].predict(at)[0], axis=-2)
 
 
 def test_batch_acquisition_returns_batches_of_right_size() -> None:
@@ -222,7 +222,7 @@ def test_batch_acquisition_returns_batches_of_right_size() -> None:
     num_query_points = 3
     ego = BatchAcquisitionRule(
         num_query_points,
-        _BatchModelMinusMeanMaximumSingleBuilder().using(OBJECTIVE))
+        _BatchModelMinusMeanMaximumSingleBuilder())
     dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
     query_point, _ = ego.acquire(
         search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
@@ -236,7 +236,7 @@ def test_batch_acquisition_finds_minimum() -> None:
     num_query_points = 4
     ego = BatchAcquisitionRule(
         num_query_points,
-        _BatchModelMinusMeanMaximumSingleBuilder().using(OBJECTIVE))
+        _BatchModelMinusMeanMaximumSingleBuilder())
     dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
     query_point, _ = ego.acquire(
         search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -198,6 +198,19 @@ class _BatchModelMinusMeanMaximumSingleBuilder(BatchAcquisitionFunctionBuilder):
         return lambda at: -tf.reduce_max(model[OBJECTIVE].predict(at)[0], axis=-2)
 
 
+def test_batch_acquisition_returns_batches_of_right_size() -> None:
+    search_space = Box(tf.constant([-2.2, -1.0]), tf.constant([1.3, 3.3]))
+    num_query_points = 3
+    ego = BatchAcquisitionRule(
+        num_query_points,
+        _BatchModelMinusMeanMaximumSingleBuilder())
+    dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
+    query_point, _ = ego.acquire(
+        search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
+    )
+    assert query_point.shape == [num_query_points, 2]
+
+
 def test_batch_acquisition_finds_minimum() -> None:
     search_space = Box(tf.constant([-2.2, -1.0]), tf.constant([1.3, 3.3]))
     num_query_points = 4
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -213,7 +213,7 @@ class _BatchModelMinusMeanMaximumSingleBuilder(BatchAcquisitionFunctionBuilder):
 
     def prepare_acquisition_function(
         self, dataset: Dataset, model: ProbabilisticModel
-    ) -> AcquisitionFunction:
+    ) -> BatchAcquisitionFunction:
         return lambda at: -tf.reduce_max(model.predict(at)[0], axis=-2)
 
 
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -24,10 +24,14 @@ from trieste.acquisition.rule import (
     ThompsonSampling,
     TrustRegion,
     OBJECTIVE,
+    BatchAcquisitionRule,
 )
 from trieste.data import Dataset
 from trieste.models import ProbabilisticModel
 from trieste.space import SearchSpace, DiscreteSearchSpace, Box
+from trieste.acquisition import SingleModelAcquisitionBuilder
+
+from trieste.acquisition.function import AcquisitionFunction
 
 from tests.util.misc import one_dimensional_range, zero_dataset
 from tests.util.model import QuadraticWithUnitVariance
@@ -186,3 +190,33 @@ def test_trust_region_for_unsuccessful_local_to_global_trust_region_reduced() ->
     npt.assert_array_less(current_state.eps, previous_state.eps)  # current TR smaller than previous
     assert current_state.is_global
     npt.assert_array_almost_equal(current_state.acquisition_space.lower, lower_bound)
+
+
+class _BatchModelMinusMeanMaximumSingleBuilder(SingleModelAcquisitionBuilder):
+    def prepare_acquisition_function(
+        self, dataset: Dataset, model: ProbabilisticModel
+    ) -> AcquisitionFunction:
+        return lambda at: -tf.reduce_max(model.predict(at)[0], axis=-2)
+
+
+def test_batch_acquisition_returns_batches_of_right_size() -> None:
+    search_space = Box(tf.constant([-2.2, -1.0]), tf.constant([1.3, 3.3]))
+    num_query_points = 3
+    ego = BatchAcquisitionRule(num_query_points, _BatchModelMinusMeanMaximumSingleBuilder().using(OBJECTIVE))
+    dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
+    query_point, _ = ego.acquire(
+        search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
+    )
+    assert query_point.shape == [num_query_points, 2]
+
+
+def test_batch_acquisition_finds_minimum() -> None:
+    search_space = Box(tf.constant([-2.2, -1.0]), tf.constant([1.3, 3.3]))
+    expected_minimum = tf.constant([0., 0.])
+    num_query_points = 4
+    ego = BatchAcquisitionRule(num_query_points, _BatchModelMinusMeanMaximumSingleBuilder().using(OBJECTIVE))
+    dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
+    query_point, _ = ego.acquire(
+        search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
+    )
+    npt.assert_allclose(query_point - expected_minimum[None, :], 0., atol=1e-3)
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -198,9 +198,8 @@ class _BatchModelMinusMeanMaximumSingleBuilder(BatchAcquisitionFunctionBuilder):
         return lambda at: -tf.reduce_max(model[OBJECTIVE].predict(at)[0], axis=-2)
 
 
-def test_batch_acquisition_returns_batches_of_right_size_and_finds_minimum() -> None:
+def test_batch_acquisition_finds_minimum() -> None:
     search_space = Box(tf.constant([-2.2, -1.0]), tf.constant([1.3, 3.3]))
-    expected_minimum = tf.constant([0., 0.])
     num_query_points = 4
     ego = BatchAcquisitionRule(
         num_query_points,
@@ -209,4 +208,5 @@ def test_batch_acquisition_returns_batches_of_right_size_and_finds_minimum() -> 
     query_point, _ = ego.acquire(
         search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
     )
+
     npt.assert_allclose(query_point, [[0., 0.]] * num_query_points, atol=1e-3)
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict
+from typing import Dict, Mapping
 
 import pytest
 import numpy.testing as npt
@@ -25,6 +25,7 @@ from trieste.acquisition.rule import (
     TrustRegion,
     OBJECTIVE,
     BatchAcquisitionRule,
+    BatchAcquisitionFunction,
 )
 from trieste.data import Dataset
 from trieste.models import ProbabilisticModel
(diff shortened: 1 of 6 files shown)


trieste/acquisition/rule.py
```diff
@@ -294,3 +294,61 @@ def acquire(
...
+        Return the batch of query points that optimizes the acquisition function produced by `builder` (see
+        :meth:`__init__`).
+        :param search_space: The global search space over which the optimization problem
+            is defined.
+        :param datasets: The known observer query points and observations.
+        :param models: The models of the specified ``datasets``.
+        :param state: Unused.
+        :return: The batch of points to query, and `None`.
+        """
+        expanded_search_space = search_space ** self._num_query_points
```
vpicheny, line 343: this generates the following error during tests: 
Unsupported operand types for ** ("SearchSpace" and "int")

I don't know how to resolve this, @joelberkeley any idea?

trieste/acquisition/rule.py
```diff
@@ -294,3 +294,61 @@ def acquire(
...
+
+        if not num_query_points > 0:
+            raise ValueError(
+                f"Number of query points must be greater than 0, got {num_query_points}"
+            )
+
+        self._num_query_points = num_query_points
+
+        if builder is None:
+            builder = ExpectedImprovement().using(OBJECTIVE)  # this is a placeholder for now
```
vpicheny: We do not have a default acquisition function for batches for now. This may be resolved as soon as we have the mc-EI.

tests/util/model.py
```diff
@@ -57,7 +57,7 @@ class QuadraticWithUnitVariance(GaussianMarginal):
     """
     def __init__(self):
         self._model = CustomMeanWithUnitVariance(
-            lambda x: tf.reduce_sum(x ** 2, axis=1, keepdims=True)
+            lambda x: tf.reduce_sum(x ** 2, axis=-1, keepdims=True)
```
vpicheny, line 60: the original function assumed a 2D tensor, while we need higher order ones

trieste/acquisition/rule.py
joelberkeley: that's cos `__pow__` is only defined on `Box`, not on all `SearchSpace`s. You can either restrict this rule to only work with `Box`es for now (with `class BatchAcquisitionRule(AcquisitionRule[None, Box]):` and
```
 def acquire(
        self,
        search_space: Box,
        ...
```
or you can implement `__pow__` on `DiscreteSearchSpace` too (probably in a separate PR)

trieste/acquisition/rule.py
joelberkeley: maybe leave this without a default for this PR?

trieste/acquisition/rule.py
joelberkeley: i could implement `__pow__` on `DiscreteSearchSpace` for you?

vpicheny: I'd got for the latter - if you have a clear idea how to implement pow for discrete spaces, go for it! I don't know how to do it efficiently.

trieste/acquisition/rule.py
vpicheny: OK - then should I replace those lines with a warning saying that for now the acq builder must be provided?

joelberkeley: nope, just change `Optional[AcquisitionFunctionBuilder] = None` to `AcquisitionFunctionBuilder` and remove the mention of the default from the docstring

trieste/acquisition/rule.py
```diff
@@ -294,3 +294,61 @@ def acquire(
         state_ = TrustRegion.State(acquisition_space, eps, y_min, is_global)
 
         return point, state_
+
+
+class BatchAcquisitionRule(AcquisitionRule[None, SearchSpace]):
+    """ Implements a acquisition rule for a batch of query points """
+
+    def __init__(self, num_query_points: int,
+                 builder: Optional[AcquisitionFunctionBuilder] = None):
```
joelberkeley: this won't work for non-batch `AcquisitionFunction`s, and therefore also for `AcquisitionFunctionBuilder`s. If we're going to keep batch and non-batch separate, I assumed we'll need a new `BatchAcquisitionFunctionBuilder` type alias and `BatchAcquisitionFunction` interface.

trieste/space.py
```diff
@@ -171,3 +172,25 @@ def discretize(self, num_samples: int) -> DiscreteSearchSpace:
...
+            return NotImplemented(
+                f"Bounds of both boxes must have the same dtype, got {self.lower.dtype} and"
+                f" {box.lower.dtype}"
+            )
+
+        expanded_lower_bound = tf.concat([self._lower, box.lower], axis=-1)
+        expanded_upper_bound = tf.concat([self._upper, box.upper], axis=-1)
+        return Box(expanded_lower_bound, expanded_upper_bound)
+
+    def __pow__(self, other: int):
```
joelberkeley:
```suggestion
    def __pow__(self, other: int) -> Box:
```

tests/util/model.py
vpicheny: you can resolve this if you are happy with this change

trieste/acquisition/rule.py
joelberkeley: you can copy-paste the `AcquisitionFunction` and `AcquisitionFunctionBuilder` definition, and just add a bit about the shapes for `AcquisitionFunction` and `BatchAcquisitionFunction`. No tests needed

vpicheny: could this be part of another PR?

joelberkeley: not really, cos you'll need a `BatchAcquisitionFunctionBuilder` to test this rule with

trieste/acquisition/rule.py
```diff
@@ -294,3 +296,55 @@ def acquire(
...
+    """ Implements a acquisition rule for a batch of query points """
+
+    def __init__(self, num_query_points: int,
+                 builder: BatchAcquisitionFunctionBuilder):
+        """
+        :param num_query_points: The number of points to acquire.
+        :param builder: The acquisition function builder to use.
+            :class:`BatchAcquisitionRule` will attempt to **maximise** the corresponding
+            acquisition function.
+            with tag `OBJECTIVE`.
```
joelberkeley:
```suggestion
```

trieste/acquisition/function.py
```diff
@@ -26,6 +26,13 @@
 
 AcquisitionFunction = Callable[[QueryPoints], tf.Tensor]
 """ Type alias for acquisition functions. """
+BatchAcquisitionFunction = Callable[[QueryPoints], tf.Tensor]
+""" 
+Type alias for batch acquisition functions. 
+
+While an AcquisitionFunction handles query points of shape [..., D] and returns [..., 1] values,
+BatchAcquisitionFunction handles batches of query points of shape [..., B, D] and returns [..., 1] values.
+"""
```
joelberkeley: yes, this is what I had in mind. You could even put the bit about `AcquisitionFunction`s in the `AcquisitionFunction` docstring

trieste/acquisition/rule.py
```diff
@@ -294,3 +296,55 @@ def acquire(
...
+        if not num_query_points > 0:
+            raise ValueError(
+                f"Number of query points must be greater than 0, got {num_query_points}"
+            )
+
+        self._num_query_points = num_query_points
+        self._builder = builder
+
+    def _vectorize_batch_acquisition(self, acquisition_function):
+        return lambda at: acquisition_function(tf.reshape(at, at.shape[:-1].as_list() + [self._num_query_points, -1]))
```
joelberkeley, line 323:
```suggestion
        return lambda at: acquisition_function(
            tf.reshape(at, at.shape[:-1].as_list() + [self._num_query_points, -1])
        )
```
so it fits on the page?

joelberkeley, line 324:
```suggestion
        return lambda at: acquisition_function(tf.reshape(at, at.shape[:-1] + [self._num_query_points, -1]))
```
you can add `TensorShape` and `list`

trieste/acquisition/rule.py
```diff
@@ -294,3 +296,55 @@ def acquire(
...
+
+        if not num_query_points > 0:
+            raise ValueError(
+                f"Number of query points must be greater than 0, got {num_query_points}"
+            )
+
+        self._num_query_points = num_query_points
+        self._builder = builder
+
+    def _vectorize_batch_acquisition(self, acquisition_function):
```
joelberkeley:
```suggestion
    def _vectorize_batch_acquisition(self, acquisition_function: AcquisitionFunction) -> BatchAcquisitionFunction:
```

trieste/acquisition/function.py
```diff
@@ -493,3 +500,17 @@ def prepare_acquisition_function(
         eta = tf.reduce_min(tf.boolean_mask(mean, is_feasible), axis=0)
 
         return lambda at: expected_improvement(objective_model, eta, at) * constraint_fn(at)
+
+
+class BatchAcquisitionFunctionBuilder(ABC):
+    """ An :class:`BatchAcquisitionFunctionBuilder` builds a batch acquisition function. """
```
```suggestion
    """ A :class:`BatchAcquisitionFunctionBuilder` builds a batch acquisition function. """
```

tests/unit/acquisition/test_rule.py
```diff
@@ -186,3 +190,33 @@ def test_trust_region_for_unsuccessful_local_to_global_trust_region_reduced() ->
     npt.assert_array_less(current_state.eps, previous_state.eps)  # current TR smaller than previous
     assert current_state.is_global
     npt.assert_array_almost_equal(current_state.acquisition_space.lower, lower_bound)
+
+
+class _BatchModelMinusMeanMaximumSingleBuilder(SingleModelAcquisitionBuilder):
+    def prepare_acquisition_function(
+        self, dataset: Dataset, model: ProbabilisticModel
+    ) -> AcquisitionFunction:
+        return lambda at: -tf.reduce_max(model.predict(at)[0], axis=-2)
```
joelberkeley: this will need to be a `BatchAcquisitionFunctionBuilder` rather than a `SingleModelAcquisitionBuilder`

trieste/space.py
```diff
@@ -44,6 +46,21 @@ def __contains__(self, value: TensorType) -> Union[bool, tf.Tensor]:
             from this :class:`SearchSpace`.
         """
 
+    @abstractmethod
+    def __mul__(self: Type[SP], other: SP) -> SP:
```
joelberkeley: i got this wrong when i first wrote it
```suggestion
    def __mul__(self: SP, other: SP) -> SP:
```

trieste/space.py
```diff
@@ -44,6 +46,21 @@ def __contains__(self, value: TensorType) -> Union[bool, tf.Tensor]:
...
+        """
+        TODO
+        :param other:
+        :return:
+        """
```
joelberkeley:
```suggestion
        """        
        :param other: A search space of the same type as this search space.
        :return: The Cartesian product of this search space with ``other``.
        """
```

trieste/space.py
```diff
@@ -44,6 +46,21 @@ def __contains__(self, value: TensorType) -> Union[bool, tf.Tensor]:
...
+    @abstractmethod
+    def __mul__(self: SP, other: SP) -> SP:
+        """
+        TODO
+        :param other:
+        :return:
+        """
+
+    @abstractmethod
+    def __pow__(self: Type[SP], other: int) -> SP:
```
joelberkeley:
```suggestion
    def __pow__(self: SP, other: int) -> SP:
        """
        Return the Cartesian product of ``other`` instances of this search space. For example, for an exponent of `3`, and search space `s`, this is `s ** 3`, which is equivalent to `s * s * s`.
        
        :param other: The number of instances of this search space to multiply. Must be strictly positive.
        :return: The Cartesian product of ``other`` instances of this search space.
        :raise ValueError: If the exponent ``other`` is less than 1.
        """
```

trieste/space.py
```diff
@@ -91,6 +108,24 @@ def sample(self, num_samples: int) -> tf.Tensor:
 
         return tf.random.shuffle(self._points)[:num_samples, :]
 
+    def __mul__(self, dss: DiscreteSearchSpace) -> DiscreteSearchSpace:
+        """
+        Return the Cartesian product of the two :class:`DiscreteSearchSpace`
```
joelberkeley:
```suggestion
        Return the Cartesian product of the two :class:`DiscreteSearchSpace`\ s. For example:
        
            >>> sa = DiscreteSearchSpace(tf.constant([[0, 1], [2, 3]]))
            >>> sb = DiscreteSearchSpace(tf.constant([[4, 5, 6], [7, 8, 9]]))
            >>> (sa * sb).points.numpy()
            array([[0, 1, 4, 5, 6],
                   [0, 1, 7, 8, 9],
                   [2, 3, 4, 5, 6],
                   [2, 3, 7, 8, 9]])

```
modulo some formatting

trieste/space.py
```diff
@@ -44,6 +46,21 @@ def __contains__(self, value: TensorType) -> Union[bool, tf.Tensor]:
...
+    def __mul__(self: SP, other: SP) -> SP:
+        """
+        TODO
+        :param other:
+        :return:
+        """
+
+    @abstractmethod
+    def __pow__(self: Type[SP], other: int) -> SP:
+        space = self
```
joelberkeley, line 67:
```suggestion
        if other < 1:
            raise ValueError("message here please")

        space = self
```

trieste/space.py
```diff
@@ -91,6 +108,24 @@ def sample(self, num_samples: int) -> tf.Tensor:
 
         return tf.random.shuffle(self._points)[:num_samples, :]
 
+    def __mul__(self, dss: DiscreteSearchSpace) -> DiscreteSearchSpace:
```
joelberkeley: mypy may error if this isn't
```suggestion
    def __mul__(self, other: DiscreteSearchSpace) -> DiscreteSearchSpace:
```

joelberkeley: and the same for Box

trieste/acquisition/rule.py
```diff
@@ -294,3 +296,54 @@ def acquire(
...
+
+        if not num_query_points > 0:
+            raise ValueError(
+                f"Number of query points must be greater than 0, got {num_query_points}"
+            )
+
+        self._num_query_points = num_query_points
+        self._builder = builder
+
+    def _vectorize_batch_acquisition(self, acquisition_function: AcquisitionFunction) -> BatchAcquisitionFunction:
```
joelberkeley: wrong way round, sorry
```suggestion
    def _vectorize_batch_acquisition(self, acquisition_function: BatchAcquisitionFunction) -> AcquisitionFunction:
```

trieste/acquisition/function.py
```diff
@@ -493,3 +502,17 @@ def prepare_acquisition_function(
...
         return lambda at: expected_improvement(objective_model, eta, at) * constraint_fn(at)
+
+
+class BatchAcquisitionFunctionBuilder(ABC):
+    """ A :class:`BatchAcquisitionFunctionBuilder` builds a batch acquisition function. """
+
+    @abstractmethod
+    def prepare_acquisition_function(
+        self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
+    ) -> AcquisitionFunction:
```
```suggestion
    ) -> BatchAcquisitionFunction:
```

trieste/space.py
```diff
@@ -44,6 +46,29 @@ def __contains__(self, value: TensorType) -> Union[bool, tf.Tensor]:
...
+    def __pow__(self: SP, other: int) -> SP:
+        """
+        Return the Cartesian product of ``other`` instances of this search space. For example, for an exponent of `3`, and search space `s`, this is `s ** 3`, which is equivalent to `s * s * s`.
+        
+        :param other: The number of instances of this search space to multiply. Must be strictly positive.
+        :return: The Cartesian product of ``other`` instances of this search space.
+        :raise ValueError: If the exponent ``other`` is less than 1.
+        """
+        if other < 1:
+            raise ValueError("message here please")
```
joelberkeley: message here please

tests/util/model.py
joelberkeley: i _think_ this is right, and I hope so, cos a lot of tests use it

trieste/acquisition/rule.py
```diff
@@ -294,3 +298,54 @@ def acquire(
...
+        if not num_query_points > 0:
+            raise ValueError(
+                f"Number of query points must be greater than 0, got {num_query_points}"
+            )
+
+        self._num_query_points = num_query_points
+        self._builder = builder
+
+    def _vectorize_batch_acquisition(self, acquisition_function: BatchAcquisitionFunction) -> AcquisitionFunction:
+        return lambda at: acquisition_function(tf.reshape(at, at.shape[:-1].as_list() + [self._num_query_points, -1]))
```
joelberkeley, line 324: not something we need to change here, but I think this is an example where higher-rank points won't work, which illustrates just how tricky it is to guarantee we support such cases

tests/unit/acquisition/test_rule.py
```diff
@@ -186,3 +189,37 @@ def test_trust_region_for_unsuccessful_local_to_global_trust_region_reduced() ->
     npt.assert_array_less(current_state.eps, previous_state.eps)  # current TR smaller than previous
     assert current_state.is_global
     npt.assert_array_almost_equal(current_state.acquisition_space.lower, lower_bound)
+
+
+class _BatchModelMinusMeanMaximumSingleBuilder(BatchAcquisitionFunctionBuilder):
+    def prepare_acquisition_function(
+        self, dataset: Mapping[str, Dataset], model: Mapping[str, ProbabilisticModel]
+    ) -> BatchAcquisitionFunction:
+        return lambda at: -tf.reduce_max(model[OBJECTIVE].predict(at)[0], axis=-2)
```
joelberkeley, line 198: does this take the negative of the maximum predictive mean of the batch? does it assume the optimizer only queries one point at a time i.e. that `at` is `[1, B, D]`?

tests/unit/acquisition/test_rule.py
```diff
@@ -186,3 +189,37 @@ def test_trust_region_for_unsuccessful_local_to_global_trust_region_reduced() ->
...
+    expected_minimum = tf.constant([0., 0.])
+    num_query_points = 4
+    ego = BatchAcquisitionRule(
+        num_query_points,
+        _BatchModelMinusMeanMaximumSingleBuilder())
+    dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
+    query_point, _ = ego.acquire(
+        search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
+    )
+    npt.assert_allclose(query_point - expected_minimum[None, :], 0., atol=1e-3)
```
joelberkeley, line 225: If you don't use broadcasting in `assert_allclose`, you can merge these two tests into one
```suggestion
def test_batch_acquisition_finds_minimum() -> None:
    search_space = Box(tf.constant([-2.2, -1.0]), tf.constant([1.3, 3.3]))
    num_query_points = 
    ego = BatchAcquisitionRule(
        num_query_points,
        _BatchModelMinusMeanMaximumSingleBuilder())
    dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
    query_point, _ = ego.acquire(
        search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
    )
    npt.assert_allclose(query_point, [[0., 0.]] * num_query_points, atol=1e-3)
```

joelberkeley: if you don't use broadcasting in `assert_allclose`, you can merge these two tests into one
```suggestion
    npt.assert_allclose(query_point, [[0., 0.]] * num_query_points, atol=1e-3)
```

tests/unit/acquisition/test_rule.py
vpicheny: yes, that's what it returns.
the maximum predictive mean is the "worst" mean over the batch (when you minimize) so that maximising this acquisition will result with a batch where all points have a low predictive mean.
This works for at is [N, B, D]. Predict returns [N, B, 1] and we reduce along the -2 axis, which delivers [N, 1].

joelberkeley: i'm confused because you're indexing at 0, which I would have thought would index N

tests/unit/acquisition/test_rule.py
```diff
@@ -186,3 +189,37 @@ def test_trust_region_for_unsuccessful_local_to_global_trust_region_reduced() ->
...
+    search_space = Box(tf.constant([-2.2, -1.0]), tf.constant([1.3, 3.3]))
+    num_query_points = 3
+    ego = BatchAcquisitionRule(
+        num_query_points,
+        _BatchModelMinusMeanMaximumSingleBuilder())
+    dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
+    query_point, _ = ego.acquire(
+        search_space, {OBJECTIVE: dataset}, {OBJECTIVE: QuadraticWithUnitVariance()}
+    )
+    assert query_point.shape == [num_query_points, 2]
```
joelberkeley, line 211: this test is redundant now you've changed the `assert_allclose` in the other

tests/unit/acquisition/test_rule.py
vpicheny: [0] is to the get the mean only and not the variance. axis=-2 squashes the B dimension.

tests/unit/acquisition/test_rule.py
vpicheny: this looks more like a unit test than the other one. Why not keep both? or add `assert query_point.shape == [num_query_points, 2]` to the second one? or would `assert_allclose` let us know whereas it's a value problem or a shape problem?

joelberkeley: `assert_allclose` does shape checks too
```
>>> np.testing.assert_allclose(tf.constant([[1, 1], [1, 1]]), [[1, 1]])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/joel.berkeley/.venv/trieste/joelberkeley/dev/lib/python3.7/site-packages/numpy/testing/_private/utils.py", line 1533, in assert_allclose
    verbose=verbose, header=header, equal_nan=equal_nan)
  File "/home/joel.berkeley/.venv/trieste/joelberkeley/dev/lib/python3.7/site-packages/numpy/testing/_private/utils.py", line 765, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

(shapes (2, 2), (1, 2) mismatch)
 x: array([[1, 1],
       [1, 1]], dtype=int32)
 y: array([[1, 1]])
```

tests/unit/acquisition/test_rule.py
joelberkeley: oh ok, i missed that

joelberkeley (approved):

PR #72, opened 2020-11-27
Changed to MinValueEntropySearch

henrymoss:
Hi 
I have updated MES to be referred to as MinValueEntropySearch, as that is what we are actually doing.

I also made all the recommended changes from my Gumbel sampler bug fix from a couple of weeks back. I thought I had merged that pull request, but the Gumbel sampler changes seemed not to make it into the repo.

Thanks

Henry


trieste/acquisition/function.py
```diff
@@ -137,9 +137,9 @@ def expected_improvement(model: ProbabilisticModel, eta: tf.Tensor, at: QueryPoi
     return (eta - mean) * normal.cdf(eta) + variance * normal.prob(eta)
 
 
-class MaxValueEntropySearch(SingleModelAcquisitionBuilder):
+class MinValueEntropySearch(SingleModelAcquisitionBuilder):
     """
-    Builder for the max-value entropy search acquisition function (for function minimisation)
+    Builder for the min-value entropy search acquisition function (for function minimisation)
```
joelberkeley, line 142: can you add something explaining this is the Max value entropy search algorithm modified for function minimisation?

PR #74, opened 2020-11-27
use tensorflow probability lbfgs optimizer

### LINKED ISSUE #117: Investigate performance of acquisition optimizer
As a user, I want the acquisition optimizer to be as efficient as possible. We currently use gpflow's Scipy optimizer with tfp's bijector for bounds. However, @joelberkeley has found in #74 that tfp's `lbfgs_minimize` takes fewer steps (though each of those steps take longer). We may alternatively be able to use gpflow's Scipy with bounds since it is L-BFGS-B. This may be faster than using the tfp bijector.

Task is to investigate the performance impact of each option and implement the "best" (having decided what "best" means).

joelberkeley:
This PR changes the acquisition optimizer from scipy's L-BFGS-B to TensorFlow Probability's L-BFGS. This changes the performance of the integration tests, and thus presumably trieste in general (individual BO steps take longer, but fewer steps are needed). This also has the benefit of moving down the dependency tree.

Q: We already map the constrained optimization problem to an unconstrained one (using tfp bijectors), so L-BFGS-B is unnecessary and L-BFGS will suffice. Is this true? If not, do we indeed need the -B version?

### DIFF
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -35,8 +35,8 @@ from trieste.utils.objectives import BRANIN_MINIMIZERS, BRANIN_MINIMUM, branin, 
 @pytest.mark.parametrize(
     "num_steps, acquisition_rule",
     [
-        (20, EfficientGlobalOptimization()),
-        (15, TrustRegion()),
+        (19, EfficientGlobalOptimization()),
+        (12, TrustRegion()),
         (17, ThompsonSampling(500, 3)),
     ],
 )
--- a/trieste/acquisition/optimizer.py
+++ b/trieste/acquisition/optimizer.py
@@ -20,9 +20,9 @@ from __future__ import annotations
 from functools import singledispatch
 from typing import Callable, TypeVar
 
-import gpflow
 import tensorflow as tf
 import tensorflow_probability as tfp
+from absl import logging
 
 from ..space import Box, DiscreteSearchSpace, SearchSpace
 from ..type import TensorType
@@ -70,11 +70,21 @@ def _box(space: Box, target_func: AcquisitionFunction) -> TensorType:
     initial_point = optimize(trial_search_space, target_func)
 
     bijector = tfp.bijectors.Sigmoid(low=space.lower, high=space.upper)
-    variable = tf.Variable(bijector.inverse(initial_point))
 
-    def _objective() -> TensorType:
-        return -target_func(bijector.forward(variable))
+    def _objective(x: TensorType) -> TensorType:
+        return tf.squeeze(-target_func(bijector.forward(x)), axis=-1)
 
-    gpflow.optimizers.Scipy().minimize(_objective, (variable,))
+    lbfgs_result = tfp.optimizer.lbfgs_minimize(
+        lambda x: tfp.math.value_and_gradient(_objective, x),
+        # we copy the tensor to work around a bug in tensorflow_probability,
+        #   see https://github.com/tensorflow/probability/issues/1182
+        tf.identity(bijector.inverse(initial_point)),
+    )
 
-    return bijector.forward(variable)
+    if not lbfgs_result.converged:
+        tf.print(
+            "L-BFGS optimization produced result, but failed to converge.",
+            output_stream=logging.WARN,
+        )
+
+    return bijector.forward(lbfgs_result.position)


trieste/acquisition/_optimizer.py
```diff
@@ -62,11 +62,16 @@ def _box(space: Box, target_func: TensorMapping) -> QueryPoints:
...
+        return -target_func(bijector.forward(x))[..., 0]
 
-    gpflow.optimizers.Scipy().minimize(_objective, (variable,))
+    lbfgs_result = tfp.optimizer.lbfgs_minimize(
+        lambda x: tfp.math.value_and_gradient(_objective, x),
+        tf.Variable(bijector.inverse(initial_point))
+    )
+
+    if not lbfgs_result.converged:
+        raise RuntimeError("L-BFGS optimization failed to converge.")
```
joelberkeley: should we raise if it didn't converge?

trieste/acquisition/_optimizer.py
```diff
@@ -62,11 +63,16 @@ def _box(space: Box, target_func: TensorMapping) -> QueryPoints:
     initial_point = optimize(trial_search_space, target_func)
 
     bijector = tfp.bijectors.Sigmoid(low=space.lower, high=space.upper)
-    variable = tf.Variable(bijector.inverse(initial_point))
 
-    def _objective() -> tf.Tensor:
-        return -target_func(bijector.forward(variable))
+    def _objective(x: tf.Variable) -> tf.Tensor:
+        return -target_func(bijector.forward(x))[..., 0]
```
st--: If you're expecting the last dimension to be 1, the following makes that assumption clear in code:
```suggestion
        return -tf.squeeze(target_func(bijector.forward(x)), axis=-1)
```

trieste/acquisition/_optimizer.py
joelberkeley: apparently not, because points can still be good points even if optimization didn't converge

trieste/acquisition/_optimizer.py
```diff
@@ -62,11 +63,19 @@ def _box(space: Box, target_func: TensorMapping) -> QueryPoints:
...
 
-    def _objective() -> tf.Tensor:
-        return -target_func(bijector.forward(variable))
+    def _objective(x: tf.Variable) -> tf.Tensor:
+        return tf.squeeze(-target_func(bijector.forward(x)), axis=-1)
 
-    gpflow.optimizers.Scipy().minimize(_objective, (variable,))
+    lbfgs_result = tfp.optimizer.lbfgs_minimize(
+        lambda x: tfp.math.value_and_gradient(_objective, x),
+        tf.Variable(bijector.inverse(initial_point)),
```
joelberkeley, line 72: that this needs to be a `Variable` is a bug in tfp afaict, see https://github.com/tensorflow/probability/issues/1182

trieste/acquisition/_optimizer.py
joelberkeley: yeah while the original is slightly more readable imo, `tf.squeeze` is safer e.g. in case the rest of the code changes

trieste/acquisition/_optimizer.py
```diff
@@ -62,11 +63,21 @@ def _box(space: Box, target_func: TensorMapping) -> QueryPoints:
...
+    def _objective(x: tf.Tensor) -> tf.Tensor:
+        return tf.squeeze(-target_func(bijector.forward(x)), axis=-1)
 
-    gpflow.optimizers.Scipy().minimize(_objective, (variable,))
+    lbfgs_result = tfp.optimizer.lbfgs_minimize(
+        lambda x: tfp.math.value_and_gradient(_objective, x),
+        # we stop the gradient computation to work around this bug in tensorflow_probability
+        # https://github.com/tensorflow/probability/issues/1182
+        tf.stop_gradient(bijector.inverse(initial_point)),
+    )
```
joelberkeley-secondmind, line 75: todo work out if this is actually a bug, or if whether the `lbfgs_optimize` docs are misleading

trieste/acquisition/_optimizer.py
```diff
@@ -62,11 +63,21 @@ def _box(space: Box, target_func: TensorMapping) -> QueryPoints:
...
-        return -target_func(bijector.forward(variable))
+    def _objective(x: tf.Tensor) -> tf.Tensor:
+        return tf.squeeze(-target_func(bijector.forward(x)), axis=-1)
 
-    gpflow.optimizers.Scipy().minimize(_objective, (variable,))
+    lbfgs_result = tfp.optimizer.lbfgs_minimize(
+        lambda x: tfp.math.value_and_gradient(_objective, x),
+        # we copy the tensor to work around a bug in tensorflow_probability,
+        #   see https://github.com/tensorflow/probability/issues/1182
+        tf.identity(bijector.inverse(initial_point)),
```
joelberkeley: note that while we're now explicitly working around a bug, the bug is not new in this PR (the bug is in tfp bijectors). It just wasn't breaking any tests before

joelberkeley:
**Very** rough performance tests suggest tfp optimizer is much slower. Until a proper perf check suggests otherwise, this change isn't useful

vpicheny:
For the record: we are NOT using L-BFGS-B at the moment, only L-BFGS.

PR #77, opened 2020-11-28
expand acquisition function type alias docstrings

joelberkeley:


vpicheny (approved):
those docstrings are very nice

PR #80, opened 2020-11-30
add search space tests; minor refactors

joelberkeley:


tests/unit/test_space.py
```diff
@@ -21,6 +22,28 @@
...
+    def limit(self) -> tf.Tensor:
+        return self._limit
+
+    def sample(self, num_samples: int) -> tf.Tensor:
+        return tf.random.uniform([num_samples], maxval=self._limit)
+
+    def __contains__(self, point: tf.Tensor) -> tf.Tensor:
+        return 0 <= point < self._limit
+
+    def __mul__(self, other: Line) -> Line:
```
awav: What does it mean to multiply a line with a line?

tests/unit/test_space.py
```diff
@@ -21,6 +22,28 @@
 from trieste.space import SearchSpace, DiscreteSearchSpace, Box
 
 
+class Line(SearchSpace):
+    def __init__(self, limit: float):
+        self._limit = tf.constant(limit)
```
awav: `tf.constant` is unnecessary, unless you specify a type

tests/unit/test_space.py
```diff
@@ -21,6 +22,28 @@
...
+
+    def __contains__(self, point: tf.Tensor) -> tf.Tensor:
+        return 0 <= point < self._limit
+
+    def __mul__(self, other: Line) -> Line:
+        return Line(self.limit * other.limit)
+
+
+def test_search_space___pow__() -> None:
+    assert (Line(5.0) ** 3).limit == 5 ** 3
```
awav: What are you testing here?

tests/unit/test_space.py
```diff
@@ -21,6 +22,28 @@
 from trieste.space import SearchSpace, DiscreteSearchSpace, Box
 
 
+class Line(SearchSpace):
```
awav: What is Line? Do you mean R space?

trieste/space.py
```diff
@@ -128,20 +134,22 @@ def __mul__(self, other: DiscreteSearchSpace) -> DiscreteSearchSpace:
...
+        if self._points.dtype is not other.points.dtype:
             return NotImplemented
 
-        N = self._points.shape[0]
-        M = other._points.shape[0]
-        tile_self = tf.tile(tf.expand_dims(self.points, 1), [1, M, 1])
-        tile_dss = tf.tile(tf.expand_dims(other.points, 0), [N, 1, 1])
-        cartesian_product = tf.concat([tile_self, tile_dss], axis=2)
+        dim_self = self._points.shape[-1]
+        dim_other = other.points.shape[-1]
```
awav: I'd prefer keeping N, let's say N0 and N1, from a context it is clear what you mean by them.

trieste/space.py
```diff
@@ -128,20 +134,22 @@ def __mul__(self, other: DiscreteSearchSpace) -> DiscreteSearchSpace:
...
+        tile_self = tf.tile(tf.expand_dims(self.points, 1), [1, len(other.points), 1])
+        tile_other = tf.tile(tf.expand_dims(other.points, 0), [len(self._points), 1, 1])
```
awav:
```suggestion
        tile_self = tf.tile(self.points[:, None], [1, other.points.shape[0], 1])
        tile_other = tf.tile(other.points[None], [self._points.shape[0], 1, 1])
```

tests/unit/test_space.py
```diff
@@ -99,6 +122,51 @@ def test_discrete_search_space_sampling_raises_when_too_many_samples_are_request
         search_space.sample(num_samples)
 
 
+def test_discrete_search_space___mul___points_is_the_concatenation_of_original_points() -> None:
```
awav, line 134: Random thought: I remember we had a discussion about tests naming, but I don't remember what was the conclusion. I find it extremely difficult to read functions with long names, in the current case, wouldn't it better to have a short docstring for describing what function does instead of the 100 char function?

tests/unit/test_space.py
```diff
@@ -99,6 +122,51 @@ def test_discrete_search_space_sampling_raises_when_too_many_samples_are_request
...
+    dss1 = DiscreteSearchSpace(tf.constant([[-1.0, -1.4], [-1.5, -3.6], [-0.5, -0.6]]))
+    dss2 = DiscreteSearchSpace(tf.constant([[1.0, 1.4], [1.5, 3.6]]))
+
+    product = dss1 * dss2
+
+    all_expected_points = tf.constant(
```
awav, line 140: You can use np.ndarray, and np.testing.assert_equal instead of custom `assert any(for ... in` loop.

tests/unit/test_space.py
joelberkeley-secondmind: indeed

tests/unit/test_space.py
joelberkeley-secondmind: cartesian product, so some part of R^2 (or a line isomorphic to that space)

tests/unit/test_space.py
joelberkeley-secondmind: i've actually slightly forgotten how this works. Something like a line with limit 10 is isomorphic to a line of limit 2 times a line of limit 5. I'll document it when I remember

tests/unit/test_space.py
joelberkeley-secondmind: that a line cubed multiplies three lines together (not two or four)

tests/unit/test_space.py
joelberkeley-secondmind: in most cases, yes, but there's no guarantee of ordering here so all we can say is that each point exists in the other tensor

trieste/space.py
joelberkeley-secondmind: hmmm ... i changed it from that because i didn't find it clear. I'll look for a way that has the best of both

trieste/space.py
joelberkeley-secondmind: why so?

tests/unit/test_space.py
joelberkeley-secondmind: i don't know. it's not really part of this PR though so do you want to make a ticket for it?

tests/unit/test_space.py
joelberkeley-secondmind: someone might down the line want to do `line.limit * [1, 2, 3]`, which would, much to their confusion, give `[1, 2, 3, 1, 2, 3, 1, 2, 3]` rather than `tf.Tensor([3, 6, 9])`

converting to `tf.Tensor` removes any ambiguity

joelberkeley-secondmind:
but first I should say @awav that this PR is incomplete, so it may be best to hold off on an in-depth review for now

tests/unit/test_space.py
joelberkeley: my main concern with using docstrings for tests is them becoming an unmaintainable mess. It's pretty difficult keeping on top of docstrings for the source code

tests/unit/test_space.py
joelberkeley: OK, I don't know if Line did make sense, but I've change it now to Integers, which represents the discrete search space on a line of the integers. The product is simply the integer points on a grid, and so on. It uses the fact that the product of two sets of integers is isomorphic to set of integers whose size is the product of the size of the two sets. i.e. (1, 2) x (1, 2, 3) = (1, 1), (1, 2) ... (2, 3) ~ 1, 2, 3, 4, 5, 6

trieste/space.py
joelberkeley: i've removed them and just named their product

joelberkeley:
@awav ok this PR's ready now

joelberkeley:
@vpicheny test

trieste/space.py
```diff
@@ -135,44 +139,47 @@ def __mul__(self, other: DiscreteSearchSpace) -> DiscreteSearchSpace:
...
             return NotImplemented
 
-        N = self._points.shape[0]
-        M = other._points.shape[0]
-        tile_self = tf.tile(tf.expand_dims(self.points, 1), [1, M, 1])
-        tile_dss = tf.tile(tf.expand_dims(other.points, 0), [N, 1, 1])
-        cartesian_product = tf.concat([tile_self, tile_dss], axis=2)
-
-        return DiscreteSearchSpace(tf.reshape(cartesian_product, [N * M, -1]))
+        tile_self = tf.tile(self.points[:, None], [1, len(other.points), 1])
```
vpicheny, line 151: this works only with rank 2 tensors; are we happy with this / asserting the ranks in the first place?

tests/unit/test_space.py
```diff
@@ -11,18 +11,47 @@
...
 
+class Integers(SearchSpace):
+    def __init__(self, exclusive_limit: int):
+        assert exclusive_limit > 0
+        self.limit: Final[int] = exclusive_limit
+
+    def sample(self, num_samples: int) -> tf.Tensor:
+        return tf.random.shuffle(tf.range(self.limit))[0]
+
+    def __contains__(self, point: tf.Tensor) -> tf.Tensor:
```
vpicheny, line 37: what happens if you test it with point=0.1?

tests/unit/test_space.py
```diff
@@ -11,18 +11,47 @@
...
 import numpy.testing as npt
 import pytest
 import tensorflow as tf
+from typing_extensions import Final
 
 from tests.util.misc import TF_DEBUGGING_ERROR_TYPES, ShapeLike, various_shapes
 from trieste.space import Box, DiscreteSearchSpace, SearchSpace
 
 
+class Integers(SearchSpace):
```
vpicheny, line 29: I actually don't understand why we need this extra space for our tests.

joelberkeley: `__pow__` is defined on `SearchSpace`, so this lets us know that `__pow__` will work correctly if someone subclasses `SearchSpace` instead of using dss or box

trieste/space.py
joelberkeley: the class `__init__` asserts rank 2

joelberkeley: (which I'll mention isn't a long term plan - just until we've made it all work with higher rank)

tests/unit/test_space.py
joelberkeley: good catch. probably not the right thing

joelberkeley: ok, fixed I think

vpicheny (approved):

PR #93, opened 2020-12-03
add support for batch monte carlo acquisition

### LINKED ISSUE #40: MC-based acquisition function supporting reparametrization trick

joelberkeley-secondmind:


trieste/acquisition/function.py
```diff
@@ -36,66 +40,59 @@
...
-class AcquisitionFunctionBuilder(ABC):
-    """ An :class:`AcquisitionFunctionBuilder` builds an acquisition function. """
+class Builder(ABC, Generic[F]):
+    """ Generic builder of things from data and models. """
 
     @abstractmethod
-    def prepare_acquisition_function(
-        self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
-    ) -> AcquisitionFunction:
+    def build(self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]) -> F:
```
joelberkeley-secondmind: i remember we changed this from `build` to `prepare_acquisition_function` a while back. This refactor means the `_acquisition_function` bit no longer makes sense, and `prepare` was never really the right word anyway. Can discuss other names if you like

trieste/acquisition/function.py
```diff
@@ -512,15 +501,158 @@ def prepare_acquisition_function(
...
+
+    def sample(at: tf.Tensor) -> tf.Tensor:
+        # todo do we need this shape check?
+        tf.debugging.assert_shapes([(at, [..., sample_size, batch_size, '_'])])
+
+        # todo check this is the same as what victor wrote
+        mean, cov = model.predict(at, full_cov=True)  # [N, B, L], [N, L, B, B]
+        identity = tf.eye(batch_size, dtype=cov.dtype)  # [B, B]
+        cov_cholesky = tf.linalg.cholesky(cov + jitter * identity)  # [N, L, B, B]
+        return mean + tf.transpose(cov_cholesky @ eps, [3, 0, 2, 1])  # [S, N, B, L]
```
vpicheny: I noticed you changed the computations here. Are you certain the math is still right? the previous version was taken from gpflow

trieste/acquisition/function.py
```diff
@@ -512,15 +501,158 @@ def prepare_acquisition_function(
...
+        mean, _ = model.predict(dataset.query_points)
+        eta = tf.reduce_min(mean, axis=0)
+
+        sampler = batch_latent_sampler_with_reparametrisation(
+            dataset,
+            model,
+            self._num_samples,
+            self._num_query_points,
+            # todo this feels like a hack (that we're pretending to support higher-rank)
+            dataset.observations.shape[-1:]
```
vpicheny: This is incorrect I think. We are looking for the number of latent models; we can have a single output but multiple latents. This is the case of the HetGP model (one GP for the mean, another for the variance).

trieste/acquisition/function.py
```diff
@@ -512,15 +501,158 @@ def prepare_acquisition_function(
...
+        )  # [S, N, B, L] (deferred)
+
+        def batch_ei(at: QueryPoints) -> tf.Tensor:
+            samples = tf.squeeze(sampler(at), axis=-1)  # [S, N, B]
+            improvement = tf.maximum(eta - samples, 0.0)  # [S, N, B]
+            batch_improvement = tf.reduce_max(improvement, axis=-1)  # [S, N]
+            ei = tf.reduce_mean(batch_improvement, axis=0, keepdims=True)  # [1, N]
+            return tf.transpose(ei)  # [N, 1]  # todo transpose
+
+        return batch_ei
```
vpicheny: sorry for the stupid question, but previously `prepare_acquisition_function` returned a callable function (at: lambda ...). How is this still valid?

trieste/acquisition/function.py
vpicheny: it feels to me that we should restrict this `sampler` to `GPflowPredictor`, and update this class to register the number of latent models

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -161,6 +169,127 @@ def create_bo_model(data):
 )
 plt.show()
 
+# %% [markdown]
+# Now, we will show how to solve the same problem, but when point are queried in (synchronous) batches instead of one at a time.
+
+# %%
+
+
+class BatchExpectedConstrainedImprovement(BatchAcquisitionFunctionBuilder):
```
vpicheny: I guess this still needs updating

trieste/acquisition/function.py
```diff
@@ -512,15 +501,158 @@ def prepare_acquisition_function(
...
+        mean, cov = model.predict(at)   # [N, L], [N, L]
+        return mean + tf.math.sqrt(cov) * eps[:, None, :]  # [S, N, L]
+
+    return sample
+
+
+# todo the batch size problem
+def batch_latent_sampler_with_reparametrisation(
+    dataset: Dataset,
+    model: ProbabilisticModel,
```
vpicheny: how about the case of multiple models? cf inequality notebook

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -161,6 +169,127 @@ def create_bo_model(data):
...
+        pof = constraint_fn(objective_dataset.query_points)
+        is_feasible = pof >= self._min_feasibility_probability
+
+        if not tf.reduce_any(is_feasible):
+            return constraint_fn
+
+        mean, _ = objective_model.predict(objective_dataset.query_points)
+        eta = tf.reduce_min(tf.boolean_mask(mean, is_feasible), axis=0)
+
+        eps = {tag:
```
vpicheny: writing this loop here is probably something we want to avoid. Can we find a way to include it in the `sampler`?

trieste/acquisition/function.py
joelberkeley-secondmind: `batch_ei` is a function (or callable) that takes a tensor as input and returns a tensor as output. `lambda x: batch_ei(x)` is another (longer) way of writing it. `lamda x: ...` is really just an unnamed function with some syntactical limitations

vpicheny: oh right, sorry that was indeed a stupid question

trieste/acquisition/function.py
joelberkeley-secondmind: restricting to `GPflowPredictor` is a significant design change that I want to avoid for a number of reasons

we should chat about this

docs/notebooks/inequality_constraints.pct.py
joelberkeley-secondmind: yes

trieste/acquisition/function.py
```diff
@@ -512,15 +501,158 @@ def prepare_acquisition_function(
...
+            samples = tf.squeeze(sampler(at), axis=-1)  # [S, N, B]
+            improvement = tf.maximum(eta - samples, 0.0)  # [S, N, B]
+            batch_improvement = tf.reduce_max(improvement, axis=-1)  # [S, N]
+            ei = tf.reduce_mean(batch_improvement, axis=0, keepdims=True)  # [1, N]
+            return tf.transpose(ei)  # [N, 1]  # todo transpose
+
+        return batch_ei
+
+
+class MonteCarloExpectedImprovement(SingleModelBuilder[AcquisitionFunction]):
```
vpicheny: move this to the montecarlo notebook

trieste/acquisition/function.py
```diff
@@ -512,15 +501,158 @@ def prepare_acquisition_function(
...
+        def acquisition_function(at: tf.Tensor) -> tf.Tensor:
+            samples = tf.squeeze(sampler(at), axis=-1)  # [S, N]
+            improvement = tf.math.maximum(eta - samples, 0.)  # [S, N]
+            ei = tf.math.reduce_mean(improvement, axis=0, keepdims=True)  # [1, N]
+            return tf.transpose(ei)  # [N, 1]  # todo transpose
+
+        return acquisition_function
+
+
+class MonteCarloNegativeLowerConfidenceBound(SingleModelBuilder[AcquisitionFunction]):
```
vpicheny: remove this now

docs/notebooks/inequality_constraints.pct.py
joelberkeley-secondmind: seems like the same as https://github.com/secondmind-labs/trieste/pull/93#discussion_r536160461

trieste/acquisition/function.py
joelberkeley-secondmind: agreed to do in another PR

trieste/acquisition/function.py
```diff
@@ -528,15 +509,134 @@ def prepare_acquisition_function(
...
+            shape=[self._target_dimension, None, sample_size],
+        )
+        self._jitter = jitter
+
+    def sample(self, at: tf.Tensor) -> tf.Tensor:
+        mean, cov = self._model.predict_full_cov(at)  # [N, B, L], [N, L, B, B]
+
+        batch_size = mean.shape[-1]
+
+        if tf.size(self._eps) == 0:
```
joelberkeley-secondmind, line 599: todo this logic assumes target_dimension, batch_size and sample_size are all non-zero, which is a reasonable assumption, but we should add error checking for it

docs/notebooks/montecarlo.pct.py
```diff
@@ -0,0 +1,210 @@
...
+    gpr = gpflow.models.GPR(astuple(data), kernel, noise_variance=1e-5)
+    set_trainable(gpr.likelihood, False)
+
+    return {OBJECTIVE: trieste.models.create_model({
+        "model": gpr,
+        "optimizer": gpflow.optimizers.Scipy(),
+        "optimizer_args": {"options": dict(maxiter=100)},
+    })}
+
+model = build_model(initial_data[OBJECTIVE])
```
vpicheny: move all of the above to utils?

docs/notebooks/montecarlo.pct.py
```diff
@@ -0,0 +1,210 @@
...
+
+# %% [markdown]
+# # Try an alternative acquisition function
```
vpicheny: Change title

docs/notebooks/montecarlo.pct.py
```diff
@@ -0,0 +1,210 @@
...
+
+# %% [markdown]
+# By default, Trieste uses Expected Improvement (EI) as its acquisition function when performing Bayesian optimization. However, many alternative acqusition functions have been developed. One such alternative is Max-value Entropy Search (MES), which approximates the distribution of current estimate of the global minimum and tries to decrease its entropy with each optimization step.
```
vpicheny: Change text

docs/notebooks/montecarlo.pct.py
```diff
@@ -0,0 +1,210 @@
...
+
+# %% [markdown]
+# We plot these two acquisition functions across our search space. Areas with high acquisition function scores (i.e bright regions) are those rated as promising locations for the next evaluation of our objective function. We see that EI wishes to continue exploring the search space, whereas MES wants to focus resources on evaluating a specific region.
```
vpicheny: Change text. The important points are : 
- both surfaces look similar
- the MC-EI is smooth because we use the reparametrization trick

joelberkeley: we're not including the notebooks in this PR are we? It will take longer if we do

vpicheny: also, this paragraph is misplaced, it should appear later. We need a paragraph here that explains that we create a MC-EI acquisition for the sake of demonstrating how to do it

docs/notebooks/montecarlo.pct.py
```diff
@@ -0,0 +1,210 @@
...
+
+plot_bo_points(
+    dataset.query_points.numpy(),
+    ax=ax[0, 0],
+    num_init=len(initial_query_points),
+    idx_best=arg_min_idx,
+)
+
+# %% [markdown]
+# We now re-initialize the model, define a new batch acquisition rule and run the whole BO loop for 10 steps.
```
vpicheny: add here that batch-EI is not analytical and we need MC

docs/notebooks/montecarlo.pct.py
```diff
@@ -0,0 +1,210 @@
...
+        def batch_ei(at):
+            samples = tf.squeeze(sampler.sample(at), axis=-1)  # [S, N, B]
+            improvement = tf.maximum(eta - samples, 0.0)  # [S, N, B]
+            batch_improvement = tf.reduce_max(improvement, axis=-1)  # [S, N]
+            ei = tf.reduce_mean(batch_improvement, axis=0, keepdims=True)  # [1, N]
+            return tf.transpose(ei)  # [N, 1]  # todo transpose
+
+        return batch_ei
+
+# %%
```
vpicheny: explain that points will be queried 3 at a time

docs/notebooks/montecarlo.pct.py
```diff
@@ -0,0 +1,210 @@
...
+bmcei = BatchMonteCarloExpectedImprovement().using(OBJECTIVE)
+batch_acq_rule = trieste.acquisition.rule.BatchAcquisitionRule(3, bmcei)
+bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+
+result = bo.optimize(15, initial_data, model, acquisition_rule=batch_acq_rule)
+
+if result.error is not None: raise result.error
+
+dataset = result.datasets[OBJECTIVE]
+
```
vpicheny: say that results look roughly like the one-at-a-time

trieste/models/model_interfaces.py
```diff
@@ -37,6 +37,17 @@ def predict(self, query_points: QueryPoints) -> Tuple[ObserverEvaluations, Tenso
...
+    @abstractmethod
+    def predict_full_cov(self, query_points: QueryPoints) -> Tuple[ObserverEvaluations, TensorType]:
+        """
+        Return the predicted mean and full covariance of the latent function(s) at the specified
+        ``query_points``.
+
+        :param query_points: The points at which to make predictions.
+        :return: The predicted mean and full covariance.
+        """
+        raise NotImplementedError
```
vpicheny, line 55: I don't think that all probabilistic models can provide a full_cov estimate. But maybe that's OK for now since we don't have a counter-example?

trieste/models/model_interfaces.py
```diff
@@ -300,6 +314,9 @@ def update(self, dataset: Dataset):
     def predict(self, query_points: QueryPoints) -> Tuple[ObserverEvaluations, TensorType]:
         return self.model.predict_y(query_points)
 
+    def predict_full_cov(self, query_points: QueryPoints) -> Tuple[ObserverEvaluations, TensorType]:
+        return self.model.predict_y(query_points, full_cov=True)
```
vpicheny: mmmmm not sure this is the right call! I would remove those lines so that predict_full_cov always returns samples from the latent models.

trieste/models/model_interfaces.py
joelberkeley: hmmm ... that's awkward. That's another reason to make models generic (which would mean you could specify only GPflow models

joelberkeley: it would also be solved really nicely by my experimental model interface

joelberkeley: it's OK I guess, we'll just have to admit we don't support prob models that can't, but I'm ok to add that as a feature later

trieste/models/model_interfaces.py
joelberkeley: ok. Presumably we should then also change the `predict` above so it's also latent?

vpicheny: no no no! the other predict has to be y

joelberkeley: how about we rename `predict_full_cov` to `predict_latent_full_cov` for clarity?

trieste/models/model_interfaces.py
joelberkeley: and not all models have latent models either. This is awkward ...

trieste/acquisition/function.py
```diff
@@ -11,80 +11,84 @@
...
+  - If the argument type is `BatchQueryPoints`, the `AcquisitionFunction` maps each batch of `B`
+    query points to a single value that describes how useful it would be evaluate all these points
+    together. With leading dimensions, this means the function takes input shape `[..., B, D]` and
+    returns shape `[..., 1]`.
 """
 
 
-class AcquisitionFunctionBuilder(ABC):
-    """ An :class:`AcquisitionFunctionBuilder` builds an acquisition function. """
+class AcquisitionBuilder(ABC, Generic[T]):
```
vpicheny: why this change? how do we know we're building a function and not a rule? besides, this looks like some work outside this PR

joelberkeley: simply cos the names were getting long. Can revert it

joelberkeley: what do you reckon?

joelberkeley: what do you reckon? We'd then have a `MonteCarloIndependentAcquisitionFunctionBuilder`, and if we add single-model variants, that would be a `SingleModelMonteCarloIndependentAcquisitionFunctionBuilder`. We could use `MC` instead of `MonteCarlo`, or find a shorter way of saying `SingleModel`

trieste/acquisition/function.py
```diff
@@ -528,15 +532,187 @@ def prepare_acquisition_function(
...
+            tf.reshape([], [sample_size, 0]), shape=[sample_size, None], dtype=tf.float64
+        )  # [S, 0]
+        self._model = model
+
+    def sample(self, at: QueryPoints) -> tf.Tensor:
+        """
+        Return approximate samples from the `model` specified at :meth:`__init__`. Multiple calls to
+        :meth:`sample`, for any given :class:`IndependentReparametrizationSampler` and ``at``, will
+        produce the exact same samples. Calls to :meth:`sample` on *different*
+        :class:`IndependentReparametrizationSampler` instances will produce different samples.
```
joelberkeley: @vpicheny you mentioned the implementation wasn't very obvious, and I think you said you'd like me to document how it works. Is this what you meant, or did you mean in finer detail?

joelberkeley: (see also the class docstring)

trieste/acquisition/function.py
vpicheny: I'd go for MC - might also use Idpt

trieste/acquisition/function.py
```diff
@@ -528,15 +532,187 @@ def prepare_acquisition_function(
...
 
 
-class BatchAcquisitionFunctionBuilder(ABC):
-    """ A :class:`BatchAcquisitionFunctionBuilder` builds a batch acquisition function. """
+class IndependentReparametrizationSampler:
+    r"""
+    This sampler employs the *reparameterization trick* to approximate samples from a
+    :class:`ProbabilisticModel`\ 's predictive distribution as
+
+    .. math:: x \mapsto \mu(x) + \epsilon \sigma(x)
```
vpicheny: sigma(x) is misleading, it looks like the covariance matrix while it's the Cholesky decomposition. Use L(x) instead?

trieste/acquisition/function.py
vpicheny: actually the not obvious bit for me is that self._eps is empty at first and populated the first time sample is called, then it is fixed.

trieste/acquisition/function.py
```diff
@@ -528,15 +532,187 @@ def prepare_acquisition_function(
...
+            self._eps.assign(
+                tf.random.normal(
+                    [mean.shape[-1], batch_size, self._sample_size], dtype=tf.float64
+                )  # [L, B, S]
+            )
+
+        identity = tf.eye(batch_size, dtype=cov.dtype)  # [B, B]
+        cov_cholesky = tf.linalg.cholesky(cov + jitter * identity)  # [..., L, B, B]
+
+        variance_contribution = cov_cholesky @ tf.cast(self._eps, cov.dtype)  # [..., L, B, S]
```
vpicheny, line 621: gpflow uses something different - probably for a good reason? We need to get someone's eyes on this.

joelberkeley: what do you mean by "uses something different"?

trieste/acquisition/function.py
joelberkeley: it's essentially a lazy constant. i'll think about how to say that without exposing implementation

trieste/acquisition/function.py
joelberkeley: are you sure? this is the independent case, and we're literally doing `tf.sqrt(cov)` with diagonal cov

trieste/acquisition/function.py
joelberkeley: is Idpt a standard abbrev? cos Ind and Indpt show up on a google search

trieste/acquisition/function.py
vpicheny: `       
 jittermat = (
            tf.eye(D, batch_shape=mean_shape[:-1], dtype=default_float()) * default_jitter()
        )  # [..., N, D, D]

        eps_shape = tf.concat([mean_shape, [S]], 0)

        eps = tf.random.normal(eps_shape, dtype=default_float())  # [..., N, D, S]

        chol = tf.linalg.cholesky(cov + jittermat)  # [..., N, D, D]

        samples = mean[..., None] + tf.linalg.matmul(chol, eps)  # [..., N, D, S]

        samples = leading_transpose(samples, [..., -1, -3, -2])  # [..., S, N, D]`

joelberkeley: can you be more specific? I've refactored it a little for clarity, replaced `tf.linalg.matmul` with `@`, dropped the `batch_shape` cos I think broadcasting does it for you. Was there something in particular?

trieste/models/model_interfaces.py
```diff
@@ -29,23 +30,40 @@ class ProbabilisticModel(tf.Module, ABC):
...
+
+        :param query_points: The points at which to make predictions, of shape [..., D].
+        :return: The mean and variance of the independent marginal distributions at each point in
+            ``query_points``. For a predictive distribution with event shape E, the mean and
+            variance will both have shape [...] + E.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def predict_joint(self, query_points: TensorType) -> Tuple[TensorType, TensorType]:
```
joelberkeley, line 48: i'm not decided on the names yet. I'd prefer `predict` to be the main method, and `predict_independent` to be the convenience method

tests/unit/acquisition/test_function.py
```diff
@@ -496,4 +502,73 @@ def _build_with_sampler(
...
+    _Acq(100).prepare_acquisition_function(data, _IndOnlyDimTwoOutputModel())
+
+
+@pytest.mark.parametrize("sample_size", [0, -2])
+def test_batch_reparametrization_sampler_raises_for_invalid_sample_size(sample_size: int) -> None:
+    with pytest.raises((ValueError, tf.errors.InvalidArgumentError)):
+        BatchReparametrizationSampler(sample_size, QuadraticWithUnitRBF())
+
+
+# todo this one uses one output dim, while we should test with more than one
```
joelberkeley: i've got a draft fix that tests this with two output dims. The mean is 0.8% off (which is too high), but the covariance is correct. I'm not sure whether to add this to the PR yet. It would mean testing the utility model in more depth

joelberkeley: i'll probably do this, it's not that much work

joelberkeley: it would also mean `_IndOnlyDimTwoOutputModel` above could be replaced, which would give us consistency across the tests

tests/util/model.py
```diff
@@ -35,63 +37,50 @@ def optimize(self, dataset: Dataset) -> None:
...
 
-@random_seed
-def test_gaussian_marginal_sample() -> None:
-    class _Sum(GaussianMarginal):
-        def predict(self, query_points: TensorType) -> Tuple[TensorType, TensorType]:
-            mean = tf.reduce_sum(query_points, axis=1, keepdims=True)
-            return mean, tf.ones_like(mean)
 
-    samples = _Sum().sample(tf.constant([[0.0, 1.0], [2.0, 3.0], [4.0, 5.0]]), 100_000)
+class QuadraticWithUnitRBF(GaussianProcess):
```
joelberkeley: todo rename

tests/util/model.py
```diff
@@ -35,63 +37,50 @@ def optimize(self, dataset: Dataset) -> None:
...
-
+        mean, cov = self.predict_joint(query_points[..., None, :])
+        return tf.squeeze(mean, -2), tf.squeeze(cov, [-2, -1])
+
+    def predict_joint(self, query_points: TensorType) -> Tuple[TensorType, TensorType]:
+        mean = tf.concat([f(query_points) for f in self._mean_functions], axis=-1)
+        cov = tf.concat(
+            [k.tensor(query_points, query_points, 1, 1)[..., None, :, :] for k in self._kernels],
+            axis=-3,
+        )
```
joelberkeley: todo tidy

tests/unit/acquisition/test_function.py
```diff
@@ -452,13 +443,14 @@ def _build_with_sampler(
...
 
-    data = Dataset(tf.zeros([0, 1]), tf.zeros([0, 2]))
+    data = Dataset(tf.zeros([0, 2], dtype=tf.float64), tf.zeros([0, 2], dtype=tf.float64))
     _Acq(20_000).prepare_acquisition_function(
         {"foo": data, "bar": data, "baz": data},
         {
-            "foo": _ArbitraryDimTwoOutputModel(-1),
-            "bar": _ArbitraryDimTwoOutputModel(0),
-            "baz": _ArbitraryDimTwoOutputModel(1),
+            # todo make these different
```
joelberkeley: todo

tests/unit/acquisition/test_function.py
```diff
@@ -495,5 +489,69 @@ def _build_with_sampler(
...
+
+    samples_mean = tf.reduce_mean(samples, axis=-3)
+    samples_covariance = tf.transpose(
+        tfp.stats.covariance(samples, sample_axis=-3, event_axis=-2), [0, 3, 1, 2]
+    )
+
+    model_mean, model_cov = model.predict_joint(xs)
+
+    npt.assert_allclose(samples_mean, model_mean, rtol=0.02)
+    npt.assert_allclose(samples_covariance, model_cov, rtol=0.04)
```
joelberkeley, line 521: it took some real searching to find a batch size that would give this level of accuracy, and it's not even that good. I'm not sure what I need to do to improve it. I probably need to look again at the kernel (I added another dimension since I got it working before)

joelberkeley: ah, rbf's not suitable for quadratic curves

tests/unit/acquisition/test_function.py
```diff
@@ -122,24 +122,21 @@ def test_expected_improvement(
     x_range = tf.cast(x_range, dtype=tf.float64)
     xs = tf.reshape(tf.stack(tf.meshgrid(x_range, x_range, indexing="ij"), axis=-1), (-1, 2))
 
-    class _Model(GaussianMarginal):
-        kernel = tfp.math.psd_kernels.MaternFiveHalves(variance_scale, length_scale=0.25).apply
+    kernel = tfp.math.psd_kernels.MaternFiveHalves(variance_scale, length_scale=0.25)
```
vpicheny, line 125: stupid question but why is the lengthscale value hard coded but not the variance?

tests/unit/acquisition/test_function.py
vpicheny: With that many samples, the mean should be very accurate. Have you looked at what happens if you use 1e7 samples? Not for the actual test that would take too long but to make sure there's not error.

One possible explanation is that the covariance matrix contains values very close to zero (which would trigger a warning although the approximation is good). This would happen if the spread of xs is much larger than the lengthscale. Or conversely, the covariance matrix is ill-conditioned because the lengthscale is very large. Then the sampling introduces some numerical errors.

trieste/acquisition/function.py
```diff
@@ -392,12 +393,16 @@ def __init__(self, sample_size: int, model: ProbabilisticModel):
         tf.debugging.assert_positive(sample_size)
 
         self._sample_size = sample_size
+
+        # _eps is essentially a lazy constant, declared here but initialised only on the first call
+        # to sample
```
vpicheny: I'm glad we have this explanation now :-)

trieste/models/model_interfaces.py
vpicheny: as discussed, I like the current choice

trieste/models/model_interfaces.py
```diff
@@ -29,23 +30,40 @@ class ProbabilisticModel(tf.Module, ABC):
...
+        """
+        :param query_points: The points at which to make predictions, of shape [..., B, D].
+        :return: The mean and covariance of the joint marginal distribution at each batch of points
+            in ``query_points``. For a predictive distribution with event shape E, the mean will
```
vpicheny, line 52: why E and not L?

vpicheny (approved):

tests/unit/acquisition/test_function.py
joelberkeley-secondmind: cos length scale is to do with the shape of the function, or at least that's what I thought until we spoke about rbf and quadratic

trieste/models/model_interfaces.py
joelberkeley-secondmind: **E**vent shape

trieste/models/model_interfaces.py
joelberkeley-secondmind: looking back at this, I still don't follow why we're doing f for one and y for the other (never mind that `sample` is also sampling f atm)

trieste/models/model_interfaces.py
joelberkeley-secondmind: also cos that's how tfp did it and I think it makes sense

trieste/acquisition/function.py
```diff
@@ -485,6 +498,10 @@ def __init__(self, sample_size: int):
         tf.debugging.assert_positive(sample_size)
         self._sample_size = sample_size
 
+    def __repr__(self) -> str:
+        """"""
+        return f"SingleModelMCIndAcquisitionFunctionBuilder({self._sample_size!r})"
```
ChrisMorter: Maybe use `{type(self).__name__}` here, so that this is robust to any changes to the class name?

trieste/acquisition/function.py
```diff
@@ -519,3 +539,78 @@ def prepare_acquisition_function(
...
+        # _eps is essentially a lazy constant, declared here but initialised only on the first call
+        # to sample
+        self._eps = tf.Variable(
+            tf.ones([0, 0, sample_size], dtype=tf.float64), shape=[None, None, sample_size]
+        )  # [0, 0, S]
+        self._model = model
+
+    def __repr__(self) -> str:
+        """"""
+        return f"BatchReparametrizationSampler({self._sample_size!r}, {self._model!r})"
```
ChrisMorter, line 567: Same comment here regarding using `{type(self).__name__}`

trieste/acquisition/function.py
```diff
@@ -519,3 +539,78 @@ def prepare_acquisition_function(
...
+        """
+        :param sample_size: The number of samples for each batch of points. Must be positive.
+        :param model: The model to sample from.
+        :raise ValueError (or InvalidArgumentError): If ``sample_size`` is not positive.
+        """
+        tf.debugging.assert_positive(sample_size)
+
+        self._sample_size = sample_size
+
+        # _eps is essentially a lazy constant, declared here but initialised only on the first call
```
ChrisMorter: I know what you mean, but arguably the `_eps` is initialised here since it has been assigned a value, even if it is a dummy value. Perhaps something like "declared and assigned a dummy value here, the actual value is assigned on the first call to sample" would be clearer?

trieste/acquisition/function.py
```diff
@@ -519,3 +539,78 @@ def prepare_acquisition_function(
...
+        """
+        tf.debugging.assert_rank_at_least(at, 2)
+        tf.debugging.assert_greater_equal(jitter, 0.0)
+
+        mean, cov = self._model.predict_joint(at)  # [..., B, L], [..., L, B, B]
+        batch_size = at.shape[-2]
+
+        if tf.size(self._eps) == 0:
+            self._eps.assign(
+                tf.random.normal(
```
ChrisMorter, line 613: Will this not be wrong if `sample` is called with different batch sizes?

trieste/acquisition/function.py
```diff
@@ -392,12 +393,16 @@ def __init__(self, sample_size: int, model: ProbabilisticModel):
         tf.debugging.assert_positive(sample_size)
 
         self._sample_size = sample_size
+
+        # _eps is essentially a lazy constant, declared here but initialised only on the first call
```
ChrisMorter: Same comment as in `BatchReparametrizationSampler` regarding use of the term "initialised".

trieste/acquisition/function.py
joelberkeley-secondmind: i used to do that until someone suggested i didn't. I'll leave it as is for now, because I've done it like this across the code base so it's consistent. I don't really have a preference in general

trieste/acquisition/function.py
joelberkeley-secondmind: this comment is mainly for @vpicheny (and other researchers) , so to some extent it depends which he finds more helpful

trieste/acquisition/function.py
joelberkeley-secondmind: yes. Hadn't thought of that at all

joelberkeley-secondmind: @vpicheny is it right that we'd only expect someone to use this with a fixed batch size?

joelberkeley-secondmind: @vpicheny is it reasonable to require a fixed batch shape for a given sampler for the whole BO loop?

joelberkeley-secondmind: i'm pretty sure it is, so i'll just make that change

trieste/acquisition/function.py
joelberkeley-secondmind: how about
```
        # _eps is essentially a lazy constant. It is declared and assigned an empty tensor here, and
        # populated on the first call to sample
```

trieste/acquisition/function.py
vpicheny: yes that seems reasonable

joelberkeley-secondmind: @ChrisMorter , fyi victor and I decided to say batch size is fixed for now, and to make a ticket to allow variable batch size (which is probably possible by adding more N(0, 1) samples or subsampling for bigger/smaller batch sizes)

here's the ticket https://github.com/secondmind-labs/trieste/issues/132

trieste/acquisition/function.py
joelberkeley-secondmind: i've actually just noticed these abstract classes shouldn't have `__repr__`s anyway, so I'll delete them

trieste/acquisition/function.py
joelberkeley-secondmind: btw the check for the batch size is added above

ChrisMorter: Ok, that seems reasonable.

ChrisMorter (approved):
LGTM

PR #94, opened 2020-12-10
monte carlo acquisition ind

joelberkeley:


trieste/acquisition/function.py
```diff
@@ -528,6 +528,54 @@ def prepare_acquisition_function(
...
+        produce the exact same samples. Calls to :meth:`sample` on *different*
+        :class:`IndependentReparametrizationSampler` instances will produce different samples.
+
+        :param at: Where to sample the predictive distribution, with shape `[..., _]`.
+        :return: The samples, of shape `[..., S, L]`, where `S` is the `sample_size` and `L` is the
+            number of latent model dimensions.
+        :raise ValueError (or InvalidArgumentError): If ``at`` is a scalar.
+        """
+        tf.debugging.assert_rank_at_least(at, 1)
+        mean, cov = self._model.predict(at[..., None, :])  # [..., 1, L], [..., 1, L]
```
vpicheny, line 570: why do we need the extra dimension?

vpicheny: Well I guess it doesn't matter if it's here or 5 lines below.

joelberkeley: yeah it's slightly shorter here

trieste/acquisition/function.py
```diff
@@ -528,6 +528,54 @@ def prepare_acquisition_function(
...
+
+        if tf.size(self._eps) == 0:
+            self._eps.assign(
+                tf.random.normal([self._sample_size, mean.shape[-1]], dtype=tf.float64)
+            )  # [S, L]
+
+        return mean + tf.sqrt(cov) * tf.cast(self._eps, cov.dtype)  # [..., S, L]
```
vpicheny, line 577: actually the gpflow convention is to add [..., S,1, L]
don't know if we should follow it or not. On the plus side the ranks would be identical to the batch case

joelberkeley: yeah that's what the shape would be if we merged batch and non-batch. Could do that here

trieste/acquisition/function.py
```diff
@@ -528,6 +528,54 @@ def prepare_acquisition_function(
...
+        :param sample_size: The number of samples to take at each point. Must be positive.
+        :param model: The model to sample from.
+        :raise ValueError (or InvalidArgumentError): If ``sample_size`` is not positive.
+        """
+        tf.debugging.assert_positive(sample_size)
+
+        self._sample_size = sample_size
+        self._eps = tf.Variable(
+            tf.ones([sample_size, 0], dtype=tf.float64), shape=[sample_size, None]
+        )  # [S, 0]
```
ChrisMorter, line 553: NIT: Is the size not [S, None]?

trieste/acquisition/function.py
joelberkeley: do you know why that's the convention?

Whether we merge with batch or not, we should be consistent within trieste

tests/unit/acquisition/test_function.py
```diff
@@ -326,3 +328,58 @@ def prepare_acquisition_function(
...
+    x: tf.Tensor,
+) -> None:
+    model = QuadraticWithUnitVariance()
+    sample_size = 10_000
+    samples = IndependentReparametrizationSampler(sample_size, model).sample(x)
+    statistic, _pvalue = stats.kstest(
+        tf.squeeze(samples).numpy(),
+        lambda arr: tfp.distributions.Normal(*model.predict(x)).cdf(arr).numpy(),
+    )
+    assert statistic < 1.36 / tf.sqrt(float(sample_size))
```
ChrisMorter: NIT: Maybe just use np.sqrt to avoid converting to a float?

tests/unit/acquisition/test_function.py
```diff
@@ -326,3 +328,58 @@ def prepare_acquisition_function(
...
+@pytest.mark.parametrize("x", tf.linspace([-10.0], [10.0], 100))
+def test_independent_reparametrization_sampler_samples_approximate_expected_distribution(
+    x: tf.Tensor,
+) -> None:
+    model = QuadraticWithUnitVariance()
+    sample_size = 10_000
+    samples = IndependentReparametrizationSampler(sample_size, model).sample(x)
+    statistic, _pvalue = stats.kstest(
+        tf.squeeze(samples).numpy(),
+        lambda arr: tfp.distributions.Normal(*model.predict(x)).cdf(arr).numpy(),
```
ChrisMorter: I think `tfp.distributions.Normal(*model.predict(x))` can be moved outside the lambda, naming it something like `expected_distributions` would probably help readability too.

tests/unit/acquisition/test_function.py
```diff
@@ -326,3 +328,58 @@ def prepare_acquisition_function(
...
+    xs = tf.linspace([-10.0], [10.0], 100)
+    npt.assert_allclose(sampler.sample(xs), sampler.sample(xs))
+
+
+@random_seed
+def test_independent_reparametrization_sampler_samples_are_distinct_for_new_instances() -> None:
+    sampler1 = IndependentReparametrizationSampler(100, QuadraticWithUnitVariance())
+    sampler2 = IndependentReparametrizationSampler(100, QuadraticWithUnitVariance())
+    xs = tf.linspace([-10.0], [10.0], 100)
+    npt.assert_array_less(1e-9, tf.abs(sampler2.sample(xs) - sampler1.sample(xs)))
```
ChrisMorter, line 394: So it looks like this is asserting that no pair of samples are within 1e-9, I see that's true (nearly always), but seems like a weird way to assert that the samples are different? Is it sufficient to just check that _some_ of them are different?

tests/unit/acquisition/test_function.py
ChrisMorter: 1.36 is the 95% level for the KS test? Maybe best to extract this as a named variable for readability?

ChrisMorter (approved):
Looks good to me, just a few minor readability related comments

trieste/acquisition/function.py
joelberkeley: i guess it depends if the shape comments are 'static'  cf. `self._eps.shape` or 'runtime' cf. `tf.shape(self._eps)` (to use the terms loosely). I don't know if we have a convention there

tests/unit/acquisition/test_function.py
joelberkeley: it seems a cheap win to be able to assert they're all different, and as you say, it would be almost impossible for any two samples to agree

tests/unit/acquisition/test_function.py
joelberkeley: will do

tests/unit/acquisition/test_function.py
joelberkeley: will do

tests/unit/acquisition/test_function.py
joelberkeley: used `math.sqrt`

trieste/acquisition/function.py
vpicheny: Let's do `[..., S,1, L]` then! so we can merge batch and non-batch later. 

I guess the gpflow convention is precisely to unify independent / batch.

joelberkeley: agreed in slack that `[..., S, L]` for now

PR #97, opened 2020-12-13
fix model deepcopy

joelberkeley:
This PR changes a call to gpflow's model copying functionality to use Python's stdlib `copy` module to copy models. This fixes a type mistake (that `gpflow.utilities.deepcopy` requires a `tf.Module` which a `ProbabilisticModel` isn't). This also has the side effect of reducing dependencies and limitations on interfaces


awav:
I don't understand the reasoning for this PR: `tf.Module` is a natural choice for storing models and making snapshots. Most importantly, it simply works right now. I'm not sure that all these 300+ lines of code are better than a *single line* with harmless `tf.Module` dependency. And don't forget that a new model interface would have to re-define a `__deepcopy__` method.

joelberkeley-secondmind:
Bear in mind I started working on this before you reintroduced `tf.Module` into `ProbabilisticModel`. You said you did that to fix deepcopy issues, so this fixes that problem without the need for adding `tf.Module`.

Whether `ProbabilisticModel` needs `tf.Module` is a fair question, but a separate one to what this PR attempts to solve. As to whether it does, we must bear in mind that `ProbabilisticModel` is by design a static, untrainable thing.

As for whether `TrainableProbabilisticModel` needs `tf.Module`, the API already has an `optimize` method on it which provides a higher-level way for users to implement their optimization routine. If users want to use `tf.Module`, they can do that very easily. On deeper inspection, not even `GPflowPredictor` needs to be a `tf.Module` with these changes.

> And don't forget that a new model interface would have to re-define a `__deepcopy__` method.

That depends. If it encounters the same problems with tfp's bijectors as gpflow's models, yes, though that can be easily fixed by pulling out the `__deepcopy__` on `GaussianProcessRegression` as a utility mixin "fix" until tfp's problem is fixed (in which case, it may be best to migrate that to GPflow itself). Of course in general, `__deepcopy__` is only needed to customize deepcopies, not enable them.

But, as you suggest, why not use `tf.Module`? Because it's not needed anywhere and doesn't provide a minimal interface. In addition, given there's no obvious need for it for anything to work, it's worth remembering that it's typed as `Any`, which means all type checking on models is lost.

awav:
The `tf.Module` is a simple solution to a temporary problem.

> In addition, given there's no obvious need for it for anything to work

It makes standard deepcopy from gpflow to work out of the box :)

More crucially: TF2.4 is out, the TFP is fixed and out, I think it would be more helpful to spend time on switching versions, instead of adding 300+ lines of code and removing them in a month :)

joelberkeley:
it's good to hear there's a fix available, but we're not going to pin to tf 2.4 for a long while yet: atm we're at `tensorflow>=2.1,!=2.2.0,!=2.3.0`

`gpflow.utilities.deepcopy` will still work with the above.

While it's only ~70 more lines, I can make it a shorter and simpler PR

joelberkeley:
btw the implementation for `SparseVariational` is an optimization. It's not needed so could be left out of this PR

joelberkeley:
this is also useful because it works with object containing `tf.Module`s, not just the `tf.Module` itself

hstojic (approved):
Ok, I see this has been a heated debate. 

I think I would agree with Artem that this is perhaps not needed as soon as Trieste moves to latest/recent TF versions and it would be easier to switch with the older version of the code. (On that note, what is needed currently to make the switch to TF 2.4?)

That being said, its not a big issue to merge this in, since we would be able to remove after switching to TF 2.4 and then we won't have any additional maintenance. Hence, I'll approve it.

Could you provide in the `module_deepcopy` some links to TF which show the bug and solution in the newest version?

joelberkeley:
Migrating to 2.4 would fix the problem, yes. We are on 2.1 so it will be a while before trieste moves to 2.4, the latest version of TensorFlow.

> Could you provide in the module_deepcopy some links to TF which show the bug and solution in the newest version?

Sorry, I don't quite understand. Do you mean a github issue? And wrt newest version, is that newest version of tf, or trieste?

hstojic:
Yes, GH issue, but some other link if unavailable
So its easier to track down what's the underlying issue

Sent from ProtonMail mobile

PR #104, opened 2020-12-19
optimize deepcopy

### LINKED ISSUE #88: Minimize memory footprint of tracked history in optimizer
As a user, I want saved state in the `BayesianOptimizer` to use minimal memory footprint.

This can be achieved by implementing `__deepcopy__` on commonly copied objects.

We could let the user specify what to copy, such as just some parts of the models, but that's more involved and may change the API

joelberkeley:
This PR implements deepcopy optimizations for various classes that are (or may be) deep copied when saving state

### DIFF
--- a/tests/unit/test_space.py
+++ b/tests/unit/test_space.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
 from typing import List, Tuple
 
 import numpy.testing as npt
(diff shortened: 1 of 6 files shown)


tests/unit/test_space.py
```diff
@@ -99,6 +100,32 @@ def test_discrete_search_space_sampling_raises_when_too_many_samples_are_request
...
+    assert all(point in dss1 for point in res.points[:, :2])
+    assert all(point in dss2 for point in res.points[:, 2:])
+
+
+def test_discrete_search_space_product_raises_if_points_have_different_types() -> None:
+    dss1 = DiscreteSearchSpace(_points_in_2D_search_space())
+    dss2 = DiscreteSearchSpace(tf.constant([[1.0, 1.4], [-1.5, 3.6]], tf.float64))
+
+    with pytest.raises(TypeError):
+        _ = dss1 * dss2
```
joelberkeley, line 121: these two tests are just moved from further down

ChrisMorter (approved):
LGTM

PR #106, opened 2020-12-21
add astuple method to Dataset

joelberkeley:
according to [this SO post](https://stackoverflow.com/q/51802109/5986907) `dataclasses.astuple` deepcopies dataclass attributes. This is unnecessary and expensive for large datasets. This PR adds an `astuple` method instead.

Note we could instead make `Dataset` a `NamedTuple` with a custom `__new__` but that would change the meaning of `__len__` so I didn't do that in this PR


sam-willis (approved):
LGTM

PR #108, opened 2020-12-22
move imports to body in notebooks

joelberkeley-secondmind:
When I learnt how to use numpy, pandas etc., the tutorials only imported libraries when it used them, rather than at the start of the notebook. This made it clear what is needed when and why. In this PR I've done that to the notebooks.

### DIFF
--- a/docs/notebooks/failure_ego.pct.py
+++ b/docs/notebooks/failure_ego.pct.py
@@ -2,11 +2,9 @@
 # # EGO with a failure region
 
 # %%
-import gpflow
 import numpy as np
 import tensorflow as tf
 
-gpflow.config.set_default_jitter(1e-5)
 np.random.seed(1234)
 tf.random.set_seed(1234)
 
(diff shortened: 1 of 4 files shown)


docs/notebooks/failure_ego.pct.py
```diff
@@ -17,27 +17,9 @@
...
-from trieste.models.model_interfaces import VariationalGaussianProcess
-from trieste.models import ModelSpec
-from gpflow.models import VGP
-from gpflow.optimizers import NaturalGradient
 
-from util.plotting_plotly import plot_function_plotly, plot_gp_plotly, add_bo_points_plotly
-from util.plotting import plot_gp_2d, plot_function_2d, plot_bo_points
-
-# %%
-gpflow.config.set_default_jitter(1e-5)
```
joelberkeley, line 24: this particular line change is outside the scope of this PR. I couldn't resist, cos it makes the intro to all the notebooks the same. I checked and the notebook results are the same

vpicheny (approved):
LGTM

PR #110, opened 2020-12-22
add bibtex formatting to references

### LINKED ISSUE #92: Prettify bibligraphic references in docstrings
As a user, I want to be able to easily read the bibliographic references in docstrings so that I can get context easily and find the source easily, and so that reading the documentation is more pleasant.

It would be particularly nice if the result included a clickable link to the reference

Currently, bibliographic references in docstrings simply paste the reference in the docstring. It works but it's crude and not very pretty.

One solution is to use [sphinxcontrib-bibtex](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/)

joelberkeley-secondmind:
~**DO NOTE MERGE** need to check [sphinxcontrib-bibtex license](https://github.com/mcmtroffaes/sphinxcontrib-bibtex/blob/develop/LICENSE.rst) before merging~ it's 2-clause BSD, so ok

This PR enables and adds bibtex formatting to bibliographic references in docstrings.


trieste/__init__.py
```diff
@@ -16,6 +16,11 @@
 models (see :mod:`~trieste.models`), and data sets (see :mod:`~trieste.data`). The
 :mod:`~trieste.acquisition` package provides a selection of acquisition algorithms and the
 functionality to define your own.
+
+Bibliography
+============
+
+.. bibliography::
```
joelberkeley-secondmind, line 23: i'm not sure where is best to put the bibliography
- relevant references in each module
- in the library root (as I've done here)
- on the home page
- in a separate "bibliography" page (I couldn't get this to work)

trieste/acquisition/function.py
```diff
@@ -336,28 +286,10 @@ def probability_of_feasibility(
 
 class ExpectedConstrainedImprovement(AcquisitionFunctionBuilder):
     """
-    Builder for the _expected constrained improvement_ acquisition function defined in
-    Gardner, 2014. The acquisition function computes the expected improvement from the best
+    Builder for the *expected constrained improvement* acquisition function defined in
```
joelberkeley-secondmind, line 289: i saw this was causing warning in the sphinx output

trieste/__init__.py
joelberkeley: i think in the library root is OK for now. We can move it later if we want

vpicheny (approved):

PR #112, opened 2020-12-25
support lists as bound args for box

joelberkeley:
The most important thing about this PR is that it assumes a default float type for trieste of `tf.float64`. The plan is to use f64 for now, as the global default, and consider what to do with f32 later

### DIFF
--- a/docs/notebooks/util/plotting.py
+++ b/docs/notebooks/util/plotting.py
@@ -16,10 +16,11 @@ import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib import cm
 
+from trieste.type import TensorType
 from trieste.utils import to_numpy
 
 
-def create_grid(mins, maxs, grid_density=20):
+def create_grid(mins: TensorType, maxs: TensorType, grid_density=20):
     """
     Creates a regular 2D grid of size grid_density*grid_density between mins and maxs.
     :param mins: list of 2 lower bounds
@@ -63,8 +64,8 @@ def plot_surface(xx, yy, f, ax, contour=False, alpha=1.0):
 
 def plot_function_2d(
     obj_func,
-    mins,
-    maxs,
+    mins: TensorType,
+    maxs: TensorType,
     grid_density=20,
     contour=False,
     log=False,
@@ -86,6 +87,8 @@ def plot_function_2d(
     :param ylabel:
     :param figsize:
     """
+    mins = to_numpy(mins)
+    maxs = to_numpy(maxs)
 
     # Create a regular grid on the parameter space
     Xplot, xx, yy = create_grid(mins=mins, maxs=maxs, grid_density=grid_density)
@@ -250,8 +253,8 @@ def plot_regret(
 
 def plot_gp_2d(
     model,
-    mins,
-    maxs,
+    mins: TensorType,
+    maxs: TensorType,
     grid_density=20,
     contour=False,
     xlabel=None,
@@ -270,6 +273,8 @@ def plot_gp_2d(
     :param ylabel: optional string
     :param figsize:
     """
+    mins = to_numpy(mins)
+    maxs = to_numpy(maxs)
 
     # Create a regular grid on the parameter space
     Xplot, xx, yy = create_grid(mins=mins, maxs=maxs, grid_density=grid_density)
--- a/docs/notebooks/util/plotting.py
+++ b/docs/notebooks/util/plotting.py
@@ -15,6 +15,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib import cm
+import tensorflow as tf
 
 from trieste.type import TensorType
 from trieste.utils import to_numpy
@@ -22,12 +23,14 @@ from trieste.utils import to_numpy
 
 def create_grid(mins: TensorType, maxs: TensorType, grid_density=20):
     """
-    Creates a regular 2D grid of size grid_density*grid_density between mins and maxs.
+    Creates a regular 2D grid of size `grid_density^2` between mins and maxs.
     :param mins: list of 2 lower bounds
     :param maxs: list of 2 upper bounds
     :param grid_density: scalar
     :return: Xplot [grid_density**2, 2], xx, yy from meshgrid for the specific formatting of contour / surface plots
     """
+    tf.debugging.assert_shapes([(mins, [2]), (maxs, [2])])
+
     xspaced = np.linspace(mins[0], maxs[0], grid_density)
     yspaced = np.linspace(mins[1], maxs[1], grid_density)
     xx, yy = np.meshgrid(xspaced, yspaced)
@@ -77,8 +80,8 @@ def plot_function_2d(
     """
     2D/3D plot of an obj_func for a grid of size grid_density**2 between mins and maxs
     :param obj_func: a function that returns a n-array given a [n, d] array
-    :param mins: list of 2 lower bounds
-    :param maxs: list of 2 upper bounds
+    :param mins: 2 lower bounds
+    :param maxs: 2 upper bounds
     :param grid_density: positive integer for the grid size
     :param contour: Boolean. If False, a 3d plot is produced
     :param log: Boolean. If True, the log transformation (log(f - min(f) + 0.1)) is applied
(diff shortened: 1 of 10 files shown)


trieste/space.py
```diff
@@ -159,35 +159,43 @@ class Box(SearchSpace):
     closed bounded intervals in :math:`\mathbb{R}`.
     """
 
-    def __init__(self, lower: TensorType, upper: TensorType):
-        """
-        :param lower: The lower (inclusive) bounds of the box.
-        :param upper: The upper (inclusive) bounds of the box.
-        :raise ValueError: If ``lower`` and ``upper`` have different shapes. Or if ``upper`` is not
+    def __init__(self, lower: Union[tf.Tensor, List[float]], upper: Union[tf.Tensor, List[float]]):
```
joelberkeley, line 162: I've restricted to `tf.Tensor` for simplicity

trieste/space.py
```diff
@@ -152,42 +152,56 @@ def __mul__(self, other: DiscreteSearchSpace) -> DiscreteSearchSpace:
...
     :math:`\mathbb{R}^D`. Mathematically it is equivalent to the Cartesian product of :math:`D`
     closed bounded intervals in :math:`\mathbb{R}`.
     """
 
-    def __init__(self, lower: TensorType, upper: TensorType):
-        """
-        :param lower: The lower (inclusive) bounds of the box.
-        :param upper: The upper (inclusive) bounds of the box.
-        :raise ValueError: If ``lower`` and ``upper`` have different shapes. Or if ``upper`` is not
+    def __init__(self, lower: Vector, upper: Vector):
```
joelberkeley, line 169: I've restricted to `tf.Tensor` for simplicity. The float type check doesn't work with numpy arrays

trieste/space.py
```diff
@@ -155,42 +155,58 @@ def __deepcopy__(self, memo: Dict[int, object]) -> DiscreteSearchSpace:
...
```
joelberkeley, line 169: overloads would be preferable from a code point of view, but the docs don't render overloads for `__init__` so I've used a `TypeVar` instead

docs/notebooks/util/plotting.py
```diff
@@ -16,10 +16,11 @@
...
 from matplotlib import cm
 
+from trieste.type import TensorType
 from trieste.utils import to_numpy
 
 
-def create_grid(mins, maxs, grid_density=20):
+def create_grid(mins: TensorType, maxs: TensorType, grid_density=20):
     """
     Creates a regular 2D grid of size grid_density*grid_density between mins and maxs.
```
sam-willis:
```suggestion
    Creates a regular 2D grid of size `grid_density^2` between mins and maxs.
```

docs/notebooks/util/plotting.py
```diff
@@ -16,10 +16,11 @@
 import numpy as np
 from matplotlib import cm
 
+from trieste.type import TensorType
 from trieste.utils import to_numpy
 
 
-def create_grid(mins, maxs, grid_density=20):
+def create_grid(mins: TensorType, maxs: TensorType, grid_density=20):
```
sam-willis, line 23: if this can only be a 2 element TensorType, then maybe you should assert that below.

sam-willis, line 24: if this can only be a 2 element TensorType, then maybe you should assert that below.

docs/notebooks/util/plotting.py
```diff
@@ -63,8 +64,8 @@ def plot_surface(xx, yy, f, ax, contour=False, alpha=1.0):
 
 def plot_function_2d(
     obj_func,
-    mins,
-    maxs,
+    mins: TensorType,
```
sam-willis, line 70: If this is a tensor-type then it's not a list (see docstring below)

docs/notebooks/util/plotting.py
```diff
@@ -250,8 +253,8 @@ def plot_regret(
 
 def plot_gp_2d(
     model,
-    mins,
-    maxs,
+    mins: TensorType,
```
sam-willis, line 259: same

trieste/space.py
```diff
@@ -155,42 +155,58 @@ def __deepcopy__(self, memo: Dict[int, object]) -> DiscreteSearchSpace:
...
-        if lower.dtype is not upper.dtype:
-            raise TypeError(
-                f"Lower and upper bounds must have the same dtype, got {lower.shape} and"
-                f" {upper.shape}"
-            )
+        if isinstance(lower, list):
+            self._lower = tf.cast(lower, dtype=tf.float64)
+            self._upper = tf.cast(upper, dtype=tf.float64)
+        else:
+            assert isinstance(lower, tf.Tensor) and isinstance(upper, tf.Tensor)  # to help mypy
```
sam-willis: you could try `tf.convert_to_tensor(lower)` etc here, that way you'd support numpy arrays (it'll use the array dtype)

trieste/space.py
```diff
@@ -155,42 +155,58 @@ def __deepcopy__(self, memo: Dict[int, object]) -> DiscreteSearchSpace:
...
+            raise ValueError(f"Bounds must have shape [D] for positive D, got {tf.shape(lower)}.")
 
-        if lower.dtype is not upper.dtype:
-            raise TypeError(
-                f"Lower and upper bounds must have the same dtype, got {lower.shape} and"
-                f" {upper.shape}"
-            )
+        if isinstance(lower, list):
+            self._lower = tf.cast(lower, dtype=tf.float64)
+            self._upper = tf.cast(upper, dtype=tf.float64)
```
sam-willis, line 195: what if lower is a list but upper is a tensor?

sam-willis (approved):
Looks okay, but I think you should handle numpy arrays before you handle lists (as tensorflow has easy numpy conversion, so you may plausibly be moving between the two).

Also `tf.convert_to_tensor` handles a lot of the mess of dealing with TensorTypes, the only issue is you have to choose to either let it infer the type (which you might not want to do for lists) or to take a fixed dtype.

docs/notebooks/util/plotting.py
joelberkeley: there is a lot of stuff in these files that needs fixing, a lot. I have another PR which completely refactors them, so I don't think it's worth adding things here. Sound reasonable?

trieste/space.py
joelberkeley: the type hints indicate that they should be the same type

sam-willis: that's true

docs/notebooks/util/plotting.py
joelberkeley-secondmind: i don't suggest reviewing these files - they're not in a good way and I've completely refactored them elsewhere. I'll make the changes you've suggested

sam-willis: Can't really say without seeing the other PR, but in general I don't support the argument that because there's lots of stuff that needs fixing you shouldn't fix things reviewers point out. 

That said all my comments are non-critical is why I approved the PR (so you can merge if you want).

joelberkeley-secondmind: yeah i think i was just being lazy, you got me

trieste/space.py
joelberkeley-secondmind: musings:

not supporting numpy was semi-intentional, because I can see it being quite a maintenance overhead. However, `TensorType` includes `ndarray`, so something would need to change.

I like the idea of being able to provide `ndarray` support, I'm just wondering how one tests that w/o duplicating every test. I'll think more about this

joelberkeley-secondmind: this would be less of a problem if tf and numpy had type hints. np will very soon, and tf in some unknown amount of time

joelberkeley-secondmind: for now, I've updated the code locally to support numpy. just looking at the tests

docs/notebooks/util/plotting.py
sam-willis: haha, no worries. As I said, optional comment, but thanks for making the changes.

joelberkeley-secondmind:
@sam-willis in case you're interested, I spoke with victor about numpy support and I think our approach is, try to support it, but don't bother testing for it cos it's too much effort, and document that that's our approach

PR #114, opened 2021-01-06
fix trust region tests

joelberkeley:


tests/unit/acquisition/test_rule.py
```diff
@@ -185,7 +185,7 @@ def test_trust_region_for_unsuccessful_local_to_global_trust_region_reduced() ->
     eps = 0.5 * (search_space.upper - search_space.lower) / 10
     previous_y_min = dataset.observations[0]
     is_global = False
-    acquisition_space = Box(dataset.query_points - eps, dataset.query_points + eps)
+    acquisition_space = Box(dataset.query_points[1] - eps, dataset.query_points[1] + eps)
```
vpicheny: actually it should be dataset.query_points[0]

tests/unit/acquisition/test_rule.py
```diff
@@ -162,7 +162,7 @@ def test_trust_region_for_successful_local_to_global_trust_region_increased() ->
...
```
vpicheny: my bad it should be dataset.query_points[0]

joelberkeley: tests don't seem to care :s we definitely need more acquisition rule tests

tests/unit/acquisition/test_rule.py
joelberkeley: are you certain?

vpicheny (approved):

tests/unit/acquisition/test_rule.py
joelberkeley: it's cos the acquisition space is only used in the algorithm to work out its size
```
            tf.reduce_prod(
                state.acquisition_space.upper - state.acquisition_space.lower
            )
```
and we don't test the `query_point` returned

PR #116, opened 2021-01-07
various VGP fixes

joelberkeley:
This fixes a type bug in VGP, adds docstrings to the class, and does some tidying


sam-willis (approved):
Why are you adding in blank docstrings?

trieste/models/model_interfaces.py
```diff
@@ -154,6 +155,7 @@ def __repr__(self) -> str:
 
     @property
     def model(self) -> SVGP:
+        """"""
```
sam-willis: Do you not want the model property to appear in the docs?

joelberkeley-secondmind: ah, ok, so this is a mistake. it should be on `__repr__`. will fix

PR #118, opened 2021-01-12
various optimizer fixes

joelberkeley:


trieste/models/model_interfaces.py
```diff
@@ -63,16 +63,27 @@ def update(self, dataset: Dataset) -> None:
 
     @abstractmethod
     def optimize(self, dataset: Dataset) -> None:
-        """ Optimize the model parameters. """
+        """
+        Optimize the model parameters with respect to the specified ``dataset``.
```
vpicheny: not sure this is the right wording. I mostly use "optimize objective wrt input variables".

trieste/models/model_interfaces.py
```diff
@@ -63,16 +63,27 @@ def update(self, dataset: Dataset) -> None:
...
+
+        :param dataset: The data with which to optimize the model.
```
vpicheny: train instead of optimize?

vpicheny (approved):

trieste/models/model_interfaces.py
joelberkeley: sure

trieste/models/model_interfaces.py
joelberkeley: good point. To include dataset parameter too, what about
```
Optimize the model objective with respect to (hyper)parameters given the specified ``dataset``.
```

PR #126, opened 2021-01-28
Pareto feature 

satrialoka:
### Feature Implemented

This is a feature for Pareto functionality needed for Multi-objective BO, the code was ported from [GPFlowopt's pareto class](https://github.com/GPflow/GPflowOpt/blob/master/gpflowopt/pareto.py) , which is used for storing pareto sets and calculating hypervolume for hypervolume-based acquisition function which will be ported by @TsingQAQ . Initial plan is to put the pareto.py at the `trieste.utils.pareto`. 

### Approach

Inside pareto.py some class and functions are implemented:

1. `Pareto` class: the core class in the pareto.py. it construct the pareto set by stores the pareto set and calculate the cell bounds covering the non-dominated region which achieved by two methods: `Pareto.bounds_2d` for objectives = 2 and `Pareto.divide_conquer_nd` for arbitrary number of objectives. It also has `hypervolume` method to calculate the hypervolume given a reference point (Should be equal or bigger than the anti-ideal point of the Pareto set). 
2. `BoundedVolumes` class: the data structure that stores the cell bounds
3. `non_dominated_sort` function: computes non-dominated points (pareto front) and its degree of dominance
4.  `atleast2d` function: utility function to fix the dimensionality like the `numpy.atleast_2d()` function
5. `to_Dataset` function: used for convert `Mapping[str, Dataset]` to `Dataset`, by concatenating the output of each Dataset inside mapping. (its also assumed that the observed point of the Datasets inside `Mapping[str, Dataset]` is the same)

## Status

Need review

## Changes

- Fix type
- Docstrings updated

## Notes

The Hypervolume BO implementation is based on this paper: [Fast calculation of multi-objective probability of improvement and expected improvement criteria for Pareto optimization](https://link.springer.com/article/10.1007/s10898-013-0118-2)


joelberkeley-secondmind:
Thanks, can you give an outline of what this PR intends to achieve, and if it's implementing any papers, links to those papers

joelberkeley-secondmind:
Also, you mention it's needed for multi-objective BO. It would help if you could describe what the specific goal there is, what the design for that could be and why, and how this enables that design. Since this sounds like a rather significant piece of work, it may be best to video chat about this.

satrialoka:
Okay. I forgot to mention, this is a port from gpflowopt will write some more details and relevant link next week.

trieste/utils/pareto.py
```diff
@@ -0,0 +1,310 @@
...
+
+        self.front = tf.zeros([0, self.data.observations.shape[1]])
+        # Initialize
+        self.update()
+
+    @staticmethod
+    def _is_test_required(smaller: tf.Tensor) -> tf.Tensor:
+        """
+        Tests if a point augments or dominates the Pareto set.
+        :param smaller: a boolean ndarray storing test point < Pareto front
```
vpicheny: tf tensor?

satrialoka: Ah yes, that's from the old code that uses numpy ndarray, I'll check all the docstring again and fix it accordingly. 
thanks for pointing this out.

satrialoka:
> 
> 
> Thanks, can you give an outline of what this PR intends to achieve, and if it's implementing any papers, links to those papers

Please check the updated first comment

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+            ),
+            tf.float32,
+        ),
+        axis=1,
+    )
+
+    return tf.boolean_mask(observations, dominance == 0), dominance
+
+
+class BoundedVolumes:
```
joelberkeley-secondmind, line 61: this looks very similar to `Box`. Can we use that?

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+from trieste.data import Dataset
+from trieste.type import TensorType
+
+jitter = gpflow.config.default_jitter()
+
+
+def atleast2d(tensor: tf.Tensor) -> tf.Tensor:
+    if tf.rank(tensor) <= 1:
+        return tf.expand_dims(tensor, axis=0)
+    return tensor
```
joelberkeley-secondmind, line 16: what's the use case for this? we should be certain of the shapes of tensors we're using

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+        data_query_points.append(datasets[tag].query_points)
+        data_observations.append(datasets[tag].observations)
+
+    assert tf.math.reduce_all(
+        [tf.math.equal(input_data, data_query_points[0]) for input_data in data_query_points]
+    ), ValueError("Input data is not the same for each objective")
+
+    mo_dataset = Dataset(data_query_points[0], tf.concat(data_observations, axis=1))
+
+    return mo_dataset
```
joelberkeley-secondmind, line 34: it looks like this is designed to allow users to express their multi-objective problem using multiple datasets. In particular, it's only used to reorganise function arguments. Is that right?

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+        self.ub = tf.zeros([0, outdim], dtype=dtype)
+
+    def size(self) -> tf.Tensor:
+        """
+        :return: volume of each bounded volume
+        """
+        return tf.math.reduce_prod(self.ub - self.lb, axis=0, keepdims=True)
+
+
+class Pareto:
```
joelberkeley-secondmind, line 98: could this be immutable?

tests/unit/test_pareto.py
```diff
@@ -0,0 +1,177 @@
...
+# limitations under the License.
+
+import numpy as np
+import numpy.testing as npt
+import pytest
+import tensorflow as tf
+
+from trieste.data import Dataset
+from trieste.utils.pareto import Pareto, non_dominated_sort
+
```
joelberkeley-secondmind, line 22: could this be in the same style as the rest of the tests? `test_dominated_sort` matches that style, but `TestPareto` uses a different style

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+            )
+
+    def hypervolume(self, reference: tf.Tensor) -> tf.Tensor:
+        """
+        Autoflow method to calculate the hypervolume indicator
+        The hypervolume indicator is the volume of the dominated region.
+        :param reference: reference point to use
+            Should be equal or bigger than the anti-ideal point of the Pareto set
+            For comparing results across runs the same reference point must be used
+        :return: hypervolume indicator (the higher the better)
```
joelberkeley-secondmind, line 294:
```suggestion
        Autoflow method to calculate the hypervolume indicator
        The hypervolume indicator is the volume of the dominated region.

        :param reference: reference point to use
            Should be equal or bigger than the anti-ideal point of the Pareto set
            For comparing results across runs the same reference point must be used
        :return: hypervolume indicator (the higher the better)
```
paragraph needed for docs to render correctly

joelberkeley-secondmind: applies to all docstrings

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def size(self) -> tf.Tensor:
+        """
+        :return: volume of each bounded volume
+        """
+        return tf.math.reduce_prod(self.ub - self.lb, axis=0, keepdims=True)
+
+
+class Pareto:
+    def __init__(self, datasets: Mapping[str, Dataset], threshold: tf.Tensor = 0):
```
joelberkeley-secondmind, line 99: what if these datasets include data that doesn't correspond to an objective e.g. constraint data?

trieste/utils/pareto.py
joelberkeley-secondmind: and where in the loop would this be updated?

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+    def _is_test_required(smaller: tf.Tensor) -> tf.Tensor:
+        """
+        Tests if a point augments or dominates the Pareto set.
+        :param smaller: a boolean tf.Tensor storing test point < Pareto front
+        :return: True if the test point dominates or augments the Pareto front (boolean)
+        """
+        # if and only if the test point is at least in one dimension smaller
+        # for every point in the Pareto set
+        idx_dom_augm = tf.reduce_any(smaller, axis=1)
+        is_dom_augm = tf.reduce_all(idx_dom_augm)
```
joelberkeley-secondmind, line 131: there are a lot of very short names. It would help readability if they were more descriptive

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+    def _update_front(self) -> tf.Tensor:
+        """
+        Calculate the non-dominated set of points based on the latest data.
+        The stored Pareto set is sorted on the first objective in ascending order.
+        :return: boolean, whether the Pareto set has actually changed since the last iteration
+        """
+        current = self.front
+        pf, _ = non_dominated_sort(self.data)
+        self.front = tf.gather_nd(pf, tf.expand_dims(tf.argsort(pf[:, 0]), axis=1))
+        return not np.array_equal(current.numpy(), self.front.numpy())
```
joelberkeley-secondmind, line 144: array equality is a tricky one. How do we handle points that are effectively equal but not exactly due to precision errors?

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def divide_conquer_nd(self):
+        """
+        Divide and conquer strategy to compute the cells covering the non-dominated region.
+        Generic version: works for an arbitrary number of objectives.
+        """
+        outdim = self.data.observations.shape[1]
+
+        # The divide and conquer algorithm operates on a pseudo Pareto set
+        # that is a mapping of the real Pareto set to discrete values
```
joelberkeley-secondmind, line 186: comments degrade over time, and can very quickly become counter-productive. Can we leave them out?

joelberkeley-secondmind:
There's quite a lot of new code in this PR. Is there any way we could break it up into smaller meaningful PRs?

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+def non_dominated_sort(datasets: Dataset) -> TensorType:
+    """
+    Computes the non-dominated set for a set of data points
+    :param dataset: data points
+    :return: tuple of the non-dominated set and the degree of dominance,
+        dominances gives the number of dominating points for each data point
+    """
+    observations = datasets.observations
+    extended = tf.tile(tf.expand_dims(observations, 0), [observations.shape[0], 1, 1])
+    swapped_ext = tf.einsum("ij...->ji...", extended)
```
joelberkeley-secondmind, line 46: this assumes `observations` has rank 2, so to avoid einsum, we can also write
```
tf.transpose(tf.tile(xs[None, ...], [len(xs), 1, 1]), [1, 0, 2])
```
or even
```
tf.reshape(tf.tile(xs, [1, len(xs)]), [len(xs), len(xs), -1])
```

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+    dominance = tf.reduce_sum(
+        tf.cast(
+            tf.logical_and(
+                tf.reduce_all(extended <= swapped_ext, axis=2),
+                tf.reduce_any(extended < swapped_ext, axis=2),
+            ),
+            tf.float32,
```
joelberkeley-secondmind, line 53: why the cast?

trieste/utils/pareto.py
joelberkeley-secondmind, line 46: this assumes `observations` has rank 2, so to avoid einsum (which is slow), we can also write
```
tf.transpose(extended, [1, 0, 2])
```
or even, using `len(dataset.observations) == len(dataset)`,
```
len_data = len(dataset)
tf.reshape(tf.tile(dataset.observations, [1, len_data]), [len_data, len_data, -1])
```

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+    assert tf.math.reduce_all(
+        [tf.math.equal(input_data, data_query_points[0]) for input_data in data_query_points]
+    ), ValueError("Input data is not the same for each objective")
+
+    mo_dataset = Dataset(data_query_points[0], tf.concat(data_observations, axis=1))
+
+    return mo_dataset
+
+
+def non_dominated_sort(datasets: Dataset) -> TensorType:
```
joelberkeley-secondmind, line 37: this doesn't actually sort the data afaict, so perhaps `non_dominated` would be more accurate?

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+        # Initialize
+        self.update()
+
+    @staticmethod
+    def _is_test_required(smaller: tf.Tensor) -> tf.Tensor:
+        """
+        Tests if a point augments or dominates the Pareto set.
+        :param smaller: a boolean tf.Tensor storing test point < Pareto front
+        :return: True if the test point dominates or augments the Pareto front (boolean)
+        """
```
joelberkeley-secondmind, line 127: if a method has docstrings, it will appear in the docs. I'm guessing that's not intentional here

trieste/utils/pareto.py
joelberkeley-secondmind, line 98: since this is a Pareto set, might `ParetoSet` be a more descriptive name?

satrialoka:
> 
> 
> There's quite a lot of new code in this PR. Is there any way we could break it up into smaller meaningful PRs?

Thank you for the code review. before fixing other things I think it is better to discuss this first.
One way that comes into my mind is to split it to (sequentially):

1. Non dominated sort function
2. Pareto set without cell bound calculations. the cell bounds calculation is needed for specific algos that will be ported by Jixiang.
3. Cell bound computation both the divide and conquer and 2d approach
4. Hypervolume method

or maybe we can put 2 and 3 into one PR (this will be correspond  with the current unit test).
what do you think?

joelberkeley:
Non dominated sort sounds like a good place to start. For the others, it would help to know more about @TsingQAQ 's work, since I'll need to know what these functionalities are needed for to know if they deliver that functionality

TsingQAQ:
Please find the current `Pareto.py` based implementation of [`hvpi`](https://github.com/TsingQAQ/trieste/blob/e8138c32e5e8f26ef09be0f1b368972ce50c0d76/trieste/acquisition/function.py#L377)

There is also a [notebook](https://github.com/TsingQAQ/trieste/blob/hvpi/docs/notebooks/multi_objective_hvpi.pct.py) in the same branch

joelberkeley:
only `.front` and `.bounds` are used in @TsingQAQ 's work atm. That suggests to me that we don't need any of the other public methods (that is, they don't need to be public). This also mean we have no need to update `Pareto`. Is that right?

satrialoka:
seems it can be changed to private methods with @TsingQAQ implementation. still, update method is used [in the constructor](https://github.com/satrialoka/trieste/blob/dd1c93159c97172163605ad5be49caf541c344c5/trieste/utils/pareto.py#L119) (maybe we can change it with other name). in the original [Gpflowopt implementation](https://github.com/GPflow/GPflowOpt/blob/3d86bcc000b0367f19e9f03f4458f5641e5dde60/gpflowopt/acquisition/hvpoi.py#L93) the `update` method are called at each BO iteration and the output of the models predictions are used instead of the observed value. Not sure if @TsingQAQ  implementation also do this, because i cannot find `update` being called either.

TsingQAQ:
Indeed, currently the whole `pareto` class is just reinitialized in each iteration.  
Agree with @satrialoka, use `update` might be a better choice, since as long as the front is not changing, the bound will not get recalculated, might be more computationally efficient.

joelberkeley:
the acquisition package uses only immutable constructs, and I'm more than keen to maintain that design. If we want to pass state between acquisition steps, we could create a dedicated acquisition rule and builder, but this would have a number of its own downsides. How likely is it that the Pareto front will be unchanged between any two steps?

> seems it can be changed to private methods with @TsingQAQ implementation. still, update method is used [in the constructor](https://github.com/satrialoka/trieste/blob/dd1c93159c97172163605ad5be49caf541c344c5/trieste/utils/pareto.py#L119) (maybe we can change it with other name)

I'm in favour of moving the content of `update` to `__init__`, and avoiding mutable state which comes with a number of drawbacks. I think this should be straightfoward here

> in the original [Gpflowopt implementation](https://github.com/GPflow/GPflowOpt/blob/3d86bcc000b0367f19e9f03f4458f5641e5dde60/gpflowopt/acquisition/hvpoi.py#L93) the `update` method are called at each BO iteration and the output of the models predictions are used instead of the observed value. Not sure if @TsingQAQ implementation also do this, because i cannot find `update` being called either.

if you wanted it to use `model.predict(dataset.query_points)` is that not already possible? Which ever we go with, it would be a small amount of work for users to make a new builder that does the opposite. This is similar to how EI works.

TsingQAQ:
>How likely is it that the Pareto front will be unchanged between any two steps  

Speak from experiments (in the above mentioned notebook):
- for the bi-objective vlmop2 cases: The Pareto front get *changed* for 16 iter / 20 total iter; 44 iter / 50 iter. 
- for the 3-objective dtlz2: The Pareto front get *changed* for 46 iter / 50 total iter;  64 iter / 80 total iter.

trieste/utils/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+
+        # Start divide and conquer until we processed all cells
+        while dc:
+            # Process test cell
+            cell = dc.pop()
+            arr = tf.range(outdim)
+            idx_lb = tf.gather_nd(pf_ext_idx, tf.stack((cell[0], arr), -1))
+            idx_ub = tf.gather_nd(pf_ext_idx, tf.stack((cell[1], arr), -1))
+            lb = tf.gather_nd(pf_ext, tf.stack((idx_lb, arr), -1))
+            ub = tf.gather_nd(pf_ext, tf.stack((idx_ub, arr), -1))
```
joelberkeley, line 224: a general note: `gather_nd` is slow, and you've used it liberally in this PR. If you're able to find faster functions, that'd be good

trieste/utils/pareto.py
satrialoka: `tf.reduce_sum` cant work on boolean returned by `tf.logical_and`. should I use `tf.math.count_nonzero` or keep the cast and use gpflow default float?

joelberkeley: oh yes, of course. might an integer type be suitable?

trieste/utils/pareto.py
satrialoka: kind of. it's used to reorganize multiple datasets into one `Datasets` with multiple outputs. we come up with this because trieste can't handle two different models with one Dataset with multiple y(output) CMIIW. [link](https://github.com/secondmind-labs/trieste/blob/0e8f34e87b9e1a0ec2ea6fd2fd857eff49c202ab/trieste/bayesian_optimizer.py#L233)

joelberkeley:
i'm going to close this PR as the work is being spread across several PRs

PR #131, opened 2021-02-01
default_role for sphinx

st--:
Makes it more likely to end up with links. See https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -121,3 +121,5 @@ html_theme_options = {
     "show_prev_next": False,
     "github_url": "https://github.com/secondmind-labs/trieste",
 }
+
+default_role = "any"  # turn all ` ` into links when possible


joelberkeley-secondmind:
thanks. were there any particular links that you noticed this adds?

separately, this produces a lot of warnings in the doc build. I've been trying to fix these and am down to two (partly so I can add the doc build to CI at some point). Do you know some way of preventing the warnings?

st--:
E.g. the cross-references between AcquisitionFunction and BatchAcquisitionFunction in https://secondmind-labs.github.io/trieste/autoapi/trieste/acquisition/index.html, or the references in https://secondmind-labs.github.io/trieste/autoapi/trieste/bayesian_optimizer/index.html to OBJECTIVE.

Warnings should be for one of two reasons:
1. No reference is found. Replace ``` `text` ``` with ` ``text`` ` for setting in mono-type without linking.
2. Multiple references are found. Disambiguate by e.g. adding a rule such as ```:class:`text` ```.

joelberkeley:
I'm not so sure about this, as it's making incorrect links (see the reference to `model` in `acquisition.SingleModelMCIndAcquisitionFunctionBuilder`, which refers to a function parameter but is being linked to an attribute of a completely unrelated dataclass). I guess this is the trade off with automagic. Re the references to `[Batch]AcquisitionFunction` and `OBJECTIVE`, these can be enabled with `:const:` and fully qualified names

joelberkeley:
actually, one sec, that might not be a problem

st--:
You should be able to get always-link-free monotype by using double backticks as in ``` ``model`` ```. And yes, this doesn't give you anything you couldn't have already, it just makes it easier to have cross-links in docstrings without adding the rules and fully-qualified names all the time. Take it or leave it:)

joelberkeley:
i'm interested in this, but want to make sure the docs remain accurate, so for now i've moved it to a github issue #134

PR #133, opened 2021-02-03
add batch EI acquisition function

### LINKED ISSUE #41: MC-based batch-EI

joelberkeley:
Adds the batch expected improvement acquisition function

### DIFF
--- a/tests/unit/acquisition/test_function.py
+++ b/tests/unit/acquisition/test_function.py
@@ -22,6 +22,7 @@ import tensorflow as tf
 import tensorflow_probability as tfp
 
 from tests.util.misc import (
+    TF_DEBUGGING_ERROR_TYPES,
     ShapeLike,
     mk_dataset,
     quadratic,
@@ -29,13 +30,13 @@ from tests.util.misc import (
     random_seed,
     various_shapes,
     zero_dataset,
-    TF_DEBUGGING_ERROR_TYPES,
 )
 from tests.util.model import GaussianProcess, QuadraticMeanAndRBFKernel, rbf
 from trieste.acquisition import SingleModelBatchAcquisitionBuilder
 from trieste.acquisition.function import (
     AcquisitionFunction,
     AcquisitionFunctionBuilder,
+    BatchMonteCarloExpectedImprovement,
     BatchReparametrizationSampler,
     ExpectedConstrainedImprovement,
     ExpectedImprovement,
@@ -48,7 +49,6 @@ from trieste.acquisition.function import (
     expected_improvement,
     lower_confidence_bound,
     probability_of_feasibility,
-    BatchMonteCarloExpectedImprovement,
 )
 from trieste.data import Dataset
 from trieste.models import ProbabilisticModel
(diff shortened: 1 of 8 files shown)


trieste/acquisition/function.py
```diff
@@ -625,3 +666,59 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> tf.Tenso
...
+        )
+
+        eta = tf.reduce_min(mean, axis=0)
+        sampler = BatchReparametrizationSampler(self._sample_size, model)
+
+        def batch_ei(at: TensorType) -> TensorType:
+            samples = tf.squeeze(sampler.sample(at, jitter=self._jitter), axis=-1)  # [..., S, B]
+            min_sample_per_batch = tf.reduce_min(samples, axis=-1)  # [..., S]
+            batch_improvement = tf.maximum(eta - min_sample_per_batch, 0.0)  # [..., S]
+            return tf.reduce_mean(batch_improvement, axis=-1, keepdims=True)  # [..., 1]
```
joelberkeley, line 727: you might want to double check this, as I've reordered it to be faster (found samples min before maximum)

vpicheny: LGTM

tests/unit/acquisition/test_function.py
```diff
@@ -546,20 +584,76 @@ def test_batch_reparametrization_sampler_samples_are_distinct_for_new_instances(
...
+
+@random_seed
+def test_batch_monte_carlo_expected_improvement() -> None:
+    xs = tf.random.uniform([3, 5, 7, 2], dtype=tf.float64)
+    model = QuadraticMeanAndRBFKernel()
+
+    mean, cov = model.predict_joint(xs)
+    mvn = tfp.distributions.MultivariateNormalFullCovariance(tf.linalg.matrix_transpose(mean), cov)
+    mvn_samples = mvn.sample(10_000)
+    expected = tf.reduce_mean(tf.reduce_max(tf.maximum(0.09 - mvn_samples, 0.0), axis=-1), axis=0)
```
vpicheny: are we sure that eta=0.09 for acq?

vpicheny: I would rather use `eta = tf.reduce_min(mean, axis=0)` instead of the hard-coded 0.09 above

vpicheny (approved):
looks good!

tests/unit/acquisition/test_function.py
joelberkeley: `QuadraticMeanAndRBFKernel` is a `GaussianProcess`, which outputs precisely the mean you tell it to. In this case, that's the quadratic function, so 0.3 ** 2

joelberkeley: I could make that clearer by changing the 0.09 in the data below to have some noise on it, to make it clear they're not the same thing?

vpicheny: Don't add noise! this is fine. I just find it hard to follow, there's no visible connexion between the model mean and 0.09. 
I may just not be used to the way tests ought to be written

joelberkeley: i could
```
min_predictive_mean_at_known_points = 0.09
... (min_predictive_mean_at_known_points - mvn_samples) ...
```

vpicheny: I like that

PR #138, opened 2021-02-07
update mypy

joelberkeley:


sam-willis (approved):
I've never seen these new style annotations before. They look much nicer, but I can't tell you if there's any pitfalls (because I'm completely ignorant). 

But you should merge if you're confident this works.

PR #140, opened 2021-02-08
add py.typed

### LINKED ISSUE #103: Ensure type information is packaged with installable
As a user, I want to be able to type check my code that calls into trieste. It's not clear at the moment if this is currently possible, given there is no py.typed file.

See the python PEP for more info https://www.python.org/dev/peps/pep-0561/

jesnie:
`py.typed` tells `mypy` to follow imports into this package, allowing users of the package to find more bugs with `mypy`.

Further reading:
https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages
https://www.python.org/dev/peps/pep-0561/#packaging-type-information

Fixes #103

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,9 @@ setup(
     long_description_content_type="text/markdown",
     url="https://github.com/secondmind-labs/trieste",
     packages=find_packages(include=("trieste*",)),
+    package_data={
+        "trieste": ["py.typed"],
+    },
     classifiers=[
         "Programming Language :: Python :: 3.7",
         "License :: OSI Approved :: Apache Software License",


joelberkeley:
thanks jesper. Did you test it works?

joelberkeley:
ok i've tested it works when installing from source. Do we reckon that's enough?

jesnie:
> thanks jesper. Did you test it works?

Yes, by installing `trieste` into a virtual environment and running `mypy` against this script:

```python
from trieste.space import Box

x: Box = "foo"
```

jesnie:
> ok i've tested it works when installing from source. Do we reckon that's enough?

I don't know how else to test it...

joelberkeley (approved):

joelberkeley:
the other way would be to package trieste into a wheel and install that, but I doubt that'll be any different since they're bpth using setup.py

PR #141, opened 2021-02-08
add build dependencies to pyproject.toml

joelberkeley:
This PR completes the pyproject.toml file according to the [python docs](https://packaging.python.org/tutorials/packaging-projects/). This is needed to follow those instructions for the next trieste release


jesnie (approved):
LGTM

This seems to work.

PR #149, opened 2021-02-15
add batch EI example to tutorials

vpicheny:
This PR adds a section at the end of the `expected_improvement` notebook to show how to use batch-EI. After running the loop, it shows the obtained GP and query points, and compares regret with the one-point-at-a-time run.

### DIFF
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -207,7 +207,7 @@ plot_bo_points(
 # Sometimes it is practically convenient to query several points at a time. We can do this in `trieste` using `BatchMonteCarloExpectedImprovement`. Note that this acquisition is computed using a Monte-Carlo method (so it requires a `sample_size`), but with a reparametrisation trick, which makes it deterministic. To use it, we change the acquisition rule and specify the number of query points (`num_query_points`) to be chosen, then observed simultaneously:
 
 # %%
-qei = trieste.acquisition.BatchMonteCarloExpectedImprovement(sample_size=100)
+qei = trieste.acquisition.BatchMonteCarloExpectedImprovement(sample_size=1000)
 batch_rule = trieste.acquisition.rule.BatchAcquisitionRule(num_query_points=3, builder=qei.using(OBJECTIVE))
 
 batch_bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -204,7 +204,7 @@ plot_bo_points(
 # %% [markdown]
 # ## Batch-sequential strategy
 #
-# Sometimes it is practically more efficient to query several points at a time. We can do this in `trieste` using `BatchMonteCarloExpectedImprovement`. Note that this acquisition is computed using a Monte-Carlo method, but with a reparametrisation trick, which makes it deterministic. To use, it we change the acquisition rule:
+# Sometimes it is practically convenient to query several points at a time. We can do this in `trieste` using `BatchMonteCarloExpectedImprovement`. Note that this acquisition is computed using a Monte-Carlo method (so it requires a `sample_size`), but with a reparametrisation trick, which makes it deterministic. To use it, we change the acquisition rule and specify the number of query points (`num_query_points`) to be chosen, then observed simultaneously:
 
 # %%
 qei = trieste.acquisition.BatchMonteCarloExpectedImprovement(sample_size=100)
@@ -218,7 +218,7 @@ batch_query_points = batch_dataset.query_points.numpy()
 batch_observations = batch_dataset.observations.numpy()
 
 # %% [markdown]
-# We can visualise again the model and query points.
+# We can visualise again the GP model and query points.
 
 # %%
 fig = plot_gp_plotly(
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -242,7 +242,7 @@ fig = add_bo_points_plotly(
 fig.show()
 
 # %% [markdown]
-# We can also compare the regrets between the purely sequential approach and the batch one. 
+# We can also compare the regret between the purely sequential approach and the batch one. 
 
 # %%
 _, ax = plt.subplots(1, 2)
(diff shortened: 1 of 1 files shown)


docs/notebooks/expected_improvement.pct.py
```diff
@@ -201,6 +201,54 @@ def build_model(data):
     idx_best=arg_min_idx,
 )
 
+# %% [markdown]
+# ## Batch-sequential strategy
```
joelberkeley, line 205: is this notebook is getting a little long? remember we also have MES in here (at least we will do when we test it)

docs/notebooks/expected_improvement.pct.py
```diff
@@ -201,6 +201,54 @@ def build_model(data):
...
+#
+# Sometimes it is practically convenient to query several points at a time. We can do this in `trieste` using `BatchMonteCarloExpectedImprovement`. Note that this acquisition is computed using a Monte-Carlo method (so it requires a `sample_size`), but with a reparametrisation trick, which makes it deterministic. To use it, we change the acquisition rule and specify the number of query points (`num_query_points`) to be chosen, then observed simultaneously:
```
joelberkeley: when you say

> We can do this in `trieste` using `BatchMonteCarloExpectedImprovement`

it might be helpful to explain more about what's going on here: that we use the batch acquisition rule and a batch acquisition function such as batch EI (to make clear we don't have to use batch EI for batch BO). Together, these can recommend a number of points. The optimizer then queries the observer at all these points

docs/notebooks/expected_improvement.pct.py
```diff
@@ -201,6 +201,54 @@ def build_model(data):
...
+    idx_best=arg_min_idx,
+    fig=fig,
+    figrow=1,
+    figcol=1,
+)
+
+fig.show()
+
+# %% [markdown]
+# We can also compare the regrets between the purely sequential approach and the batch one.
```
joelberkeley:
```suggestion
# We can also compare the regret between the purely sequential approach and the batch one. 
```
might read better

docs/notebooks/expected_improvement.pct.py
```diff
@@ -201,6 +201,54 @@ def build_model(data):
...
+
+batch_bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+
+batch_result = batch_bo.optimize(5, initial_data, model, acquisition_rule=batch_rule)
+batch_dataset = batch_result.try_get_final_datasets()[OBJECTIVE]
+batch_query_points = batch_dataset.query_points.numpy()
+batch_observations = batch_dataset.observations.numpy()
+
+# %% [markdown]
+# We can visualise again the GP model and query points.
```
joelberkeley:
```suggestion
# We can again visualise the GP model and query points.
```

docs/notebooks/expected_improvement.pct.py
joelberkeley: I'm also a little confused by

> Sometimes it is practically convenient to query several points at a time.

it's not immediately clear to me that you mean query the observer, rather than query the acquisition function. Could you also clarify "practically convenient"?

docs/notebooks/expected_improvement.pct.py
vpicheny: Indeed it's long. I suggest to keep it this way for now, but once MES is here, we split it in half, so the first one only contains the basics and the second has MES and qEI, and compares the regrets of both with EI.

docs/notebooks/expected_improvement.pct.py
vpicheny: I meant that in some cases you can query several points at a time with little extra cost (for instance if you can parallelize the calls to an expensive observer).

docs/notebooks/expected_improvement.pct.py
```diff
@@ -201,6 +201,54 @@ def build_model(data):
...
+# %% [markdown]
+# ## Batch-sequential strategy
+#
+# Sometimes it is practically convenient to query several points at a time. We can do this in `trieste` using `BatchMonteCarloExpectedImprovement`. Note that this acquisition is computed using a Monte-Carlo method (so it requires a `sample_size`), but with a reparametrisation trick, which makes it deterministic. To use it, we change the acquisition rule and specify the number of query points (`num_query_points`) to be chosen, then observed simultaneously:
+
+# %%
+qei = trieste.acquisition.BatchMonteCarloExpectedImprovement(sample_size=1000)
+batch_rule = trieste.acquisition.rule.BatchAcquisitionRule(num_query_points=3, builder=qei.using(OBJECTIVE))
+
+batch_bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
```
vpicheny: @joelberkeley do we need to redefine that?

docs/notebooks/expected_improvement.pct.py
```diff
@@ -201,6 +201,54 @@ def build_model(data):
...
+
+batch_result = batch_bo.optimize(5, initial_data, model, acquisition_rule=batch_rule)
```
vpicheny: rebuild model?

docs/notebooks/expected_improvement.pct.py
joelberkeley: no

docs/notebooks/expected_improvement.pct.py
joelberkeley: yes ... just `build_model(...)` again

docs/notebooks/expected_improvement.pct.py
joelberkeley: is it worth adding that? or will it be obvious to most readers?

vpicheny: I think this is obvious

docs/notebooks/expected_improvement.pct.py
```diff
@@ -201,6 +201,54 @@ def build_model(data):
...
+
+# %%
+qei = trieste.acquisition.BatchMonteCarloExpectedImprovement(sample_size=1000)
+batch_rule = trieste.acquisition.rule.BatchAcquisitionRule(num_query_points=3, builder=qei.using(OBJECTIVE))
+
+model = build_model(initial_data[OBJECTIVE])
+batch_result = bo.optimize(5, initial_data, model, acquisition_rule=batch_rule)
+batch_dataset = batch_result.try_get_final_datasets()[OBJECTIVE]
+batch_query_points = batch_dataset.query_points.numpy()
+batch_observations = batch_dataset.observations.numpy()
```
joelberkeley: i've approved, but perhaps move this down a cell, so users can see this `numpy`-ing is plotting specific and they don't need to do it if they don't want?

joelberkeley (approved):

docs/notebooks/expected_improvement.pct.py
joelberkeley: entirely up to you

joelberkeley:
nice. it would be good to hear how you found working with the batch API

PR #153, opened 2021-02-16
revert CodeQL analysis

joelberkeley:
reverts #146 which was a trial


vpicheny:
why did we try and why do we not want it anymore?

joelberkeley:
@vpicheny we tried it so james robson could test it. It's possible we'll use it in future, but it was rushed in (I didn't actually look at what it did when we merged). For now I've made a ticket to consider it #154

vpicheny (approved):

PR #168, opened 2021-02-22
add Gramacy and Lee toy objective

joelberkeley:


tests/integration/test_bayesian_optimization.py
```diff
@@ -73,9 +68,8 @@ def build_model(data: Dataset) -> GaussianProcessRegression:
     best_y = dataset.observations[arg_min_idx]
     best_x = dataset.query_points[arg_min_idx]
 
-    argmin = tf.cast(BRANIN_GLOBAL_ARGMIN, tf.float64)
-    relative_argmin_err = tf.abs((best_x - argmin) / argmin)
+    relative_minimizer_err = tf.abs((best_x - BRANIN_MINIMIZERS) / BRANIN_MINIMIZERS)
```
vpicheny, line 71: Branin has several minimizers, we would care here about best_x being close to one of them, not all of them, right?

joelberkeley: indeed, that's what we're doing. btw I've just renamed things and changed dtypes in this file

vpicheny (approved):

PR #173, opened 2021-02-25
Notebook inequality constraints fixes

tensorlicious:
fix references in inequality_contraints notebook
clarify description in `inequality_constraints.pct.py`
fix inequality condition for constraint in last plot

### DIFF
--- a/docs/notebooks/inequality_constraints.pct.py
+++ b/docs/notebooks/inequality_constraints.pct.py
@@ -11,7 +11,7 @@ tf.random.set_seed(1793)
 # %% [markdown]
 # ## The problem
 #
-# In this tutorial, we replicate one of the results of <cite data-cite="gardner14"/>, specifically their synthetic experiment "simulation 1", which consists of an objective function with a single constraint, defined over a two-dimensional input domain. We'll start by defining the problem parameters.
+# In this tutorial, we replicate one of the results of [[GKZ+14](https://secondmind-labs.github.io/trieste/autoapi/trieste/index.html#id4)], specifically their synthetic experiment "simulation 1", which consists of an objective function with a single constraint, defined over a two-dimensional input domain. We'll start by defining the problem parameters.
 
 # %%
 import gpflow
@@ -108,7 +108,7 @@ initial_models = {
 # %% [markdown]
 # ## Define the acquisition process
 #
-# We can construct the _expected constrained improvement_ acquisition function defined in <cite data-cite="gardner14"/>, where they use the probability of feasibility wrt the constraint model.
+# We can construct the _expected constrained improvement_ acquisition function defined in [[GKZ+14](https://secondmind-labs.github.io/trieste/autoapi/trieste/index.html#id4)], where they use the probability of feasibility with respect to the constraint model.
 
 # %%
 pof = trieste.acquisition.ProbabilityOfFeasibility(threshold=Sim.threshold)
--- a/docs/notebooks/inequality_constraints.pct.py
+++ b/docs/notebooks/inequality_constraints.pct.py
@@ -11,7 +11,7 @@ tf.random.set_seed(1793)
 # %% [markdown]
 # ## The problem
 #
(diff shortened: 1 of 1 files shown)


docs/notebooks/inequality_constraints.pct.py
```diff
@@ -11,7 +11,7 @@
 # %% [markdown]
 # ## The problem
 #
-# In this tutorial, we replicate one of the results of <cite data-cite="gardner14"/>, specifically their synthetic experiment "simulation 1", which consists of an objective function with a single constraint, defined over a two-dimensional input domain. We'll start by defining the problem parameters.
+# In this tutorial, we replicate one of the results of [[GKZ+14](https://secondmind-labs.github.io/trieste/autoapi/trieste/index.html#id4)], specifically their synthetic experiment "simulation 1", which consists of an objective function with a single constraint, defined over a two-dimensional input domain. We'll start by defining the problem parameters.  The constraint is satisfied when `constraint(input_data) <= threshold`.
```
joelberkeley: these hyperlinks will link to the bibliography on the remote site, while they should point to the bibliography in the current doc build. This is so that when we build the docs locally, it points to the right place

I assume you did this so the links work when run as notebooks?

tensorlicious: Yes, open to alternatives, but it wasn't working as it was and was also breaking the formatting in the notebook.

tensorlicious: is it sufficient to hardlink to the paper:
http://proceedings.mlr.press/v32/gardner14.html
?

joelberkeley: that would lose the benefit the sphinxcontrib-bibtex plugin, which I'm not keen to lose so quickly. I'm looking at the syntax for the `<cite data-cite=...>`. Perhaps there's a way of doing that differently that will work nicely.

I've prioritised the docs website over the notebooks , having assumed that's what people will see first and will use more often

joelberkeley: this page shows how to do citations in jupyter https://nbviewer.jupyter.org/github/ipython/nbconvert-examples/blob/master/citations/Tutorial.ipynb

joelberkeley: `<cite data-cite="gardner14">Gardner et al.</cite>` keeps the link as is in the docs, but shows _`Gardner et al.`_ in the notebook, which is an improvement but not perfect

joelberkeley: you can do the whole thing with
```
<cite data-cite="gardner14">[Gardner et al.](http://proceedings.mlr.press/v32/gardner14.html)</cite>
```

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -239,13 +239,14 @@ def batch_efi(at):
...
+# Finally, we compare the regret from the non-batch strategy (left panel) to the regret from the batch strategy (right panel).
+# In the following plots each marker represents a query point. The x-axis is the index of the query point (where the first queried point has index 0), and the y-axis is the observed value. The vertical blue line denotes the end of initialisation/start of optimisation. Green points satisfy the constraint, red points do not.
 
 # %%
 from util.plotting import plot_regret
 
-mask_fail = constraint_data.observations.numpy() >= Sim.threshold
-batch_mask_fail = batch_constraint_data.observations.numpy() >= Sim.threshold
+mask_fail = constraint_data.observations.numpy() > Sim.threshold
+batch_mask_fail = batch_constraint_data.observations.numpy() > Sim.threshold
```
joelberkeley, line 249: @vpicheny is this right?

vpicheny: yes, fail is when > threshold.

vpicheny: it propagates to plot regret in that way:
`safe_obs_values[mask_fail] = np.max(obs_values)`

so when mask = 1, ignore the value for the regret plot

joelberkeley (approved):

Issue #57, opened 2020-11-15
Comprehensive unit tests


joelberkeley:
While everything has _some_ level of testing, unit test coverage for the toolbox is not comprehensive. Missing tests are mostly either unhappy path tests incl. error handling, or integration tests used instead of unit tests.

The first task is to go through the code, look at the tests, and check what already has comprehensive test coverage.

Comprehensive unit tests exist for (list updated as tests are checked/added):

- acquisition
  - [x] combination #57
  - function
    - [x] mes (except for full functional form ... see #52 )
    - [x] everything except mes #62 (note we need to update the `QuadraticMeanAndRBFKernel` utility to have non-trivial minimum and variance, and update the tests respectively, for these to be good tests, see #65 )
  - rule
    - [ ] EfficientGlobalOptimization
    - [ ] ThompsonSampling
    - [ ] TrustRegion (rather than testing this more, we could pull it out into a separate thing and the current tests will cover it, and then we'll also get the benefit of using it everywhere)
- models
  - [ ] config
  - [ ] model_interfaces
  - [ ] optimizer
- utils
  - [x] misc
  - [x] objectives
  - [ ] pareto
- [x] bayesian_optimizer #17 
- [x] data #60
- [x] observer #58
- [x] space #59 #80 
- [x] type

uri-granta:
Unit test coverage has been improved and we now track code coverage. More work to be done here but this issue is no longer useful for tracking it.

PR #174, opened 2021-02-25
add tests for AcquisitionFunctionBuilder reducers & refactor

joelberkeley:
This PR completes the test coverage for the `AcquisitionFunctionBuilder` reducers, and refactors the remaining tests to have lighter dependencies. It's part of #57

### DIFF
--- a/trieste/acquisition/combination.py
+++ b/trieste/acquisition/combination.py
@@ -26,9 +26,9 @@ from .function import AcquisitionFunction, AcquisitionFunctionBuilder
 
 class Reducer(AcquisitionFunctionBuilder):
     r"""
-    A :class:`Reducer` builds an :data:`~trieste.acquisition.AcquisitionFunction` whose output is
+    A :class:`Reducer` builds an :const:`~trieste.acquisition.AcquisitionFunction` whose output is
     calculated from the outputs of a number of other
-    :data:`~trieste.acquisition.AcquisitionFunction`\ s. How these outputs are composed is defined
+    :const:`~trieste.acquisition.AcquisitionFunction`\ s. How these outputs are composed is defined
     by the method :meth:`_reduce`.
     """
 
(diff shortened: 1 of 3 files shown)


trieste/acquisition/combination.py
```diff
@@ -24,24 +24,20 @@
...
+    A :class:`Reducer` builds an `~trieste.acquisition.AcquisitionFunction` whose output is
+    calculated from the outputs of a number of other `~trieste.acquisition.AcquisitionFunction`\ s.
+    How these outputs are composed is defined by the method :meth:`_reduce`.
     """
 
     def __init__(self, *builders: AcquisitionFunctionBuilder):
         r"""
         :param \*builders: Acquisition function builders. At least one must be provided.
-        :raise ValueError: If no builders are specified.
+        :raise ValueError (or InvalidArgumentError): If no builders are specified.
```
tensorlicious: can a `ValueError` be raised by this function?

joelberkeley: no it appears not. tf.debugging seems to raise InvalidArgumentError except for assert_shapes and other shape-based checks, where it can also raise ValueError. I'll change this to InvalidArgumentError

trieste/acquisition/combination.py
```diff
@@ -26,24 +26,21 @@
 
 class Reducer(AcquisitionFunctionBuilder):
     r"""
-    A :class:`Reducer` builds an :func:`~trieste.acquisition.AcquisitionFunction` whose output is
+    A :class:`Reducer` builds an :data:`~trieste.acquisition.AcquisitionFunction` whose output is
```
ChrisMorter: Is `:data:` intentional?

ChrisMorter (approved):
LGTM

trieste/acquisition/combination.py
joelberkeley: it is: `:func:` is for actual functions, `:data:` is the best option I could find for type aliases. I think it's a generic "this is a fixed module-level thing".

joelberkeley: `:const:` was another option, can't recall why I chose `:data:`

joelberkeley: oh, actually `:data:` is for module-level variables, so I'll use `:const:`

PR #177, opened 2021-03-01
add multi-output model wrapper

### LINKED ISSUE #176: Multioutput model wrapper for independent models
As a user, I want to be able to stack models so that I can model multiple outputs with a single model, when those outputs all have a common set of query points, and the outputs dtypes that can be combined into a single tensor (i.e. that the outputs and query points can be put in a single `Dataset`). This is so that I can do multi-objective optimization without ambiguity about which datasets and models are objectives, which are not, and which order these outputs are in when used e.g. in constructing the Pareto sets.

joelberkeley:
This PR allows users to combine independent models under a single `TrainableProbabilisticModel` interface so that multi-objective data can be contained within a single `Dataset` and trieste know how to handle it.

### DIFF
--- a/trieste/models/model_interfaces.py
+++ b/trieste/models/model_interfaces.py
@@ -102,7 +102,11 @@ class ModelStack(TrainableProbabilisticModel):
 
     **Note:** Only supports vector outputs (i.e. with event shape [E]). Outputs for any two models
     are assumed independent. Each model may itself be single- or multi-output, and any one
-    multi-output model may have dependence between its outputs.
+    multi-output model may have dependence between its outputs. When we speak of *event size* in
+    this class, we mean the output dimension for a given :class:`TrainableProbabilisticModel`,
+    whether that is the :class:`ModelStack` itself, or one of the subsidiary
+    :class:`TrainableProbabilisticModel`\ s within the :class:`ModelStack`. Of course, the event
+    size for a :class:`ModelStack` will be the sum of the event sizes of each subsidiary model.
     """
 
     def __init__(
--- a/trieste/models/model_interfaces.py
+++ b/trieste/models/model_interfaces.py
@@ -94,6 +94,95 @@ class TrainableProbabilisticModel(ProbabilisticModel):
         raise NotImplementedError
 
 
+class ModelStack(TrainableProbabilisticModel):
+    r"""
+    A :class:`ModelStack` is a wrapper around a number of :class:`TrainableProbabilisticModel`\ s.
+    It combines the outputs of each model for predictions and sampling, and delegates training data
+    to each model for updates and optimization.
+
+    **Note:** Only supports vector outputs (i.e. with event shape [E]). Outputs for any two models
+    are assumed independent. Each model may itself be single- or multi-output, and any one
+    multi-output model may have dependence between its outputs.
+    """
+
+    def __init__(
+        self,
+        model_with_event_size: tuple[TrainableProbabilisticModel, int],
+        *models_with_event_sizes: tuple[TrainableProbabilisticModel, int],
+    ):
+        r"""
+        The order of individual models specified at :meth:`__init__` determines the order of the
+        :class:`ModelStack` output dimensions.
+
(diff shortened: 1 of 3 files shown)


trieste/models/model_interfaces.py
```diff
@@ -94,6 +94,95 @@ def optimize(self, dataset: Dataset) -> None:
...
+        Optimize all the wrapped models on their corresponding data. The data for each model is
+        extracted by splitting the observations in ``dataset`` along the event axis according to the
+        event sizes specified at :meth:`__init__`.
+
+        :param dataset: The query points and observations for *all* the wrapped models.
+        """
+        observations = tf.split(dataset.observations, self._event_sizes, axis=-1)
+
+        for model, obs in zip(self._models, observations):
+            model.optimize(Dataset(dataset.query_points, obs))
```
henrymoss, line 187: Not needed for this PR, but I think we should allow a bit more flexibility over the model optimiser here. i.e similar to how we discussed with the acquisition optimiser.  It is very often the case that we might want to at least control the number of random restarts, especially in the early BO steps where there is little data.

trieste/models/model_interfaces.py
```diff
@@ -94,6 +94,95 @@ def optimize(self, dataset: Dataset) -> None:
...
+
+    def update(self, dataset: Dataset) -> None:
+        """
+        Update all the wrapped models on their corresponding data. The data for each model is
+        extracted by splitting the observations in ``dataset`` along the event axis according to the
+        event sizes specified at :meth:`__init__`.
+
+        :param dataset: The query points and observations for *all* the wrapped models.
+        """
+        observations = tf.split(dataset.observations, self._event_sizes, axis=-1)
```
henrymoss, line 171: I'm trying to understand what the dataset will look like here.

 So dataset.observations will have an additional dimension that corresponds to different events. So we have observations for each individual event? What exactly is an event here?

trieste/models/model_interfaces.py
```diff
@@ -94,6 +94,95 @@ def optimize(self, dataset: Dataset) -> None:
...
+    ):
+        r"""
+        The order of individual models specified at :meth:`__init__` determines the order of the
+        :class:`ModelStack` output dimensions.
+
+        :param model_with_event_size: The first model, and the size of its output events.
+            **Note:** This is a separate parameter to ``models_with_event_sizes`` simply so that the
+            method signature requires at least one model. It is not treated specially.
+        :param \*models_with_event_sizes: The other models, and sizes of their output events.
+        """
```
henrymoss, line 125: How likely is it that model orders could get swapped within an optimisation pipeline? If not very likely then this is fine.

I guess we could have this as a dict, with explicit names for models?, i.e. similar to the .using(TAG) idea.

trieste/models/model_interfaces.py
joelberkeley: the optimizer is configurable for each individual model. See the `optimizer` argument to the constructors for the GPflow model wrappers. Is this what you mean?

trieste/models/model_interfaces.py
joelberkeley: A ModelStack would be defined once when setting up the BO loop. It wouldn't be deconstructed (indeed there's no way to get the models out once they're in) so there's no chance for it to be mixed up. The user would have to make sure they put the models in the same order as they've put their observations in the observer

trieste/models/model_interfaces.py
joelberkeley: An event is a subset of the observations. For example, if I have 3 objectives, and I have a GPR for the first and SVGP for the last two, an event could be an observation of the first objective, or an observation of the last two objectives. The event sizes would be 1 and 2.

joelberkeley: I've used event in other places in trieste to mean all the observations, and sometimes to mean a query point. Mainly I use it to mean a set of observations which are not independent.

henrymoss (approved):

PR #184, opened 2021-03-03
Pareto class for two objectives

### LINKED ISSUE #179: Pareto set: two objectives
As a user, I want functionality available to calculate the cell bounds of a Pareto front for the case of two objectives, so that I can easily find the cell bounds when defining multi-objective acquisition functionality

satrialoka:
### Feature Implemented

This is a feature for Pareto functionality needed for Multi-objective BO, the code was ported from [GPFlowopt's pareto class](https://github.com/GPflow/GPflowOpt/blob/master/gpflowopt/pareto.py). This PR is small PR derived from previously large PR #126 
 
### Approach

Inside pareto.py  is implemented:

1. `Pareto` class:  Stores a Pareto set and calculates the cell bounds covering the non-dominated region. currently, in this PR only 2d method `_bounds_2d` for calculating the cell bounds is present.
2. `BoundedVolumes` class: Stores the index of the cell bounds

## Status

Need review

## Notes
In the original implementation on GPFlowOpt, the Pareto class is immutable, this makes it possible to not recalculate the bounds if the Pareto front unchanged [code](https://github.com/satrialoka/trieste/blob/dd1c93159c97172163605ad5be49caf541c344c5/trieste/utils/pareto.py#L168).

### DIFF
--- a/trieste/pareto.py
+++ b/trieste/pareto.py
@@ -16,6 +16,7 @@
 from typing import Tuple
 
 import tensorflow as tf
+from typing_extensions import Final
 
 from .type import TensorType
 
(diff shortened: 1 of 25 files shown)


trieste/pareto.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __init__(self, lower_idx: tf.Tensor, upper_idx: tf.Tensor):
+        """
+        Construct bounded volumes.
+
+        :param lower_idx: the lowerbounds index of the volumes
+        :param upper_idx: the upperbounds index of the volumes
+        """
+
+        assert tf.reduce_all(lower_idx.shape == upper_idx.shape)
```
joelberkeley: do we know more about the shapes here? should they be vectors? If so
```python
tf.debugging.assert_shapes([(lower_idx, ["N"]), (upper_idx, ["N"])])
```
may be more appropriate.

On a more general note, I believe `assert` is more for internal code correctness, while `raise ValueError` or `raise TypeError` etc. are more suitable for user input validation. `tf.debugging.assert_xyx` is like the latter

trieste/pareto.py
```diff
@@ -0,0 +1,132 @@
...
+        dtype = self.lower_idx.dtype
+        outdim = self.lower_idx.shape[1]
+        self.lower_idx = tf.zeros([0, outdim], dtype=dtype)
+        self.upper_idx = tf.zeros([0, outdim], dtype=dtype)
+
+    def size(self) -> tf.Tensor:
+        """
+        :return: volume of each bounded volume
+        """
+        return tf.math.reduce_prod(self.upper_idx - self.lower_idx, axis=0, keepdims=True)
```
joelberkeley: these three methods aren't used in this PR. Are they needed yet?

trieste/pareto.py
```diff
@@ -0,0 +1,132 @@
...
+
+        pf, _ = non_dominated(observations)
+        self.front: Final[TensorType] = tf.gather_nd(pf, tf.argsort(pf[:, :1], axis=0))
+        self.bounds: Final[BoundedVolumes] = self._bounds_2d(self.front)
+
+    @staticmethod
+    def _bounds_2d(front: TensorType) -> BoundedVolumes:
+        """
+        :param front: pareto front of current observations
+        """
```
joelberkeley: adding a docstring will make the method show up in the docs. We don't want that here do we?

trieste/pareto.py
```diff
@@ -0,0 +1,132 @@
+# Copyright 2020 The Trieste Contributors
```
joelberkeley: btw since you're not using Box, you could leave the file in utils. I don't really mind

satrialoka: okay

trieste/pareto.py
satrialoka: not yet for this PR, might be used in the divide and conquer method. I'll remove it

trieste/pareto.py
satrialoka: done

trieste/pareto.py
satrialoka: removed

trieste/pareto.py
satrialoka: removed

trieste/pareto.py
satrialoka: Done

trieste/utils/pareto.py
```diff
@@ -41,3 +45,63 @@ def non_dominated(observations: TensorType) -> TensorType:
     )
 
     return tf.boolean_mask(observations, dominance == 0), dominance
+
+
+class BoundedVolumes:
+    """
+    A :class:`BoundedVolumes store the index of the Pareto front to form lower and upper
```
```suggestion
    A :class:`BoundedVolumes` store the index of the Pareto front to form lower and upper
```

joelberkeley (approved):

trieste/utils/pareto.py
```diff
@@ -41,3 +45,63 @@ def non_dominated(observations: TensorType) -> TensorType:
...
+        """
+        Construct bounded volumes.
+
+        :param lower_idx: the lowerbounds index of the volumes
+        :param upper_idx: the upperbounds index of the volumes
+        """
+
+        tf.debugging.assert_shapes([(lower_idx, ["N", "D"]), (upper_idx, ["N", "D"])])
+        self.lower_idx = lower_idx
+        self.upper_idx = upper_idx
```
```suggestion
        self.lower_idx: Final[TensorType] = lower_idx
        self.upper_idx: Final[TensorType] = upper_idx
```

PR #185, opened 2021-03-04
Acquisition function based on the predictive variance

icouckuy:
This acquisition function calculates the determinant of the predictive covariance matrix for each set of batch points. For a batch of size 1 this is the same as returning the predictive variance.

This acquisition function is useful in the context of Bayesian active learning.

Notebook and tests in preparation.


TsingQAQ:
Am trying to add some more code to help this PR to be more completed. Though simple and intuitive, am not totally sure which paper of this exactly is linked with. 

@henrymoss Was thinking if we take a log of the det, could we somehow just interpret as the entropy of multivariate normal, or DPP-MAX, or even GIBBON? Is this helps improving this PR?

Also, am not totally sure what kind of test/notebooks we need here, please point out if there is a suggestion :pray:

PR #186, opened 2021-03-04
allow custom acquisition optimizers

joelberkeley:


trieste/acquisition/optimizer.py
```diff
@@ -11,42 +11,47 @@
...
+# todo name this more precisely. This may mean making the dss and box variants their own functions.
+@singledispatch
+def optimize(
+    space: Union[Box, DiscreteSearchSpace], target_func: AcquisitionFunction
+) -> TensorType:
+    """
+    :param space: The space of points over which to search, for points with shape S.
+    :param target_func: The function to maximise, with input shape [...] + S and output shape
+        [..., 1].
+    :return: The point in ``space`` that maximises ``target_func``, with shape [1] + S.
```
joelberkeley-secondmind, line 49: does this only support S = [D]?

trieste/acquisition/rule.py
```diff
@@ -30,29 +30,29 @@
...
 )
+from .optimizer import AcquisitionOptimizer, optimize
 
 S = TypeVar("S")
 """ Unbound type variable. """
 
-SP = TypeVar("SP", bound=SearchSpace, contravariant=True)
-""" Contravariant type variable bound to :class:`SearchSpace`. """
+SP_contra = TypeVar("SP_contra", bound=SearchSpace, contravariant=True)
+""" Contravariant type variable bound to :class:`~trieste.space.SearchSpace`. """
```
henrymoss, line 47: What does this do?

trieste/acquisition/rule.py
```diff
@@ -87,46 +87,55 @@ def acquire(
...
+            optimization problem is defined.
         :param datasets: The known observer query points and observations.
         :param models: The models of the specified ``datasets``.
         :param state: Unused.
         :return: The single point to query, and `None`.
         """
         acquisition_function = self._builder.prepare_acquisition_function(datasets, models)
-        point = _optimizer.optimize(search_space, acquisition_function)
-        return point, None
+        return self._optimizer(search_space, acquisition_function), None
```
henrymoss, line 138: This is good. Exactly what I thought we should have. It will allow me to add in awkward search spaces and new optimisers easily.

trieste/acquisition/rule.py
```diff
@@ -220,6 +229,7 @@ def __init__(
         builder: AcquisitionFunctionBuilder | None = None,
         beta: float = 0.7,
         kappa: float = 1e-4,
+        optimizer: AcquisitionOptimizer[Box] | None = None,
     ):
         """
```
henrymoss, line 234: For EGO, this was AcquisitionOptimizer[SP_contra]. Does this mean our trust region implementation is less flexible in terms of the search spaces it supports?

trieste/acquisition/rule.py
```diff
@@ -316,27 +333,45 @@ def acquire(
...
+        optimizer: AcquisitionOptimizer[SP_contra] | None = None,
+    ):
         """
         :param num_query_points: The number of points to acquire.
         :param builder: The acquisition function builder to use. :class:`BatchAcquisitionRule` will
             attempt to **maximise** the corresponding acquisition function.
+        :param optimizer: The optimizer with which to optimize the acquisition function built by
+            ``builder``. **Note:** This must be able to optimize its target function with respect to
+            a *single point*, not a batch of points. The acquisition rule will convert the batch
+            acquisition function to an independent acquisition function. Additionally, this
```
henrymoss: Maybe clarify what "an independent acquisition function" means here.

trieste/acquisition/optimizer.py
```diff
@@ -11,42 +11,47 @@
...
-    Return the point in ``space`` (with shape S) that maximises the function ``target_func``, as the
-    single entry in a 1 by S tensor.
 
-    ``target_func`` must satisfy the following:
+AcquisitionOptimizer = Callable[[SP, AcquisitionFunction], TensorType]
+"""
+Type alias for a function that returns the single point that maximizes an acquisition function over
+a search space. For a search space with points of shape S, and acquisition function with input shape
+[...] + S output shape [..., 1], the :data:`AcquisitionOptimizer` return shape should be [1] + S.
+"""
```
henrymoss, line 41: In BO, input shape is normally referred to as d (i.e dimension). Is there a reason why you haven't done this here?

trieste/acquisition/optimizer.py
```diff
@@ -11,42 +11,47 @@
...
-    :return: The point in ``space`` that maximises ``target_func``.
+@singledispatch
+def optimize(space: Box | DiscreteSearchSpace, target_func: AcquisitionFunction) -> TensorType:
+    """
+    :param space: The space of points over which to search, for points with shape [D].
+    :param target_func: The function to maximise, with input shape [..., D] and output shape
+        [..., 1].
+    :return: The point in ``space`` that maximises ``target_func``, with shape [1, D].
     """
-    raise TypeError(f"No optimize implementation found for space of type {type(space)}")
```
henrymoss, line 51: This is great for future compatibility!

trieste/acquisition/optimizer.py
```diff
@@ -60,7 +65,7 @@ def _discrete_space(space: DiscreteSearchSpace, target_func: TensorMapping) -> T
 
 
 @optimize.register
-def _box(space: Box, target_func: TensorMapping) -> TensorType:
+def _box(space: Box, target_func: AcquisitionFunction) -> TensorType:
     trial_search_space = space.discretize(20 * tf.shape(space.lower)[-1])
     initial_point = optimize(trial_search_space, target_func)
```
henrymoss, line 70: Victor and I were discussing this. It may be worth increasing the size of this initialisation (by quite a large amount), i.e. something that scales at a higher rate than linearly with input dimension.

Moreover, as we are using tensorflow, calculating the acquisition function across a much larger set of points is essentially the same cost as this smaller initialisation. In the past, I have used thousands of restarts for even smallish problems. 

This would be a great thing to test with our first COCO run?

tests/util/misc.py
```diff
@@ -81,6 +81,17 @@ def zero_dataset() -> Dataset:
...
+    :param query_point_shape: The shape of a *single* query point.
+    :param observation_shape: The shape of a *single* observation.
+    :return: An empty dataset with points of the specified shapes, and dtype `tf.float64`.
+    """
+    qp = tf.zeros(tf.TensorShape([0]) + query_point_shape, tf.float64)
+    obs = tf.zeros(tf.TensorShape([0]) + observation_shape, tf.float64)
+    return Dataset(qp, obs)
+
+
 def one_dimensional_range(lower: float, upper: float) -> Box:
```
henrymoss, line 95: What is this for?

docs/notebooks/expected_improvement.pct.py
```diff
@@ -216,8 +217,10 @@ def build_model(data):
 # Here we use the `BatchMonteCarloExpectedImprovement` function. Note that this acquisition function is computed using a Monte-Carlo method (so it requires a `sample_size`), but with a reparametrisation trick, which makes it deterministic.
 
 # %%
+from trieste.acquisition.rule import BatchAcquisitionRule
+
 qei = trieste.acquisition.BatchMonteCarloExpectedImprovement(sample_size=1000)
-batch_rule = trieste.acquisition.rule.BatchAcquisitionRule(
+batch_rule: BatchAcquisitionRule[Box] = BatchAcquisitionRule(
     num_query_points=3, builder=qei.using(OBJECTIVE)
 )
```
henrymoss, line 226: This looks good for now. But will hopefully be replaced soonish with a singe EGO rule.

trieste/acquisition/rule.py
joelberkeley: yes

trieste/acquisition/optimizer.py
joelberkeley: nope, will change

tests/util/misc.py
joelberkeley: which bit? `empty_dataset`?

trieste/acquisition/optimizer.py
joelberkeley: great! any bit in particular?

trieste/acquisition/rule.py
joelberkeley: if you're asking what is contravariance, have a read of these

https://stackoverflow.com/questions/61568462/python-typing-what-does-typevara-b-covariant-true-mean
https://www.python.org/dev/peps/pep-0483/#covariance-and-contravariance
https://ds12.github.io/scala-class/slides/images/lecture1/variance.png

joelberkeley: happy to explain more

trieste/acquisition/rule.py
joelberkeley: victor wants to do trust region for discrete spaces at some point

trieste/acquisition/rule.py
joelberkeley: how about a link to the `AcquisitionFunction` definition, where it explains it in detail?

docs/notebooks/expected_improvement.pct.py
joelberkeley: yeah I'm also not happy with having to type hint these. The fix (short of Python implementing look-ahead type inference ... a name I just made up) is to pass the search space to the acquisition rule constructor, rather than `acquire`. This would mean the search space isn't passed to the `BayesianOptimizer`, Victor says that wouldn't be intuitive for users. Maybe he's right. There are ways round that anyway. I might do that while we're doing this work.

trieste/acquisition/optimizer.py
joelberkeley: sounds good

joelberkeley: shall we do that in another PR?

trieste/acquisition/rule.py
henrymoss: Okay. Good to know!

trieste/acquisition/rule.py
henrymoss: Yeah, that would do it.

trieste/acquisition/optimizer.py
henrymoss: For when I add in spaces of strings and we have a genetic algorithm that supports maximisation over the string space, for example.

henrymoss: Basically, we can plug and play in new spaces and their optimisers

trieste/acquisition/optimizer.py
henrymoss: Yeah sure. I can do this. Maybe discuss in standup

tests/util/misc.py
henrymoss: Yep? Couldn't see where it was used. Is it just for testing?

joelberkeley: yeah I've used it in `test_efficient_global_optimization`. There are lots of tests elsewhere I'd use this too

docs/notebooks/expected_improvement.pct.py
joelberkeley: moreover, passing the search space to the constructor would probably mean we don't need `SP_contra`

henrymoss: True!

tests/util/misc.py
henrymoss: Makes sense then!

trieste/acquisition/optimizer.py
joelberkeley: ah, that's why - D is the dimension, but S is a shape, which could be [D] or something more complex if working with higher-rank input spaces

henrymoss (approved):

PR #190, opened 2021-03-05
add multi-objective toy function 

TsingQAQ:
# Tasks
This is a PR for adding multi-objective toy function (VLMOP2 [1], DTLZ1 [2], DTLZ2 [2])
## Feature Implemented
- a multi-objective test function class
- 3 multi-objective toy function
- methods for generating reference pareto front 

## Tests
- Test raise on incorrect input shape
-  Test generating reference pareto front correctly lies on ideal front
- Numerical test covoered for synthetic function on specific input.

Suggestions expected here. 

## Reference
[1] [Multiobjective evolutionary algorithm test suites](https://dl.acm.org/doi/pdf/10.1145/298151.298382?casa_token=7MeY-IgAy2MAAAAA:ta0LirTYLyyCHBxJz14-62trCQUerEmiy-ZAHUwRx6hG8wPROi6BcIHpGoZUEnuA3nr51wBX3CI)
[2] [Scalable multi-objective optimization test problems](https://ieeexplore.ieee.org/abstract/document/1007032?casa_token=3wLvQUY-H30AAAAA:hUmGG5qSAY683hZsRIxW2dgEL0VeHsHqGrMCw2aTPFG_zA7lVsXjSrWR7Uww4pLGUWXvZROb)


trieste/utils/objectives.py
```diff
@@ -126,6 +128,143 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+        :cite: fonseca1995multiobjective
+        True pareto lies on -1/sqrt(2)-1/sqrt(2) and x1=x2
+        """
+        tf.debugging.assert_greater(n, 0)
+        _x = tf.linspace([-1 / tf.sqrt(2.0)], [1 / tf.sqrt(2.0)], n)
+        return vlmop2(tf.concat([_x, _x], axis=1))
+
+
+def vlmop2(x: TensorType) -> TensorType:
+    tf.debugging.assert_equal(x.shape[-1], 2, "vlmop2 only allow 2d input")
```
joelberkeley: `assert_shapes` is best for shape checking

trieste/utils/objectives.py
```diff
@@ -126,6 +128,143 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+        """
+
+    @abstractmethod
+    def gen_pareto_optimal_points(self, n: int):
+        """
+        Generate `n` pareto optimal points
+        """
+
+
+class VLMOP2(MultiObjectiveTestProblem):
```
joelberkeley: are there more descriptive names for these? Or are these well-known names?

trieste/utils/objectives.py
```diff
@@ -126,6 +128,143 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+    """
+    The VLMOP2 function, typically evaluated over :math:`[-2, 2]^2`. See
+    :cite:`van1999multiobjective`  for details.
```
joelberkeley: :cite:\`foo\` requires foo to be in docs/refs.bib

trieste/utils/objectives.py
```diff
@@ -126,6 +128,143 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+    def prepare_benchmark(self):
+        return vlmop2
+
+    def gen_pareto_optimal_points(self, n: int):
+        """
+        :cite: fonseca1995multiobjective
+        True pareto lies on -1/sqrt(2)-1/sqrt(2) and x1=x2
+        """
+        tf.debugging.assert_greater(n, 0)
+        _x = tf.linspace([-1 / tf.sqrt(2.0)], [1 / tf.sqrt(2.0)], n)
```
joelberkeley: underscores in local variable names isn't standard elsewhere in trieste, so for readability can we avoid them?

trieste/utils/objectives.py
```diff
@@ -126,6 +128,143 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+
+    f = None
+    for i in range(m):
+        xM = x[:, m - 1 :]
+        y = 1 + g(xM)
+        y *= 1 / 2 * tf.reduce_prod(x[:, : m - 1 - i], axis=1, keepdims=True)
+        if i > 0:
+            y *= 1 - x[:, m - i - 1, tf.newaxis]
+        f = y if f is None else tf.concat([f, y], 1)
+    return f
```
joelberkeley: it would be good if these functions were compatible with tf.function. In particular, `tf.function` doesn't like local mutability with `tf.Tensor`. Would it be difficult to do that? You could use broadcasting or `tf.Variable`

joelberkeley:
I've left some comments - not complete yet though. Before that, do you mean to use these in tests, notebooks, or general usage?

trieste/utils/objectives.py
TsingQAQ: I think *VLMOP2* is known in multi-objective optimization community, 'V', 'L' just stands for the author's name of [1], which is a common function naming strategy in this community. 

Suggestions of a more clear naming approach is appreciated.

trieste/utils/objectives.py
TsingQAQ: okay

TsingQAQ:
> I've left some comments - not complete yet though. Before that, do you mean to use these in tests, notebooks, or general usage?

@joelberkeley  It is mainly used for notebooks and general usage for multi-objective problem, like [branin function](https://github.com/secondmind-labs/trieste/blob/79d7a484472861762f78a08c21c600f9e64d8f49/trieste/utils/objectives.py#L31).

trieste/utils/objectives.py
TsingQAQ: Seems I didn't see a direct way of using broadcasting here, correct me if I missed anything as I was not very familiar with checking the compatibility of `tf.function`. 

Alternatively, is it a good idea of using `tf.map_fn` here as below? 
```
def dtlz1_tffver(x: TensorType, m: int, k: int, d: int) -> TensorType:
    tf.debugging.assert_equal(
        x.shape[-1], d, f"input x dim: {x.shape[-1]} is not align with pre-specified dim: {d}"
    )

    def g(xM):
        return 100 * (
            k
            + tf.reduce_sum(
                (xM - 0.5) ** 2 - tf.cos(20 * math.pi * (xM - 0.5)), axis=-1, keepdims=True
            )
        )

    def fm(input):
        x, i = input
        y = 1 + g(x[:, m - 1:])
        y *= 1 / 2 * tf.reduce_prod(x[:, : m - 1 - i], axis=1, keepdims=True)
        if i > 0:
            y *= 1 - x[:, m - i - 1, tf.newaxis]
        return y

    test = tf.map_fn(fm, (tf.tile(x[tf.newaxis, ...], [m, 1, 1]),
                          tf.range(m, dtype=tf.int32)), dtype=(x.dtype, tf.int32),
                          fn_output_signature=x.dtype)
    return tf.concat(tf.unstack(test), 1)
```

trieste/utils/objectives.py
henrymoss: As the format for these synthetic functions is different to the non-MO ones, it might be cleaner to have it as a separate file?

tests/unit/utils/test_objectives.py
```diff
@@ -72,6 +76,39 @@ def test_logarithmic_goldstein_price_no_function_values_are_less_than_global_min
...
+        (DTLZ2(5, 2), tf.constant([[0.3, 0.1]])),
+    ],
+)
+def test_func_raises_specified_input_dim_not_alin_with_actual_input_dim(
+    obj_inst: MultiObjectiveTestProblem, actual_x: TensorType
+):
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        obj_inst.prepare_benchmark()(actual_x)
+
+
```
henrymoss: Can we also test a specific input and see that it gives the expected outputs? i.e. similar to the tests for the non-MO synthetic functions.

TsingQAQ: Some numerical tests added. 
Nevertheless, the expected results in the test are hand derived, a bit unintuitive for maintaining I guess.

trieste/utils/objectives.py
TsingQAQ: It has now been moved into `mo_objectives.py`, while the test are kept in the same file.

trieste/utils/objectives.py
henrymoss: Was this resolved?

TsingQAQ: Nope, I was not aware of an easy way to use broadcasting here, the only alternative way came in my mind is something as above, still expect some suggestions here.

henrymoss: Perhaps @uri-granta has some suggestions? I'm afraid I don't.

 I'm not sure that it matters if the synthetic objective functions are not fully optimized w.r.t speed. Any real-world objective functions certainly will not be. Kind of the point of BO!

henrymoss:
Can you bring this one up to date please @TsingQAQ ?

tests/unit/utils/test_objectives.py
```diff
@@ -79,6 +81,104 @@ def test_logarithmic_goldstein_price_no_function_values_are_less_than_global_min
     )
 
 
+@pytest.mark.parametrize(
+    "test_x, expected",
+    [
+        (
+            tf.constant([[0.0, 0.0]]),
+            tf.convert_to_tensor([[1.0 - tf.math.exp(-1.0), 1.0 - tf.math.exp(-1.0)]]),
```
henrymoss: why do we have all the tf.convert_to_tensor ?

tests/unit/utils/test_objectives.py
```diff
@@ -79,6 +81,104 @@ def test_logarithmic_goldstein_price_no_function_values_are_less_than_global_min
...
+@pytest.mark.parametrize(
+    "obj_inst, input_dim, num_obj, gen_pf_num",
+    [
+        (DTLZ1(3, 2), 3, 2, 1000),
+        (DTLZ1(5, 3), 5, 3, 1000),
+        (DTLZ2(3, 2), 3, 2, 1000),
+        (DTLZ2(12, 6), 12, 6, 1000),
+    ],
+)
+def test_gen_pareto_front_is_equal_to_math_defined(
```
henrymoss, line 177: Also should vlomp also have some tests in here?

tests/unit/utils/test_objectives.py
```diff
@@ -79,6 +81,104 @@ def test_logarithmic_goldstein_price_no_function_values_are_less_than_global_min
...
+        tf.debugging.assert_near(tf.norm(pfs, axis=1), 1.0, rtol=1e-6)
+
+
+@pytest.mark.parametrize(
+    "obj_inst, actual_x",
+    [
+        (DTLZ1(3, 2), tf.constant([[0.3, 0.1]])),
+        (DTLZ2(5, 2), tf.constant([[0.3, 0.1]])),
+    ],
+)
```
henrymoss: vlmop?

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,164 @@
...
+import tensorflow as tf
+
+from ..type import TensorType
+
+
+class MultiObjectiveTestProblem(ABC):
+    r"""Base class for test multi-objective test functions.
+    between a provided point and the closest point on the true pareto front.
+    """
+
```
henrymoss: "base class for synthetic multi-objective test functions"

Not sure you need that second line?

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,164 @@
...
+    @abstractmethod
+    def prepare_benchmark(self):
+        """
+        Evaluate Original problem
```
henrymoss: Evaluate original problem.

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,164 @@
...
+    transl = 1 / tf.sqrt(2.0)
+    y1 = 1 - tf.exp(-1 * tf.reduce_sum((x - transl) ** 2, axis=1))
+    y2 = 1 - tf.exp(-1 * tf.reduce_sum((x + transl) ** 2, axis=1))
+    return tf.stack([y1, y2], axis=1)
+
+
+class DTLZ(MultiObjectiveTestProblem):
+    """
+    DTLZ series multi-objective test functions.
+    :cite: deb2002scalable
```
henrymoss: See :cite:`deb2002scalable` for details.

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,164 @@
...
+            num_objective,
+            f"input dimension {input_dim}"
+            f"  must be greater than function objective numbers {num_objective}",
+        )
+        self.dim = input_dim
+        self.M = num_objective
+        self.k = self.dim - self.M + 1
+        self.bounds = [[0] * input_dim, [1] * input_dim]
+        super().__init__()
+
```
henrymoss: Do we need this?

tests/unit/utils/test_objectives.py
TsingQAQ: CMIIW here, it seems `tf.constant` does not support list of `Tensor`s, which is generated because of `tf.math.exp`, so the `convert_to_tensor` is used instead for converting compatibility on tensors.
The line 89 is written as an expression (thus having `tf.math.exp`) not exact number is more for interpretability, as it is the math definition of the synthetic function.

tests/unit/utils/test_objectives.py
TsingQAQ: It seems it is rather difficult for vlmop have test for pareto front. As this test is mainly used for testing if the algorithm generated pareto front is equivalent to math defined (e.g., a hyperplane for DTLZ1 and hypersphere for DTLZ2, which means the PF in objective space have some math defined relationship that can be checked). 
While for vlmop2, according to the citation, the pareto front for vlmop is just defined according to input distributions (i.e., "True pareto lies on -1/sqrt(2)-1/sqrt(2) and x1=x2"), which also conveys the PF do not have any analytical expression, hence might be hard to test.

trieste/utils/mo_objectives.py
TsingQAQ: The second line shouldn't be there, will remove.

trieste/utils/mo_objectives.py
TsingQAQ: Not necessary, will remove.

tests/unit/utils/test_objectives.py
joelberkeley: you can write this
```python
[[1.0, 1.0]] - tf.exp(-1.0)
```

henrymoss: I think you could just put the exact numbers in (like you do in the next test). Means that this will be much neater and not take up 25 lines of code.

TsingQAQ: Changed to exact numbers for neater expression.

henrymoss:
This looks good. Once @uri-granta takes a look, we can merge.

Feel free to start another PR with the notebook :)

TsingQAQ:
Just add a `seed` param fix. Actually am not totally sure whether we need this seed in `gen_pareto_optimal_points`. As it seems the operational-level seed, (i.e., `tf.random.xxx(seed=)`) is suggested used in combination with `tf.random.set_seed()`, but if one only specify the global-level seed, the random sequences generated by `tf.random` can also be fully reproducible.

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
+# Copyright 2020 The Trieste Contributors
```
uri-granta: Might it be clearer to name the file multi_objectives.py?

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+from abc import ABC, abstractmethod
+from functools import partial
+
+import tensorflow as tf
+
+from ..type import TensorType
+
+
+class MultiObjectiveTestProblem(ABC):
+    """base class for synthetic multi-objective test functions"""
```
uri-granta:
```suggestion
    """Base class for synthetic multi-objective test functions."""
```

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+
+    @abstractmethod
+    def prepare_benchmark(self):
```
uri-granta: This needs a return type, and a docstring! Is this returning an objective function of the form `Callable[[TensorType], TensorType]`? 

Also, is the word "prepare" necessary here? And is this a "benchmark" or just the "objective"?

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Evaluate original problem.
+        """
+
+    @abstractmethod
+    def gen_pareto_optimal_points(self, n: int):
```
uri-granta: This also needs a return type, and ideally a :param parameter description in the docstring.

And again, is "gen" necessary?

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Generate `n` pareto optimal points.
```
uri-granta:
```suggestion
        Generate `n` Pareto optimal points.
```

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+
+
+class VLMOP2(MultiObjectiveTestProblem):
+    """
+    The VLMOP2 function, typically evaluated over :math:`[-2, 2]^2`. See
```
uri-granta:
```suggestion
    The VLMOP2 function, typically evaluated over :math:`[-2, 2]^2`.
```

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+        :cite: fonseca1995multiobjective
+        True pareto lies on -1/sqrt(2)-1/sqrt(2) and x1=x2
+        """
+        tf.debugging.assert_greater(n, 0)
+        _x = tf.linspace([-1 / tf.sqrt(2.0)], [1 / tf.sqrt(2.0)], n)
+        return vlmop2(tf.concat([_x, _x], axis=1))
+
+
+def vlmop2(x: TensorType) -> TensorType:
+    tf.debugging.assert_shapes([(x, ("N", 2))], message="vlmop2 only allow 2d input")
```
uri-granta: Needs a doscstring.

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+
+
+class VLMOP2(MultiObjectiveTestProblem):
+    """
+    The VLMOP2 function, typically evaluated over :math:`[-2, 2]^2`. See
+    See :cite:`van1999multiobjective` for details.
+    """
+
+    bounds = [[-2.0] * 2, [2.0] * 2]
```
uri-granta: Are bounds and dim supposed to be a required part of every MultiObjectiveTestProblem? If so they should be specified and documented in the ABC to ensure that everything implements them (you may need to mark them as @abstractmethod @property to do this).

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+            f"input dimension {input_dim}"
+            f"  must be greater than function objective numbers {num_objective}",
+        )
+        self.dim = input_dim
+        self.M = num_objective
+        self.k = self.dim - self.M + 1
+        self.bounds = [[0] * input_dim, [1] * input_dim]
+
+
+class DTLZ1(DTLZ):
```
uri-granta: Needs class docstring.

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+    def prepare_benchmark(self):
+        return partial(dtlz1, m=self.M, k=self.k, d=self.dim)
+
+    def gen_pareto_optimal_points(self, n: int, seed=None):
```
uri-granta: Seed should really be a parameter in the abstractmethod too, as well as in VLMOP2, even if it's not always used. That way, a program knows how to call gen_pareto_optimal_points on a MultiObjectiveTestProblem without having to know which specific type of MultiObjectiveTestProblem it is.

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+        tf.debugging.assert_greater_equal(self.M, 2)
+        rnd = tf.random.uniform([n, self.M - 1], minval=0, maxval=1, seed=seed)
+        strnd = tf.sort(rnd, axis=-1)
+        strnd = tf.concat([tf.zeros([n, 1]), strnd, tf.ones([n, 1])], axis=-1)
+        return 0.5 * (strnd[..., 1:] - strnd[..., :-1])
+
+
+def dtlz1(x: TensorType, m: int, k: int, d: int) -> TensorType:
```
uri-granta: Docstring

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+    tf.debugging.assert_shapes(
```
uri-granta: This should probably also assert that m is greater than 0.

trieste/utils/mo_objectives.py
```diff
@@ -0,0 +1,161 @@
...
+
+    f = None
+    for i in range(m):
+        xM = x[:, m - 1 :]
+        y = 1 + g(xM)
+        y *= 1 / 2 * tf.reduce_prod(x[:, : m - 1 - i], axis=1, keepdims=True)
+        if i > 0:
+            y *= 1 - x[:, m - i - 1, tf.newaxis]
+        f = y if f is None else tf.concat([f, y], 1)
+    return f
```
uri-granta: I'm surprised this is passing the type tests, as currently if you pass m = 0 then the result is None and not a TensorType. I'll try and figure out why (and possibly also a neater way of doing this loop).

uri-granta (commented):
Sorry for the slow response! Here are a few comments/questions, though I might look it again tomorrow morning with a fresh pair of eyes.

trieste/utils/mo_objectives.py
TsingQAQ: Yes, changed

trieste/utils/mo_objectives.py
TsingQAQ: Maybe not necessary, used `objective` instead.

trieste/utils/mo_objectives.py
TsingQAQ: Type and doc added. 
For the "gen", if speaking from the sampling of distribution perspective maybe it is good to say so?  (Actually this name is borrowed form [here](https://github.com/pytorch/botorch/blob/5a5173b6ad89fd52aa66e5ed942208b25ea4b03b/botorch/test_functions/multi_objective.py#L188)).  
While indeed not all the Pareto optimal points are "generated" (as for instance the VLMOP2's are deterministically chosen), hope at least for now it makes some sense, please point out if it is better to remove still.

trieste/utils/mo_objectives.py
TsingQAQ: Is it possible to maybe elaborate more on this point?
The things are for some synthetic function the dim and bounds are used as a class variable as one could specify it in advance, while for some it needs to be specified during instantiation as input arg. 
Is that mean we would like them to be as property with some `getter` and `setter`?

trieste/utils/mo_objectives.py
uri-granta: Happy to leave it in.

trieste/utils/mo_objectives.py
joelberkeley: `TensorType` is effectively `Any`

trieste/utils/mo_objectives.py
uri-granta: I meant that if all classes have a dim and bounds then we should specify and document this in the MultiObjectiveTestProblem ABC. But Python doesn't let you add abstract attributes, only abstract methods or properties. So you could have:

````python
class MultiObjectiveTestProblem(ABC):

    @property
    @abstractmethod
    def dim(self) -> int:
        """ The dimension """
````

And then the classes could implement it, either statically:

```python
class VLMOP2(MultiObjectiveTestProblem):
    dim = 2
````

Or dynamically:

````python
class DTLZ(MultiObjectiveTestProblem):
    def __init__(self, input_dim: int, ...):
        ....
        self._dim = input_dim

    @property
    def val(self) -> int:
        return self._dim
````

This way the type checker can catch a class that forgets to do this.

File: trieste/utils/mo_objectives.py (hunk 26 above)
Comment (uri-granta) [reply trieste/utils/mo_objectives.py]: Yes, I've just discovered that! What's the reason for that (given that np.ndarray and tf.Tensor are normal-looking classes)?

Comment (joelberkeley) [reply trieste/utils/mo_objectives.py]: tf and np have no type hints (well np does now but you need both to have type hints to avoid the `Any`ness) - don't know the mechanics of it

Comment (joelberkeley) [reply trieste/utils/mo_objectives.py]: might be because you also need py.typed file to state a library has type hints

Comment (uri-granta) [reply trieste/utils/mo_objectives.py]: Yes, hadn't realised that these libraries didn't come with stubs. Looks like mypy introduced built in stubs in 1.20 (https://numpy.org/doc/stable/reference/typing.html) but as you say that's not much use without tensorflow stubs.

File: trieste/utils/mo_objectives.py (hunk 21 above)
Comment (TsingQAQ) [reply trieste/utils/mo_objectives.py]: Many thanks! I am not totally sure if the `bounds` is a must need as well as `dim`, but for now think it is good to add as well.

File: trieste/utils/multi_objectives.py [hunk 27]
```diff
@@ -0,0 +1,225 @@
...
+    """
+    Base class for synthetic multi-objective test functions.
+    Prepares the synthetic function and generates pareto optimal points.
+    The latter can be used as a reference of performance measure of certain
+    multi-objective optimization algorithms.
+    """
+
+    @property
+    @abstractmethod
+    def dim(self):
```
uri-granta: These should have return types:

```suggestion
    def dim(self) -> int:
```

TsingQAQ: thanks, sorry should add this earlier.
Btw do we also need a return type for `bounds`? is that `ListN` or maybe `SequenceN` ?

uri-granta: `ListN`is good (I haven't merged the change to `SequenceN` yet, and I'm happy to wait until this is merged in.)

trieste/utils/multi_objectives.py
```diff
@@ -0,0 +1,225 @@
...
+    def objective(self) -> Callable[[TensorType], TensorType]:
+        """
+        Get the synthetic test function.
+        :return: A callable synthetic function
+        """
+
+    @abstractmethod
+    def gen_pareto_optimal_points(self, n: int, seed=None) -> tf.Tensor:
+        """
+        Generate `n` Pareto optimal points.
```
uri-granta, line 62: There should be a blank line before the :param here (and also in objective before the :return, and in vlmop2 and dtlz1). This currently produces an invisible warning in the docs build, but since these warnings sometimes result in bad documentation we're planning on turning them into errors that get checked as part of the pre-merge quality checks (and which can be run locally using tox).

trieste/utils/objectives.py
uri-granta: I'm still new to tensorflow, but wouldn't something like this, using `tf.TensorArray`, make it compatible with tf.function?

```python
def dtlz2(x: TensorType, m: int, d: int) -> TensorType:

    def g(xM):
        z = (xM - 0.5) ** 2
        return tf.reduce_sum(z, axis=1, keepdims=True)

    ta = tf.TensorArray(x.dtype, size=m)
    for i in tf.range(m):
        y = 1 + g(x[:, m - 1 :])
        for j in tf.range(m - 1 - i):
            y *= tf.cos(math.pi / 2 * x[:, j, tf.newaxis])
        if i > 0:
            y *= tf.sin(math.pi / 2 * x[:, m - 1 - i, tf.newaxis])
        ta = ta.write(i, y)
    f = some_sort_of_transposition_to_make_the_right_shape(ta.stack())
    return f
```

trieste/utils/multi_objectives.py
```diff
@@ -0,0 +1,225 @@
...
+    The DTLZ1 synthetic function.
+    :param x: The points at which to evaluate the function, with shape [..., d].
+    :param m: The objective numbers.
+    :param k: The input dimensionality for g.
+    :param d: The dimensionality of the synthetic function.
+    :return : The function values at ``x``, with shape [..., m].
+    :raise ValueError (or InvalidArgumentError): If ``x`` has an invalid shape.
+    """
+    tf.debugging.assert_shapes(
+        [(x, ("N", d))],
```
uri-granta: Stupid question: is the shape of x `(..., d)` or `(N, d)`? (i.e. is it rank 2 or rank >=2?)

uri-granta (commented):
Thanks for addressing all of the comments. Added just a couple more, and then I think it should be good to merge!

trieste/utils/multi_objectives.py
TsingQAQ: Thanks, aware that `ListN` maybe more of used in test so currently used `list[list[float]]` for bounds.

trieste/utils/multi_objectives.py
TsingQAQ: The blank line has been added.

trieste/utils/multi_objectives.py
TsingQAQ: This rank issue earlier is indeed not considered, now the code has been modified with the input shape of x to be `(..., d)`, little extra test code is also added here for x's rank > 2.

henrymoss:
Is this ready to go? @uri-granta @TsingQAQ ?

TsingQAQ:
> Is this ready to go? @uri-granta @TsingQAQ ?

Seems most of the comments have been resolved except for the bit nasty code on querying observations in objective,  if this is ok then I think it is ready.

uri-granta (approved):
I'm happy with the code as is.

TsingQAQ:
Will take a final look at the objective function code maybe before merge.

trieste/utils/multi_objectives.py
```diff
@@ -0,0 +1,232 @@
...
+
+def dtlz1(x: TensorType, m: int, k: int, d: int) -> TensorType:
+    """
+    The DTLZ1 synthetic function.
+
+    :param x: The points at which to evaluate the function, with shape [..., d].
+    :param m: The objective numbers.
+    :param k: The input dimensionality for g.
+    :param d: The dimensionality of the synthetic function.
+    :return : The function values at ``x``, with shape [..., m].
```
uri-granta: (I think the docs check is failing because of this, and the similar one further down)

```suggestion
    :return: The function values at ``x``, with shape [..., m].
```

trieste/utils/multi_objectives.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def dtlz2(x: TensorType, m: int, d: int) -> TensorType:
+    """
+    The DTLZ2 synthetic function.
+
+    :param x: The points at which to evaluate the function, with shape [..., d].
+    :param m: The objective numbers.
+    :param d: The dimensionality of the synthetic function.
+    :return : The function values at ``x``, with shape [..., m].
```
uri-granta:
```suggestion
    :return: The function values at ``x``, with shape [..., m].
```

trieste/utils/multi_objectives.py
TsingQAQ: Thanks, this seems not very clear to learn from error message like:
`/home/runner/work/trieste/trieste/docs/autoapi/trieste/utils/multi_objectives/index.rst:147:Field list ends without a blank line; unexpected unindent.`

trieste/utils/multi_objectives.py
```diff
@@ -0,0 +1,234 @@
...
+    ta = tf.TensorArray(x.dtype, size=m)
+    for i in tf.range(m):
+        y = 1 + g(x[..., m - 1 :])
+        for j in tf.range(m - 1 - i):
+            y *= tf.cos(math.pi / 2 * x[..., j, tf.newaxis])
+        if i > 0:
+            y *= tf.sin(math.pi / 2 * x[..., m - 1 - i, tf.newaxis])
+        ta = ta.write(i, y)
+
+    return tf.squeeze(tf.concat(tf.split(ta.stack(), m, axis=0), axis=-1), axis=0)
```
TsingQAQ, line 234: The list has now has been replaced by `tf.TensorArray` as suggested, perhaps could bother @uri-granta for a final look?

uri-granta: Looks good to me, though since I'm new to tensorflow tensor manipulation still makes my head hurt. @henrymoss does the conversion from an array of tensors to a concatenated tensor look ok to you?

henrymoss: I think so. Can't think of a neater way. 

@TsingQAQ  get this up to date, then we can merge :) I can review your Notebook tomorrow morning.

TsingQAQ: Updated, could we merge it and move forward?

PR #194, opened 2021-03-08
Pareto set: Hypervolume

### LINKED ISSUE #181: Pareto set: hypervolume
As a user, I want functionality available to calculate the hypervolume of a Pareto front given the cell bounds, so that I can easily find the hypervolume when defining multi-objective acquisition functionality

satrialoka:
### Feature Implemented

This is a feature for Pareto functionality needed for Multi-objective BO, the code was ported from [GPFlowopt's pareto class](https://github.com/GPflow/GPflowOpt/blob/master/gpflowopt/pareto.py). This PR is small PR derived from previously large PR #126 and resolve #181
 
### Approach

Inside pareto.py `Pareto` class, this function is implemented:

1. `hypervolume_indicator` function: calculates hypervolume indicator given reference point 

## Status

Need review

## Notes

### DIFF
--- a/tests/unit/test_pareto.py
+++ b/tests/unit/test_pareto.py
@@ -113,7 +113,7 @@ def test_pareto_2d_bounds() -> None:
     )
 
 
-def test_hypervolume() -> None:
+def test_pareto_hypervolume_indicator() -> None:
     objectives = tf.constant(
         [
             [0.9575, 0.4218],
--- a/tests/unit/test_pareto.py
+++ b/tests/unit/test_pareto.py
@@ -11,11 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
 
 import numpy.testing as npt
 import pytest
 import tensorflow as tf
 
+from tests.util.misc import TF_DEBUGGING_ERROR_TYPES, NList
 from trieste.utils.pareto import Pareto, non_dominated
 
 
@@ -113,14 +115,35 @@ def test_pareto_2d_bounds() -> None:
     )
 
 
-def test_pareto_hypervolume_indicator() -> None:
-    objectives = tf.constant(
-        [
-            [-1.0, -0.5],
-            [-0.5, -1.0],
-        ]
-    )
+@pytest.mark.parametrize("reference", [0.0, [0.0], [[0.0]]])
+def test_pareto_hypervolume_indicator_raises_for_reference_with_invalid_shape(
+    reference: NList[float]
+) -> None:
+    pareto = Pareto(tf.constant([[-1.0, -0.6], [-0.8, -0.7], [-0.6, -1.1]]))
 
-    pareto_2d = Pareto(objectives)
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        pareto.hypervolume_indicator(tf.constant(reference))
 
-    npt.assert_allclose(pareto_2d.hypervolume_indicator(tf.constant([0.0, 0.0])), 0.75, rtol=1e-2)
+
+@pytest.mark.parametrize("reference", [[0.1, -0.65], [-0.7, -0.1]])
+def test_pareto_hypervolume_indicator_raises_for_reference_below_anti_ideal_point(
+    reference: list[float]
+) -> None:
+    pareto = Pareto(tf.constant([[-1.0, -0.6], [-0.8, -0.7], [-0.6, -1.1]]))
+
+    with pytest.raises(tf.errors.InvalidArgumentError):
+        pareto.hypervolume_indicator(tf.constant(reference))
+
+
+@pytest.mark.parametrize("objectives, reference, expected", [
+    ([[1.0, 0.5]], [2.3, 2.0], 1.95),
+    ([[-1.0, -0.6], [-0.8, -0.7], [-0.6, -1.1]], [0.1, -0.1], 0.92),
+    (  # reference point is equal to one pareto point in one dimension
+        [[-1.0, -0.6], [-0.8, -0.7], [-0.6, -1.1]], [0.1, -0.6], 0.37
+    ),
+])
+def test_pareto_hypervolume_indicator(
+    objectives: list[list[float]], reference: list[float], expected: float
+) -> None:
+    pareto = Pareto(tf.constant(objectives))
+    npt.assert_allclose(pareto.hypervolume_indicator(tf.constant(reference)), expected)
--- a/tests/unit/test_pareto.py
+++ b/tests/unit/test_pareto.py
@@ -117,7 +117,7 @@ def test_pareto_2d_bounds() -> None:
 
 @pytest.mark.parametrize("reference", [0.0, [0.0], [[0.0]]])
 def test_pareto_hypervolume_indicator_raises_for_reference_with_invalid_shape(
-    reference: NList[float]
+    reference: NList[float],
 ) -> None:
     pareto = Pareto(tf.constant([[-1.0, -0.6], [-0.8, -0.7], [-0.6, -1.1]]))
 
@@ -127,7 +127,7 @@ def test_pareto_hypervolume_indicator_raises_for_reference_with_invalid_shape(
 
 @pytest.mark.parametrize("reference", [[0.1, -0.65], [-0.7, -0.1]])
 def test_pareto_hypervolume_indicator_raises_for_reference_below_anti_ideal_point(
-    reference: list[float]
+    reference: list[float],
 ) -> None:
     pareto = Pareto(tf.constant([[-1.0, -0.6], [-0.8, -0.7], [-0.6, -1.1]]))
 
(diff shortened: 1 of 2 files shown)


trieste/utils/pareto.py
```diff
@@ -105,3 +105,32 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
         upper = tf.concat([range_ + 1, pf_ext_idx[::-1, 1:][: pf_ext_idx[-1, 0]]], axis=-1)
 
         return BoundedVolumes(lower, upper)
+
+    def hypervolume_indicator(self, reference: tf.Tensor) -> tf.Tensor:
+        """
+        Method to calculate the hypervolume indicator
+        The hypervolume indicator is the volume of the dominated region.
```
joelberkeley: You don't need "method to .."

tests/unit/test_pareto.py
```diff
@@ -111,3 +111,22 @@ def test_pareto_2d_bounds() -> None:
...
+            [0.9572, 0.6557],
+            [0.4854, 0.0357],
+            [0.8003, 0.8491],
+            [0.1419, 0.9340],
+        ]
+    )
+
+    pareto_2d = Pareto(objectives)
+
+    npt.assert_allclose(pareto_2d.hypervolume_indicator(tf.constant([2.0, 2.0])), 3.3878, rtol=1e-2)
```
joelberkeley: how did you calculate this?

trieste/utils/pareto.py
```diff
@@ -105,3 +105,32 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+
+        :param reference: a reference point to use, with shape [D].
+            Should be equal or bigger than the anti-ideal point of the Pareto set
+            for comparing results across runs, the same reference point must be used.
```
joelberkeley: Is this meant to be
```
            Should be equal or bigger than the anti-ideal point of the Pareto set.
            For comparing results across runs, the same reference point must be used.
```
or
```
            Should be equal or bigger than the anti-ideal point of the Pareto set
            for comparing results across runs. The same reference point must be used.
```

trieste/utils/pareto.py
```diff
@@ -105,3 +105,32 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        :return: hypervolume indicator (the higher the better)
```
joelberkeley: `(the higher the better)` adds context that might not be relevant for a given user. Do you think it's necessary?

trieste/utils/pareto.py
```diff
@@ -105,3 +105,32 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
         upper = tf.concat([range_ + 1, pf_ext_idx[::-1, 1:][: pf_ext_idx[-1, 0]]], axis=-1)
 
         return BoundedVolumes(lower, upper)
+
+    def hypervolume_indicator(self, reference: tf.Tensor) -> tf.Tensor:
```
```suggestion
    def hypervolume_indicator(self, reference: TensorType) -> TensorType:
```

trieste/utils/pareto.py
```diff
@@ -105,3 +105,32 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        """
+        Method to calculate the hypervolume indicator
+        The hypervolume indicator is the volume of the dominated region.
+
+        :param reference: a reference point to use, with shape [D].
```
joelberkeley, line 114: could you add an assert_shapes that checks the shape of this - particularly that D matches the `front` or `bounds`

trieste/utils/pareto.py
joelberkeley: why must we use the same reference point? what happens if you don't use the same reference point?

trieste/utils/pareto.py
joelberkeley, line 114: Can you expand on what a reference point is? From the perspective of someone who doesn't know much about multi-objective, or is used to a different name

tests/unit/test_pareto.py
```diff
@@ -111,3 +111,22 @@ def test_pareto_2d_bounds() -> None:
     npt.assert_allclose(
         pareto_2d.front, tf.constant([[0.1419, 0.9340], [0.1576, 0.7922], [0.4854, 0.0357]])
     )
+
+
+def test_hypervolume() -> None:
```
joelberkeley: the naming convention for tests for methods in trieste is usually `test_<class>_<method>` with `_<what we're testing>` at the end if there's more than one test for that method. This makes it clear what exactly is under test
```suggestion
def test_pareto_hypervolume_indicator() -> None:
```

trieste/utils/pareto.py
satrialoka: removed

tests/unit/test_pareto.py
satrialoka: I am using the same value as they use in https://github.com/GPflow/GPflowOpt/blob/3d86bcc000b0367f19e9f03f4458f5641e5dde60/testing/unit/test_pareto.py#L63

trieste/utils/pareto.py
satrialoka: The first one, I've updated the docstring.

trieste/utils/pareto.py
satrialoka: it's not necessary I think. removed.

trieste/utils/pareto.py
satrialoka: done

trieste/utils/pareto.py
satrialoka: it won't be comparable, since this reference point is used as a point bound to calculate the hypervolume. this is only the case if we want to compare/benchmarking the hypervolume result across runs.

trieste/utils/pareto.py
satrialoka: this term is specific for hypervolume-based Pareto optimization. mathematically, reference point = max(Pareto front set)+epsilon. epsilon is some constant. visually it is explained in Figure 2 of this article https://link.springer.com/content/pdf/10.1007/s10898-013-0118-2.pdf . 
not sure what to add In the docstring. what if "a bound point to calculate the volume"

tests/unit/test_pareto.py
joelberkeley: do you know how they came across this value?

joelberkeley: the reason I ask is because it's difficult to tell by looking at this test whether it's right or not

satrialoka: sorry for late response, I just confirmed it with Ivo. so in the gpflowopt they just try the algorithm (fully with acquisition function and BO steps) on some benchmark function. and see if everything works correctly and use both the pareto and the hypervolume of that pareto (3.3878) given [2, 2] ref point for the test. 
It seems strange way to do it actually.

satrialoka: one of the option that we can use as Ivo suggest is drawing random points from upper right part of unit circle.  if the reference point is [1,1], the hypervolume indicator should be close to 0.5*0.5 (the area of upper right square) - 1/4 of the circle area.

joelberkeley: would it not be possible to simply use some hand-chosen pareto points and work out the hypervolume by hand? To take a really simply case, a pareto front of ([-1, -0.5], [-0.5, -1]) with reference [0, 0] would give a hypervolume of 0.75 (do check that - I'm not certain I've got hypervolume yet). Probably worth using more interesting points and reference, but you get my drift

joelberkeley: we'd also need tests for both an empty pareto front and for a single point. If the expected behaviour for an empty point is undefined, the method should error

trieste/utils/pareto.py
```diff
@@ -105,3 +105,40 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+                (self.bounds.upper_idx, ["N", "D"]),
+                (self.front, ["M", "D"]),
+                (reference, ["D"]),
+            ]
+        )
+
+        min_pfront = tf.reduce_min(self.front, 0, keepdims=True)
+        pseudo_pfront = tf.concat((min_pfront, self.front, reference[None]), 0)
+        D = tf.shape(pseudo_pfront)[1]
+        N = tf.shape(self.bounds.upper_idx)[0]
```
joelberkeley: since you've asserted the shape above, you could write, if you want,
```python
N, D = tf.shape(self.bounds.upper_idx)
```

tests/unit/test_pareto.py
satrialoka: > 
> 
> would it not be possible to simply use some hand-chosen pareto points and work out the hypervolume by hand? To take a really simply case, a pareto front of ([-1, -0.5], [-0.5, -1]) with reference [0, 0] would give a hypervolume of 0.75 (do check that - I'm not certain I've got hypervolume yet). Probably worth using more interesting points and reference, but you get my drift

I don't have any interesting points in mind right now. but your sample indeed pretty easy to visualize and think of. I'll add it. If there are interesting points in the future, we can always add them as additional test points.

satrialoka: > 
> 
> we'd also need tests for both an empty pareto front and for a single point. If the expected behaviour for an empty point is undefined, the method should error

okay

joelberkeley: how about I make some tests and send them your way? We should add the tests with the code

satrialoka: okay

trieste/utils/pareto.py
joelberkeley: how about "Defines the upper bound of the hypervolume"?

trieste/utils/pareto.py
```diff
@@ -105,3 +105,39 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
 
         return BoundedVolumes(lower, upper)
+
+    def hypervolume_indicator(self, reference: TensorType) -> TensorType:
+        """
+        Calculate the hypervolume indicator
+        The hypervolume indicator is the volume of the dominated region.
+
+        :param reference: a reference point to use, with shape [D].
+            Should be equal or bigger than the anti-ideal point of the Pareto set.
```
joelberkeley, line 116: victor mentioned it's useful to be able to specify a reference point with elements less than those of the anti-ideal point in some dimensions. That this can be used when treating objectives as constraints, or when you want to focus on a particular section of the front.

This algorithm doesn't appear to work in that case. Is that a limitation of the algorithm or implementation?

joelberkeley: actually, let's just stick with it as it is, and error if it's not greater than or equal to the anti-ideal point. I think that's just
```
tf.debugging.assert_greater_equal(reference, self.front)
```

joelberkeley: we can add the other functionality later

satrialoka: okay, maybe better to open an issue for it.

trieste/utils/pareto.py
satrialoka: sounds good

trieste/utils/pareto.py
joelberkeley: yeah. I can do that

joelberkeley (approved):

trieste/utils/pareto.py
satrialoka: thanks

PR #202, opened 2021-03-15
Add analytic Ehvi acquisition function

TsingQAQ:
A draft PR for analytic hyper-volume expected improvement with some extra functionality. 

### New Features  
- Analytic EHVI based on [1]'s expression
- A `get_reference_point` function for calculating reference point 
- ------
~~- A draft notebook for EHVI usage demo~~
~~- A `plot_bo_points_in_obj_space` function for plotting bo points in objective space~~

### Reference
- [1] [Yang, K., Emmerich, M., Deutz, A., & Bäck, T. (2019). Efficient computation of expected hypervolume improvement using box decomposition algorithms. Journal of Global Optimization, 75(1), 3-34.](https://link.springer.com/article/10.1007/s10898-019-00798-7)


trieste/acquisition/multiobjective/analytic.py
```diff
@@ -0,0 +1,161 @@
...
+
+from ...data import Dataset
+from ...models import ProbabilisticModel
+from ...type import TensorType
+from ...utils.pareto import Pareto
+from ..function import AcquisitionFunction
+from .function import HypervolumeAcquisitionBuilder, get_reference_point
+
+
+class Expected_Hypervolume_Improvement(HypervolumeAcquisitionBuilder):
```
joelberkeley:
```suggestion
class ExpectedHypervolumeImprovement(HypervolumeAcquisitionBuilder):
```
class names use camel case in Python

trieste/acquisition/multiobjective/analytic.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    Builder for the :func:`hv_probability_of_improvement` acquisition function
+    refer yang2019efficient
+    """
+
+    def __init__(self, ref_pt_calc_method: Union[str, Callable] = "default"):
```
joelberkeley: you're not using this argument anywhere, so can I suggest we leave customizing this to another PR?

trieste/acquisition/multiobjective/analytic.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        :param ref_pt_calc_method the method of calculating the reference point,
+         either default or a callable
+        """
+        self._ref_pt_calc_method = ref_pt_calc_method
+
+    def __repr__(self) -> str:
+        return f"Expected Hypervolume Improvement({self._ref_pt_calc_method!r})"
```
joelberkeley:
```suggestion
        return f"ExpectedHypervolumeImprovement({self._ref_pt_calc_method!r})"
```
the result of `repr(x)` should, as much as possible, look like a call to the class constructor. For example, `eval(repr(ExpectedHypervolumeImprovement()))` would actually create an `ExpectedHypervolumeImprovement`

trieste/acquisition/multiobjective/analytic.py
```diff
@@ -0,0 +1,161 @@
...
+        return lambda at: self._acquisition_function(model, at, _pf, _reference_pt)
+
+    @staticmethod
+    def _acquisition_function(
+        model: ProbabilisticModel,
+        at: TensorType,
+        pareto: Pareto,
+        ref_pt: TensorType,
+    ) -> tf.Tensor:
+        return expected_hv_improvement(model, at, pareto, ref_pt)
```
joelberkeley:
```suggestion
        return lambda at: expected_hv_improvement(model, at, _pf, _reference_pt)
```
i know we do this elsewhere, but we never actually use `_acquisition_function`. Meanwhile, I think it makes the code harder to read. What do you think?

trieste/acquisition/multiobjective/function.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class HypervolumeAcquisitionBuilder(SingleModelAcquisitionBuilder):
+    @abstractmethod
+    def _calculate_ref_pt(self, pareto: Pareto, ref_pt_calc_method="default"):
+        """
+        calculate the reference point for hypervolme calculation
+        :param pareto: Pareto class
+        :param ref_pt_calc_method
+        """
```
joelberkeley: did you have a particular use case for this?

tests/util/model.py
```diff
@@ -86,3 +86,12 @@ def __init__(self, kernel_amplitude: float | TensorType | None = None):
...
         return "QuadraticMeanAndRBFKernel()"
+
+
+class LinearMeanAndRBFKernel(GaussianProcess):
+    def __init__(self, kernel_amplitude: float | TensorType | None = None):
+        kernel = tfp.math.psd_kernels.ExponentiatedQuadratic(kernel_amplitude)
+        super().__init__([linear], [kernel])
+
+    def __repr__(self) -> str:
+        return "LinearMeanAndRBFKernel()"
```
joelberkeley: can I suggest that, since this is only used in one place, that we just have `GaussianProcess([linear], [tfp.math.psd_kernels.ExponentiatedQuadratic(kernel_amplitude)])` where it's used?

tests/util/misc.py
```diff
@@ -116,6 +116,20 @@ def quadratic(x: tf.Tensor) -> tf.Tensor:
...
+    The multi-dimensional linear function.
+
+    :param x: A tensor whose last dimension is of length greater than zero.
+    :return: The sum :math:`x` of ``x``.
+    :raise ValueError: If ``x`` is a scalar or has empty trailing dimension.
+    """
+    if x.shape == [] or x.shape[-1] == 0:
+        raise ValueError(f"x must have non-empty trailing dimension, got shape {x.shape}")
+
+    return tf.reduce_sum(x, axis=-1, keepdims=True)
```
joelberkeley: I'm not clear on the benefit to making this available to all the tests. `quadratic` is a simple function with a single global minimum. Is there something `linear` offers that we'll want to use frequently?

joelberkeley:
Thanks a lot for this @TsingQAQ . I've had a chance to look at some of the high-level design, and left some comments. I'll look at the meat of the PR (the algorithms) when I can.

docs/notebooks/multiobjective_ehvi.py
```diff
@@ -0,0 +1,154 @@
+# -*- coding: utf-8 -*-
```
joelberkeley: the notebooks should be in another PR, though I realise it's probably easier for you to leave them in this branch until we've finalised things a little more, so no need to move them yet.

trieste/acquisition/multiobjective/__init__.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2020 The Trieste Contributors
```
joelberkeley: Can we have the acquisition function in the same module as the others? I do agree it's worth splitting the acquisition functions up, but I want to do that in one go so we can choose which ones go together

joelberkeley:
btw I haven't forgotten that you've opened this PR at this point to make it easier to discuss how we represent the cell bounds. I'll get to that when I can

trieste/acquisition/multiobjective/function.py
TsingQAQ: Was earlier think about this because one may have different acquisition function based on `HypervolumeAcquisitionBuilder`, but maybe for now it is better to remove this part.

tests/util/model.py
TsingQAQ: Sounds great. 
It is mainly used for test of EHVI, just would like to have a different model for different objective function (e.g., the usage [here](https://github.com/secondmind-labs/trieste/blob/11fbe880967d88a0477ff69de0e0c8be006be633/tests/unit/acquisition/test_ehvi.py#L36)), so that the predicted pareto front can be more than one point (as otherwise there might be only one dominated point).

tests/util/misc.py
TsingQAQ: I was not aware of the usage of `quadratic` in that way and originally thought `misc.py` might just be the place to put a new model for use. 
It is assumed to be often used for test of multi-objective related acquisition function, the reason, as mentioned above, is it might be better to have different model for different objective function. The reason to choose linear is just heuristically thought it can provide a model that is enough different than the `QuadraticMeanAndRBFKernel` test model, and it is not ristricted to 2 dimension input as the  `GaussianProcess([branin], [kernel])`. 
But this can also be moved directly into the test for multi-objective related function if this is not useful for others, would this move a better choice at the moment?

tests/util/model.py
joelberkeley: ok. Recall other more interesting functions are available in `trieste.utils.objectives`

henrymoss:
Right then! Lets sort out EHVI. Can you bring the branch up to date?  

I'm think this PR should just be the code +tests. Save the notebook for another PR (perhaps one after we get the synthetic benchmarks in?)

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
         return lambda at: expected_improvement(objective_model, eta)(at) * constraint_fn(at)
 
 
+class ExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder):
+    """
+    Builder for the :func:`expected_hv_improvement` acquisition function
```
```suggestion
    Builder for the :func:`expected_hv_improvement` acquisition function.
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+    The implementation of the acquisition function largely
+    follows :cite:`yang2019efficient`
+    """
+
+    def __repr__(self) -> str:
+        return "ExpectedHypervolumeImprovement()"
```
uri-granta, line 524: For some reason we currently include an empty docstring for all the other \_\_repr\_\_ methods in this file. I'll investigate why, but in the meantime we might as well be consistent.

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+        :param dataset: The data from the observer. Must be populated.
+        :param model: The model over the specified ``dataset``.
+        :return: The expected_hv_improvement function.
+        """
+        tf.debugging.assert_positive(len(dataset), message="Dataset must be populated.")
+        mean, _ = model.predict(dataset.query_points)
+
+        _pf = Pareto(mean)
+        _reference_pt = get_reference_point(_pf.front)
+        return lambda at: expected_hv_improvement(model, _pf, _reference_pt)(at)
```
uri-granta: Can't you just write this instead?

```suggestion
        return expected_hv_improvement(model, _pf, _reference_pt)
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+def expected_hv_improvement(
+    model: ProbabilisticModel,
+    pareto: Pareto,
+    reference_point: TensorType,
+) -> AcquisitionFunction:
+    r"""
+    HV calculation using Eq. 44 of :cite:`yang2019efficient` paper
+    The expected hypervolume calculation is performed in the non-dominated region, which,
+    provided the partitioned cell, can be reformulated as a combination of several one
+    dimensional generalized expected improvement based on each of the partitioned cell.
```
uri-granta: Do you mean "each of the partitioned cells" or "each of the partitioned slices" maybe?

```suggestion
    dimensional generalized expected improvements based on each of the partitioned cell.
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+    The generlized expected improvement can be further divided into 2 types for easier
+    math operation, which corresponding to Psi function and nu function calculation.
```
```suggestion
    math operation, corresponding to Psi function and nu function calculations.
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+    Mainly referring Eq. 44, and Eq. 45
+    Note:
+    1. Since in Trieste we do not assume the use of a certain non-dominated partition algorithm.
```
```suggestion
    1. Since in Trieste we do not assume the use of a certain non-dominated partition algorithm,
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+       we do not assume the last dimension partitioned cell has only one (lower) bound
+       (which is used in the :cite:`yang2019efficient` paper), this is not equally efficient as
```
```suggestion
       (which is used in the :cite:`yang2019efficient` paper). This is not as efficient as
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+        the original paper, but is applicable to different non-dominated partition algorithm
+    2. The Psi and nu function in the original paper is defined for a maximization problem, to
```
```suggestion
    2. The Psi and nu function in the original paper is defined for a maximization problem. To
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+        return lambda at: expected_hv_improvement(model, _pf, _reference_pt)(at)
+
+
+def expected_hv_improvement(
+    model: ProbabilisticModel,
+    pareto: Pareto,
+    reference_point: TensorType,
+) -> AcquisitionFunction:
+    r"""
+    HV calculation using Eq. 44 of :cite:`yang2019efficient` paper
```
```suggestion
    HV calculation using Eq. 44 of :cite:`yang2019efficient` paper.
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+    math operation, which corresponding to Psi function and nu function calculation.
+    Mainly referring Eq. 44, and Eq. 45
+    Note:
+    1. Since in Trieste we do not assume the use of a certain non-dominated partition algorithm.
+       we do not assume the last dimension partitioned cell has only one (lower) bound
+       (which is used in the :cite:`yang2019efficient` paper), this is not equally efficient as
+        the original paper, but is applicable to different non-dominated partition algorithm
+    2. The Psi and nu function in the original paper is defined for a maximization problem, to
+       make use of the same notation for easier following, we inverse our problem (as maximization)
+       to make use of the same equation
```
```suggestion
       to make use of the same equation.
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+    provided the partitioned cell, can be reformulated as a combination of several one
+    dimensional generalized expected improvement based on each of the partitioned cell.
+    The generlized expected improvement can be further divided into 2 types for easier
+    math operation, which corresponding to Psi function and nu function calculation.
+    Mainly referring Eq. 44, and Eq. 45
+    Note:
+    1. Since in Trieste we do not assume the use of a certain non-dominated partition algorithm.
+       we do not assume the last dimension partitioned cell has only one (lower) bound
+       (which is used in the :cite:`yang2019efficient` paper), this is not equally efficient as
+        the original paper, but is applicable to different non-dominated partition algorithm
```
```suggestion
        the original paper, but is applicable to different non-dominated partition algorithm.
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+    model: ProbabilisticModel,
+    pareto: Pareto,
+    reference_point: TensorType,
+) -> AcquisitionFunction:
+    r"""
+    HV calculation using Eq. 44 of :cite:`yang2019efficient` paper
+    The expected hypervolume calculation is performed in the non-dominated region, which,
+    provided the partitioned cell, can be reformulated as a combination of several one
+    dimensional generalized expected improvement based on each of the partitioned cell.
+    The generlized expected improvement can be further divided into 2 types for easier
```
```suggestion
    The generalized expected improvement can be further divided into 2 types for easier
```

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+        return tf.reduce_sum(
+            tf.reduce_sum(tf.reduce_prod(factor_combinations, axis=-1), axis=-1),
+            axis=-1,
+            keepdims=True,
+        )
+
+    return acquisition
+
+
+def get_reference_point(front: TensorType) -> TensorType:
```
uri-granta: This could probably be made a private function for now (i.e. renaming it to `_get_reference_point`)? Though it should still have a description of the parameter in the docstring.

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+        candidate_std = tf.expand_dims(candidate_std, 1)  # [..., 1, out_dim]
+
+        lb_points, ub_points = pareto.hypercell_bounds(
+            tf.constant([-inf] * candidate_mean.shape[-1], dtype=x.dtype), reference_point
+        )
+
+        neg_lb_points, neg_ub_points = -ub_points, -lb_points
+
+        neg_ub_points = tf.minimum(
+            neg_ub_points, 1e10
```
uri-granta: Why do we need to truncate the upper bound?

trieste/acquisition/function.py
```diff
@@ -509,6 +512,156 @@ def prepare_acquisition_function(
...
+            neg_lb_points, neg_lb_points, neg_candidate_mean, candidate_std
+        )  # [..., num_cells, out_dim]
+
+        psi_lb2ub = tf.maximum(psi_lb - psi_ub, 0.0)  # [..., num_cells, out_dim]
+        nu_contrib = nu(neg_lb_points, neg_ub_points, neg_candidate_mean, candidate_std)
+
+        # get stacked factors of Eq. 45
+        # [2^m, indices_at_dim]
+        cross_index = tf.constant(
+            list(product(*[[0, 1] for _ in range(reference_point.shape[-1])]))
```
uri-granta: It is actually possible to generate this directly in tf via:

```
dim = reference_point.shape[-1]
cross_index = tf.reshape(tf.stack(tf.meshgrid(*[[0,1]]*dim), axis=-1), (-1,dim))
```

but your way is probably clearer. Still, you can simplify it further to:

```suggestion
            list(product(*[[0, 1]] * reference_point.shape[-1]))
```

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +472,153 @@ def prepare_acquisition_function(
...
+        (
+            1,
+            200_000,
+            tf.constant([[0.3, 0.2], [0.2, 0.22], [0.1, 0.25], [0.0, 0.3]]),
+            2,
+            2.0,
+            0.01,
+            1e-2,
+        ),
+        (2, 50_000, tf.constant([[0.0, 0.0]]), 2, 1.0, 0.01, 1e-2),
```
uri-granta: Would be nice to also have a test with input_dim 3 or even 4?

(Also for parametrized tests with lots of parameters like here it can often be useful to use [pytest.param](https://docs.pytest.org/en/stable/reference.html#pytest-param) to assign the cases text ids, to make it easier to see what each case is testing.)

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +472,153 @@ def prepare_acquisition_function(
...
+    _prt = Pareto(model_pred_observation)
+    ehvi = expected_hv_improvement(model, _prt, get_reference_point(_prt.front))
+
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        ehvi(at)
+
+
+@random_seed
+@pytest.mark.parametrize(
+    "input_dim, num_samples_per_point, existing_observations, obj_num, variance_scale ,rtol, atol",
```
uri-granta:
```suggestion
    "input_dim, num_samples_per_point, existing_observations, obj_num, variance_scale, rtol, atol",
```

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +472,153 @@ def prepare_acquisition_function(
...
+    )  # num_cells x B
+    splus_idx = tf.expand_dims(tf.cast(splus_valid, dtype=ub_points.dtype), -1)
+    splus_lb = tf.tile(lb_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+    splus_lb = tf.maximum(splus_lb, tf.expand_dims(predict_samples, 1))
+    splus_ub = tf.tile(ub_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+    splus = tf.concat([splus_idx, splus_ub - splus_lb], axis=-1)
+
+    ehvi_approx = tf.transpose(
+        tf.reduce_sum(tf.reduce_prod(splus, axis=-1), axis=1, keepdims=True)
+    )  #
```
uri-granta:
```suggestion
    )
```

uri-granta (commented):
Looks good to me from an engineering perspective.

trieste/acquisition/function.py
joelberkeley: iirc, if you include a docstring, it will appear in the built docs, even if it's private, so you probably don't want a docstring at all

trieste/acquisition/function.py
joelberkeley: it's so as not to show the default docstring for `__repr__`, which doesn't add anything and pollutes the docs

trieste/acquisition/function.py
TsingQAQ: Indeed, will change.

trieste/acquisition/function.py
TsingQAQ: Yes, will modify the doc to make it more clear.

trieste/acquisition/function.py
TsingQAQ: The truncation is more for numerical stability, as the the gradient of CDF of `inf` will cause `nan` issue IIRC. So a naive fix is applied here like truncating the upper bound to avoid the direct operation on `inf`. Please suggest on this point if any better way could be of use.

trieste/acquisition/function.py
TsingQAQ: The doc string here is more used during reviewing stage for easier follow, I think yes it is better to remove them once there is no issue on it before merging.

TsingQAQ: If it is ok I would prefer not a private `get_reference_point`, the reason is:

-  this is a generic functionality that likely to be used by many mo acquisition function (e.g., the qehvi at least). 

- And this method (implemented in this `get_reference_point`) is not a bundle to specific acquisition function, as one could specify any other reference point calculation method for mo acquisition function.

tests/unit/acquisition/test_function.py
TsingQAQ: Added `input_dim` 3 and 4 test case, also add a case with 3 obj as now the partition for obj > 2 is supported by #207.
Test cases with the text ids are also used.

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+        )  # [..., num_cells, out_dim]
+        psi_lb = Psi(
+            neg_lb_points, neg_lb_points, neg_candidate_mean, candidate_std
+        )  # [..., num_cells, out_dim]
+
+        psi_lb2ub = tf.maximum(psi_lb - psi_ub, 0.0)  # [..., num_cells, out_dim]
+        nu_contrib = nu(neg_lb_points, neg_ub_points, neg_candidate_mean, candidate_std)
+
+        # get stacked factors of Eq. 45
+        # [2^m, indices_at_dim]
```
TsingQAQ: These code comments (e.g., shape hints etc) are mainly used for reviewing, maybe such comments are expected to be removed later if a cleaner code is expected?

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+        """
+        :param dataset: The data from the observer. Must be populated.
+        :param model: The model over the specified ``dataset``.
+        :return: The expected_hv_improvement function.
+        """
+        tf.debugging.assert_positive(len(dataset), message="Dataset must be populated.")
+        mean, _ = model.predict(dataset.query_points)
+
+        _pf = Pareto(mean)
+        _reference_pt = get_reference_point(_pf.front)
```
henrymoss, line 538: If you are already passing _pf to expected_hv_improvement, why do you also need to pass _reference_pt ?

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+        return expected_hv_improvement(model, _pf, _reference_pt)
+
+
+def expected_hv_improvement(
+    model: ProbabilisticModel,
+    pareto: Pareto,
+    reference_point: TensorType,
+) -> AcquisitionFunction:
+    r"""
+    HV calculation using Eq. 44 of :cite:`yang2019efficient` paper.
```
henrymoss: HV --> Hyper-volume (HV)

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+    The expected hypervolume improvement calculation in the non-dominated region
+    can be decomposed into sub-calculations based on each partitioned cell.
+    This sub-scalculation can be further reformulated as a combination of two types of
+    generalized expected improvements for easier math operation, corresponding to Psi
+    function and nu function calculations.
+    Mainly referring Eq. 44, and Eq. 45
```
henrymoss: For easier calculation, this sub-calculation can be reformulated as a combination of two generalized expected improvements, corresponding to Psi (Eq. 44) and Nu (Eq. 45) function calculations, respectively.

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+    Note:
+    1. Since in Trieste we do not assume the use of a certain non-dominated partition algorithm,
+       we do not assume the last dimension partitioned cell has only one (lower) bound
+       (which is used in the :cite:`yang2019efficient` paper). This is not as efficient as
+        the original paper, but is applicable to different non-dominated partition algorithm.
+    2. The Psi and nu function in the original paper is defined for a maximization problem. To
+       make use of the same notation for easier following, we inverse our problem (as maximization)
+       to make use of the same equation.
```
henrymoss: As the Psi and nu function in the original paper are defined for maximization problems, we inverse our minimisation problem (to also be a maximisation), allowing use of the original notation and equations.

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+    :return The expected_hv_improvement acquisition function modified for objective
+        minimisation. This function will raise :exc:`ValueError` or
+        :exc:`~tf.errors.InvalidArgumentError` if used with a batch size greater than one.
+    """
+
+    def acquisition(x: TensorType) -> TensorType:
+        tf.debugging.assert_shapes(
+            [(x, [..., 1, None])],
+            message="This acquisition function only supports batch sizes of one.",
+        )
```
henrymoss, line 576: For now ;)

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+        normal = tfp.distributions.Normal(
+            loc=tf.zeros(shape=1, dtype=x.dtype), scale=tf.ones(shape=1, dtype=x.dtype)
+        )
+
+        def Psi(a, b, mean, std) -> TensorType:
+            """
```
henrymoss: You should probably add some type checking here, i.e. a: Tensortype etc.

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+            :param b: shape [num_cells, out_dim] integral upper bounds
+            :param mean: shape [..., out_dim]
+            :param std: shape [..., out_dim]
+            """
+            return std * normal.prob((b - mean) / std) + (mean - a) * (
+                1 - normal.cdf((b - mean) / std)
+            )
+
+        def nu(lb, ub, mean, std) -> TensorType:
+            """
```
henrymoss: As above

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+            :param ub: shape [..., num_cells, out_dim] slice upper bound
+            :param mean: shape [..., out_dim]
+            :param std: shape [..., out_dim]
+            """
+            return (ub - lb) * (1 - normal.cdf((ub - mean) / std))
+
+        candidate_mean, candidate_var = model.predict(tf.squeeze(x, -2))
+        candidate_std = tf.sqrt(candidate_var)
+
+        # calc ehvi assuming maximization
```
henrymoss: In trieste, we have tried to not have any comment lines as they can get separated from the code they refer to. Try to keep discussion at the start of the function (where we discuss the input/outputs) or inline (like you did with shapes below). This comment is unnecessary anyway,

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+
+        lb_points, ub_points = pareto.hypercell_bounds(
+            tf.constant([-inf] * candidate_mean.shape[-1], dtype=x.dtype), reference_point
+        )
+
+        neg_lb_points, neg_ub_points = -ub_points, -lb_points
+
+        neg_ub_points = tf.minimum(
+            neg_ub_points, 1e10
+        )  # this maximum: 1e10 is heuristically chosen
```
henrymoss: Don't need this comment

trieste/acquisition/function.py
henrymoss: I'm happy with shape hints inline with their vectors.

trieste/acquisition/function.py
henrymoss: Say " clip to improve numerical stability"

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
...
+
+        # Take the cross product of psi_diff and nu across all outcomes
+        # [..., num_cells, 2(operation_num, refer Eq. 45), num_obj]
+        stacked_factors = tf.concat(
+            [tf.expand_dims(psi_lb2ub, -2), tf.expand_dims(nu_contrib, -2)], axis=-2
+        )
+
+        # [..., num_cells, 2^m, 2(operation_num), num_obj]
+        factor_combinations = tf.linalg.diag_part(tf.gather(stacked_factors, cross_index, axis=-2))
+
```
henrymoss: Could lines 632-643 (and the long operation in the return statement of 647) be made into a function? 

You could then define this function with an inline comment  describing what it is does (i.e eq. 45 ...)

See line 210 for MinValueEntropySearch, where we keep the code interpretable by defining functions.

trieste/acquisition/function.py
henrymoss: keep  get_reference_point public as it will be useful for other acquisition functions. I propose putting it with the Pareto.py though.

trieste/acquisition/function.py
```diff
@@ -509,6 +512,155 @@ def prepare_acquisition_function(
         return lambda at: expected_improvement(objective_model, eta)(at) * constraint_fn(at)
 
 
+class ExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder):
+    """
+    Builder for the :func:`expected_hv_improvement` acquisition function.
```
henrymoss: I think I have misunderstood something here. But are we not in the multi-model case when doing Multi-objective BO? Perhaps @vpicheny can explain?

henrymoss (commented):
Reviewed function.py

Will tackle the tests later. 

Looks good though!

trieste/acquisition/function.py
vpicheny: you can either do multiobj with a single multi-output model or a set of single-output models (or both...).

vpicheny: I would also expect a multi-model by default

henrymoss: I don't think we support the set of single-output models case at the moment though, as we only allow a single objective function tag.

This isn't necessarily an issue. We can stack single output models using the model stacker to get a single model for all the objectives. (which is what @TsingQAQ  does in his tests) .

Perhaps we leave the multi-model case for constraints? I think this is what Joel wanted.

joelberkeley: the reason is that this implementation (and perhaps EHVI in general) assumes all relevant observations can be combined in the same tensor. Thus the corresponding data can be combined in the same dataset. It seemed like an unnecessary freedom involving too much boilerplate and other complexity to allow multiple datasets

henrymoss: Thanks @joelberkeley 

I agree completely. So @TsingQAQ no changes required here.

vpicheny: sounds good!

trieste/acquisition/function.py
TsingQAQ: The reason is more for a future extension consideration: it might also be good to allow user specify his self defined `get_reference_point`. As for the reference point setting, to my best of knowledge there is no unique agreed solution. Hence, in future if a user has specified his own `modified_get_reference_point`, it might be good to proceed the reference point at this acq prepare stage and feed into the acquisition function for calculation & optimization.

vpicheny: indeed choosing the reference point is an open question and this flexibility is important

henrymoss: Yep that makes sense!

trieste/acquisition/function.py
TsingQAQ: Is there any plan to make one acq for EHVI with arbitrary batch size? 
We have a qehvi in prepare indeed, but it is written as another acquisition function.

trieste/acquisition/function.py
TsingQAQ: Rearranged into a new function (start at line 604).

trieste/acquisition/function.py
henrymoss: I guess we might merge the two, but thats to discuss another time.

trieste/acquisition/function.py
```diff
@@ -509,6 +512,161 @@ def prepare_acquisition_function(
...
+                tf.gather(stacked_factors, cross_index, axis=-2)
+            )  # [..., num_cells, 2^d, 2(operation_num), num_obj]
+
+            return tf.reduce_sum(tf.reduce_prod(factor_combinations, axis=-1), axis=-1)
+
+        candidate_mean, candidate_var = model.predict(tf.squeeze(x, -2))
+        candidate_std = tf.sqrt(candidate_var)
+
+        neg_candidate_mean = -tf.expand_dims(candidate_mean, 1)  # [..., 1, out_dim]
+        candidate_std = tf.expand_dims(candidate_std, 1)  # [..., 1, out_dim]
```
henrymoss, line 633: Could the new function not include (almost) all the lines below here? i.e. dealing with all the upper bound stuff?

trieste/acquisition/function.py
```diff
@@ -509,6 +512,161 @@ def prepare_acquisition_function(
...
+    r"""
+    expected Hyper-volume (HV) calculating using Eq. 44 of :cite:`yang2019efficient` paper.
+    The expected hypervolume improvement calculation in the non-dominated region
+    can be decomposed into sub-calculations based on each partitioned cell.
+    For easier calculation, this sub-calculation can be reformulated as a combination
+    of two generalized expected improvements, corresponding to Psi (Eq. 44) and Nu (Eq. 45)
+    function calculations, respectively.
+    Note:
+    1. Since in Trieste we do not assume the use of a certain non-dominated partition algorithm,
+       we do not assume the last dimension partitioned cell has only one (lower) bound
```
henrymoss: last dimension "of the " partitioned ...

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +471,175 @@ def prepare_acquisition_function(
...
+    _prt = Pareto(model_pred_observation)
+    ehvi = expected_hv_improvement(model, _prt, get_reference_point(_prt.front))
+
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        ehvi(at)
+
+
+@random_seed
+@pytest.mark.parametrize(
+    "input_dim, num_samples_per_point, existing_observations, obj_num, variance_scale ,rtol, atol",
```
henrymoss: rtol and atol seem to be the same for all test, so do we need that in the pytest.mark.parametrize

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +471,175 @@ def prepare_acquisition_function(
...
+        axis=-1,
+    )  # num_cells x B
+    splus_idx = tf.expand_dims(tf.cast(splus_valid, dtype=ub_points.dtype), -1)
+    splus_lb = tf.tile(lb_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+    splus_lb = tf.maximum(splus_lb, tf.expand_dims(predict_samples, 1))
+    splus_ub = tf.tile(ub_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+    splus = tf.concat([splus_idx, splus_ub - splus_lb], axis=-1)
+
+    ehvi_approx = tf.transpose(tf.reduce_sum(tf.reduce_prod(splus, axis=-1), axis=1, keepdims=True))
+    ehvi_approx = tf.reduce_mean(ehvi_approx, axis=-1)
```
henrymoss: Whats going on here?

henrymoss: i.e. what is splus?

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +471,175 @@ def prepare_acquisition_function(
     npt.assert_allclose(eci(x), ei(x) * pof(x))
 
 
+class _PseudoTrainableQuadratic(QuadraticMeanAndRBFKernel, PseudoTrainableProbModel):
+    pass
```
joelberkeley: you shouldn't need `PseudoTrainableProbModel` when testing acquisition functions, because acquisition functions don't train the model. You can just use `QuadraticMeanAndRBFKernel`

docs/refs.bib
```diff
@@ -84,3 +84,14 @@ @article{Couckuyt2012
...
+
+@article{yang2019efficient,
+  title={Efficient computation of expected hypervolume improvement using box decomposition algorithms},
+  author={Yang, Kaifeng and Emmerich, Michael and Deutz, Andr{\'e} and B{\"a}ck, Thomas},
+  journal={Journal of Global Optimization},
+  volume={75},
+  number={1},
+  pages={3--34},
+  year={2019},
+  publisher={Springer}
```
joelberkeley:
```suggestion
    title={Efficient computation of expected hypervolume improvement using box decomposition algorithms},
    author={Yang, Kaifeng and Emmerich, Michael and Deutz, Andr{\'e} and B{\"a}ck, Thomas},
    journal={Journal of Global Optimization},
    volume={75},
    number={1},
    pages={3--34},
    year={2019},
    publisher={Springer}
```

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +471,175 @@ def prepare_acquisition_function(
...
+    def __init__(
+        self,
+        *,
+        kernel_amplitude: float | TensorType | None = None,
+    ):
+        kernel = tfp.math.psd_kernels.ExponentiatedQuadratic(kernel_amplitude)
+        super().__init__([lambda x: tf.reduce_sum(x, axis=-1, keepdims=True)], [kernel])
+
+    def __repr__(self) -> str:
+        return "_PseudoTrainableLinear()"
```
joelberkeley, line 488: similarly here: you don't need this class, you can just use
```python
def linear_kernel_gp(kernel_amplitude: float | TensorType | None = None) -> GaussianProcess:
    return GaussianProcess(
        [lambda x: tf.reduce_sum(x, axis=-1, keepdims=True)],
        [tfp.math.psd_kernels.ExponentiatedQuadratic(kernel_amplitude)],
    )
```
since you don't need to inherit from `PseudoTrainableProbModel`

trieste/acquisition/function.py
```diff
@@ -509,6 +512,161 @@ def prepare_acquisition_function(
...
+
+        def Psi(a: TensorType, b: TensorType, mean: TensorType, std: TensorType) -> TensorType:
+            """
+            Generic Expected Improvement defined at Eq. 19 in :cite:`yang2019efficient`,
+            calculate the expected improvement on reference a, integrating from b
+            :param a: shape [num_cells, out_dim] expected improvement reference point
+            :param b: shape [num_cells, out_dim] integral upper bounds
+            :param mean: shape [..., out_dim]
+            :param std: shape [..., out_dim]
+            """
```
joelberkeley: docstrings for inner functions will never reach the docs site (trieste don't use them elsewhere, indeed you haven't used them above). Just sth you might want to know. I'd not bother with them

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +471,175 @@ def prepare_acquisition_function(
...
+    expected = expected_hv_improvement(model, _prt, get_reference_point(_prt.front))(xs)
+    npt.assert_allclose(acq_fn(xs), expected)
+
+
+@pytest.mark.parametrize("at", [tf.constant([[0.0], [1.0]]), tf.constant([[[0.0], [1.0]]])])
+def test_ehvi_raises_for_invalid_batch_size(at: TensorType) -> None:
+    num_obj = 2
+    train_x = tf.constant([[-2.0], [-1.5], [-1.0], [0.0], [0.5], [1.0], [1.5], [2.0]])
+
+    model = ModelStack(*[(_mo_test_models[_](), 1) for _ in range(num_obj)])
```
joelberkeley: general comment: underscore in python (mostly) refers to a throwaway variable, so whilst it's not incorrect, it reads odd to then use it

trieste/acquisition/function.py
```diff
@@ -509,6 +512,161 @@ def prepare_acquisition_function(
         return lambda at: expected_improvement(objective_model, eta)(at) * constraint_fn(at)
 
 
+class ExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder):
+    """
+    Builder for the :func:`expected_hv_improvement` acquisition function.
```
joelberkeley: this is a builder for the expected hypervolume improvement acquisition function as opposed to a builder for `expected_hv_improvement`, as that's not the acquisition function. Similarly below, `prepare_acquisition_function` returns the expected hypervolume improvement acquisition function rather than `The expected_hv_improvement function`

tests/unit/acquisition/test_function.py
TsingQAQ: Here is a bit dirty indeed, basically this is another way of calculating HVI, the code is mainly inheriting the [gpflowopt implementation](https://github.com/GPflow/GPflowOpt/blob/master/gpflowopt/acquisition/hvpoi.py) with a modification to allow MC based acq calculating.

The main idea here for calculating ehvi is that, for a certain (partitioned) cell: 

- `splus_valid`: denote if this predicting point (or `predict_samples` in this case) can contribute hypervolume improvement in the cell (it can only contribute if it can dominate the upper bound point (i.e., worst point) of the cell, which is what `>` and `reduce_all` does).
- `splus_ub`, variable storing the upper bound of the partitioned cell
- `splus_lb`, variable storing the dimensional-wise max(lower bound, predict point)
- Line 633: `splus_ub - splus_lb` is a way of representing the (dimensional-wise) improvement, which can be used to calculate the HVI in this cell
- Line 636: average through MC sample

- splus: a variable containing validity label: `splus_valid`, as well as the contribution part:`splus_ub - splus_lb`. Then we can make use of it to calculate the HVI given that it is a valide contribution.

A maybe more interpetable example can be: 
so consider a simplest case with only 1 cell (2obj): lb: [-inf, -inf],  ub: [0, 0], 
1. for pred [-0.5, -0.5], as it dominates [0, 0], `splus_valid` = 1 means it is a valid point (can contribute on this cell)
the `splus_lb` will record something like max([-inf, -inf], [-0.5, -0.5]), and then use the splus_ub - splus_lb can finally be used for calculating its hvi.
2. for pred [-0.5, 1], or [1, 1], `splus_valid`=0, means it does not contributed in this cell.

tests/unit/acquisition/test_function.py
TsingQAQ: When I earlier directly make use of `QuadraticMeanAndRBFKernel`, I got some type checking error (shown [here](https://github.com/secondmind-labs/trieste/runs/2344041151)) as `ModelStack` receive `TrainableProbabilisticModel`,  is there anything that am not doing properlly?

joelberkeley: you're right, I hadn't thought of that. Perhaps it would be good to have a version of `ModelStack` for non-trainable `ProbabilisticModel`s. Something to think about after this PR

joelberkeley: aha, I've just remembered this is how I wrote `GaussianProcess`. IIRC the mean and kernel lists are one for each output dimension, so you can do sth like
```python
def _mo_test_model(*kernel_amplitudes: float | TensorType | None = None) -> GaussianProcess:
    means = [quadratic, lambda x: tf.reduce_sum(x, axis=-1, keepdims=True), quadratic]
    kernels = [tfp.math.psd_kernels.ExponentiatedQuadratic(k) for k in kernel_amplitudes]
    return GaussianProcess(means, kernels)
```
You don't even need a `ModelStack` for the tests

I'm saying this so people know what `GaussianProcess` can do

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +471,175 @@ def prepare_acquisition_function(
...
+_mo_test_models = (
+    _PseudoTrainableQuadratic,
+    _PseudoTrainableLinear,
+    _PseudoTrainableQuadratic,
+)
+
+
+def test_ehvi_builder_raises_for_empty_data() -> None:
+    num_obj = 3
+    dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, 1]))
```
joelberkeley: couple of points
  - `empty_dataset` is a helper for this kind of thing. Can save a few key presses.
  - also output shapes don't match
```suggestion
    dataset = Dataset(tf.zeros([0, 2]), tf.zeros([0, num_obj]))
```

tests/unit/acquisition/test_function.py
TsingQAQ: Ah I see, was earlier bit trapped with only using `ModelStack` there.
This is already much neater than the current cumbersome test model, will use sth like this indeed, thanks!

tests/unit/acquisition/test_function.py
TsingQAQ: The rtol and atol is more borrowing from the test of `test_expected_improvement`, but was not very sure how to properly set them through different tests especially with a coupled effect on mc sample num & variance scale (kernel amplitude). 

So guess if a same rtol and atol is ok, it can be removed from `pytest.mark.parametrize`, as the current code has modified.

tests/unit/acquisition/test_function.py
TsingQAQ: Removed the `TrainableProbabilisticModel` by using a modified `_mo_test_model` here (Line 468) instead for mo test.

trieste/acquisition/function.py
TsingQAQ: Okay, then it has been mostly removed.

tests/unit/acquisition/test_function.py
TsingQAQ: Changed.

tests/unit/acquisition/test_function.py
henrymoss: Could you perhaps add a couple of inline comments about whats going on? 

Basically a comment for each of the bullet points above

henrymoss: Then I think we can merge. Next we can do the synthetic function PR and then get a notebook in?

TsingQAQ: Comments have been added.

TsingQAQ: The synthetic function PR → notebook is a good order.

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +465,147 @@ def prepare_acquisition_function(
...
+    # calc MC approx EHVI
+    splus_valid = tf.reduce_all(
+        tf.tile(ub_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+        > tf.expand_dims(predict_samples, axis=1),
+        axis=-1,  # validity label representing whether predict_samples can contribute hvi in cell
+    )  # [f_samples, num_cells,  B]
+    splus_idx = tf.expand_dims(tf.cast(splus_valid, dtype=ub_points.dtype), -1)
+    splus_lb = tf.tile(lb_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+    splus_lb = tf.maximum(  # dimensional wise max of cell lower bound & predict samples
+        splus_lb, tf.expand_dims(predict_samples, 1)
```
henrymoss: max of lower bounds and predict_samples

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +465,147 @@ def prepare_acquisition_function(
...
+    ref_pt = get_reference_point(_pareto.front)
+    lb_points, ub_points = _pareto.hypercell_bounds(
+        tf.constant([-math.inf] * ref_pt.shape[-1]), ref_pt
+    )
+
+    # calc MC approx EHVI
+    splus_valid = tf.reduce_all(
+        tf.tile(ub_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+        > tf.expand_dims(predict_samples, axis=1),
+        axis=-1,  # validity label representing whether predict_samples can contribute hvi in cell
```
henrymoss: can predict_samples contribute to hvi in cell

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +465,147 @@ def prepare_acquisition_function(
...
+    )  # [f_samples, num_cells,  B]
+    splus_idx = tf.expand_dims(tf.cast(splus_valid, dtype=ub_points.dtype), -1)
+    splus_lb = tf.tile(lb_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+    splus_lb = tf.maximum(  # dimensional wise max of cell lower bound & predict samples
+        splus_lb, tf.expand_dims(predict_samples, 1)
+    )
+    splus_ub = tf.tile(ub_points[tf.newaxis, :, tf.newaxis, :], [num_samples_per_point, 1, N, 1])
+    splus = tf.concat(  # concatenate validity labels & dimensional wise (possible) improvement
```
henrymoss: concatenate validity labels and possible improvements

tests/unit/acquisition/test_function.py
```diff
@@ -461,6 +465,147 @@ def prepare_acquisition_function(
...
+        [splus_idx, splus_ub - splus_lb], axis=-1
+    )
+    # calc hvi: in case hvi contributing label is valid:
+    # calculate the product of dimensional wise improvement
+    # sum up hvi through all partitioned cell to get the hvi in the non-dominated region
```
henrymoss: calculate hyper-volume improvement over the non-dominated cells

henrymoss (approved):

PR #204, opened 2021-03-17
Hartmann test functions

henrymoss:
I have added some additional test functions to test the (slightly) higher dimensional capabilities of Trieste. 

We now have the 3-d and 6-d Hartmann functions. These are very popular benchmarks, and I have used them a lot in my own work.


henrymoss:
Woops! Still had all the logs from the MES branch. Ignore them

trieste/utils/objectives.py
```diff
@@ -126,6 +126,103 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+            [2329.0, 4135.0, 8307.0, 3736.0, 1004.0, 9991.0],
+            [2348.0, 1451.0, 3522.0, 2883.0, 3047.0, 6650.0],
+            [4047.0, 8828.0, 8732.0, 5743.0, 1091.0, 381.0],
+        ],
+        dtype=tf.float64,
+    )
+    P = P * 10 ** (-4)
+
+    inner_sum = -tf.reduce_sum(A * tf.math.square(tf.expand_dims(x, 1) - P), -1)
+    return -tf.reduce_sum(a * tf.math.exp(inner_sum), -1)
```
joelberkeley: is this the same as for `hartmann_3`? If so, how about we pull out the common bits into a module-private function?

trieste/utils/objectives.py
```diff
@@ -126,6 +126,103 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+"""
+The global minimum for the :func:`hartmann_3` function, with shape [1] and dtype
+float64.
+"""
+
+
+def hartmann_6(x: TensorType) -> TensorType:
+    """
+    The Hartmann 6 test function over :math:`[0, 1]^6`. This function has
+    6 local and one global minima.
```
joelberkeley: there are references for this in https://www.sfu.ca/~ssurjano/hart6.html

do you think it's worth adding one?

trieste/utils/objectives.py
```diff
@@ -126,6 +126,103 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+    The Hartmann 3 test function over :math:`[0, 1]^3`. This function has 3 local
+    and one global minima.
+
+    :param x: The points at which to evaluate the function, with shape [..., 3].
+    :return: The function values at ``x``, with shape [..., 1].
+    :raise ValueError (or InvalidArgumentError): If ``x`` has an invalid shape.
+    """
+    tf.debugging.assert_shapes([(x, (..., 3))])
+
+    a = tf.constant([1.0, 1.2, 3.0, 3.2], dtype=tf.float64)
```
joelberkeley: if you make these variables lists, the function will also work with other floating point types

trieste/utils/objectives.py
```diff
@@ -126,6 +126,103 @@ def logarithmic_goldstein_price(x: TensorType) -> TensorType:
...
+            [3689.0, 1170.0, 2673.0],
+            [4699.0, 4387.0, 7470.0],
+            [1091.0, 8732.0, 5547.0],
+            [381.0, 5743.0, 8828.0],
+        ],
+        dtype=tf.float64,
+    )
+    P = P * 10 ** (-4)
+
+    inner_sum = -tf.reduce_sum(A * tf.math.square(tf.expand_dims(x, 1) - P), -1)
```
joelberkeley: curious - why `square` over `** 2`

trieste/utils/objectives.py
henrymoss: IS that a function with an underscore before its name?

trieste/utils/objectives.py
henrymoss: Can I add just the URL or do you want a proper reference in the docs file?

trieste/utils/objectives.py
henrymoss: Nice trick!

trieste/utils/objectives.py
henrymoss: No reason. I will change, as ** 2 is more readable

trieste/utils/objectives.py
joelberkeley: no need to add the URL - that's already in the module docstring. I was thinking a proper reference, but I'll leave this up to you. I'm pretty ambivalent

trieste/utils/objectives.py
joelberkeley: yes. no docstring either else it will appear in the docs, and no tests

henrymoss: I think this is fine as it is?

tests/unit/utils/test_objectives.py
```diff
@@ -65,21 +60,33 @@ def test_objective_maps_minimizers_to_minimum(
 
 
 def test_branin_no_function_values_are_less_than_global_minimum() -> None:
-    npt.assert_array_less(tf.squeeze(BRANIN_MINIMUM) - 1e-6, branin(_unit_grid_2d()))
+    samples = Box([0.0, 0.0], [1.0, 1.0]).sample(1000)
```
joelberkeley, line 63: `tf.random.uniform` is more powerful, and would be consistent with similar random sampling across other tests in trieste

trieste/utils/objectives.py
joelberkeley: seems like a good opportunity for DRY - the two functions only differ in A and P

joelberkeley (approved):

PR #207, opened 2021-03-18
Pareto: Divide and conquer cells

### LINKED ISSUE #180: Pareto set: n objectives
As a user, I want functionality available to calculate the cell bounds of a Pareto front for the case of any number of objectives, so that I can easily find the cell bounds when defining multi-objective acquisition functionality

This should be valid for any number of objectives, including 1 and 2

satrialoka:
### Feature Implemented

This is a feature for Pareto functionality needed for Multi-objective BO, the code was ported from [GPFlowopt's Pareto class](https://github.com/GPflow/GPflowOpt/blob/master/gpflowopt/pareto.py). This PR is small PR derived from previously large PR #126 this PR resolved #180
 
### Approach

Inside pareto.py a method is implemented:

1. `divide_conquer_nd` method for calculating the box decomposition 

## Status
Need Review

## Changes
add some test

## Notes
the points for the test are drawn from the unit cube and to confirm the validity, I inspect it by visualizing the points and the resulting hypercube.
The algorithm is based on this paper: https://ieeexplore.ieee.org/document/6256586


trieste/utils/pareto.py
```diff
@@ -106,6 +113,166 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+                (int(pf_ext_idx.shape[0]) - 1) * tf.ones(number_of_objectives, dtype=tf.int32),
+            ],
+            axis=0,
+        )[None]
+
+        total_size = tf.reduce_prod(max_pf - min_pf)
+        # Start divide and conquer until we processed all cells
+
+        loop_cond = lambda dc, lower, upper: dc.shape[0] > 0
+        loop_body = lambda dc, lower, upper: self._while_body(
```
satrialoka: lambda function doesn't pass flake8, not sure if using `def` will make it more readable.

satrialoka: plug in the lambda function on the tf.while_loop instead

trieste/utils/pareto.py
```diff
@@ -73,16 +74,22 @@ class Pareto:
     The latter is needed for certain multiobjective acquisition functions.
     """
 
-    def __init__(self, observations: TensorType):
+    def __init__(self, observations: TensorType, generic_strategy: bool = False):
         """
         :param observations: The observations for all objectives, with shape [N, 2].
```
```suggestion
        :param observations: The observations for all objectives, with shape [N, D].
```

trieste/utils/pareto.py
```diff
@@ -106,6 +113,164 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        upper = tf.concat([upper, idx_ub[None]], axis=0)
+        return lower, upper
+
+    def _rejected_test_body(self, cell, lb, ub, dc, total_size, threshold):
+
+        dc_dist = cell[1] - cell[0]
+        hc_size = tf.math.reduce_prod(ub - lb, axis=0, keepdims=True)
+
+        # Only divide when it is not an unit cell
+        # and the volume is above the approx. threshold
```
joelberkeley: comments drift over time and become counterproductive. Can we leave them out?

trieste/utils/pareto.py
```diff
@@ -106,6 +113,164 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        # :return: True if the test point dominates or augments the Pareto front (boolean)
+
+        # if and only if the test point is at least in one dimension smaller
+        # for every point in the Pareto set
+        idx_dom_augm = tf.reduce_any(smaller, axis=1)
+        is_dom_augm = tf.reduce_all(idx_dom_augm)
+
+        return is_dom_augm
+
+    def divide_conquer_nd(self, front: TensorType, threshold: TensorType = 0) -> TensorType:
```
joelberkeley: any reason this is public, and if so, why is the 2D version not public?

trieste/utils/pareto.py
```diff
@@ -73,16 +74,22 @@ class Pareto:
     The latter is needed for certain multiobjective acquisition functions.
     """
 
-    def __init__(self, observations: TensorType):
+    def __init__(self, observations: TensorType, generic_strategy: bool = False):
```
joelberkeley: what's the use case for `generic_strategy`? Why would a user want to use the generic strategy when the 2D strategy works (and is presumably quicker?)?

trieste/utils/pareto.py
```diff
@@ -106,6 +113,164 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        lower,
+        upper,
+        number_of_objectives,
+        pf_ext,
+        pf_ext_idx,
+        jitter,
+        front,
+        total_size,
+        threshold,
+    ):
```
joelberkeley: if you make this function local to `divide_conquer_nd` it will keep the namespace of `Pareto` clearer. You'll also be able to use local variables like `number_of_objectives` without using a `lambda` which should make the code easier to read

tests/unit/test_pareto.py
```diff
@@ -145,10 +165,104 @@ def test_pareto_hypervolume_indicator_raises_for_reference_below_anti_ideal_poin
...
+            [0.8003, 0.8491],
+            [0.1419, 0.9340],
+        ]
+    )
+
+    pareto = Pareto(objectives, generic_strategy=True)
+
+    npt.assert_allclose(
+        pareto.bounds.lower_idx, tf.constant([[3, 0], [2, 0], [1, 2], [0, 2], [0, 0]])
+    )
```
joelberkeley:
```suggestion
    npt.assert_allclose(
        pareto.bounds.lower_idx, [[3, 0], [2, 0], [1, 2], [0, 2], [0, 0]]
    )
```
lists will do for numpy testing asserts. It might also mean you can fit them on one line, and is less to read

tests/unit/test_pareto.py
```diff
@@ -145,10 +165,104 @@ def test_pareto_hypervolume_indicator_raises_for_reference_below_anti_ideal_poin
...
+            [0.9572, 0.6557],
+            [0.4854, 0.0357],
+            [0.8003, 0.8491],
+            [0.1419, 0.9340],
+        ]
+    )
+
+    pareto = Pareto(objectives, generic_strategy=True)
+
+    npt.assert_allclose(
```
joelberkeley: these should be exactly equal not `allclose`

tests/unit/test_pareto.py
```diff
@@ -145,10 +165,104 @@ def test_pareto_hypervolume_indicator_raises_for_reference_below_anti_ideal_poin
...
+    objectives = tf.constant(
+        [
+            [0.9575, 0.4218],
+            [0.9649, 0.9157],
+            [0.1576, 0.7922],
+            [0.9706, 0.9595],
+            [0.9572, 0.6557],
+            [0.4854, 0.0357],
+            [0.8003, 0.8491],
+            [0.1419, 0.9340],
```
joelberkeley: it will be easier to see what the test is testing with simpler numbers. With long numbers, there's more to read which makes test failures harder to debug, and also makes refactoring harder. A future developer might well wonder: are these numbers so detailed for a reason?

trieste/utils/pareto.py
```diff
@@ -106,6 +113,164 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        is_dom_augm = tf.reduce_all(idx_dom_augm)
+
+        return is_dom_augm
+
+    def divide_conquer_nd(self, front: TensorType, threshold: TensorType = 0) -> TensorType:
+        # Divide and conquer strategy to compute the cells covering the non-dominated region.
+        # Generic version: works for an arbitrary number of objectives.
+
+        len_front, number_of_objectives = front.shape
+        jitter = gpflow.config.default_jitter()
```
joelberkeley: we don't use gpflow's jitter in trieste, so as to avoid mutable globals. Instead you can use a default argument like [in `BatchReparametrizationSampler`](https://github.com/secondmind-labs/trieste/blob/develop/trieste/acquisition/function.py#L601). I'm not sure how you'd make this make sense for the 2D case (where `jitter` would be an argument but unused), maybe you can think of sth

trieste/utils/pareto.py
```diff
@@ -106,6 +113,164 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        test_rejected = self._is_test_required((lb + jitter) < front)
+        dc = tf.cond(
+            tf.logical_and(test_rejected, tf.logical_not(test_accepted)),
+            lambda: self._rejected_test_body(cell, lb, ub, dc, total_size, threshold),
+            lambda: dc,
+        )
+
+        return dc, lower, upper
+
+    def _accepted_test_body(self, lower, upper, idx_lb, idx_ub):
```
joelberkeley: these functions will need type hints

trieste/utils/pareto.py
```diff
@@ -106,6 +113,164 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+
+        # Only divide when it is not an unit cell
+        # and the volume is above the approx. threshold
+        not_unit_cell = tf.reduce_any(dc_dist > 1)
+        vol_above_thresh = tf.reduce_all((hc_size[0] / total_size) > threshold)
+        dc = tf.cond(
+            tf.logical_and(not_unit_cell, vol_above_thresh),
+            lambda: self._divide_body(dc, dc_dist, cell),
+            lambda: tf.identity(dc),
+        )
```
joelberkeley: was there a particular reason to use `tf.cond` over
```python
dc = (
    self._divide_body(dc, dc_dist, cell)
    if tf.logical_and(not_unit_cell, vol_above_thresh)
    else tf.identity(dc)
)
```

trieste/utils/pareto.py
```diff
@@ -106,6 +113,164 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+
+        len_front, number_of_objectives = front.shape
+        jitter = gpflow.config.default_jitter()
+        lower = tf.zeros([0, number_of_objectives], dtype=tf.int32)
+        upper = tf.zeros([0, number_of_objectives], dtype=tf.int32)
+
+        # Extend front with the ideal and anti-ideal point
+        min_pf = tf.reduce_min(front, axis=0, keepdims=True) - 1
+        max_pf = tf.reduce_max(front, axis=0, keepdims=True) + 1
+        pf_ext = tf.concat(
```
joelberkeley: can we have longer names? `pf_ext`, for example, is pretty cryptic

trieste/utils/pareto.py
satrialoka: Yes. I'll remove them.

trieste/utils/pareto.py
satrialoka: just add a choice to user to compares those two. but it seems not necessary to add this. I'll remove it

tests/unit/test_pareto.py
satrialoka: not for a reason, just sampled it from unit cube. I'll change it to shorter numbers

trieste/utils/pareto.py
satrialoka: no actually, I'll update it

trieste/utils/pareto.py
satrialoka: Just thinking about it again. it is needed to test if generic_strategy also valid for 2d case. if we remove it we should also remove 2d test case for divide_conquer_nd. what do you think?

tests/unit/test_pareto.py
satrialoka: I replaced all integer related values with `assert_array_equal`

trieste/utils/pareto.py
satrialoka: Ah. I was thinking to replace all python if-else with `tf.cond`. it's not necessary here right?

joelberkeley: i might have said that. I can't remember. It's not important so we can leave it as is

trieste/utils/pareto.py
joelberkeley: I'd say remove it. I believe it's not a good idea to expose functionality for the sake of testing - you'll end up having to maintain a larger API.

on this subject, I'm not sure about `divide_conquer_nd` being public at all. It's already called at initialisation and the result exposed via `.bounds`

satrialoka: okay

satrialoka: removed it

trieste/utils/pareto.py
satrialoka: done adding type hints

trieste/utils/pareto.py
```diff
@@ -101,10 +108,159 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+            return divide_conquer_cells, lower_result, upper_result
+
+        _, lower_result, upper_result = tf.while_loop(
+            lambda divide_conquer_cells, lower_result, upper_result: divide_conquer_cells.shape[0]
+            > 0,
+            lambda divide_conquer_cells, lower_result, upper_result: while_body(
+                divide_conquer_cells,
+                lower_result,
+                upper_result,
+            ),
```
joelberkeley: no need for the `lambda` here
```suggestion
            while_body,
```

satrialoka: ah yeah missed that

trieste/utils/pareto.py
satrialoka: changed it to private

trieste/utils/pareto.py
```diff
@@ -101,10 +108,155 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+                    cell, lower, upper, divide_conquer_cells, total_size, threshold
+                ),
+                lambda: divide_conquer_cells,
+            )
+
+            return divide_conquer_cells, lower_result, upper_result
+
+        _, lower_result, upper_result = tf.while_loop(
+            lambda divide_conquer_cells, lower_result, upper_result: divide_conquer_cells.shape[0]
+            > 0,
```
joelberkeley: this reads a little like it's two arguments. is it ok to put it on one line e.g. by using `len(divide_conquer_cells)` or `lambda divide_conquer_cells, _lower, _upper: ...`

trieste/utils/pareto.py
```diff
@@ -101,10 +108,155 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        return BoundedVolumes(lower_result, upper_result)
+
+    @staticmethod
+    def _is_test_required(smaller: TensorType) -> TensorType:
+        idx_dom_augm = tf.reduce_any(smaller, axis=1)
+        is_dom_augm = tf.reduce_all(idx_dom_augm)
+
+        return is_dom_augm
+
+    def _divide_conquer_nd(self, front: TensorType, threshold: TensorType = 0) -> TensorType:
```
```suggestion
    def _divide_conquer_nd(self, front: TensorType, threshold: TensorType | float = 0) -> TensorType:
```
so the default makes sense?

trieste/utils/pareto.py
```diff
@@ -101,10 +108,155 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+    ):
+        edge_size, idx = tf.reduce_max(divide_conquer_cells_dist), tf.argmax(
+            divide_conquer_cells_dist
+        )
+        edge_size1 = int(tf.round(tf.cast(edge_size, dtype=tf.float32) / 2.0))
+        edge_size2 = edge_size - edge_size1
+
+        upper = tf.unstack(tf.identity(cell[1]))
+        upper[idx] = upper[idx] - edge_size1
+        upper = tf.stack(upper)
```
joelberkeley: this mutation is probably not going to play nice with `tf.function`. Is it possible to make `edge_size1` into an appropriately shaped tensor and subtract it straight from `tf.identity(cell[1])`?

joelberkeley: similar for `lower` later

trieste/utils/pareto.py
```diff
@@ -101,10 +108,155 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+
+    def _divide_body(
+        self,
+        divide_conquer_cells: TensorType,
+        divide_conquer_cells_dist: TensorType,
+        cell: TensorType,
+    ):
+        edge_size, idx = tf.reduce_max(divide_conquer_cells_dist), tf.argmax(
+            divide_conquer_cells_dist
+        )
```
```suggestion
        edge_size = tf.reduce_max(divide_conquer_cells_dist)
        idx = tf.argmax(divide_conquer_cells_dist)
```
perhaps, since you're already going over the line?

trieste/utils/pareto.py
```diff
@@ -101,10 +108,155 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+            axis=0,
+        )[None]
+
+        total_size = tf.reduce_prod(max_front - min_front)
+
+        def while_body(
+            divide_conquer_cells: TensorType,
+            lower_result: TensorType,
+            upper_result: TensorType,
+        ):
```
```suggestion
        ) -> tuple[TensorType, TensorType, TensorType]:
```

trieste/utils/pareto.py
```diff
@@ -101,10 +108,155 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+            lambda: tf.identity(divide_conquer_cells),
+        )
+        return divide_conquer_cells
+
+    def _divide_body(
+        self,
+        divide_conquer_cells: TensorType,
+        divide_conquer_cells_dist: TensorType,
+        cell: TensorType,
+    ):
```
```suggestion
    ) -> TensorType:
```

trieste/utils/pareto.py
```diff
@@ -101,10 +108,155 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+
+    def _rejected_test_body(
+        self,
+        cell: TensorType,
+        lower: TensorType,
+        upper: TensorType,
+        divide_conquer_cells: TensorType,
+        total_size: TensorType,
+        threshold: TensorType,
+    ):
```
```suggestion
    ) -> TensorType:
```

trieste/utils/pareto.py
```diff
@@ -101,10 +108,155 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+            loop_vars=[divide_conquer_cells, lower_result, upper_result],
+            shape_invariants=[
+                tf.TensorShape([None, 2, number_of_objectives]),
+                tf.TensorShape([None, number_of_objectives]),
+                tf.TensorShape([None, number_of_objectives]),
+            ],
+        )
+        return BoundedVolumes(lower_result, upper_result)
+
+    def _accepted_test_body(self, lower_result, upper_result, lower_idx, upper_idx):
```
```suggestion
    def _accepted_test_body(
        self,
        lower_result: TensorType,
        upper_result: TensorType,
        lower_idx: TensorType,
        upper_idx: TensorType
    ) -> tuple[TensorType, TensorType]:
```

trieste/utils/pareto.py
satrialoka: fixed

trieste/utils/pareto.py
satrialoka: agree

trieste/utils/pareto.py
satrialoka: yeah thats better

trieste/utils/pareto.py
joelberkeley: in a manner of speaking

trieste/utils/pareto.py
satrialoka: missed it sorry

trieste/utils/pareto.py
satrialoka: sorry. missed the two way merge I seems.

trieste/utils/pareto.py
satrialoka: got this error
`ERROR test_pareto.py - TypeError: unsupported operand type(s) for |: '_GenericAlias' and 'type'`
will check it

trieste/utils/pareto.py
satrialoka: is it make sense if we only pass the jitter to divide_conquer_nd? and not passing it to 2d case? so it's only param on divide_conquer_nd and Pareto constructor.

joelberkeley: that makes sense, but you'll need it in `__init__`. On reflection, I don't think that's so bad as long as it has a default, which it will

trieste/utils/pareto.py
joelberkeley: add `from __future__ import annotations` above all the other imports: this is an advanced mypy feature that requires this import

satrialoka: okay 👌🏼

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        total_size = tf.reduce_prod(max_front - min_front)
+
+        def while_body(
+            divide_conquer_cells: TensorType,
+            lower_result: TensorType,
+            upper_result: TensorType,
+        ) -> Tuple[TensorType, TensorType, TensorType]:
+            divide_conquer_cells = tf.unstack(divide_conquer_cells, axis=0)
+            cell = divide_conquer_cells[-1]
+            divide_conquer_cells = tf.cond(
```
joelberkeley: this also might not work as expected with `tf.function` (re-assigning to `divide_conquer_cells`), but should be as simple as using a new variable name?

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+
+            test_accepted = self._is_test_required((upper - jitter) < front)
+            lower_result, upper_result = tf.cond(
+                test_accepted,
+                lambda: self._accepted_test_body(lower_result, upper_result, lower_idx, upper_idx),
+                lambda: (lower_result, upper_result),
+            )
+
+            test_rejected = self._is_test_required((lower + jitter) < front)
+            divide_conquer_cells = tf.cond(
```
joelberkeley: here's another re-assignment

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+                tf.logical_and(test_rejected, tf.logical_not(test_accepted)),
+                lambda: self._rejected_test_body(
+                    cell, lower, upper, divide_conquer_cells, total_size, threshold
+                ),
+                lambda: divide_conquer_cells,
+            )
+
+            return divide_conquer_cells, lower_result, upper_result
+
+        _, lower_result, upper_result = tf.while_loop(
```
joelberkeley: here's another re-assignment

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        self,
+        cell: TensorType,
+        lower: TensorType,
+        upper: TensorType,
+        divide_conquer_cells: TensorType,
+        total_size: TensorType,
+        threshold: TensorType,
+    ) -> TensorType:
+
+        divide_conquer_cells_dist = cell[1] - cell[0]
```
joelberkeley, line 238: here's another re-assignment - I'm not quite sure what's happening here. You've got `divide_conquer_cells` as an argument then you immediately re-assign to it

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+    ) -> TensorType:
+        edge_size = tf.reduce_max(divide_conquer_cells_dist)
+        idx = tf.argmax(divide_conquer_cells_dist)
+        edge_size1 = int(tf.round(tf.cast(edge_size, dtype=tf.float32) / 2.0))
+        edge_size2 = edge_size - edge_size1
+
+        upper = tf.unstack(tf.identity(cell[1]))
+        upper[idx] = upper[idx] - edge_size1
+        upper = tf.stack(upper)
+        divide_conquer_cells = tf.concat(
```
joelberkeley: here's another re-assignment

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+            [divide_conquer_cells, tf.stack([tf.identity(cell[0]), upper], axis=0)[None]], axis=0
+        )
+        lower = tf.unstack(tf.identity(cell[0]))
+        lower[idx] = lower[idx] + edge_size2
+        lower = tf.stack(lower)
+        divide_conquer_cells = tf.concat(
```
joelberkeley: here's another re-assignment

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        total_size: TensorType,
+        threshold: TensorType,
+    ) -> TensorType:
+
+        divide_conquer_cells_dist = cell[1] - cell[0]
+        hc_size = tf.math.reduce_prod(upper - lower, axis=0, keepdims=True)
+
+        not_unit_cell = tf.reduce_any(divide_conquer_cells_dist > 1)
+        vol_above_thresh = tf.reduce_all((hc_size[0] / total_size) > threshold)
+        divide_conquer_cells = tf.cond(
```
joelberkeley, line 237: here's another re-assignment

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+    def _accepted_test_body(
+        self,
+        lower_result: TensorType,
+        upper_result: TensorType,
+        lower_idx: TensorType,
+        upper_idx: TensorType,
+    ) -> Tuple[TensorType, TensorType]:
+        lower_result = tf.concat([lower_result, lower_idx[None]], axis=0)
+        upper_result = tf.concat([upper_result, upper_idx[None]], axis=0)
+        return lower_result, upper_result
```
joelberkeley: here's another re-assignment

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+            )
+
+            arr = tf.range(number_of_objectives)
+            lower_idx = tf.gather_nd(pseudo_front_idx, tf.stack((cell[0], arr), -1))
+            upper_idx = tf.gather_nd(pseudo_front_idx, tf.stack((cell[1], arr), -1))
+            lower = tf.gather_nd(pseudo_front, tf.stack((lower_idx, arr), -1))
+            upper = tf.gather_nd(pseudo_front, tf.stack((upper_idx, arr), -1))
+
+            test_accepted = self._is_test_required((upper - jitter) < front)
+            lower_result, upper_result = tf.cond(
```
joelberkeley: here's another re-assignment

trieste/utils/pareto.py
```diff
@@ -101,10 +110,161 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        )[None]
+
+        total_size = tf.reduce_prod(max_front - min_front)
+
+        def while_body(
+            divide_conquer_cells: TensorType,
+            lower_result: TensorType,
+            upper_result: TensorType,
+        ) -> Tuple[TensorType, TensorType, TensorType]:
+            divide_conquer_cells = tf.unstack(divide_conquer_cells, axis=0)
```
joelberkeley: here's another re-assignment

trieste/utils/pareto.py
satrialoka: DEFAULT is defined in utils, circular import occurred, seems I need to move pareto.py back outside utils

trieste/utils/pareto.py
satrialoka: is there rank1 version of `tf.sparse.SparseTensor` ?

trieste/utils/pareto.py
satrialoka: I see. didn't know that. hmm. is it possible to fix all these reassignments using a new variable name? if so, I am thinking what name should those be.

trieste/utils/pareto.py
joelberkeley: I don't know, but this might work (with some tweaking)
```
tf.concat([tf.zeros([idx, ...]), edge_size1 * tf.ones([...]), tf.zeros([len(cell[1]) - idx, ...])], axis=...)
```

satrialoka: I'll try it

trieste/pareto.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+class Pareto:
+    """
+    A :class:`Pareto` Construct a Pareto set.
+    Stores a Pareto set and calculates the cell bounds covering the non-dominated region.
+    The latter is needed for certain multiobjective acquisition functions.
+    """
+
+    def __init__(self, observations: TensorType, jitter: float = DEFAULTS.JITTER):
```
joelberkeley:
```suggestion
    def __init__(self, observations: TensorType, *, jitter: float = DEFAULTS.JITTER):
```
for consistency

trieste/utils/pareto.py
joelberkeley: i'll make some suggestions

trieste/utils/pareto.py
joelberkeley: might be easier, at least for reviewing's sake, to move DEFAULTS to misc.py and import DEFAULTS into `__init__.py`

satrialoka: okay i'll move DEFAULTS to misc.py . should I move back Pareto to utils?

joelberkeley: yeah that'd be great - it's difficult to follow the PR otherwise

trieste/utils/pareto.py
satrialoka: okay

trieste/utils/pareto.py
satrialoka: tried it [commit](https://github.com/secondmind-labs/trieste/pull/207/commits/b1c1b64c6136f26929fcd7187a63c40c0bc2d18b) what do you think?

trieste/pareto.py
satrialoka: added

trieste/utils/pareto.py
joelberkeley: `divide_conquer_cells_unstacked` perhaps?

trieste/utils/pareto.py
joelberkeley: you can also just add an underscore to variable names if they've already been used `lower_result_`. I wouldn't recommend it normally, because this is a complex algorithm and it's good to be as clear as possible, but I don't have time to suggest anything better

trieste/utils/pareto.py
joelberkeley: another approach is `new_divide_conquer_cells`

trieste/utils/pareto.py
joelberkeley: is this a bug?

trieste/utils/pareto.py
```diff
@@ -101,10 +110,167 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        hc_size = tf.math.reduce_prod(upper - lower, axis=0, keepdims=True)
+
+        not_unit_cell = tf.reduce_any(divide_conquer_cells_dist > 1)
+        vol_above_thresh = tf.reduce_all((hc_size[0] / total_size) > threshold)
+        divide_conquer_cells = tf.cond(
+            tf.logical_and(not_unit_cell, vol_above_thresh),
+            lambda: self._divide_body(divide_conquer_cells, divide_conquer_cells_dist, cell),
+            lambda: tf.identity(divide_conquer_cells),
+        )
+        return divide_conquer_cells
```
```suggestion
        new_divide_conquer_cells = tf.cond(
            tf.logical_and(not_unit_cell, vol_above_thresh),
            lambda: self._divide_body(divide_conquer_cells, divide_conquer_cells_dist, cell),
            lambda: tf.identity(divide_conquer_cells),
        )
        return new_divide_conquer_cells
```

trieste/utils/pareto.py
satrialoka: not quite sure on what you mean here if there is reassignment it should be at line 136? 
`divide_conquer_cells_dist` is different from `divide_conquer_cells`
could you elaborate?

joelberkeley: oh I misread this as `divide_conquer_cells` - all's fine

trieste/utils/pareto.py
satrialoka: I am thinking to use `lower_result_final` since it's "final" wrt to the method which it resides. what do you think?

joelberkeley: sounds good - you understand the algorithm so are better placed to name things

joelberkeley (approved):

trieste/utils/misc.py
```diff
@@ -171,3 +171,13 @@ def unwrap(self) -> NoReturn:
...
+
+
+class DEFAULTS:
+    """ Default constants used in Trieste. """
+
+    JITTER: Final[float] = 1e-6
+    """
+    The default jitter, typically used to stabilise computations near singular points, such as in
+    Cholesky decomposition.
+    """
```
henrymoss, line 183: Why has this been moved to misc.py?

trieste/utils/pareto.py
```diff
@@ -13,12 +13,15 @@
...
 from typing import Tuple
 
 import tensorflow as tf
 from typing_extensions import Final
 
 from ..type import TensorType
+from .misc import DEFAULTS
 
 
 def non_dominated(observations: TensorType) -> Tuple[TensorType, TensorType]:
```
henrymoss, line 27: Why are duplicate points kept?

trieste/utils/pareto.py
```diff
@@ -73,25 +76,31 @@ class Pareto:
...
+    def _get_bounds(self, front: TensorType, jitter: float) -> BoundedVolumes:
+        if front.shape[-1] > 2:
+            return self._divide_conquer_nd(front, jitter)
+        else:
+            return self._bounds_2d(front)
 
     @staticmethod
     def _bounds_2d(front: TensorType) -> BoundedVolumes:
 
-        # this assumes the Pareto set has been sorted in ascending order on the first
```
henrymoss: Say what this method does!

henrymoss: i.e. what its calculating

trieste/utils/pareto.py
```diff
@@ -101,10 +110,168 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+        idx_dom_augm = tf.reduce_any(smaller, axis=1)
+        is_dom_augm = tf.reduce_all(idx_dom_augm)
+
+        return is_dom_augm
+
+    def _divide_conquer_nd(
+        self, front: TensorType, jitter: float, threshold: TensorType | float = 0
+    ) -> TensorType:
+        # Divide and conquer strategy to compute the cells covering the non-dominated region.
+        # Generic version: works for an arbitrary number of objectives.
```
henrymoss: Does this has a reference?

henrymoss: I'm finding it hard to work out what is going on in this method.

trieste/utils/misc.py
satrialoka: We need this in Pareto.py. if we kept this this in __init__ the circular import is occured. Joel suggest to move it to misc.py 
detail: https://github.com/secondmind-labs/trieste/pull/207#discussion_r605186454

trieste/utils/pareto.py
satrialoka: do you mean the outputs of `non_dominated `function?

henrymoss: yep

trieste/utils/pareto.py
satrialoka: the reference is: https://ieeexplore.ieee.org/document/6256586 you can check the description of the algorithm at section: C. Branch-and-bound procedure

satrialoka: My main source to port/implement this is from the gpflowopt code https://github.com/GPflow/GPflowOpt/blob/3d86bcc000b0367f19e9f03f4458f5641e5dde60/gpflowopt/pareto.py#L173

henrymoss: Its definitely worth adding the reference in the docs and in the code here. Also maybe refer it to a branch and bound rather than divide an conquer so that users can quickly work out where to read.

henrymoss: Its a large block of code and currently only has a single test ( test_pareto_divide_conquer_nd_three_dimension_case). How difficult would it be for you to provide another example (maybe 4 or 5 dimensions)?

trieste/utils/pareto.py
satrialoka: The reason is we don't want to make the function slower, and haven't found TF workaround to do it efficiently. so we decide to mention in the docstring that the duplicate points are kept. detail: https://github.com/secondmind-labs/trieste/pull/143#discussion_r575835476 
what do you think?

trieste/utils/pareto.py
satrialoka: currently I inspect the validity of the result by visualizing the hyperbounds. for d>3 seems it is not possible to do that (correct me if i am wrong). do you have any suggestion?

satrialoka: > 
> 
> Its definitely worth adding the reference in the docs and in the code here. Also maybe refer it to a branch and bound rather than divide an conquer so that users can quickly work out where to read.

Should we put it in the `Pareto` class's docstring? since `_divide_and_conquer` here is private, I am not sure if it is ok to put it in `_divide_and_conquer`

henrymoss: Yes in the Pareto class.

trieste/utils/pareto.py
henrymoss: That makes sense. Where is the presence of duplicate points taken into account later on?

trieste/utils/pareto.py
henrymoss: Once you have this and have brought the branch up to date then I think we are ready to merge! I'll do a final check then :)

trieste/utils/pareto.py
satrialoka: I am afraid I don't have an answer to that for now. it might be in the test of an acquisition function algorithm when duplicate points arise frequently. but I don't have any concrete example for that for now.

trieste/utils/pareto.py
satrialoka: okay. please check the latest commit. not sure if I did it right.

henrymoss: Nice! I'm just going to ask @uri-granta  to give this a quick look over please. It would be good to get an engineers perspective on this! 

We can then merge this PR and https://github.com/secondmind-labs/trieste/pull/217 and focus on the EHVI acquisition function itself.

satrialoka: Nice.

trieste/utils/__init__.py
```diff
@@ -17,18 +17,7 @@
 from typing_extensions import Final
```
uri-granta, line 17: Can remove this import now (would be nice if the format checks caught this automatically).

trieste/utils/pareto.py
```diff
@@ -71,27 +74,37 @@ class Pareto:
     A :class:`Pareto` Construct a Pareto set.
     Stores a Pareto set and calculates the cell bounds covering the non-dominated region.
     The latter is needed for certain multiobjective acquisition functions.
+
+    For hypervolume-based multiobjective optimisation with n>2 objectives, this class
+    implement branch and bound procedure algorithm. a divide and conquer method introduced
```
uri-granta:
```suggestion
    implements branch and bound procedure algorithm. a divide and conquer method introduced
```

trieste/utils/pareto.py
```diff
@@ -71,27 +74,37 @@ class Pareto:
...
 
-    def __init__(self, observations: TensorType):
+    def __init__(self, observations: TensorType, *, jitter: float = DEFAULTS.JITTER):
         """
-        :param observations: The observations for all objectives, with shape [N, 2].
+        :param observations: The observations for all objectives, with shape [N, D].
         :raise ValueError (or InvalidArgumentError): If ``observations`` has an invalid shape.
         """
-        tf.debugging.assert_shapes([(observations, [None, 2])])
+        tf.debugging.assert_rank(observations, 2)
```
uri-granta, line 88: (Do we still want to assert that `shape[-1] >= 2`?)

trieste/utils/pareto.py
```diff
@@ -101,10 +114,168 @@ def _bounds_2d(front: TensorType) -> BoundedVolumes:
...
+            loop_vars=[divide_conquer_cells, lower_result, upper_result],
+            shape_invariants=[
+                tf.TensorShape([None, 2, number_of_objectives]),
+                tf.TensorShape([None, number_of_objectives]),
+                tf.TensorShape([None, number_of_objectives]),
+            ],
+        )
+        return BoundedVolumes(lower_result_final, upper_result_final)
+
+    def _accepted_test_body(
```
uri-granta, line 217: The following few methods could all be marked `@classmethod` or `@staticmethod`. Also might make sense to move _is_test_required down here with the other releated methods.

uri-granta (approved):
Looks good to me from an engineering perspective. Good variable names!

trieste/utils/pareto.py
satrialoka: For me personally, it might be nice to have that assertion.

trieste/utils/__init__.py
satrialoka: got this error when I remove it : `trieste/utils/pareto.py:67: error: Name 'Final' is not defined`

trieste/utils/pareto.py
satrialoka: should we use `tf.debugging.assert_greater_equal` ? not sure how to do this using ` tf.debugging.assert_shapes`

henrymoss:
Once @uri-granta's commented are addressed and the test pass, I'm happy to merge this :)

henrymoss:
Perfect. @satrialoka  do you want to merge? Then @TsingQAQ  can get  #217 up to date.

satrialoka:
yeah I think we're ready to merge

henrymoss:
Can you do it @satrialoka ?

PR #208, opened 2021-03-23
One rule to rule them all

henrymoss:
@joelberkeley  and I have refactored the batch design within Trieste. 

Batch BO is now achieved through the EfficientGlobalOptimization class through a (optional) batch size input.

All acquisition functions are defined over an additional input dimension representing batch size. If the acquisition function does not support batch BO, then an error is raised saying that it does not support batches of size >1.

Lots of tests have been added.


docs/notebooks/failure_ego.pct.py
```diff
@@ -156,7 +156,7 @@ def optimize(self, dataset):
 
 class ProbabilityOfValidity(SingleModelAcquisitionBuilder):
     def prepare_acquisition_function(self, dataset, model):
-        return lambda at: lower_confidence_bound(model, 0.0, at)
+        return lambda at: lower_confidence_bound(model, 0.0)(at)
```
joelberkeley:
```suggestion
        return lower_confidence_bound(model, 0.0)
```

trieste/acquisition/rule.py
```diff
@@ -63,19 +57,19 @@ def acquire(
 
         **Type hints:**
           - The global search space must be a :class:`~trieste.space.SearchSpace`. The exact type
-            of :class:`~trieste.space.SearchSpace` depends on the specific
-            :class:`AcquisitionRule`.
+                of :class:`~trieste.space.SearchSpace` depends on the specific
+                :class:`AcquisitionRule`.
```
joelberkeley: why the indentations?

trieste/acquisition/optimizer.py
```diff
@@ -65,16 +75,25 @@ def _discrete_space(space: DiscreteSearchSpace, target_func: AcquisitionFunction
...
+    space: Box, target_func: AcquisitionFunction, num_query_points: int = 1
+) -> TensorType:
+    expanded_space = space ** num_query_points
+    trial_search_space = expanded_space.discretize(
+        tf.minimum(2000, 500 * tf.shape(expanded_space.lower)[-1])
+    )
 
-    bijector = tfp.bijectors.Sigmoid(low=space.lower, high=space.upper)
+    def vectorized_target_func(at):
+        return target_func(tf.reshape(at, at.shape[:-1].as_list() + [num_query_points, -1]))
```
joelberkeley: if this stuff is a separate function, it can be used with both `optimize_continuous` and `optimize_discrete`

joelberkeley: it also means we don't need to tinker with `optimize_continuous` and `optimize_discrete` in this PR

joelberkeley: and lets users define optimizers without needing to worry about batch dims. They can just wrap their optimizer in a convenient batchifier

joelberkeley: i'm not clear on this one: but it might also make the greedy-batch optimizer much easier to add

trieste/acquisition/rule.py
henrymoss: black did it

joelberkeley: that's odd. black shouldn't touch docstrings

joelberkeley: ok that's not really true having checked, but it's never modified these lines before and the version of black is pinned. It also look wrong.

joelberkeley: how are you running black?

trieste/acquisition/optimizer.py
henrymoss: Please elaborate. Do you have a clear idea how to do this?

trieste/acquisition/rule.py
henrymoss: just from command line. as "black ."

vpicheny: also makes this PR appear very large unfortunately!

trieste/acquisition/optimizer.py
```diff
@@ -65,16 +75,25 @@ def _discrete_space(space: DiscreteSearchSpace, target_func: AcquisitionFunction
...
     variable = tf.Variable(bijector.inverse(initial_point))
 
     def _objective() -> TensorType:
-        return -target_func(bijector.forward(variable))
+        return -vectorized_target_func(bijector.forward(variable))
 
     gpflow.optimizers.Scipy().minimize(_objective, (variable,))
-
-    return bijector.forward(variable)
+    points = bijector.forward(variable)
```
vpicheny: maybe not part of this PR, but since we're using Scipy it would be better to use L-BFGS-B and avoid using a bijector.

trieste/acquisition/optimizer.py
```diff
@@ -35,23 +35,33 @@
...
 
 
 @optimize.register
-def _discrete_space(space: DiscreteSearchSpace, target_func: AcquisitionFunction) -> TensorType:
+def optimize_discrete(
+    space: DiscreteSearchSpace, target_func: AcquisitionFunction, num_query_points: int = 1
+) -> TensorType:
+
+    if num_query_points > 1:
+        raise ValueError("Batch optimisation not yet supported for discrete search spaces")
```
vpicheny: actually, it should! if I'm not mistaken we can apply the same operation to discrete spaces: `space ** num_query_points`

trieste/acquisition/optimizer.py
```diff
@@ -65,16 +75,25 @@ def _discrete_space(space: DiscreteSearchSpace, target_func: AcquisitionFunction
...
+def optimize_continuous(
+    space: Box, target_func: AcquisitionFunction, num_query_points: int = 1
+) -> TensorType:
+    expanded_space = space ** num_query_points
+    trial_search_space = expanded_space.discretize(
+        tf.minimum(2000, 500 * tf.shape(expanded_space.lower)[-1])
+    )
 
-    bijector = tfp.bijectors.Sigmoid(low=space.lower, high=space.upper)
+    def vectorized_target_func(at):
```
vpicheny: vectorized is an easily overloaded term, replace the name with something more obvious? eg `target_func_with_vectorized_inputs`

tests/unit/acquisition/test_function.py
```diff
@@ -131,8 +98,8 @@ def test_expected_improvement_builder_builds_expected_improvement_using_best_fro
     )
     model = QuadraticMeanAndRBFKernel()
     acq_fn = ExpectedImprovement().prepare_acquisition_function(dataset, model)
-    xs = tf.linspace([-10.0], [10.0], 100)
-    expected = expected_improvement(model, tf.constant([0.0]), xs)
+    xs = tf.linspace([[-10.0]], [[10.0]], 100)
+    expected = expected_improvement(model, tf.constant([0.0]))(xs)
```
vpicheny, line 102: shouldn't xs be a rank 3 tensor now?

tests/unit/acquisition/test_function.py
```diff
@@ -198,14 +173,22 @@ def test_min_value_entropy_search_raises_for_gumbel_samples_with_invalid_shape(
...
+        mes(at)
 
 
 def test_min_value_entropy_search_returns_correct_shape() -> None:
     model = QuadraticMeanAndRBFKernel()
     gumbel_samples = tf.constant([1.0])
-    query_at = tf.linspace([-10.0], [10.0], 5)
-    evals = min_value_entropy_search(model, gumbel_samples, query_at)
+    query_at = tf.linspace([[-10.0]], [[10.0]], 5)
+    evals = min_value_entropy_search(model, gumbel_samples)(query_at)
```
vpicheny, line 191: same here, query_at should have rank 3?

vpicheny:
There are lots of places in `test_function` where at or query_at does not have a batch dimension. I have tagged a couple of those but not all.

tests/unit/acquisition/test_rule.py
```diff
@@ -49,9 +46,14 @@
 from trieste.type import TensorType
 
 
-def _line_search_maximize(search_space: Box, f: AcquisitionFunction) -> TensorType:
+def _line_search_maximize(
+    search_space: Box, f: AcquisitionFunction, num_query_points: int = 1
+) -> TensorType:
+    if num_query_points != 1:
+        raise ValueError("_line_search_maximizer only defined for batches of size 1")
     xs = tf.linspace(search_space.lower, search_space.upper, 10 ** 6)
```
vpicheny, line 50: only works for 1D boxes I guess? should we assert dim?

tests/unit/acquisition/test_rule.py
```diff
@@ -60,12 +62,14 @@ class NegQuadratic(AcquisitionFunctionBuilder):
...
             self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
         ) -> AcquisitionFunction:
-            return lambda x: -quadratic(x - 1)
+            return lambda x: -quadratic(tf.squeeze(x, -2) - 1)
 
     search_space = Box([-10], [10])
     ego = EfficientGlobalOptimization(NegQuadratic(), optimizer)
     data, model = empty_dataset([1], [1]), QuadraticMeanAndRBFKernel(x_shift=1)
     query_point, _ = ego.acquire(search_space, {"": data}, {"": model})
+    print("HERREEEEE")
```
vpicheny: should we keep this? ;-)

trieste/acquisition/function.py
```diff
@@ -32,26 +31,10 @@
...
-BatchAcquisitionFunction = Callable[[TensorType], TensorType]
-"""
-Type alias for batch acquisition functions.
-
-A `BatchAcquisitionFunction` maps a set of `B` query points (each of dimension `D`) to a single
+An :const:`AcquisitionFunction` maps a set of `B` query points (each of dimension `D`) to a single
 value that describes how useful it would be evaluate all these points together (to our goal of
-optimizing the objective function). Thus, with leading dimensions, a `BatchAcquisitionFunction`
+optimizing the objective function). Thus, with leading dimensions, an :const:`AcquisitionFunction`
 takes input shape `[..., B, D]` and returns shape `[..., 1]`.
```
vpicheny, line 37: I'd like to add something upfront to say that non-batch acq fct should be defined with batch 1

trieste/acquisition/function.py
```diff
@@ -121,41 +104,45 @@ def prepare_acquisition_function(
...
+        :exc:`ValueError` if used with a batch size greater than one.
     """
-    mean, variance = model.predict(at)
-    normal = tfp.distributions.Normal(mean, tf.sqrt(variance))
-    return (eta - mean) * normal.cdf(eta) + variance * normal.prob(eta)
+
+    def acquisition(x: TensorType) -> TensorType:
+        tf.debugging.assert_shapes(
+            [(x, [..., 1, None])],
+            message="Trying to use a non-batch acquisition function for batch BO",
```
vpicheny: maybe make this message less confusing by being explicit about the -2 dimension that must be 1?

trieste/acquisition/function.py
```diff
@@ -544,161 +540,21 @@ def sample(self, at: TensorType) -> TensorType:
...
 
         if tf.size(self._eps) == 0:
             self._eps.assign(
                 tf.random.normal([self._sample_size, mean.shape[-1]], dtype=tf.float64)
             )  # [S, L]
 
-        return mean + tf.sqrt(var) * tf.cast(self._eps, var.dtype)  # [..., S, L]
-
-
-class MCIndAcquisitionFunctionBuilder(AcquisitionFunctionBuilder):
```
vpicheny: why is this not needed anymore?

trieste/acquisition/rule.py
joelberkeley: do you mind reverting them?

trieste/acquisition/rule.py
```diff
@@ -94,27 +88,39 @@ def __init__(
...
+                ``builder``. This should *maximize* the acquisition function, and must be compatible
+                with the global search space. Defaults to
+                :func:`~trieste.acquisition.optimizer.optimize`.
+        :param num_query_points: The number of points to acquire.
         """
+
+        if not num_query_points > 0:
+            raise ValueError(
+                f"Number of query points must be greater than 0, got {num_query_points}"
+            )
```
joelberkeley, line 113: i'm starting to prefer `tf.debugging` where possible cos it becomes part of the graph when compiled. It also comes with free error messages. What say you?

trieste/acquisition/rule.py
```diff
@@ -94,27 +88,39 @@ def __init__(
...
+
         if builder is None:
-            builder = ExpectedImprovement().using(OBJECTIVE)
+            if num_query_points == 1:
+                builder = ExpectedImprovement().using(OBJECTIVE)
+            else:
+                raise ValueError("Need to specify a batch acquisition function")
```
joelberkeley: I'm for dropping the default. Simpler. What do you think?

trieste/acquisition/rule.py
```diff
@@ -94,27 +88,39 @@ def __init__(
...
             optimizer = optimize
 
         self._builder = builder
         self._optimizer = optimizer
+        self._num_query_points = num_query_points
 
     def __repr__(self) -> str:
         """"""
-        return f"EfficientGlobalOptimization({self._builder!r})"
+        return f"EfficientGlobalOptimization({self._num_query_points!r}, {self._builder!r})"
```
joelberkeley: I realise optimizer's already missing from this string, but the order needs to match the `__init__` args (unless using named parameters)

trieste/acquisition/optimizer.py
joelberkeley: yeah defo not part of this PR, but also important

trieste/acquisition/function.py
joelberkeley: sounds sensible

trieste/acquisition/rule.py
henrymoss: Yeah, I will swap those out for these code touched by this PR. Maybe a separate PR to fix all other occurrences?

trieste/acquisition/function.py
henrymoss: @joelberkeley ?

trieste/acquisition/function.py
joelberkeley: can we also avoid the phrase non-batch? I think it will confuse people. Everything's batch now, just some things only support batch size 1

btw this will read
```
ValueError: ... Trying to use a non-batch acquisition function for batch BO
```
which sounds like the code might be using a non-batch acquisition as a fallback due to the error.

we could be explicit with `" This acquisition function only supports batch size of one "`

trieste/acquisition/function.py
```diff
@@ -121,41 +104,45 @@ def prepare_acquisition_function(
...
     .. math:: x \mapsto \mathbb E \left[ \max (\eta - f(x), 0) \right]
 
     This function was introduced by Mockus et al, 1975. See :cite:`Jones:1998` for details.
 
     :param model: The model of the objective function.
     :param eta: The "best" observation.
-    :param at: The points for which to calculate the expected improvement.
-    :return: The expected improvement at ``at``.
+    :return: The expected improvement acquisition function. This function will raise
+        :exc:`ValueError` if used with a batch size greater than one.
```
```suggestion
        :exc:`ValueError`  or :exc:`~tf.errors.InvalidArgumentError` if used with a batch size greater than one.
```
`assert_shapes` can raise either afaik (depending on whether the function's compiled)

trieste/acquisition/rule.py
henrymoss: Yep will do, although it means I will have to fiddle the notebooks

trieste/acquisition/function.py
joelberkeley: similar in other places

tests/unit/acquisition/test_function.py
joelberkeley: tada
``` python
>>> tf.linspace([[-10.0]], [[10.0]], 100).shape
TensorShape([100, 1, 1])
```

joelberkeley: but to be pedantic, the first dimension is leading. A point of shape [1, 1] is perfectly valid for acquisition functions

trieste/acquisition/rule.py
joelberkeley: yeah for sure - they're everywhere

trieste/acquisition/function.py
joelberkeley: this is scope creep. I can move it to a separate PR. Basically it hardly does anything and we were saying how it would be nice to have fewer classes

vpicheny: yes, let's do that in a separate PR please!

tests/unit/acquisition/test_function.py
```diff
@@ -324,7 +328,7 @@ def prepare_acquisition_function(
 
     ei = ExpectedImprovement().using("foo").prepare_acquisition_function(data, models_)
 
-    at = tf.constant([[-0.1], [1.23], [-6.78]])
+    at = tf.constant([[[-0.1]], [[1.23]], [[-6.78]]])
```
vpicheny, line 341: make it rank 3?

trieste/acquisition/function.py
vpicheny: I like this proposition

trieste/acquisition/rule.py
vpicheny: well it's not very important but I find it convenient. Should we make that part of another PR, where we also try to get rid of .using(OBJECTIVE) when there is a single model?

docs/notebooks/failure_ego.pct.py
joelberkeley: i've unresolved this so I can keep track of it

tests/unit/acquisition/test_function.py
joelberkeley: i don't understand what you mean

trieste/acquisition/rule.py
henrymoss: Yep. Lets leave for now

trieste/acquisition/rule.py
henrymoss: Yep, doing now

trieste/acquisition/function.py
henrymoss: Adding them :)

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -181,8 +179,8 @@ def prepare_acquisition_function(self, datasets, models):
         }
 
         pf = trieste.acquisition.probability_of_feasibility(
-            models[CONSTRAINT], self._threshold, objective_dataset.query_points
-        )
+            models[CONSTRAINT], self._threshold
+        )(tf.expand_dims(objective_dataset.query_points,1))
```
joelberkeley:
```suggestion
        )(tf.expand_dims(objective_dataset.query_points, 1))
```

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -209,8 +207,8 @@ def batch_efi(at):
 
 num_query_points = 4
 batch_eci = BatchExpectedConstrainedImprovement(50, Sim.threshold)
-batch_rule: BatchAcquisitionRule[Box] = BatchAcquisitionRule(
-    num_query_points, batch_eci
+batch_rule: EfficientGlobalOptimization[Box] = EfficientGlobalOptimization(
+    batch_eci, num_query_points= num_query_points
```
joelberkeley:
```suggestion
    batch_eci, num_query_points=num_query_points
```

trieste/acquisition/function.py
joelberkeley: https://github.com/secondmind-labs/trieste/pull/210

tests/unit/acquisition/test_function.py
henrymoss: So that it has a batch dimension?

henrymoss:
> There are lots of places in `test_function` where at or query_at does not have a batch dimension. I have tagged a couple of those but not all.

Should be sorted now

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,91 @@
...
+    if batch_size == 1:
+        return batch_size_one_optimizer(space, target_func)
+    else:
+        return optimize_batch(batch_size_one_optimizer, space, target_func, batch_size)
+
+
+def optimize_batch(
+    batch_size_one_optimizer: AcquisitionOptimizer,
+    space: SearchSpace,
+    target_func: AcquisitionFunction,
```
joelberkeley: these arguments are no longer needed. Indeed it's all more composable if we leave them out. As it currently is, users need to specify the search space and acquisition function at the same time as the batch size. This isn't necessary and could create coupling in their code.

henrymoss: Yep this makes sense!

trieste/acquisition/rule.py
joelberkeley: it seems odd to me to say "the builder's optional" then error when you don't specify it

joelberkeley: but then perhaps it's no more odd than accepting a batch size only to error when the builder doesn't work with it

tests/unit/acquisition/test_function.py
vpicheny: yes

trieste/acquisition/function.py
```diff
@@ -339,17 +331,25 @@ def lower_confidence_bound(model: ProbabilisticModel, beta: float, at: TensorTyp
     :param model: The model of the objective function.
     :param beta: The weight to give to the standard deviation contribution of the LCB. Must not be
         negative.
-    :param at: The points at which to evaluate the LCB.
-    :return: The lower confidence bound at ``at``.
+    :return: The expected improvement function. This function will raise
```
vpicheny: change back to LCB

trieste/acquisition/function.py
```diff
@@ -304,14 +297,13 @@ def prepare_acquisition_function(
...
+            greater than one.
+        :raise ValueError: If ``beta`` is negative.
         """
-        return lambda at: self._acquisition_function(model, self._beta, at)
-
-    @staticmethod
-    def _acquisition_function(model: ProbabilisticModel, beta: float, at: TensorType) -> TensorType:
-        return -lower_confidence_bound(model, beta, at)
+        lcb = lower_confidence_bound(model, self._beta)
+        return lambda at: -lcb(at)
```
vpicheny, line 306: is this line necessary?

trieste/acquisition/function.py
```diff
@@ -392,20 +392,16 @@ def prepare_acquisition_function(
         """
         :param dataset: Unused.
         :param model: The model over the specified ``dataset``.
-        :return: The probability of feasibility acquisition function.
+        :return: The expected improvement function. This function will raise
```
vpicheny: change back to PF

trieste/acquisition/function.py
```diff
@@ -419,17 +415,23 @@ def probability_of_feasibility(
 
     :param model: The model of the objective function.
     :param threshold: The (scalar) probability of feasibility threshold.
-    :param at: The points at which to evaluate the probability of feasibility. Must have rank at
-        least two
-    :return: The probability of feasibility at ``at``.
-    :raise ValueError (or InvalidArgumentError): If arguments have the incorrect shape.
+    :return: The expected improvement function. This function will raise
```
vpicheny: change back to PF

trieste/acquisition/function.py
```diff
@@ -493,7 +497,7 @@ def prepare_acquisition_function(
             )
 
         constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)
-        pof = constraint_fn(objective_dataset.query_points)
+        pof = constraint_fn(objective_dataset.query_points[:, None, ...])
```
vpicheny, line 500: expand dims would look cleaner?

trieste/acquisition/function.py
```diff
@@ -544,161 +548,21 @@ def sample(self, at: TensorType) -> TensorType:
...
-            number of latent model dimensions.
-        :raise ValueError (or InvalidArgumentError): If ``at`` is a scalar.
+        :return: The samples, of shape `[..., S, 1, L]`, where `S` is the `sample_size` and `L` is
+            the number of latent model dimensions.
+        :raise ValueError (or InvalidArgumentError): If ``at`` has an invalid shape.
         """
-        tf.debugging.assert_rank_at_least(at, 1)
-        mean, var = self._model.predict(at[..., None, :])  # [..., 1, L], [..., 1, L]
+        tf.debugging.assert_shapes([(at, [..., 1, None])])
+        mean, var = self._model.predict(at[..., None, :, :])  # [..., 1, 1, L], [..., 1, 1, L]
```
vpicheny, line 558: shapes seem weird here

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,91 @@
 AcquisitionOptimizer = Callable[[SP, AcquisitionFunction], TensorType]
 """
 Type alias for a function that returns the single point that maximizes an acquisition function over
-a search space. For a search space with points of shape S, and acquisition function with input shape
-[...] + S output shape [..., 1], the :data:`AcquisitionOptimizer` return shape should be [1] + S.
+a search space. For a search space with points of shape [D], and acquisition function with input
```
vpicheny, line 37: either [..., D] and [...., B, D] or [D] and [B, D]?

vpicheny: actually maybe it's [..., 1, D] now?

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,91 @@
...
+    space: SearchSpace, target_func: AcquisitionFunction, batch_size: int
+) -> AcquisitionOptimizer[SP]:
     """
+    A wrapper around our :const:`AcquisitionOptimizer`s. This class chooses
+    an :const:`AcquisitionOptimizer` appropriate for the
+    problem's :class:`~trieste.space.SearchSpace` and desired batch size.
+    
     :param space: The space of points over which to search, for points with shape [D].
-    :param target_func: The function to maximise, with input shape [..., D] and output shape
+    :param target_func: The function to maximise, with input shape [..., 1, D] and output shape
```
vpicheny, line 51: isn't it [..., B, D] here?

trieste/acquisition/function.py
joelberkeley: haha. artem actually moved me to index notation, which I think is clearer, cos you can see the final shape. Not too bothered though

tests/unit/acquisition/test_function.py
joelberkeley: constrained ei doesn't support batch size > 1, and this is already has a batch dimension of size 1 afaict

trieste/acquisition/function.py
joelberkeley: do you mean splitting it in two? if so yes: it avoids evaluating everything in `lower_confidence_bound` which isn't part of the acquisition function, such as validating beta

trieste/acquisition/function.py
joelberkeley: it's to allow broadcasting with eps. Is there a cleaner way?

joelberkeley: we end up with 3 dims anyway

joelberkeley: or do you think i've got the shapes wrong?

trieste/acquisition/optimizer.py
joelberkeley: it's always [..., B, D] but in some cases B = 1

trieste/acquisition/optimizer.py
henrymoss: yep!

trieste/acquisition/function.py
vpicheny: It might be that with the batch dimension we do not need to introduce an extra dimension

vpicheny: forget it, I think I was wrong

trieste/acquisition/optimizer.py
vpicheny: unresolving, change docstring?

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,83 @@
...
+        raise NotImplementedError(
+            """ No optimizer currentely supports acquisition function
+            maximisation over this search space"""
+        )
+
+    return optimizer(space, target_func)
+
+
+def batchify(
+    batch_size_one_optimizer: AcquisitionOptimizer,
```
joelberkeley:
```suggestion
    batch_size_one_optimizer: AcquisitionOptimizer[SP],
```

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,83 @@
...
+a search space. For a search space with points of shape [D], and acquisition function with input
+shape [..., B, D] output shape [..., 1], the :const:`AcquisitionOptimizer` return shape should be
+[B, D].
 """
 
 
-@singledispatch
-def optimize(space: Box | DiscreteSearchSpace, target_func: AcquisitionFunction) -> TensorType:
+def automatic_optimizer_selector(
+    space: DiscreteSearchSpace, target_func: AcquisitionFunction
```
joelberkeley: if you make the type of `space` `Box | DiscreteSearchSpace` you won't need the exception below

trieste/acquisition/optimizer.py
henrymoss: I think this is [..., 1, D ] here now. B only comes into play one pushed though batchify

trieste/acquisition/optimizer.py
henrymoss: What does this do?

trieste/acquisition/rule.py
```diff
@@ -94,27 +88,42 @@ def __init__(
         self,
         builder: AcquisitionFunctionBuilder | None = None,
         optimizer: AcquisitionOptimizer[SP_contra] | None = None,
+        num_query_points: int = 1,
```
joelberkeley, line 98: hmmm ... i've just noticed that if the optimizers accept the batch size, users can need to specify it twice:
```python
EfficientGlobalOptimization(EI().using(""), batchify(optimize_discrete, 3), 3)
```

joelberkeley: then there's the question: how can users only specify the batch size and not all the other stuff? One option is union types, but a more principled option is a class method

trieste/acquisition/rule.py
```diff
@@ -62,12 +61,12 @@ def acquire(
...
-            :class:`AcquisitionRule`.
-          - Each :class:`AcquisitionRule` must define the type of its corresponding acquisition
-            state (if the rule is stateless, this type can be `None`). The ``state`` passed
-            to this method, and the state returned, must both be of that type.
+        - The global search space must be a :class:`~trieste.space.SearchSpace`. The exact type
+          of :class:`~trieste.space.SearchSpace` depends on the specific
+          :class:`AcquisitionRule`.
+        - Each :class:`AcquisitionRule` must define the type of its corresponding acquisition
+          state (if the rule is stateless, this type can be `None`). The ``state`` passed
+          to this method, and the state returned, must both be of that type.
```
joelberkeley: now you're going the other way

trieste/acquisition/optimizer.py
henrymoss: THe exception is more informative?

trieste/acquisition/rule.py
henrymoss: eh?

trieste/acquisition/rule.py
henrymoss: would you rather not have batch size as an input to the rule? That would work

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,81 @@
...
+    A wrapper around our :const:`AcquisitionOptimizer`s. This class wraps a
+    :const:`AcquisitionOptimizer` to allow it to optimize batch acquisition functions.
+
+    :param batch_size_one_optimizer: An optimizer that returns only batch size one, i.e. produces a
+        single point with shape [1, D].
+    :param batch_size: The number of points in the batch.
+    :return: An :const:`AcquisitionOptimizer` that will provide a batch of points with shape [B, D].
+    """
+    tf.debugging.assert_positive(batch_size)
+
```
joelberkeley, line 57: did you find
```python
if batch_size == 1:
    return batch_size_one_optimizer
```
wasn't working?

trieste/acquisition/rule.py
```diff
@@ -234,7 +251,7 @@ def __init__(
         """
         :param builder: The acquisition function builder to use. :class:`TrustRegion` will attempt
             to **maximise** the corresponding acquisition function. Defaults to
-            :class:`~trieste.acquisition.ExpectedImprovement` with tag `OBJECTIVE`.
+            class:`~trieste.acquisition.ExpectedImprovement` with tag `OBJECTIVE`.
```
joelberkeley:
```suggestion
            :class:`~trieste.acquisition.ExpectedImprovement` with tag `OBJECTIVE`.
```

trieste/acquisition/optimizer.py
henrymoss: It wasn't needed? I guess maybe it speed up things a little?

trieste/acquisition/rule.py
```diff
@@ -245,7 +262,7 @@ def __init__(
             builder = ExpectedImprovement().using(OBJECTIVE)
 
         if optimizer is None:
-            optimizer = optimize
+            optimizer = optimize_continuous
```
joelberkeley, line 268: does trust region not work with batches? Though I guess this is something for another PR

docs/notebooks/expected_improvement.pct.py
```diff
@@ -217,10 +217,10 @@ def build_model(data):
 # Here we use the `BatchMonteCarloExpectedImprovement` function. Note that this acquisition function is computed using a Monte-Carlo method (so it requires a `sample_size`), but with a reparametrisation trick, which makes it deterministic.
 
 # %%
-from trieste.acquisition.rule import BatchAcquisitionRule
+from trieste.acquisition.rule import EfficientGlobalOptimization
 
 qei = trieste.acquisition.BatchMonteCarloExpectedImprovement(sample_size=1000)
-batch_rule: BatchAcquisitionRule[Box] = BatchAcquisitionRule(
+batch_rule: EfficientGlobalOptimization[Box] = EfficientGlobalOptimization(
```
joelberkeley, line 223: **OPTIONAL** depending on the final API for `EfficientGlobalOptimization`, we may find we don't need this type hint cos the optimizer can specify the search space

trieste/acquisition/rule.py
joelberkeley: you seem to like (de-)indenting things

tests/unit/acquisition/test_optimizer.py
```diff
@@ -38,10 +46,77 @@
...
+)
+def test_optimize_batch(
+    search_space: Box, acquisition: AcquisitionFunction, maximizers: TensorType, batch_size: int
+) -> None:
+    batch_size_one_optimizer = optimize_continuous
+    batch_optimizer = batchify(batch_size_one_optimizer, batch_size)
+    points = batch_optimizer(search_space, acquisition)
+    assert points.shape == [batch_size] + search_space.lower.shape
+    for point in points:
+        tf.reduce_any(point == maximizers, axis=0)
```
joelberkeley: i missed an `assert` here. this might even need an extra `reduce_all`

trieste/acquisition/optimizer.py
joelberkeley: btw if this is just `DiscreteSearchSpace` it won't type check when called with other search spaces such as `Box`

it'll either need to be `SearchSpace` or the union above

trieste/acquisition/optimizer.py
joelberkeley: sure - don't mind

trieste/acquisition/optimizer.py
joelberkeley: say I have `optimize_continuous` which is an `AcquisitionOptimizer[Box]`, i.e. it's an optimizer over `Box` search spaces. If I call `new_optimizer = batchify(optimize_continuous, 3)`, having the same type variable `SP` on both the input and output means that I can be sure that `new_optimizer` is also an optimizer over `Box`.

trieste/acquisition/rule.py
joelberkeley: it's good to keep PRs really focussed. This helps reviewers see what's changed and why, and also helps when rebasing on new changes, where conflicts are either avoided or are easier to fix

trieste/acquisition/rule.py
joelberkeley: it's an option. I'm not sure what the best approach for the notebooks would be. For now I'd probably specify the optimizer just cos it would be quicker and we're not rolling in time

henrymoss: I think leave it as is. Its worth being able to choose batch size when defining the EGO rule. I think batch size is a property of the BO routine, not the optimizer. 

What do you think @vpicheny ?

trieste/acquisition/rule.py
henrymoss: I dont know why this is happening!

trieste/acquisition/optimizer.py
vpicheny: I see, where D can actually be B*D
Maybe it is slightly confusing that we use the same D but I don't think there's a better solution.

trieste/acquisition/rule.py
vpicheny: the batch size should definitely be defined by the rule. It is not great to have to specify it for the optimizer as well.

henrymoss: You only need to specify it for the optimizer if you choose to tell it a specific optimizer (which the vast majority of users will not need to do).

vpicheny: it's probably fine then? we probably need a good docstring to make this clear though

henrymoss: Perhaps we should always call batchify, so that a user can specify a particular type of optimizer (i.e. random search or local or genetic algorithms), but all the batching is dealt with under the hood. 

This is the way that all other BO packages work. It feels awkward to have to specify a batch optimizer. The batchness (and batch size) is just a property of the acquisition function.

henrymoss: I vote for always calling batchify (see new PR).  This greatly simplifies the specification of custom optimizers. 

This will also allow for an easy implementation of greedy batches!

tests/unit/acquisition/test_optimizer.py
henrymoss: THis will not work, as we need to test "point is approx equal to maximizers". This cannot be done while working with a set of multiple possible maximisers. 

I propose remove the branin test case here and just using the npt.assert_allclose.

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,81 @@
...
+    return optimizer
 
 
-@optimize.register
-def _discrete_space(space: DiscreteSearchSpace, target_func: AcquisitionFunction) -> TensorType:
-    target_func_values = target_func(space.points)
+def optimize_discrete(space: DiscreteSearchSpace, target_func: AcquisitionFunction) -> TensorType:
+    """
+    An :const:`AcquisitionOptimizer` for :const:'DiscreteSearchSpace' spaces and
+    batches of size of 1.
```
joelberkeley:
```suggestion
    An :const:`AcquisitionOptimizer` for :class:'DiscreteSearchSpace' spaces and
    batches of size of 1.
```
`:const:` is for constants, `:class:` is for classes

trieste/acquisition/optimizer.py
```diff
@@ -64,17 +120,25 @@ def _discrete_space(space: DiscreteSearchSpace, target_func: AcquisitionFunction
...
+
+    initial_point = optimize_discrete(trial_search_space, target_func)  # [1, D]
 
     bijector = tfp.bijectors.Sigmoid(low=space.lower, high=space.upper)
-    variable = tf.Variable(bijector.inverse(initial_point))
+    variable = tf.Variable(bijector.inverse(initial_point))  # [1, D]
 
     def _objective() -> TensorType:
-        return -target_func(bijector.forward(variable))
+        return -target_func(bijector.forward(variable[:, None, :]))  # [1]
```
joelberkeley, line 108: are you sure about this? acq fns are [..., B, D] -> [..., 1] i.e. [1, D] -> [1]

trieste/acquisition/rule.py
joelberkeley: i don't really mind, though I don't follow how this will make it easier to use greedy batches. I would have thought it will make it harder cos you can't customize which function does the batching

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,81 @@
...
+    :return: The batch of points in ``space`` that maximises ``target_func``, with shape [1, D].
+    """
+
+    if isinstance(space, Box):
+        optimizer = optimize_continuous
+    elif isinstance(space, DiscreteSearchSpace):
+        optimizer = optimize_discrete
+    else:
+        raise NotImplementedError(
+            """ No optimizer currentely supports acquisition function
```
joelberkeley, line 63:
```suggestion
            """No optimizer currently supports acquisition function
```

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,81 @@
...
+            maximisation over this search space"""
```
joelberkeley: **OPTIONAL** this error message is fragile - if someone adds an optimizer that does support that search space, it's no longer true

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,81 @@
...
-[...] + S output shape [..., 1], the :data:`AcquisitionOptimizer` return shape should be [1] + S.
+a search space. For a search space with points of shape [D], and acquisition function with input
+shape [..., B, D] output shape [..., 1], the :const:`AcquisitionOptimizer` return shape should be
+[B, D].
 """
 
 
-@singledispatch
-def optimize(space: Box | DiscreteSearchSpace, target_func: AcquisitionFunction) -> TensorType:
+def automatic_optimizer_selector(
```
joelberkeley, line 43: **OPTIONAL** btw this is doing the same as singledispatch

tests/unit/acquisition/test_optimizer.py
joelberkeley: i don't follow. This checks that all `point`s in `points` are maximizers

trieste/acquisition/optimizer.py
```diff
@@ -64,17 +120,25 @@ def _discrete_space(space: DiscreteSearchSpace, target_func: AcquisitionFunction
     return space.points[max_value_idx : max_value_idx + 1]
 
 
-@optimize.register
-def _box(space: Box, target_func: AcquisitionFunction) -> TensorType:
+def optimize_continuous(space: Box, target_func: AcquisitionFunction) -> TensorType:
+    """
+    An :const:`AcquisitionOptimizer` for :const:'Box' spaces and batches of size of 1.
```
joelberkeley:
```suggestion
    An :const:`AcquisitionOptimizer` for :class:'Box' spaces and batches of size of 1.
```

docs/notebooks/expected_improvement.pct.py
joelberkeley: OPTIONAL

tests/unit/acquisition/test_optimizer.py
joelberkeley: so i'm confident this is possible, and the branin test tests more than we get from quadratic since branin is more interesting with multiple global minima

i'll approve and leave it with you to do if you reckon it's worth it

tests/unit/acquisition/test_optimizer.py
```diff
@@ -38,10 +46,77 @@
...
+        (Box([0, 0, 0], [1, 1, 1]), _quadratic_sum([0.5, -0.5, 0.2]), ([[0.5, -0.5, 0.2]])),
+    ],
+)
+def test_automatic_optimizer_selector(
+    search_space: Box, acquisition: AcquisitionFunction, maximizers: TensorType, batch_size: int
+) -> None:
+    optimizer = automatic_optimizer_selector
+    points = optimizer(search_space, acquisition)
+    for point in points:
+        tf.reduce_any(point == maximizers, axis=0)
```
joelberkeley: similar to above

joelberkeley (approved):

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,81 @@
...
+
+    return optimizer(space, target_func)
+
+
+def batchify(
+    batch_size_one_optimizer: AcquisitionOptimizer[SP],
+    batch_size: int,
+) -> AcquisitionOptimizer[SP]:
     """
+    A wrapper around our :const:`AcquisitionOptimizer`s. This class wraps a
```
joelberkeley: "This class wraps a ... "

this isn't a class

joelberkeley:
```suggestion
    r"""
    A wrapper around our :const:`AcquisitionOptimizer`\ s. This class wraps a
```
if you pluralise a type, you need a backslash and space for it to render right, and if you have a backslash, you need `r"""`

trieste/acquisition/optimizer.py
```diff
@@ -35,24 +34,81 @@
...
 """
 
 
-@singledispatch
-def optimize(space: Box | DiscreteSearchSpace, target_func: AcquisitionFunction) -> TensorType:
+def automatic_optimizer_selector(
+    space: SearchSpace, target_func: AcquisitionFunction
+) -> TensorType:
     """
+    A wrapper around our :const:`AcquisitionOptimizer`s. This class performs
```
joelberkeley, line 47: also not a class

trieste/acquisition/optimizer.py
```diff
@@ -35,46 +35,113 @@
...
 
-    bijector = tfp.bijectors.Sigmoid(low=space.lower, high=space.upper)
-    variable = tf.Variable(bijector.inverse(initial_point))
+	bijector = tfp.bijectors.Sigmoid(low=space.lower, high=space.upper)
+	variable = tf.Variable(bijector.inverse(initial_point))  # [1, D]
 
-    def _objective() -> TensorType:
-        return -target_func(bijector.forward(variable))
+	def _objective() -> TensorType:
+		return -target_func(bijector.forward(variable[:, None, :]))  # [1]
```
joelberkeley: is your IDE set to use indents?

joelberkeley: or 8 spaces instead of 4?

henrymoss: Ive just noticed this and sorted

PR #214, opened 2021-03-30
added greedy batch optimizer

henrymoss:
I have added a greedy batch optimizer in optimizer.py .

This is a simple first step in implementing more batch acquisition functions (i.e Local Penalisation and GIBBON).

This new optimizer can be used to optimize our current batch EI acquisition function in a greedy sequential manner, i.e. we build up the batch element by element.

### DIFF
--- a/tests/unit/acquisition/test_optimizer.py
+++ b/tests/unit/acquisition/test_optimizer.py
@@ -22,6 +22,7 @@ from trieste.acquisition import AcquisitionFunction
 from trieste.acquisition.optimizer import (
     automatic_optimizer_selector,
     batchify,
+    batchify_greedy,
     optimize_continuous,
     optimize_discrete,
 )
@@ -74,6 +75,7 @@ def test_optimize_continuous(
 
 
 @random_seed
+@pytest.mark.parametrize("batchify_type", [batchify, batchify_greedy])
 @pytest.mark.parametrize("batch_size", [1, 2, 3, 5])
 @pytest.mark.parametrize(
     "search_space, acquisition, maximizer",
@@ -82,11 +84,15 @@ def test_optimize_continuous(
         (Box([-1, -1, -1], [1, 1, 1]), _quadratic_sum([0.5, -0.5, 0.2]), ([[0.5, -0.5, 0.2]])),
     ],
 )
-def test_optimize_batch(
-    search_space: Box, acquisition: AcquisitionFunction, maximizer: TensorType, batch_size: int
+def test_batchify(
+    search_space: Box,
+    acquisition: AcquisitionFunction,
+    maximizer: TensorType,
+    batch_size: int,
+    batchify_type,
 ) -> None:
     batch_size_one_optimizer = optimize_continuous
-    batch_optimizer = batchify(batch_size_one_optimizer, batch_size)
+    batch_optimizer = batchify_type(batch_size_one_optimizer, batch_size)
     points = batch_optimizer(search_space, acquisition)
     assert points.shape == [batch_size] + search_space.lower.shape
     for point in points:
--- a/trieste/acquisition/optimizer.py
+++ b/trieste/acquisition/optimizer.py
@@ -126,6 +126,9 @@ def batchify(
     """
     tf.debugging.assert_positive(batch_size)
 
+    if batch_size == 1:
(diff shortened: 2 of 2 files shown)


trieste/acquisition/optimizer.py
```diff
@@ -140,3 +143,44 @@ def target_func_with_vectorized_inputs(
...
+
+                current_batch_tiled = tf.repeat(current_batch[None, :, :], x.shape[-3], axis=-3)
+                x_with_current_batch = tf.concat([current_batch_tiled, x], -2)
+                return f(x_with_current_batch)
+
+            next_batch_point = batch_size_one_optimizer(
+                search_space, target_func_with_fixed_partial_batch
+            )  # [1, D]
+
+            current_batch = tf.concat([current_batch, next_batch_point], axis=0)  # [i+1, D]
```
joelberkeley, line 182: you might want to be careful here with mutable variables: will this work if someone wraps it with `@tf.function`?

trieste/acquisition/optimizer.py
```diff
@@ -140,3 +143,44 @@ def target_func_with_vectorized_inputs(
...
+    :param batch_size: The number of points in the batch.
+    :return: An :const:`AcquisitionOptimizer` that will provide a batch of points with shape [B, D].
+    """
+    tf.debugging.assert_positive(batch_size)
+
+    def optimizer(search_space: SP, f: AcquisitionFunction) -> TensorType:
+
+        current_batch = batch_size_one_optimizer(search_space, f)  # [1, D]
+
+        for i in range(1, batch_size):
```
joelberkeley, line 168: similar to below: do we need `tf.range`?

tests/unit/acquisition/test_optimizer.py
```diff
@@ -82,11 +84,15 @@ def test_optimize_continuous(
...
     ],
 )
-def test_optimize_batch(
-    search_space: Box, acquisition: AcquisitionFunction, maximizer: TensorType, batch_size: int
+def test_batchify(
+    search_space: Box,
+    acquisition: AcquisitionFunction,
+    maximizer: TensorType,
+    batch_size: int,
+    batchify_type,
```
joelberkeley, line 92:
```suggestion
    batchify_type: AcquisitionOptimizer[Box],
```

trieste/acquisition/optimizer.py
```diff
@@ -140,3 +143,44 @@ def target_func_with_vectorized_inputs(
         return tf.reshape(vectorized_points, [batch_size, -1])  # [B, D]
 
     return optimizer
+
+
+def batchify_greedy(
```
joelberkeley, line 148: does it make sense to rename `batchify` to explain how it differs from this kind of batching?

trieste/acquisition/optimizer.py
```diff
@@ -140,3 +143,44 @@ def target_func_with_vectorized_inputs(
...
+    batch_size_one_optimizer: AcquisitionOptimizer[SP],
+    batch_size: int,
+) -> AcquisitionOptimizer[SP]:
+    """
+    A wrapper around our :const:`AcquisitionOptimizer`s. This class wraps a
```
joelberkeley, line 153: "our" implies `batchify_greedy` isn't valid for user-defined optimizers, whereas it very much is

joelberkeley: also, this isn't a class. You've also got the same problem on the other function

tests/unit/acquisition/test_optimizer.py
```diff
@@ -74,6 +75,7 @@ def test_optimize_continuous(
 
 
 @random_seed
+@pytest.mark.parametrize("batchify_type", [batchify, batchify_greedy])
```
joelberkeley, line 78: what differentiates `batchify_greedy` from `batchify`, and can we have a test for that?

trieste/acquisition/optimizer.py
joelberkeley, line 153:
```suggestion
    r"""
    A wrapper around our :const:`AcquisitionOptimizer`\ s. This class wraps a
```
you might wanna watch out for remaining PR comments - this was in the last PR too but wasn't fixed before merge. You'll need to fix it for `batchify` too - the docstring renders all wrong there

Issue #201, opened 2021-03-11
Allow hypervolume reference point to be less than anti-ideal


joelberkeley:
As a user, I want to be able to calculate the hypervolume with a reference point that's less than the anti-ideal point. This is so that I can emphasise certain regions of the Pareto front, or constrain certain objectives (did I word this last bit right?).

**Describe alternatives you've considered**
The dominated region excluded by such a reference point could either _not_ contribute to the overall hypervolume, or it could count as negative. @vpicheny and @joelberkeley prefer the latter, which implies that all reference points beyond the Pareto front would result in zero total hypervolume, rather than negative.

TsingQAQ:
Tried to think a little how could have this functionality with existing structure of `Pareto` and come up with one approach that might help discuss. 

It seems the main functionality that will be affected is `hypervolume_indicator` and `hypercell_bounds`. 
The `hypervolume indicator` can be trivially extended by just truncating those negative hv contributions to zero (e.g., an implementation refer [here](https://github.com/TsingQAQ/trieste/blob/1c7fbb119f0b213636491e8ea93e2bd2eaaa7278/trieste/utils/pareto.py#L326)).   

As for the `hypercell_bounds`, it seems a bit more complicated by modify gathering from current stored index containing all original front. A lazy way could be to instantiate another `Pareto` which its front is only those the reference point dominated one (see [this line](https://github.com/TsingQAQ/trieste/blob/1c7fbb119f0b213636491e8ea93e2bd2eaaa7278/trieste/utils/pareto.py#L363)). this partially works but not support on the situation when the **reference point has dominated all of the front** (this is sometimes needed especially in constraint scenario, for instance when no point has satisfied the constraint), as an empty input of observations is not allowed for `Pareto`.

Here is a [draft branch](https://github.com/TsingQAQ/trieste/blob/ref_pts/trieste/utils/pareto.py) for this functionality.

PR #217, opened 2021-03-31
add `get_hypercell_bounds` functionality

TsingQAQ:
### New Features  
A functionality of getting the partitioned hypercell bounds, which can be utilized by acquisition function calculation like [here](https://github.com/TsingQAQ/trieste/blob/965f79e968c4381b058247f050f9c2c81dfdc4b1/trieste/acquisition/function.py#L600). 

### Known limitation
It is now followed the current style of `pareto` by assuming the reference point can not be less than the anti-ideal point, which can be improved for more generalized problem setting as mentioned at #201.


trieste/utils/pareto.py
```diff
@@ -148,3 +148,42 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
         hypervolume = tf.reduce_sum(tf.reduce_prod(upper - lower, 0))
 
         return tf.reduce_prod(reference[None] - min_pfront) - hypervolume
+
+    def get_hypercell_bounds(self, anti_reference: TensorType, reference: TensorType) -> Tuple:
```
```suggestion
    def hypercell_bounds(self, anti_reference: TensorType, reference: TensorType) -> Tuple:
```
partly cos we haven't used `get` anywhere else, and also cos I've never really understood the value of `get_xyz`: I mean it's a function so the `get_`-ness is captured in the signature? Dunno, anyway there's a suggestion

```suggestion
    def get_hypercell_bounds(self, anti_reference: TensorType, reference: TensorType) -> tuple[TensorType, TensorType]:
```

joelberkeley: you'll also need `from __future__ import annotations` before your import statements for this to work

trieste/utils/pareto.py
```diff
@@ -148,3 +148,42 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+                (self.bounds.upper_idx, ["N", "D"]),
+                (self.front, ["M", "D"]),
+                (reference, ["D"]),
+                (anti_reference, ["D"]),
+            ]
+        )
+
+        pseudo_pfront = tf.concat((anti_reference[None], self.front, reference[None]), axis=0)
+        N = tf.shape(self.bounds.upper_idx)[0]
+        D = tf.shape(self.bounds.upper_idx)[1]
```
```suggestion
        N, D = tf.shape(self.bounds.upper_idx)
```

TsingQAQ: Not very sure if it is proper to mention here, but it seems when I used the suggested more compact notation, it will raise an issue when running the notebook of analytic EHVI: 
```
    OperatorNotAllowedInGraphError: iterating over `tf.Tensor` is not allowed: AutoGraph did convert this function. This might indicate you are trying to use an unsupported feature.
```
I do not have an idea on this, maybe any comments?

trieste/utils/pareto.py
TsingQAQ: Is it a good idea to directly make use of `Tuple[TensorType, TensorType]`?

joelberkeley: not really - we've migrated away from using `typing.Tuple` (except for type aliases where that's not possible yet)

trieste/utils/pareto.py
joelberkeley: oh wow, that's odd. Especially since tuple unpacking defers to `__iter__`, and so does `for`, and `for` is supported by `tf.function`

joelberkeley: yeah just stick with your version

trieste/utils/pareto.py
```diff
@@ -148,3 +149,44 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+            [
+                (self.bounds.lower_idx, ["N", "D"]),
+                (self.bounds.upper_idx, ["N", "D"]),
+                (self.front, ["M", "D"]),
+                (reference, ["D"]),
+                (anti_reference, ["D"]),
+            ]
+        )
+
+        pseudo_pfront = tf.concat((anti_reference[None], self.front, reference[None]), axis=0)
```
TsingQAQ, line 355: @joelberkeley one thing not very sure here is, do we need some defensive programming to cast dtype here?

joelberkeley: it's a good point. I'm not sure what's best. The alternative would be raise an exception if the dtypes didn't match. While that's a little strict, it's what I'd expect tensorflow to do so the consistency might help. You could do that with `tf.debugging.assert_same_float_dtype`

trieste/utils/pareto.py
```diff
@@ -148,3 +149,44 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
         hypervolume = tf.reduce_sum(tf.reduce_prod(upper - lower, 0))
 
         return tf.reduce_prod(reference[None] - min_pfront) - hypervolume
+
```
joelberkeley, line 324: iirc, you'll need to make the bounds functionality private, which is going to mean you'll have to rewrite a fair few tests

TsingQAQ: I am not quite following here, do you mean the `pareto.bounds`?  If this is the case, from application perspective, it is not so far needed for acquisition function

joelberkeley: yes, making `.bounds` private (`._bounds`)

trieste/utils/pareto.py
TsingQAQ: Used `tf.debugging.assert_type` instead.
Seems `tf.debugging.assert_same_float_dtype` do not support dtype assert when Tensor's shape>1, weird.

trieste/utils/pareto.py
```diff
@@ -47,7 +48,7 @@ def non_dominated(observations: TensorType) -> Tuple[TensorType, TensorType]:
     return tf.boolean_mask(observations, dominance == 0), dominance
 
 
-class BoundedVolumes:
+class _BoundedVolumes:
```
TsingQAQ, line 52: As discussed with @satrialoka, it may be suitable to make this class internal, CMIIW.

trieste/utils/pareto.py
TsingQAQ: Made private, the test is not modified too much as there is only one test verifying `_bounds` numerical correctness, so I was hoping this wouldn't be a big issue here, CMIIW.

trieste/utils/pareto.py
henrymoss: What does this mean?

TsingQAQ: The  `BoundedVolumes` class is more only used internally to store the index of bounds.
And since we could provide the `get_hypercell_bounds` for users to directly get the partitioned cell bounds for acquisition function building, this class may not need to expose outside `pareto.py`.

henrymoss: Makes sense!

henrymoss (commented):
I'm happy with this. 

Worth running by @uri-granta  before merging though :)

trieste/utils/pareto.py
```diff
@@ -50,7 +49,7 @@ def non_dominated(observations: TensorType) -> Tuple[TensorType, TensorType]:
...
     """
     A :class:`BoundedVolumes` store the index of the Pareto front to form lower and upper
```
uri-granta:
```suggestion
    A :class:`_BoundedVolumes` store the index of the Pareto front to form lower and upper
```

trieste/utils/pareto.py
```diff
@@ -299,26 +298,69 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
         lower = tf.reshape(tf.gather_nd(pseudo_front, lower_idx), [D, N])
         hypervolume = tf.reduce_sum(tf.reduce_prod(upper - lower, 0))
 
         return tf.reduce_prod(reference[None] - min_front) - hypervolume
+
+    def hypercell_bounds(
+        self, anti_reference: TensorType, reference: TensorType
+    ) -> tuple[TensorType, TensorType]:
+        """
+        Get the partitioned hypercells lower and upper bounds
```
uri-granta:
```suggestion
        Get the partitioned hypercell's lower and upper bounds.
```

trieste/utils/pareto.py
```diff
@@ -299,26 +298,69 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+        :param anti_reference: a worst point to use with shape [D].
+            Defines the lower bound of the hypercell
+        :param reference: a reference point to use, with shape [D].
+            Defines the upper bound of the hypervolume.
+            Should be equal or bigger than the anti-ideal point of the Pareto set.
```
uri-granta:
```suggestion
            Should be equal to or bigger than the anti-ideal point of the Pareto set.
```

trieste/utils/pareto.py
```diff
@@ -299,26 +298,69 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+            For comparing results across runs, the same reference point must be used.
+        :return: lower, upper bounds or the partitioned cell
```
uri-granta:
```suggestion
        :return: lower, upper bounds of the partitioned cell
```

uri-granta (approved):
Looks good to me!

PR #219, opened 2021-04-07
Add taskipy support for running tests, generating cov reports

uri-granta:
Add a [taskipy](https://github.com/illBeRoy/taskipy) section to the pyproject.toml file, to speed up running tests for people who are using a virtual environment rather than tox. Leaving the tox config alone, so people (and github workflow) can continue using those.

To use the tasks, install the requirements into your venv:

```
pip install -r common_build/taskipy/requirements.txt -c common_build/taskipy/constraints.txt
```

You can then run the following tasks from anywhere inside the repository. In future, new tasks could be added (and existing ones modified) as needed.

- `task tests` to run all tests (including generating a code coverage report)
- `task mypy` to run mypy type checking
- `task check_format` to check (but not fix) formatting with black, flake8 and isort
- `task format` to automatically fix formatting with black, flake8 and isort

To have task automatically tab-complete to the list of configured tasks, save the following file somewhere and add `source taskipy-completion.bash` to your .bashrc file.

```bash
_task_completions()
{
    PYPROJECTDIR=$(pwd)
    while [[ "$PYPROJECTDIR" != "" && ! -e "$PYPROJECTDIR/pyproject.toml" ]]; do
        PYPROJECTDIR=${PYPROJECTDIR%/*}
    done
    PYPROJECT="$PYPROJECTDIR/pyproject.toml"
    if test -f $PYPROJECT; then
        if [ "${#COMP_WORDS[@]}" != "2" ]; then return; fi
        TASKS=$(awk '/\[/{prefix=$0; next} $1{print prefix $0}' $PYPROJECT  | grep -oP "taskipy.tasks]\K[^=]*(?==)")
        COMPREPLY=($(compgen -W "$TASKS" "${COMP_WORDS[1]}"))
    fi
}

complete -F _task_completions task
```


henrymoss:
Looks like a useful change! 

Why does 
`pip install -r common_build/taskipy/requirements.txt -c common_build/taskipy/constraints.txt`

have an apostrophe after pip in your description of the PR?

Also, it would be worth adding this information into the  contribution guidelines ( https://github.com/secondmind-labs/trieste/blob/develop/CONTRIBUTING.md ), where you should also add your name ;)

uri-granta:
The backtick after pip was just a typo, sorry! I'll have a go at updating the contribution guidelines. The only concern there is to make sure people realise that running tox is (currently) still the only way of making 100% sure your submission will succeed (since without a tool like poetry it's easy to mess up your venvs).

Should I add myself as both Admin and Maintainer?

CONTRIBUTING.md
```diff
@@ -79,6 +79,21 @@ $ flake8 .
 $ isort .
 ```
 
+#### Virtual environments and taskipy
+
+During development, it can sometimes be more convenient to set up a virtual environment to run the tests in rather than use tox. If you decide to do this, then we recommend installing [taskipy](https://github.com/illBeRoy/taskipy) by running:
```
henrymoss: rather than *using* tox

henrymoss (approved):

PR #222, opened 2021-04-09
Improve unit test coverage

uri-granta:
Add some more unit tests to improve our (already pretty good) coverage.

### DIFF
--- a/tests/unit/models/test_model_interfaces.py
+++ b/tests/unit/models/test_model_interfaces.py
@@ -452,6 +452,15 @@ def test_sparse_variational_model_attribute() -> None:
     assert sv.model is model
 
 
+def test_sparse_variational_update_updates_num_data() -> None:
+    model = SparseVariational(
+        _svgp(tf.zeros([1, 4])),
+        Dataset(tf.zeros([3, 4]), tf.zeros([3, 1])),
+    )
+    model.update(Dataset(tf.zeros([5, 4]), tf.zeros([5, 1])))
+    assert model.model.num_data == 5
+
+
 @pytest.mark.parametrize(
     "new_data",
     [Dataset(tf.zeros([3, 5]), tf.zeros([3, 1])), Dataset(tf.zeros([3, 4]), tf.zeros([3, 2]))],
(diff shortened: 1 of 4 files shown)


tests/unit/models/test_model_interfaces.py
```diff
@@ -452,6 +452,15 @@ def test_sparse_variational_model_attribute() -> None:
...
+def test_sparse_variational_update_updates_num_data() -> None:
+    model = SparseVariational(
+        _svgp(tf.zeros([1, 4])),
+        Dataset(tf.zeros([3, 4]), tf.zeros([3, 1])),
+    )
+    model.update(Dataset(tf.zeros([5, 4]), tf.zeros([5, 1])))
+    assert model.model.num_data == 5
+
+
 @pytest.mark.parametrize(
```
henrymoss, line 464: This is good for now. The SVGP stuff will be revamped in the next couple of weeks anyway :)

trieste/acquisition/function.py
```diff
@@ -199,6 +199,7 @@ def prepare_acquisition_function(
             raise ValueError("Dataset must be populated.")
 
         query_points = self._search_space.sample(num_samples=self._grid_size)
+        tf.debugging.assert_same_float_dtype([dataset.query_points, query_points])
         query_points = tf.concat([dataset.query_points, query_points], 0)
         fmean, fvar = model.predict(query_points)
         fsd = tf.math.sqrt(fvar)
```
henrymoss, line 205: Nice check. Although wont we get the same error from the line below anyway?

uri-granta: The error message in the line below is a tiny bit more cryptic in that it doesn't make it obvious that either type is ok as long as they're consistent.

henrymoss: Sure. Maybe it would be better to just force the samples to be the same dtype as the query points?

You might be able to specify their type when you sample them?

henrymoss (approved):

trieste/acquisition/function.py
uri-granta: I think an error is probably better than silently changing the sample dtype, but not too fussed either way. If you're ok, I might check this in as is.

henrymoss: Yep go for it!

PR #223, opened 2021-04-12
Fix synthetic examples output

henrymoss:
Small fix so that the output of the two Hartmann synthetic test functions returns the right dimensions (ie. [N,1] rather than N). Also added a unit test to make sure this does not happen again.

### DIFF
--- a/tests/unit/utils/test_objectives.py
+++ b/tests/unit/utils/test_objectives.py
@@ -56,7 +56,9 @@ from trieste.utils.objectives import (
 def test_objective_maps_minimizers_to_minimum(
     objective: Callable[[TensorType], TensorType], minimizers: TensorType, minimum: TensorType
 ) -> None:
-    npt.assert_allclose(objective(minimizers), tf.squeeze(minimum), rtol=1e-5)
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_objectives.py
```diff
@@ -56,7 +56,11 @@
 def test_objective_maps_minimizers_to_minimum(
     objective: Callable[[TensorType], TensorType], minimizers: TensorType, minimum: TensorType
 ) -> None:
-    npt.assert_allclose(objective(minimizers), tf.squeeze(minimum), rtol=1e-5)
+    objective_values_at_minisers = objective(minimizers)
```
uri-granta: Is minisers an intentional shortening of minimizers? Also, are we generally sticking to US spelling is there no policy on this?

uri-granta (approved):
Looks good. Pity mypy can't reason about array shapes (or at least dimensions).

tests/unit/utils/test_objectives.py
henrymoss: That is a typo. I think we are sticking to US.

PR #224, opened 2021-04-14
Run pylint and fix some errors

uri-granta:
- Add support for running the pylint static code analysis tool, including setting some sensible defaults.
- Fix some of the resulting complaints (mostly missing docstrings).
- Don't enable this as part of `tox -e format` quite yet: let this bed in for a while to see what sort of errors it throws up.


trieste/acquisition/optimizer.py
```diff
@@ -54,8 +54,6 @@ def automatic_optimizer_selector(
     :return: The batch of points in ``space`` that maximises ``target_func``, with shape [1, D].
     """
 
-    optimizer: AcquisitionOptimizer[Box | DiscreteSearchSpace]
```
uri-granta, line 57: Pylint flagged this as unnecessary. Should this be used elsewhere as a return type?

trieste/bayesian_optimizer.py
```diff
@@ -255,7 +255,7 @@ def optimize(
                     model.update(dataset)
                     model.optimize(dataset)
 
-            except Exception as error:
+            except Exception as error:  # pylint: disable=broad-except
```
uri-granta, line 258: Generally it's a bad idea to catch all `Exception`s, but it makes sense here, so I've disabled the warning just for this line.

uri-granta (commented):
One remaining error that I haven't fixed yet is regarding VariationalGaussianProcess update method, which has a jitter parameter, unlike all the other TrainableProbabilisticModel update methods. Any idea how this is intended to be used?

henrymoss:
> One remaining error that I haven't fixed yet is regarding VariationalGaussianProcess update method, which has a jitter parameter, unlike all the other TrainableProbabilisticModel update methods. Any idea how this is intended to be used?

Perhaps @vpicheny  knows?

trieste/acquisition/optimizer.py
henrymoss: I needed that to pass the type tests? If you can get rid of it, that would be great.

trieste/bayesian_optimizer.py
henrymoss: :)

trieste/acquisition/optimizer.py
uri-granta: Looks like it's no longer needed: I've removed it and the type tests are still passing.

trieste/models/optimizer.py
```diff
@@ -60,6 +60,13 @@ class Optimizer:
     """ If `True`, the optimization process will be compiled with :func:`tf.function`. """
 
     def create_loss(self, model: tf.Module, dataset: Dataset) -> LossClosure:
+        """
+        Extract a loss function for the specified `model` with the `dataset`.
+
+        :param model: The model to extract a loss function for.
+        :param dataset: The data with which to extract the loss function.
```
henrymoss: Extract seems a weird word here? Perhaps say build instead?

trieste/models/optimizer.py
```diff
@@ -156,6 +171,15 @@ def _create_scipy_optimizer(
 
 @singledispatch
 def create_loss_function(model, dataset: TrainingData, compile: bool = False) -> LossClosure:
+    """
+    Generic function for extracting a loss function for a specified `model` and `dataset`.
+    The implementations depends on the type of the model.
+
+    :param model: The model to extract a loss function for.
+    :param dataset: The data with which to extract the loss function.
+    :param compile: Whether to compile with :func:`tf.function`.
```
henrymoss: same as above (extract -> build)

henrymoss (approved):

henrymoss:
Can this be merged?

PR #226, opened 2021-04-21
Fixed inconsistensies with model sampling

henrymoss:
Currently our model wrappers are supposed to  return samples of shape [..., N, S, E] when queried at a points of shape [... N, D] (as detailed in line 70 of model_interfaces.py). E is output dimension (typically 1), and S is number of samples.

While this is the case for the mock models defined for testing and elsewhere in the docs, the Gpflow predictor wrapper (line 263 of model_interfaces.py) does not return samples with this shape, instead returning samples of shape [..., S, N, E] (the behaviour of gpflow models). 

Also, our tests pick and choose which output shape they are expecting depending on which model type they are testing (see lines 115 v.s. 434 in test_model_interfaces.py). The expected index of the sampling dimension is different in these two tests.

I have added a fix to make the gpflow wrapper match with our documentation and have fixed the Thompson sampling rule accordingly. This should now be consistent across the whole library (including tests).


henrymoss:
The Thompson sampler in the Thompson sampling rule will be updated in my next PR (where I am moving samplers), so dont worry about that too much for now.

trieste/models/model_interfaces.py
```diff
@@ -261,7 +261,10 @@ def predict_joint(self, query_points: TensorType) -> tuple[TensorType, TensorTyp
         return self.model.predict_f(query_points, full_cov=True)
 
     def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
-        return self.model.predict_f_samples(query_points, num_samples)
+        samples = self.model.predict_f_samples(query_points, num_samples)  # [..., S, N, E]
+        rank = tf.rank(samples)
```
uri-granta: Maybe worth a comment here explaining why we do this?

uri-granta (approved):
Looks good though it does raise the question of why we chooe to return sample shapes in a different format to gpflow. Are both options equally arbitrary or is one more 'reasonable'?

vpicheny:
I'm not sure I understand... in the `IndependentReparametrizationSampler` we return shapes `[..., S, 1, L]`, and in `BatchReparametrizationSampler` we return `[..., S, B, L]`,  which is consistent with gpflow.

henrymoss:
I have now made changes the other way. Now EVERYTHING (including gpflow wrapper, parameterization samples, and test dummy models) return samples following the gpflow convention. This means that I have only change the test dummy models and some of their tests.

I have also improved the documentation a little.

tests/util/model.py
```diff
@@ -49,7 +49,7 @@ def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
         mean, var = self.predict(query_points)
```
uri-granta, line 49: So if query_points is of always shape [..., N, D] should this be asserted somewhere, to catch someone passing in something with rank < 2? (comment not specific to this ProbabilisticModel)

uri-granta (approved):

PR #227, opened 2021-04-21
Local penalization

henrymoss:
Hi 

This is a draft PR (no tests) for local penalization. @uri-granta  and I have converged on a MultiAcquisitionFunctionBuilder solution (line 740 of function.py). This builder can provide multiple acquisition functions during each BO step, as required for greedy batch selection (with or without local penalization). Crucially, building the Builder allows us to get away with performing some expensive computations (like estimating Lipshitz constants) only once per BO step.

Please have a look at Line 792 (function.py) to see the local penalisation acquisition function and at Line 160-175  (rule.py) to see how our EGO rule deals with greedy and joint batch design.

### DIFF
--- a/trieste/acquisition/function.py
+++ b/trieste/acquisition/function.py
@@ -806,7 +806,8 @@ class LocallyPenalizedExpectedImprovement(MultiAcquisitionFunctionBuilder[Tensor
     For our penalization function, we implement the soft penalization strategy of :cite:`Gonzalez:2016`, 
     where an estimate of the objective function's Lipschitz constant is used to control the size 
     of penalization. The Lipschitz constant and additionl penalization parameters are estimated once 
-    when preparing  the :const:`AcquisitionFunctionMaker` (not for each individual :const:`AcquisitionFunction`).
+    when preparing the :const:`AcquisitionFunctionMaker` and reused for each 
+    individual :const:`AcquisitionFunction`) it makes.
     
     """
 
@@ -844,6 +845,7 @@ class LocallyPenalizedExpectedImprovement(MultiAcquisitionFunctionBuilder[Tensor
             grads = g.gradient(mean,sampled_points)
             grads_norm =  tf.norm(grads, axis=1)
             max_grads_norm = tf.reduce_max(grads_norm)
+            return max_grads_norm
 
         lipschitz_constant = get_lipschitz_estimate(samples)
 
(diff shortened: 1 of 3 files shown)


trieste/acquisition/function.py
```diff
@@ -717,3 +717,215 @@ def batch_ei(at: TensorType) -> TensorType:
...
+            return max_grads_norm
+
+        lipschitz_constant = get_lipschitz_estimate(samples)
+
+        if lipschitz_constant < 1e-5: # threshold to improve numerical stability for 'flat' models
+                lipschitz_constant = 10
+               
+        eta = tf.reduce_min(mean, axis=0)
+
+            def acquisition_function_maker(pending_points: TensorType=None) -> AcquisitionFunction:
```
vpicheny, line 857: indentation issue?

trieste/acquisition/function.py
```diff
@@ -717,3 +717,215 @@ def batch_ei(at: TensorType) -> TensorType:
...
+                g.watch(sampled_points)
+                mean, _ = model.predict(sampled_points)
+            grads = g.gradient(mean,sampled_points)
+            grads_norm =  tf.norm(grads, axis=1)
+            max_grads_norm = tf.reduce_max(grads_norm)
+            return max_grads_norm
+
+        lipschitz_constant = get_lipschitz_estimate(samples)
+
+        if lipschitz_constant < 1e-5: # threshold to improve numerical stability for 'flat' models
```
vpicheny, line 852: indeed this seems very arbitrary

vpicheny (commented):
It is difficult to assess if this is the right solution. The `AcquisitionFunctionMaker` introduces a lot not new code that may be puzzling for many users. Is it here for the sole purpose of avoid recomputing the Lipschitz constant? if so is this the best solution? Can we think of other cases for which such an abstraction would be useful? 
Thompson sampling with RFF is another close case where we optimise a bunch of acquisition functions so it'd be good to find commonalities. The main difference being that in TS the acq function optim can be done independently, hence in parallel, and not sequentially. 
Local penalization is also something we'd like to be able to apply broadly, e.g. to multiobjective or constrained cases. How easy would that be with the current solution?

uri-granta:
> The `AcquisitionFunctionMaker` introduces a lot not new code that may be puzzling for many users. Is it here for the sole purpose of avoid recomputing the Lipschitz constant? if so is this the best solution? Can we think of other cases for which such an abstraction would be useful?

That's the key question. To me this looks like a clean abstraction, and it certainly works well for computing the Lipschitz constant just once. Whether or not it's the right abstraction depends on how well it applies to other cases.

One alternative approach would be to generalize AcquisitionFunctionBuilder's prepare_acquisition_function method so it can take additional optional parameters, pass in the pending points each time, and calculate the Lipschitz function only when there are no pending points. That might well be simpler.

PR #229, opened 2021-04-22
Move samplers

henrymoss:
This PR tidies up our sampling functionality and reduces the size of function.py .

We now have a new samplers.py file that contains our samplers, including a DiscreteThompsonSampler and GumbelSampler that have been extracted from the ThompsonSampling rule and MinValueEntropySearch builder, respectively.

All relevant tests (aswell as some new tests) have been added in a new test_sampler.py file.

### DIFF
--- a/tests/unit/acquisition/test_sampler.py
+++ b/tests/unit/acquisition/test_sampler.py
@@ -1,0 +1,265 @@
+# Copyright 2020 The Trieste Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import annotations
+
+import math
+
+import numpy.testing as npt
+import pytest
+import tensorflow as tf
+import tensorflow_probability as tfp
+
+from tests.util.misc import TF_DEBUGGING_ERROR_TYPES, ShapeLike, quadratic, random_seed
+from tests.util.model import GaussianProcess, QuadraticMeanAndRBFKernel, rbf
+from trieste.acquisition.sampler import (
+    BatchReparametrizationSampler,
+    DiscreteThompsonSampler,
+    GumbelSampler,
+    IndependentReparametrizationSampler,
+)
+from trieste.data import Dataset
+from trieste.space import Box
+from trieste.utils.objectives import branin
+
+
+@pytest.mark.parametrize("sample_size", [0, -2])
+def test_gumbel_sampler_raises_for_invalid_sample_size(
+    sample_size: int,
+) -> None:
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        GumbelSampler(sample_size, QuadraticMeanAndRBFKernel())
+
+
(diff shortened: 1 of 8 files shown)


trieste/acquisition/function.py
```diff
@@ -11,6 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""
+This module contains acquisition function builders, which build acquisition functions
```
uri-granta: Maybe also say that this defines what an acquisition function actually is?

> This module defines acquisition functions, which estimate the utility of evaluating sets of candidate points, and contains builders for constructing such functions.

trieste/acquisition/sampler.py
```diff
@@ -0,0 +1,301 @@
...
+from ..utils import DEFAULTS
+
+
+class Sampler(ABC):
+    r"""
+    An :class:`Sampler` samples a specific quantity according to an
+    underlying :class:`ProbabilisticModel`.
+    """
+
+    @abstractmethod
```
uri-granta: If the default init is commonly all that you want then I think you'd be justified in making it not absract. That way you could remove the init functions for DiscreteThompsonSampler and GumbelSampler. But also fine as is.

FYI there is also a way to ensure that this gets called even if the subclasses forget to call `super().__init__`, but I haven't seen it used anywhere so suspect there's a good reason to avoid it.

trieste/acquisition/sampler.py
```diff
@@ -0,0 +1,301 @@
...
+    def __init__(self, sample_size: int, model: ProbabilisticModel):
+        """
+        :param sample_size: The desired number of samples.
+        :param model: The model to sample from.
```
uri-granta, line 41: Included raises comment (as below)?

trieste/acquisition/sampler.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        :param sample_size: The number of samples to take at each point. Must be positive.
+        :param model: The model to sample from.
+        :raise ValueError (or InvalidArgumentError): If ``sample_size`` is not positive.
+        """
+        super().__init__(sample_size, model)
+
+    def __repr__(self) -> str:
+        """"""
+        return f"DiscreteThompsonSampler({self._sample_size!r}, {self._model!r})"
```
uri-granta: Since the reprs all follow the same format, you could if you wish put them in the base class instead. Something like:

```
        return f"{self.__class__.__name__}({self._sample_size!r}, {self._model!r})"
```

tests/unit/acquisition/test_sampler.py
```diff
@@ -0,0 +1,265 @@
...
+    GumbelSampler,
+    IndependentReparametrizationSampler,
+)
+from trieste.data import Dataset
+from trieste.space import Box
+from trieste.utils.objectives import branin
+
+
+@pytest.mark.parametrize("sample_size", [0, -2])
+def test_gumbel_sampler_raises_for_invalid_sample_size(
```
uri-granta, line 37: Fine as is, but FYI it's possible to put related tests in a class for code clarity. For example, you could put all the gumbel sampler tests in a class TestGumpelSampler.

uri-granta (approved):
Looks much better afterwards!

trieste/acquisition/sampler.py
```diff
@@ -0,0 +1,273 @@
...
+
+    @abstractmethod
+    def sample(self, at: TensorType) -> TensorType:
+        """
+        :param at: Input points that define the sampler.
+        :return: Samples.
+        """
+
+
+class DiscreteThompsonSampler(Sampler):
```
vpicheny, line 61: does it make sense to have this class? maybe we should take this opportunity to rewrite the TS code using `BatchReparametrizationSampler`? On the other hand this would limit its use to gpflow models.

PR #230, opened 2021-04-22
Local Penalization

henrymoss:
An alternative approach for greedy batch design.

We now have a special type of acquisition function builder (as subclass of the general builder) that requires a pending_points input when it is built. This allows us to return the acquisition function that measures the utility of making a new evaluation **given** the pending points are already going to be queried.

When the builder is called with pending_points = None (i.e when collecting the first batch element), we calculate the Lipschitz constant. This is then reused for all remaining points in the batch.

This solution is less elegant but much more intuitive and doest require as much code.


trieste/acquisition/function.py
```diff
@@ -852,3 +853,202 @@ def batch_ei(at: TensorType) -> TensorType:
...
+        :param models: The models over each dataset in ``datasets``.
+        :param pending_points: Points already chosen to be in the current batch (of shape [M,D]), 
+        	where M is the number of pending points and D is the search space dimension.
+        :return: An acquisition function.
+        """
+
+
+class SingleModelGreedyAcquisitionBuilder(ABC):
+    """
+    Convenience acquisition function builder for an acquisition function (or component of a
```
uri-granta: You forgot to update the docstring.

Also it's a pity we have to duplicate code like this, but I can't think of any cleaner way.

trieste/acquisition/function.py
```diff
@@ -852,3 +853,202 @@ def batch_ei(at: TensorType) -> TensorType:
...
+	        self._lipschitz_constant = lipschitz_constant
+	        self._eta = eta
+
+
+        log_base_acquisition = tf.math.log(expected_improvement(model, eta))
+
+        if not pending_points: # no penalization required for first batch element.
+            return base_acquisition
+
+        penalization = local_penalizer(model, pending_points, lipschitz_constant, eta)
```
```suggestion
        penalization = local_penalizer(model, pending_points, self._lipschitz_constant, self._eta)
```

Also, you may want to check that these aren't None (in case the function was incorrectly called with non-empty pending points the first time).

trieste/acquisition/function.py
```diff
@@ -852,3 +853,202 @@ def batch_ei(at: TensorType) -> TensorType:
...
+	        if lipschitz_constant < 1e-5: # threshold to improve numerical stability for 'flat' models
+	                lipschitz_constant = 10
+
+	        eta = tf.reduce_min(mean, axis=0)
+
+	        self._lipschitz_constant = lipschitz_constant
+	        self._eta = eta
+
+
+        log_base_acquisition = tf.math.log(expected_improvement(model, eta))
```
```suggestion
        log_base_acquisition = tf.math.log(expected_improvement(model, self._eta))
```

and maybe move below the not pending_points check

trieste/acquisition/rule.py
```diff
@@ -93,7 +97,7 @@ class EfficientGlobalOptimization(AcquisitionRule[None, SP_contra]):
 
     def __init__(
         self,
-        builder: AcquisitionFunctionBuilder | None = None,
+        builder: AcquisitionFunctionBuilder | GreedyAcquisitionFunctionBuilder | None = None,
```
uri-granta, line 101: You don't need this change, since a GreedyAcquisitionFunctionBuilder is a type of AcquisitionFunctionBuilder.

```suggestion
        builder: AcquisitionFunctionBuilder | None = None,
```

trieste/acquisition/rule.py
```diff
@@ -154,8 +160,19 @@ def acquire(
         :param state: Unused.
         :return: The single (or batch of) points to query, and `None`.
         """
+
         acquisition_function = self._builder.prepare_acquisition_function(datasets, models)
-        return self._optimizer(search_space, acquisition_function), None
+        points = self._optimizer(search_space, acquisition_function)
+
+        if isinstance(self._builder, GreedyAcquisitionFunctionBuilder) and self._num_query_points>1:
```
uri-granta: If you intend to call `prepare_acquisition_function(datasets, models)` for the first point regardless of whether it's greedy or not, you may need to give pending_points a default value in the classes above. The alternative is to fully separate GreedyAcquisitionFunctionBuilder calls from the other ones.

uri-granta (commented):
I think this is better than the previous approach, but feel like it's a little ugly. However, I'm struggling to think of how to improve it further.

Also I don't think the **kwargs is set up quite right at the moment. I'll slack you to chat.

henrymoss:
I have a working version of LP now. It is reasonably well tested, both across unit and integration tests.  A notebook will follow in another PR.

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
             return tf.reduce_mean(batch_improvement, axis=-1, keepdims=True)  # [..., 1]
 
         return batch_ei
+
+
+class GreedyAcquisitionFunctionBuilder(ABC):
+    """
+    An :class:`GreedyAcquisitionFunctionBuilder` builds an acquisition function
```
```suggestion
    A :class:`GreedyAcquisitionFunctionBuilder` builds an acquisition function
```

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+    suitiable for greedily building batches for batch Bayesian
+    Optimisation. :class:`GreedyAcquisitionFunctionBuilder` differs
+    from :class:`AcquisitionFunctionBuilder` by requiring that a set
+    of pending points is passed to the builder.
```
uri-granta: (Maybe also note that it's called multiple times each iteration?)

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+    Builder of the acquisition function maker for greedily collecting batches by local
+    penalization.  The resulting :const:`AcquisitionFunctionMaker` takes in a set of pending
+    points and returns the expected improvment acquisiton function penalized around those points.
+
+    Local penalization allows us to perform batch Bayesian optimization with a standard (non-batch)
+    acqusition function. By iteratively building a batch of points by maximizing an acquisition
+    function that is clipped around the values correponding to locations close to the already
+    chosen (pending) points, local penalization provides diverse batches of canidadate points.
+
+    Local penalization is applied to the acquisition function multiplicatively, however, to
```
```suggestion
    Local penalization is applied to the acquisition function multiplicatively. However, to
```

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+    improve numerical stability, we perfom additive penalization in a log space.
+
+    For our penalization function, we implement the soft penalization strategy
+    of :cite:`Gonzalez:2016`, where an estimate of the objective function's Lipschitz
+    constant is used to control the size of penalization.
+
+    The Lipschitz constant and additionl penalization parameters are estimated once
+    when preparing the :const:`AcquisitionFunctionMaker` and reused for each
```
uri-granta: (this paragraph is now out of date)

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+    from :class:`AcquisitionFunctionBuilder` by requiring that a set
+    of pending points is passed to the builder.
+    """
+
+    @abstractmethod
+    def prepare_acquisition_function(
+        self,
+        datasets: Mapping[str, Dataset],
+        models: Mapping[str, ProbabilisticModel],
+        pending_points: TensorType = None,
```
uri-granta: Strictly speaking, pending points should be marked as optional:

```suggestion
        pending_points: Optional[TensorType] = None,
```

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+        :param model: The model over the specified ``dataset``.
+        :param pending_points: The points we penalize with respect to.
+        :return: The (log) expected improvement penalized with respect to the pending points.
+        This function will raise :exc:`ValueError` if its first call does not
+        have pending_points=None.
+        """
+        if len(dataset.query_points) == 0:
+            raise ValueError("Dataset must be populated.")
+
+        if (
```
uri-granta, line 898: Possibly clearer formatting as:

```python
if pending_points is None:
    # only compute ...
    sample = ....
```

Ditto for the comments on line lines 815 and 828 (and lines 134 and 173 in rule).

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+            self._lipschitz_constant = lipschitz_constant
+            self._eta = eta
+
+        if self._lipschitz_constant is None:
+            raise ValueError("Local penalization must be first called with no pending_points.")
+
+        def log_acquisition(x: TensorType) -> TensorType:
+            return tf.math.log(expected_improvement(model, self._eta)(x))
+
+        log_base_acquisition = log_acquisition
```
uri-granta: Is this line necessary? Why not just return/use `log_acquisition`?

tests/unit/acquisition/test_function.py
```diff
@@ -68,6 +71,13 @@ def prepare_acquisition_function(
         return raise_exc
 
 
+class _ArbitraryGreedySingleBuilder(SingleModelGreedyAcquisitionBuilder):
+    def prepare_acquisition_function(
+        self, dataset: Dataset, model: ProbabilisticModel
```
uri-granta:
```suggestion
        self, dataset: Dataset, model: ProbabilisticModel, pending_points: TensorType = None
```

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+        dataset: Dataset,
+        model: ProbabilisticModel,
+        pending_points: TensorType = None,
+    ) -> AcquisitionFunction:
+        """
+        :param dataset: The data from the observer.
+        :param model: The model over the specified ``dataset``.
+        :param pending_points: The points we penalize with respect to.
+        :return: The (log) expected improvement penalized with respect to the pending points.
+        This function will raise :exc:`ValueError` if its first call does not
```
uri-granta: You should put exceptions as a separate entry rather than as part of the return (though you were missing the indentation to do that anyway, which is why there was an error).

```suggestion
        :raise ValueError: if the first call does not
```

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+    when preparing the :const:`AcquisitionFunctionMaker` and reused for each
+    individual :const:`AcquisitionFunction`) it makes.
+
+    """
+
+    def __init__(self, search_space: SearchSpace, grid_size: int = 500):
+        """
+        :param search_space: The global search space over which the optimisation is defined.
+        :param grid_size: Size of the grid over which the Lipschitz constant is estimated.
+        We recommend scaling this with search space dimension.
```
uri-granta: Sphinx needs indentation here to be happy.

```suggestion
            We recommend scaling this with search space dimension.
```

trieste/acquisition/function.py
```diff
@@ -676,3 +676,245 @@ def batch_ei(at: TensorType) -> TensorType:
...
+    where :math:`z = \frac{1}{\sqrt{2\sigma^2(x')}}(L||x'-x|| + \eta - \mu(x'))`.
+
+    The penalization from a set of pending points is just product of the individual penalizations.
+    See :cite:`Gonzalez:2016` for a full derivation.
+
+    :param model: The model over the specified ``dataset``.
+    :param pending_points: The points we penalize with respect to.
+    :param lipschitz_constant: The estimated Lipschitz constant of the objective function.
+    :param eta: The estimated global minima.
+    :return: The local penalization function. This function will raise
```
uri-granta, line 987: Put exceptions in a separate :raise entry.

uri-granta (approved):
Looks good bar some small fixes (which should also cover the typing and docs errors).

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
             return tf.reduce_mean(batch_improvement, axis=-1, keepdims=True)  # [..., 1]
 
         return batch_ei
+
+
+class GreedyAcquisitionFunctionBuilder(ABC):
+    """
+    A :class:`GreedyAcquisitionFunctionBuilder` builds an acquisition function
+    suitiable for greedily building batches for batch Bayesian
```
vpicheny: suitable

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
             return tf.reduce_mean(batch_improvement, axis=-1, keepdims=True)  # [..., 1]
 
         return batch_ei
+
+
+class GreedyAcquisitionFunctionBuilder(ABC):
```
vpicheny, line 764: not sure this is for this PR but this builder could also be used in the asynchronous case I believe? (i.e. we would choose the next experiment while B-1 are still running) In that case maybe Greedy is not the right term?

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
...
+            where M is the number of pending points and D is the search space dimension.
+        :return: An acquisition function.
+        """
+
+
+class LocallyPenalizedExpectedImprovement(SingleModelGreedyAcquisitionBuilder):
+    r"""
+    Builder of the acquisition function maker for greedily collecting batches by local
+    penalization.  The resulting :const:`AcquisitionFunctionMaker` takes in a set of pending
+    points and returns the expected improvment acquisiton function penalized around those points.
```
vpicheny: acquisition

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
...
+
+    Local penalization allows us to perform batch Bayesian optimization with a standard (non-batch)
+    acqusition function. By iteratively building a batch of points by maximizing an acquisition
```
vpicheny, line 846: acquisition

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
...
+    function that is clipped around the values correponding to locations close to the already
+    chosen (pending) points, local penalization provides diverse batches of canidadate points.
```
vpicheny: candidate

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
...
+
+    Local penalization is applied to the acquisition function multiplicatively. However, to
+    improve numerical stability, we perfom additive penalization in a log space.
+
+    For our penalization function, we implement the soft penalization strategy
+    of :cite:`Gonzalez:2016`, where an estimate of the objective function's Lipschitz
+    constant is used to control the size of penalization.
+
+    The Lipschitz constant and additional penalization parameters are estimated once
+    when first preparing the acquisition function with no pending points. These estaimtes
```
vpicheny: estimates

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
...
+    are reused for all subsequent function calls.
+    """
+
+    def __init__(self, search_space: SearchSpace, grid_size: int = 500):
```
vpicheny: grid_size is misleading: we're not using a grid (search_space.sample() returns uniformly distributed samples). Plus, grid_size could be understood as `sample_size = grid_size ^ dim`

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
...
+                lipschitz_constant = 10
+
+            self._lipschitz_constant = lipschitz_constant
+            self._eta = eta
+
+        if self._lipschitz_constant is None:
+            raise ValueError("Local penalization must be first called with no pending_points.")
+
+        def log_acquisition(x: TensorType) -> TensorType:
+            return tf.math.log(expected_improvement(model, self._eta)(x))
```
vpicheny: Is it important to return values in log space? otherwise I would suggest back-transforming to EI space for consistency with classical EI (here and below). EI value could be used for debugging (positivity) and to track down convergence. Alternatively add `Log` to the function name...

trieste/acquisition/function.py
```diff
@@ -676,3 +676,243 @@ def batch_ei(at: TensorType) -> TensorType:
...
+        if len(dataset.query_points) == 0:
+            raise ValueError("Dataset must be populated.")
+
+        if (
+            pending_points is None
+        ):  # only compute penalization parameters once per optimization step
+            samples = self._search_space.sample(num_samples=self._grid_size)
+            samples = tf.concat([dataset.query_points, samples], 0)
+
+            def get_lipschitz_estimate(
```
vpicheny, line 904: slightly confusing that this is here and not part of the `PenalizationFunction`. Maybe it's not possible, but this is what would makes sense (e.g. we could use a different penalty that doesn't need the Lischitz constant)

henrymoss: I guess it could go in the penalization function. I'll see what that looks like

vpicheny (approved):
it can be merged as it is, however my two main remarks are:

- maybe the `Greedy` abstraction could be used in a different context (asynchronous batches), could be worth thinking about it before merging
- having an abstraction for the `PenalizationFunction` but computing the Lipschitz constant anyway within the LP rule isn't really logical (but I could live with that)

PR #231, opened 2021-04-22
Fix docs build and reduce warnings

### LINKED ISSUE #36: Add doc build to CI
As a developer, I want the doc build to be run on CI, so that I know my changes haven't broken the doc build.

This is very closely related to #37, but also includes docstrings and building the website

### LINKED ISSUE #228: Fix docs build issues and improve testing
**Describe the bug**
The docs build is currently failing to generate all the tutorials. This hasn't been caught since we only deploy the docs when we do a tagged release, and the notebook tests that are run each time are rudimentary and aren't failing.

**To reproduce**
1. Check out latest develop
2. Follow instructions to build docs
3. Check _build/html/tutorials.html file and compare to online docs

This is a regression, as running the same steps on the v0.4.0 tag works fine.

**Comments**
As part of the fix, we should make sure that there are tests to prevent this happening again.

uri-granta:
The docs build was temporarily broken due to a nbsphinx bug. This PR:

- updates nbsphinx to a version without the bug (see https://github.com/spatialaudio/nbsphinx/issues/549) [in fact this was done independently by another commit, so doesn't show up in the diff]
- fixes the remaining sphinx build warnings
- updates the docs build to error on warnings
- creates a new tox job for building the docs (and updates the documentation)
- creates a new github quality check job to run this job

Note that I may need to separately update the repository config to mark the new check as required for merges to develop.


tests/constraints.txt
```diff
@@ -7,17 +7,18 @@ chardet==4.0.0
...
+Deprecated==1.2.12
 dm-tree==0.1.6
 flatbuffers==1.12
 gast==0.3.3
-google-auth==1.28.1
+google-auth==1.29.0
 google-auth-oauthlib==0.4.4
 google-pasta==0.2.0
-gpflow==2.1.4
+gpflow==2.1.5
```
hstojic, line 17: I think upgrading GPflow should be left to some other PR, probably it won't break any of the code, but just in case...

tests/constraints.txt
```diff
@@ -39,11 +40,12 @@ rsa==4.7.2
...
 six==1.15.0
 tabulate==0.8.9
-tensorboard==2.4.1
+tensorboard==2.5.0
+tensorboard-data-server==0.6.0
 tensorboard-plugin-wit==1.8.0
 tensorflow==2.4.1
 tensorflow-estimator==2.4.0
-tensorflow-probability==0.12.1
+tensorflow-probability==0.12.2
```
hstojic, line 48: same as GPflow

hstojic (commented):
Looks good so far... why some notebooks didnt appear in manual docs build at the end?

tests/constraints.txt
joelberkeley: bear in mind this is the constraints for CI, not for the library itself, so this number isn't used by users

uri-granta: > Looks good so far... why some notebooks didnt appear in manual docs build at the end?

A change to docutils meant that nbsphinx was automatically ignoring notebook files with lines over some (not very big) limit! Upgrading nbsphinx fixed it.

uri-granta:
> Looks good so far... why some notebooks didnt appear in manual docs build at the end?

A change to docutils meant nbsphinx was automatically ignoring notebook files with lines longer than a (not very big) limit. Upgrading nbsphinx fixed it.

tests/constraints.txt
uri-granta: Nevertheless, it might make sense to pin the versions of key libraries in requirements.txt, at least with ~. Though I'd certainly hope that a minor point revision change to gpflow would never break anything!

joelberkeley: gpflow doesn't use semver. Patch versions are meant to be backwards compatible

docs/notebooks/expected_improvement.pct.py
```diff
@@ -1,3 +1,8 @@
+# ---
```
uri-granta: @joelberkeley You don't happen to know why this isn't working? I'm trying to mark this notebook as orphan (as described in https://nbsphinx.readthedocs.io/en/0.8.3/orphan.html) so we don't get a warning that it's not part of a TOC (it's only included as a doc link in the FAQ).

joelberkeley: no idea. I've never worked with headers for notebooks

tests/constraints.txt
uri-granta: We already pin gpflow to 2.1.x in setup.py, so I would personally be happy with this update. Also I'd be very sad if the tests didn't catch a break.

hstojic: ok, if patches should be backward compatible lets go with the upgrade

hstojic (approved):
all good, tox build successful, no warnings shown and all notebooks seem to be there

johnamcleod (approved):
Looks good to me!

PR #233, opened 2021-04-23
added workflow_dispatch to be able to manually trigger a GH action

### LINKED ISSUE #232: allow manual triggering of GH actions
Manual triggering of GH actions and choosing a branch to trigger it on is often useful, there is a way to do it by adding `workflow_dispatch`, a simple change.

See, for example: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

hstojic:
This resolves #232, simple addition of a `workflow_dispatch` event to GH actions will allow us to manually trigger actions as well - often useful when testing branches.


.github/workflows/deploy.yaml
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 on:
+  workflow_dispatch:
```
uri-granta: While it's useful to be able to run the quality checks on an arbitrary branch, I don't think we want to be able to publish the docs from one!

.github/workflows/quality-checks.yaml
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 on:
+  workflow_dispatch:
```
uri-granta, line 16: Have you tested that this works or does it need to be merged before the Run Worfklow dropdown appears in the Actions tab?

.github/workflows/deploy.yaml
hstojic: ah, true, I automatically added it to both workflows, without really thinking what the workflow is :)

.github/workflows/quality-checks.yaml
hstojic: I didn't test it here, but I used it for Bellman and it works without issues like this,,,

uri-granta: Happy for you to merge and find out (once you've removed the change to deploy.yaml)

hstojic:
removed it from the deploy workflow...ready for the review!

uri-granta (approved):
LGTM. Make sure to test it after the merge!

PR #234, opened 2021-04-26
Support arbitrary tensor-likes not just lists

### LINKED ISSUE #212: Support tuples as bounds in Box constructor
**Describe the bug**
When I use a `tuple` to define the bounds of the `Box`,  i got `tf.float32` instead of `tf.float64` tensor when sampling from it.

**To reproduce**
1. define the bound of `Box` using `tuple` instead of `list`
2. call `.sample(number_of_sample)`

Include a minimal reproducible code example if relevant.
```
import tensorflow as tf
import trieste as tr

bound_list = tr.space.Box([0., 0.], [1., 1.])
bound_tuple = tr.space.Box((0., 0.), (1., 1.))

tf.debugging.assert_type(bound_tuple.sample(5),bound_list.sample(5).dtype)
```
**Expected behaviour**
I expect the tensor type to be `tf.float64`, or at least there is a defensive code to inform the user that the bound should be `list` instead of `tuple`.

**System information**
 - OS: Windows 10
 - Python version: 3.8.6
 - Trieste version: 0.4.0
 - TensorFlow version: 2.4.0
 - GPflow version: 2.1.4

**Additional context**
The fix could be defining the `dtype=tf.float64` when calling `tf.convert_to_tensor` [here](https://github.com/secondmind-labs/trieste/blob/5db2b0af0cc9ff6fd4b1f4602fcad526c6e61be7/trieste/space.py#L202) or write defensive code to only accept `list` as an input.

uri-granta:
Update Box constructor and mk_dataset utility function to support arbitary Tensor-like inputs (e.g. tuples), not just lists.

### DIFF
--- a/tests/unit/acquisition/test_function.py
+++ b/tests/unit/acquisition/test_function.py
@@ -608,7 +608,7 @@ def test_batch_monte_carlo_expected_improvement_raises_for_empty_data() -> None:
 
 def test_batch_monte_carlo_expected_improvement_raises_for_model_with_wrong_event_shape() -> None:
     builder = BatchMonteCarloExpectedImprovement(100)
-    data = mk_dataset([[0.0, 0.0]], [[0.0, 0.0]])
+    data = mk_dataset([(0.0, 0.0)], [(0.0, 0.0)])
     matern52 = tfp.math.psd_kernels.MaternFiveHalves(
         amplitude=tf.cast(2.3, tf.float64), length_scale=tf.cast(0.5, tf.float64)
     )
--- a/tests/unit/test_pareto.py
+++ b/tests/unit/test_pareto.py
@@ -17,7 +17,7 @@ import numpy.testing as npt
 import pytest
 import tensorflow as tf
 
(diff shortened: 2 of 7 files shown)


henrymoss:
@uri-granta  why do we want this?

uri-granta:
> @uri-granta why do we want this?

There are a couple of instances where we allow users to specify tensor-like inputs using normal Python lists: e.g. [[0,0], [1,1]]. This generalises them to support other tensor-like formats, such as tuples: e.g. ((0,0), (1,1)). This is something tensorflow already supports internally, and is generally the more Pythonic way to do things. We do this using the Sequence abstract base class, which matches anything that behaves like a sequence (https://docs.python.org/3/glossary.html#term-sequence).

hstojic (approved):
To me it seems like a sensible change, and its backward compatible - looks good.

While we are on the topic of tuples, how come `tuple` instead of `Tuple` from typing is used in some places, e.g. in `models/model_interfaces`?

PR #240, opened 2021-04-29
Help with debugging docs build errors

uri-granta:
Since the API documentation is generated from inline Python docstrings, the line numbers in any API error logs refer to generated reST files, not the original Python source files! This PR:

1. Ensures we keep these generated reST files around
2. Adds a comment to the docs README explaining this.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -9,6 +9,16 @@ $ tox -e docs
 
 Open `docs/_build/html/index.html` in a browser to view the docs.
 
+### Fixing documentation build errors
+
+Any errors in the notebooks or API documentation markup will result in documentation
+build errors that must be fixed. To fix these, ensure that all new documentation is valid [reST markup](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html).
+
+Note that since the API documentation is generated from inline Python docstrings, 
+**the line numbers in any API error logs refer to generated reST files,
+not the original Python source files**. You can find these generated reST files
+at `docs/autoapi`.
+
 # License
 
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -9,6 +9,16 @@ $ tox -e docs
...
 
+### Fixing documentation build errors
+
+Any errors in the notebooks or API documentation markup will result in documentation
+build errors that must be fixed. To fix these, ensure that all new documentation is valid [reST markup](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html).
+
+Note that since the API documentation is generated from inline Python docstrings, 
+**the line numbers in any API error logs refer to generated reST files,
+not the original Python source files**. You can find these generated reST files
+at `docs/autoapi`.
```
henrymoss, line 20: No way of getting them to line up with the Python source? That would be more useful?

henrymoss (approved):

docs/README.md
uri-granta: Doesn't look like it. The sphinx-autoapi plugin converts the python scripts to documentation files, which sphinx then compiles. So when sphinx generates the error messages, it's never seen the original files.

TsingQAQ: Stupid question: does this mean one have to debug the docs locally with a Linux OS?

uri-granta: You can still debug using the error messages on the github PR, but the line numbers will be wrong (and sadly the sphinx errors are often not clear enough to make it obvious what the problem is). Running it locally should be much easier now though, as it only requires running tox, just like running the tests.

PR #241, opened 2021-05-04
Extra synthetic benchmarks + tidy

henrymoss:
I have added two extra synthetic benchmarks: a 4-d Rosenbrock and 5-d Ackley function.

As well as adding tests for this new functions, I also combined some existing tests using parametrize and moved the tests of the multi-objective synthetic functions into a separate file (like we did with MO code).

### DIFF
--- a/trieste/utils/objectives.py
+++ b/trieste/utils/objectives.py
@@ -164,6 +164,85 @@ float64.
 """
 
 
+def rosenbrock_4(x: TensorType) -> TensorType:
+    """
+    The Rosenbrock function, rescaled to have zero mean and unit variance over :math:`[0, 1]^4. See
+    :cite:`Picheny2013` for details.
+
+    This function (also known as the Banana function) is unimodal, however the minima
+    lies in a narrow valley.
+
+    :param x: The points at which to evaluate the function, with shape [..., 4].
+    :return: The function values at ``x``, with shape [..., 1].
+    :raise ValueError (or InvalidArgumentError): If ``x`` has an invalid shape.
+    """
+    tf.debugging.assert_shapes([(x, (..., 4))])
+
+    x = 15.0 * x - 5
+    unscaled_function = tf.reduce_sum(
+        (100.0 * (x[..., 1:] - x[..., :-1]) ** 2 + (1 - x[..., :-1]) ** 2), axis=-1, keepdims=True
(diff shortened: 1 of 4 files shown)


trieste/utils/objectives.py
```diff
@@ -164,6 +164,85 @@ def hartmann_3(x: TensorType) -> TensorType:
...
+    This function (also known as the Banana function) is unimodal, however the minima
+    lies in a narrow valley.
+
+    :param x: The points at which to evaluate the function, with shape [..., 4].
+    :return: The function values at ``x``, with shape [..., 1].
+    :raise ValueError (or InvalidArgumentError): If ``x`` has an invalid shape.
+    """
+    tf.debugging.assert_shapes([(x, (..., 4))])
+
+    x = 15.0 * x - 5
```
uri-granta: For some reason the type checker has decided that after this statement, x is guaranteed to be a basic Python float, which it obviously isn't. I think this might actually a bug with mypy (I've found a minimal example) so I'll raise it with them.

In the meantime, you can either use a different variable with an explicit type annotation:

```suggestion
    y : TensorType = 15.0 * x - 5
```

Or swap the mulitplicands around:

```suggestion
    x = x * 15.0 - 5
```

As an aside, you should perhaps add a comment somewhere about the scaling?

trieste/utils/objectives.py
```diff
@@ -164,6 +164,85 @@ def hartmann_3(x: TensorType) -> TensorType:
...
+
+
+ROSENBROCK_4_MINIMIZER = tf.constant([[0.4, 0.4, 0.4, 0.4]], tf.float64)
+"""
+The global minimizer for the :func:`rosenbrock_4` function, with shape [1, 4] and
+dtype float64.
+"""
+
+
+ROSENBROCK_4_MINIMUM = tf.constant([-1.01917], tf.float64)
```
uri-granta, line 195: Why not have this actually evaluate the function at the minimizer? Wouldn't it be more accurate that way?

uri-granta (approved):
LGTM

trieste/utils/objectives.py
henrymoss: I think the idea is to have this independent from our code. This means that the tests can only pass if our implementation is working.

PR #242, opened 2021-05-04
Added a random search acquisition function optimizer

henrymoss:
Added a random search acquisition function optimizer. 

This is a simple (but often effective) optimizer that we are currently missing and is useful for benchmarking.

Random search can be applied to any search space with a sample method and so will be useful for when we start to support a wider range of search space types. At the moment, each of our optimizers only work for a single search space type.

### DIFF
--- a/trieste/acquisition/optimizer.py
+++ b/trieste/acquisition/optimizer.py
@@ -130,7 +130,8 @@ def batchify(
     :param batch_size: The number of points in the batch.
     :return: An :const:`AcquisitionOptimizer` that will provide a batch of points with shape [B, D].
     """
-    tf.debugging.assert_positive(batch_size)
+    if batch_size <= 0:
+        raise ValueError(f"batch_size must be positive, got {batch_size}")
 
     def optimizer(search_space: SP, f: AcquisitionFunction) -> TensorType:
         expanded_search_space = search_space ** batch_size  # points have shape [B * D]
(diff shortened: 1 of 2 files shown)


trieste/acquisition/optimizer.py
```diff
@@ -113,6 +117,35 @@ def _objective() -> TensorType:
     return bijector.forward(variable)  # [1, D]
 
 
+def optimize_random(space: SearchSpace, target_func: AcquisitionFunction) -> TensorType:
+    """
+    A search across 10,000 random points :const:`AcquisitionOptimizer` defined for
+    any :class:'SearchSpace' with a :method:`sample` and batches of size of 1.
```
uri-granta: To reference a specific method I think it's :meth:, not :method:, but here I think just writing method the word would be fine.

trieste/acquisition/optimizer.py
```diff
@@ -113,6 +117,35 @@ def _objective() -> TensorType:
...
+    If we have a :class:'DiscreteSearchSpace' with fewer than 10,000 points,
+    then we query all the points in the space.
+
+    :param space: The space over which to search.
+    :param target_func: The function to maximise, with input shape [..., 1, D] and output shape
+        [..., 1].
+    :return: The **one** point in ``space`` that maximises ``target_func``, with shape [1, D].
+    """
+    num_samples = 10000
```
uri-granta: (maybe?)

```suggestion
    num_samples = 10_000
```

uri-granta (approved):
Looks good, though it wouldn't be much more work to remove the magic 10,000 number:

```python
def generate_random_optimizer(num_samples: int = 10_000) -> AcquisitionOptimizer:
    """ Generate an acquistion optimizer that samples random points in the search space. """

    def optimize_random(space: SearchSpace, target_func: AcquisitionFunction) -> TensorType:
        .... # as before (apart from setting num_samples)

    return optimize_random
````

which you can then use via `target_func=generate_random_optimizer()` or `target_func=generate_random_optimizer(20_000)`.

PR #245, opened 2021-05-06
Augmented Expected Improvement

henrymoss:
Implementation of the AEI acquisition function (a hack of EI to deal with very noisy optimisation tasks).

It is the same as EI but with an additional multiplicative factor that uses the variance of the observation noise (as provided by the model).

To build unit tests, I had to add a likelihood attribute to the dummy test models, i.e. mimicking the gpflow models.

### DIFF
--- a/trieste/acquisition/function.py
+++ b/trieste/acquisition/function.py
@@ -27,7 +27,7 @@ import tensorflow as tf
 import tensorflow_probability as tfp
 
 from ..data import Dataset
-from ..models import ProbabilisticModel, GPflowPredictor
+from ..models import ProbabilisticModel
 from ..space import SearchSpace
 from ..type import TensorType
 from ..utils import DEFAULTS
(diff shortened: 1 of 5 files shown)


trieste/acquisition/function.py
```diff
@@ -27,7 +27,7 @@
 import tensorflow_probability as tfp
 
 from ..data import Dataset
-from ..models import ProbabilisticModel
+from ..models import ProbabilisticModel, GaussianProcessRegression
```
vpicheny: can't see where we need GPR here

trieste/acquisition/function.py
```diff
@@ -156,6 +156,82 @@ def acquisition(x: TensorType) -> TensorType:
...
+    def acquisition(x: TensorType) -> TensorType:
+        tf.debugging.assert_shapes(
+            [(x, [..., 1, None])],
+            message="This acquisition function only supports batch sizes of one.",
+        )
+        mean, variance = model.predict(tf.squeeze(x, -2))
+        normal = tfp.distributions.Normal(mean, tf.sqrt(variance))
+        expected_improvement = (eta - mean) * normal.cdf(eta) + variance * normal.prob(eta)
+
+        noise_variance = model.likelihood.variance
```
vpicheny: to double-check: do all ProbabilisticModels have a likelihood variance?

vpicheny (approved):
very well documented and tested, thanks for the effort!

PR #248, opened 2021-05-07
update setup.py 

henrymoss:
Changed to support the most recent build of GPflow.


uri-granta (approved):
LGTM though you'll need to rerun format (as black seems to have changed)

uri-granta (commented):
Not sure what the typing-extensions issue is (and why it's only affecting the docs build). I've tried rerunning the tests to see if it resolves itself. If not, you could:

1. See whether it fails locally when you run tox -e docs
2. Assuming it does, try regenerating the constraints and see if that fixes it.
3. If not, try pinning typing-extensions to 3.7.4.3 for now and then regenerated and rerun.
4. If that still doesn't work (or beforehand, if you get frustrated), pester me

PR #250, opened 2021-05-10
Added Shekel benchmark

henrymoss:
Added extra benchmark function and tests.

### DIFF
--- a/trieste/utils/objectives.py
+++ b/trieste/utils/objectives.py
@@ -32,7 +32,6 @@ def branin(x: TensorType) -> TensorType:
     """
     The Branin-Hoo function, rescaled to have zero mean and unit variance over :math:`[0, 1]^2`. See
     :cite:`Picheny2013` for details.
-
     :param x: The points at which to evaluate the function, with shape [..., 2].
     :return: The function values at ``x``, with shape [..., 1].
     :raise ValueError (or InvalidArgumentError): If ``x`` has an invalid shape.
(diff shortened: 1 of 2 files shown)


trieste/utils/objectives.py
```diff
@@ -32,7 +32,6 @@ def branin(x: TensorType) -> TensorType:
     """
     The Branin-Hoo function, rescaled to have zero mean and unit variance over :math:`[0, 1]^2`. See
     :cite:`Picheny2013` for details.
-
```
uri-granta: Why? I would expect removing these blank lines would cause a docs build error!

trieste/utils/objectives.py
```diff
@@ -164,14 +160,55 @@ def hartmann_3(x: TensorType) -> TensorType:
 """
 
 
+def shekel_4(x: TensorType) -> TensorType:
+    """
+    The Shekel test function over :math:`[0, 1]^4`. This function has
+    ten local minima and a single global minima.
```
uri-granta:
```suggestion
    ten local minima and a single global minimum.

uri-granta (approved):
Looks good (apart from the unexpected docstring reformatting)

trieste/utils/objectives.py
uri-granta: Apparently no errors. I guess I should really try to understand the docs format at some point rather than just rely on fixing stuff from the warnings!

PR #251, opened 2021-05-11
More flexible Local Penalization.

henrymoss:
I have added functionality to specify the base acquisition function used by local penalization.

LP assumes a strictly positive acquisition function and so we map all base acquisition functions through a softplus transform (as recommended by the original authors).

### DIFF
--- a/tests/unit/acquisition/test_function.py
+++ b/tests/unit/acquisition/test_function.py
@@ -769,7 +769,7 @@ def test_locally_penalized_expected_improvement_raises_when_called_with_invalid_
     base_builder = NegativeLowerConfidenceBound()
     with pytest.raises(ValueError):
         LocalPenalizationAcquisitionFunction(
-            search_space, base_acquisition_function_builder=base_builder
+            search_space, base_acquisition_function_builder=base_builder  # type: ignore
         )
 
 
--- a/tests/unit/acquisition/test_function.py
+++ b/tests/unit/acquisition/test_function.py
@@ -769,7 +769,7 @@ def test_locally_penalized_expected_improvement_raises_when_called_before_initia
     "base_builder", [ExpectedImprovement(), MinValueEntropySearch(Box([0, 0], [1, 1]))]
 )
 def test_locally_penalized_acquisition_matches_base_acquisition(
-    base_builder: AcquisitionFunctionBuilder,
+    base_builder,
 ) -> None:
     data = Dataset(tf.zeros([3, 2], dtype=tf.float64), tf.ones([3, 2], dtype=tf.float64))
     search_space = Box([0, 0], [1, 1])
@@ -801,7 +801,7 @@ def test_locally_penalized_acquisition_matches_base_acquisition(
 )
 def test_locally_penalized_expected_improvement_combines_base_and_penalization_correctly(
     penalizer: Callable[..., PenalizationFunction],
-    base_builder: AcquisitionFunctionBuilder,
+    base_builder,
 ):
     data = Dataset(tf.zeros([3, 2], dtype=tf.float64), tf.ones([3, 2], dtype=tf.float64))
     search_space = Box([0, 0], [1, 1])
--- a/tests/unit/acquisition/test_function.py
+++ b/tests/unit/acquisition/test_function.py
@@ -42,7 +42,7 @@ from trieste.acquisition.function import (
     ExpectedConstrainedImprovement,
     ExpectedHypervolumeImprovement,
     ExpectedImprovement,
-    LocallyPenalizedExpectedImprovement,
+    LocalPenalization,
     MinValueEntropySearch,
     NegativeLowerConfidenceBound,
     NegativePredictiveMean,
@@ -729,7 +729,7 @@ def test_locally_penalized_expected_improvement_builder_raises_for_empty_data() 
     data = Dataset(tf.zeros([0, 1]), tf.ones([0, 1]))
     space = Box([0, 0], [1, 1])
     with pytest.raises(ValueError):
-        LocallyPenalizedExpectedImprovement(search_space=space).prepare_acquisition_function(
+        LocalPenalization(search_space=space).prepare_acquisition_function(
             data, QuadraticMeanAndRBFKernel()
         )
 
@@ -737,7 +737,7 @@ def test_locally_penalized_expected_improvement_builder_raises_for_empty_data() 
 def test_locally_penalized_expected_improvement_builder_raises_for_invalid_num_samples() -> None:
     search_space = Box([0, 0], [1, 1])
     with pytest.raises(ValueError):
-        LocallyPenalizedExpectedImprovement(search_space, num_samples=-5)
+        LocalPenalization(search_space, num_samples=-5)
 
 
 @pytest.mark.parametrize("pending_points", [tf.constant([0.0]), tf.constant([[[0.0], [1.0]]])])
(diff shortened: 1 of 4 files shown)


trieste/acquisition/function.py
```diff
@@ -777,11 +777,16 @@ def __init__(
         search_space: SearchSpace,
         num_samples: int = 500,
         penalizer: Callable[..., PenalizationFunction] = None,
+        base_acquisition_function_builder: AcquisitionFunctionBuilder = None,
```
uri-granta: Given that you only have one model, don't you want a (optional) SingleModelAcquisitionFunctionBuilder?

```suggestion
        base_acquisition_function_builder: Optional[SingleModelAcquisitionFunctionBuilder] = None,
```

trieste/acquisition/function.py
```diff
@@ -790,11 +795,15 @@ def __init__(
 
         if penalizer is None:
             penalizer = soft_local_penalizer
-
         self._lipschitz_penalizer = penalizer
 
+        if base_acquisition_function_builder is None:
+            base_acquisition_function_builder = ExpectedImprovement()
+        self._base_builder = base_acquisition_function_builder
```
uri-granta: You could put these all on one line as follows: (and may be able to avoid having to give a type annotation this way too)

```suggestion
        self._base_builder = ExpectedImprovement() if base_acquisition_function_builder is None else base_acquisition_function_builder
```

trieste/acquisition/function.py
```diff
@@ -790,11 +795,15 @@ def __init__(
...
+
         self._lipschitz_constant = None
         self._eta = None
+        self._base_acquisition_function = None
```
uri-granta: You'll need a type annotation here or mypy will guess wrongly that the type is always None.

```suggestion
        self._base_acquisition_function : Optional[AcquisitionFunction] = None
```

uri-granta (commented):
A few typing comments

tests/unit/acquisition/test_function.py
```diff
@@ -762,59 +764,84 @@ def test_locally_penalized_expected_improvement_raises_when_called_before_initia
...
+) -> None:
     data = Dataset(tf.zeros([3, 2], dtype=tf.float64), tf.ones([3, 2], dtype=tf.float64))
     search_space = Box([0, 0], [1, 1])
     model = QuadraticMeanAndRBFKernel()
 
-    acq_builder = LocallyPenalizedExpectedImprovement(search_space)
-    lp_ei = acq_builder.prepare_acquisition_function(data, model, None)
-    best = acq_builder._eta
-    ei = expected_improvement(model, best)
+    lp_acq_builder = LocallyPenalizedExpectedImprovement(
```
vpicheny: change here and everywhere name to `LocallyPenalizedAcquisitionFunction`?

tests/unit/acquisition/test_function.py
```diff
@@ -762,59 +764,84 @@ def test_locally_penalized_expected_improvement_raises_when_called_before_initia
...
-    npt.assert_array_equal(
-        lp_ei_values, tf.math.exp(tf.math.log(ei_values) + tf.math.log(penal_values))
+    lp_acq_values = lp_acq(xs[..., None, :])
+    base_acq_values = base_acq(xs[..., None, :])
+    penal_values = penalizer(xs[..., None, :])
+    penalized_base_acq = tf.math.exp(
+        tf.math.log(tf.math.softplus(base_acq_values)) + tf.math.log(penal_values)
     )
 
+    if isinstance(base_builder, ExpectedImprovement):
```
vpicheny, line 841: is it worth having a separate case for EI? (naive question)

trieste/acquisition/function.py
```diff
@@ -777,24 +777,33 @@ def __init__(
...
         num_samples: int = 500,
         penalizer: Callable[..., PenalizationFunction] = None,
+        base_acquisition_function_builder: Optional[SingleModelAcquisitionBuilder] = None,
     ):
         """
         :param search_space: The global search space over which the optimisation is defined.
         :param num_samples: Size of the random sample over which the Lipschitz constant
             is estimated. We recommend scaling this with search space dimension.
+        :param penalizer: The chosen penalization method (defaults to soft penalization).
+        :param base_acquisition_function_builder: Base acquisition function to be
```
vpicheny, line 790: do we need to say here or above that it only works with positive acq functions?

trieste/acquisition/function.py
```diff
@@ -839,13 +848,22 @@ def get_lipschitz_estimate(
             self._lipschitz_constant = lipschitz_constant
             self._eta = eta
```
vpicheny, line 861: what's eta in the non-EI case?

henrymoss: eta is needed for the penalizer, regardless of the base acq

tests/unit/acquisition/test_function.py
henrymoss: Mes relies on the random Gumbel sample, so there is a small difference between the different acq builds.

henrymoss:
@uri-granta, @vpicheny  and I chose a different direction with this. We now only support EI and MES (for now) and raise an error if another (potentially negative-valued) acquisition function is fed in. We decided this is preferable that scaling EI and MES, which could result in odd performance.

I also have another odd type issue?

trieste/acquisition/function.py
```diff
@@ -777,24 +778,41 @@ def __init__(
...
         penalizer: Callable[..., PenalizationFunction] = None,
+        base_acquisition_function_builder: Optional[SingleModelAcquisitionBuilder] = None,
     ):
         """
         :param search_space: The global search space over which the optimisation is defined.
         :param num_samples: Size of the random sample over which the Lipschitz constant
             is estimated. We recommend scaling this with search space dimension.
+        :param penalizer: The chosen penalization method (defaults to soft penalization).
+        :param base_acquisition_function_builder: Base acquisition function to be
+            penalized (defaults to expected improvement).
```
uri-granta: If you are going to restrict this to ExpectedImprovement and MinValueEntropySearch, it should say so in the type (make it `Union[ExpectedImprovement, MinValueEntropySearch]` and there should be a comment about why ("strictly positivie") in the param description.

trieste/acquisition/function.py
```diff
@@ -777,24 +778,41 @@ def __init__(
...
             raise ValueError(f"num_samples must be positive, got {num_samples}")
         self._num_samples = num_samples
 
-        if penalizer is None:
-            penalizer = soft_local_penalizer
+        self._lipschitz_penalizer = soft_local_penalizer if penalizer is None else penalizer
 
-        self._lipschitz_penalizer = penalizer
+        if base_acquisition_function_builder is None:
+            self._base_builder = ExpectedImprovement()
```
uri-granta: You'll need a type annotation, so mypy doesn't wrongly assume that _base_builder is always an ExpectedImprovement builder.

```suggestion
            self._base_builder: SingleModelAcquisitionBuilder = ExpectedImprovement()
```

Alternatively you could rewrite/simplify it:

```suggestion
            self._base_builder = base_acquisition_function_builder or ExpectedImprovement()
            if not isinstance(self._base_builder, (ExpectedImprovement, MinValueEntropySearch)):
                ....
```

uri-granta (approved):
LGTM. See comment re typing.

vpicheny (approved):
LGTM

PR #252, opened 2021-05-12
Uri/simplify single objective api

### LINKED ISSUE #55: Convenient API when using one model and dataset
As a user, I want an easy, intuitive, and safe way to use the toolbox when my problem is only for one model and dataset.

At the moment, users must use a single-entry dictionary, which is inconvenient and ugly boilerplate.

We have a number of candidate designs to achieve this. They are listed below. The summaries are a work in progress

uri-granta:
Simplify the API when there's just one model/dataset by:

* Add `.model` and `.dataset` properties on `Record` class for single model/dataset case.
* Add `try_get_final_dataset` and `try_get_final_model` wrappers on `OptimizationResult` class for single model/dataset case.
* Make the tag in `mk_observer` optional.
* Add new `acquire_single` method in `AcquisitionRule` for handling the single model/dataset case. Keeping this separate from `acquire` makes implementing new `AcquisitionRule`s simpler. In fact, users will typically not be exposed to this at all, as it's normally called by the optimization loop.
* Make `optimize` method in `BayesianOptimizer` also handle the single model/dataset.
* Update `AcquisitionRule` initializers to accept both full acquisition builders and single model acquisition builders (so you don't have to write `.using` each time).

I've updated the relevant tutorials to use the new API and show that everything can now be done without referencing dicts and tags. The constraints tutorials still use the tagged approach.

### DIFF
--- a/docs/notebooks/batch_optimization.pct.py
+++ b/docs/notebooks/batch_optimization.pct.py
@@ -75,7 +75,7 @@ from trieste.acquisition import BatchMonteCarloExpectedImprovement
 from trieste.acquisition.rule import EfficientGlobalOptimization
 
 batch_ei_acq = BatchMonteCarloExpectedImprovement(sample_size=1000)
-batch_ei_acq_rule: EfficientGlobalOptimization[Box] = EfficientGlobalOptimization(
+batch_ei_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=batch_ei_acq)
 points_chosen_by_batch_ei, _ = batch_ei_acq_rule.acquire(search_space, initial_data, model)
 
@@ -86,7 +86,7 @@ points_chosen_by_batch_ei, _ = batch_ei_acq_rule.acquire(search_space, initial_d
 from trieste.acquisition import LocalPenalizationAcquisitionFunction
 
 local_penalization_acq = LocalPenalizationAcquisitionFunction(search_space, num_samples=1000)
-local_penalization_acq_rule: EfficientGlobalOptimization[Box] = EfficientGlobalOptimization(
+local_penalization_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=local_penalization_acq)
 points_chosen_by_local_penalization, _ = local_penalization_acq_rule.acquire(
     search_space, initial_data, model)
@@ -140,7 +140,7 @@ cbar.set_label("EI", rotation=270)
 # %%
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
-batch_ei_rule: EfficientGlobalOptimization[Box] = EfficientGlobalOptimization(
+batch_ei_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=3, builder=batch_ei_acq
 )
 qei_result = bo.optimize(10, initial_data, model, acquisition_rule=batch_ei_rule)
@@ -149,7 +149,7 @@ qei_result = bo.optimize(10, initial_data, model, acquisition_rule=batch_ei_rule
 # and then repeat the same optimization with `LocalPenalizationAcquisitionFunction`.
 
 # %%
-local_penalization_rule: EfficientGlobalOptimization[Box] = EfficientGlobalOptimization(
+local_penalization_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=3, builder=local_penalization_acq
 )
 local_penalization_result = bo.optimize(
--- a/docs/notebooks/batch_optimization.pct.py
+++ b/docs/notebooks/batch_optimization.pct.py
@@ -77,7 +77,7 @@ from trieste.acquisition.rule import EfficientGlobalOptimization
 batch_ei_acq = BatchMonteCarloExpectedImprovement(sample_size=1000)
 batch_ei_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=batch_ei_acq)
-points_chosen_by_batch_ei, _ = batch_ei_acq_rule.acquire(search_space, initial_data, model)
+points_chosen_by_batch_ei, _ = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
 
 # %% [markdown]
 # and then do the same with `LocalPenalizationAcquisitionFunction`.
@@ -88,7 +88,7 @@ from trieste.acquisition import LocalPenalizationAcquisitionFunction
 local_penalization_acq = LocalPenalizationAcquisitionFunction(search_space, num_samples=1000)
 local_penalization_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=local_penalization_acq)
-points_chosen_by_local_penalization, _ = local_penalization_acq_rule.acquire(
+points_chosen_by_local_penalization, _ = local_penalization_acq_rule.acquire_single(
     search_space, initial_data, model)
 
 # %% [markdown]
--- a/docs/notebooks/batch_optimization.pct.py
+++ b/docs/notebooks/batch_optimization.pct.py
@@ -1,0 +1,190 @@
+# %% [markdown]
+# # Batch Bayesian Optimization with Batch Expected Improvement and Local Penalization
+
+# %% [markdown]
+# Sometimes it is practically convenient to query several points at a time. This notebook demonstrates two ways to perfom batch Bayesian optimization with `trieste`.
+
+# %%
+import numpy as np
+import tensorflow as tf
+from util.plotting import create_grid, plot_acq_function_2d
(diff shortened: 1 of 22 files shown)


docs/notebooks/batch_optimization.pct.py
```diff
@@ -51,18 +50,16 @@ def build_model(data):
...
+        "optimizer_args": {
+            "minimize_args": {"options": dict(maxiter=100)},
+        },
     }
 
 
-model_spec = build_model(initial_data[OBJECTIVE])
-models = map_values(create_model, model_spec)
+model_spec = build_model(initial_data)
+model = create_model(model_spec)
```
henrymoss, line 62: This makes sense to me!

henrymoss, line 62: This is much nicer

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -123,13 +123,13 @@ def create_bo_model(data):
...
 #
-# We can now run the optimization loop
+# We can now run the optimization loop. Since we have multiple datasets (objective and constraint), we use `optimize_multi` rather than `optimize`.
 
 # %%
 num_steps = 20
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
-data = bo.optimize(
+data = bo.optimize_multi(
```
henrymoss: Is there a way of automatically getting this behaviour, without having to remember to do this? Or perhaps just get intuitive error messages saying to use optimize_mulit?

uri-granta: Done. Note that AcquisitionRule still has both acquire and acquire_multi as combining those would mean every new AcquisitionRule would have to explicitly handle the single model case. I've added an error message to tell people to use acquire_multi if they accidentally pass in a tagged model or dataset.

henrymoss: LGTM

trieste/bayesian_optimizer.py
```diff
@@ -99,6 +129,19 @@ def try_get_final_models(self) -> Mapping[str, TrainableProbabilisticModel]:
         """
         return self.final_result.unwrap().models
 
+    def try_get_final_model(self) -> TrainableProbabilisticModel:
+        """
+        Convenience method to attempt to get the final model for a single-objective run.
+
+        :return: The final model, if the optimization completed successfully.
```
henrymoss: A multi-objective run could still have only a single model? (e..g with the modelstack) Do you mean a single-model run?

docs/notebooks/batch_optimization.pct.py
```diff
@@ -78,9 +75,9 @@ def build_model(data):
 from trieste.acquisition.rule import EfficientGlobalOptimization
 
 batch_ei_acq = BatchMonteCarloExpectedImprovement(sample_size=1000)
-points_chosen_by_batch_ei, _ = EfficientGlobalOptimization(
-    num_query_points=10, builder=batch_ei_acq.using(OBJECTIVE)
-).acquire(search_space, initial_data, models) # type: ignore
+batch_ei_acq_rule: EfficientGlobalOptimization[Box] = EfficientGlobalOptimization(
```
vpicheny: I don't understand what this line does (`[Box]`)

docs/notebooks/batch_optimization.pct.py
```diff
@@ -89,9 +86,10 @@ def build_model(data):
 from trieste.acquisition import LocalPenalizationAcquisitionFunction
 
 local_penalization_acq = LocalPenalizationAcquisitionFunction(search_space, num_samples=1000)
-points_chosen_by_local_penalization, _ = EfficientGlobalOptimization(
-    num_query_points=10, builder=local_penalization_acq.using(OBJECTIVE)
-).acquire(search_space, initial_data, models) # type: ignore
+local_penalization_acq_rule: EfficientGlobalOptimization[Box] = EfficientGlobalOptimization(
```
vpicheny: same here

docs/notebooks/expected_improvement.pct.py
```diff
@@ -28,15 +28,14 @@
 # %% [markdown]
 # ## Sample the observer over the search space
 #
-# Sometimes we don't have direct access to the objective function. We only have an observer that indirectly observes it. In _Trieste_, the observer outputs a number of datasets, each of which must be labelled so the optimization process knows which is which. In our case, we only have one dataset, the objective. We'll use _Trieste_'s default label for single-model setups, `OBJECTIVE`. We can convert a function with `branin`'s signature to a single-output observer using `mk_observer`.
+# Sometimes we don't have direct access to the objective function. We only have an observer that indirectly observes it. In _Trieste_, an observer can output a number of datasets. In our case, we only have one dataset, the objective. We can convert a function with `branin`'s signature to a single-output observer using `mk_observer`.
```
vpicheny, line 31: we need to move this explanation somehow to the constraints notebook - and maybe expand it a bit, and make sure the "how to" questions point to the constraints notebook.

trieste/acquisition/rule.py
```diff
@@ -54,12 +56,12 @@ class AcquisitionRule(ABC, Generic[S, SP_contra]):
     """ The central component of the acquisition API. """
 
     @abstractmethod
-    def acquire(
+    def acquire_multi(
```
vpicheny: if this is the main function, it should be called `acquire`? and we use `acquire_single` (or a better name) for the single model?

docs/notebooks/batch_optimization.pct.py
uri-granta: This is a (hopefully temporary) way to explain to the type checker what the search space looks like, since it can't figure it out by itself. This is why the line previously had a type: ignore comment. We do this elsewhere in the notebooks, though I would prefer to find a way to make this unnecessary.

docs/notebooks/expected_improvement.pct.py
uri-granta: I've just made a first stab at updating the constraints notebook, but will take a look at where else I can document it.

docs/notebooks/batch_optimization.pct.py
vpicheny: since these notebooks (and especially the intro one) audience include researchers, etc., maybe this would introduce too much confusion? while type: ignore is OK...

trieste/acquisition/rule.py
uri-granta: Could do, though since people are more likely to be using the single model version I thought it made more sense this way?

The only reason there are two versions is to avoid forcing people who write new Acqusition Rules to explicitly deal with the single model case each time. This way, they write the more complex `acquire_multi` for their rule, and get the simpler `acquire` for free.

docs/notebooks/batch_optimization.pct.py
uri-granta: Sure. I'll revert (and change the existing cases to type ignore too).

trieste/acquisition/rule.py
vpicheny: Not sure what people would use most, multi or single model case. @henrymoss any opinion on what should be the default?

henrymoss: Do some rules only work foor single models?

uri-granta: The ThompsonSampling rule currently assumes a single model.

henrymoss: Perhaps @vpicheny means that we should still request people write the more complex multi-model acquire as the default, but just name that acquire. With the single-model one coming for free, but named acquire_single?

vpicheny: yes that's what I meant.

vpicheny:
In general, what's important here is that most of the BO loop doesn't care if it's a single model or multiple models. It's always train, acquire, update... so most of the code should work in either case (and we don't have to duplicate anything). It looks like this is what's happening with this PR.

docs/notebooks/failure_ego.pct.py
```diff
@@ -231,7 +233,7 @@ def prepare_acquisition_function(self, dataset, model):
 
 # %%
 fig, ax = plot_gp_2d(
-    result.models[FAILURE].model,  # type: ignore
+    cast(GaussianProcessRegression, result.models[FAILURE]).model,
```
henrymoss: I think type ignores are less scary here? and much more intuitive. @vpicheny ?

trieste/acquisition/rule.py
uri-granta: Following a chat, Henry convinced me that simplifying the naming for people writing new rules is more important than simplifying it for users running single model experiments, as the latter won't typically be exposed to the acquire method anyway since it's called automatically the optimization loop. I've changed it as suggested.

uri-granta:
I think I've made all the suggested changes now. @henrymoss and @vpicheny: if you're happy, then please tick and I'll merge this.

henrymoss (approved):

vpicheny (approved):

PR #253, opened 2021-05-13
More integration tests

henrymoss:
I have added a couple of lightweight extra integration tests.

- We now test MES (worth checking as will be a building block for many future acquisition functions. )
- We now test constrained optimization (our first integration test that checks we are handling multi-model optimization problems correctly.)


uri-granta (approved):
LGTM!

PR #254, opened 2021-05-14
Add predict_y to ProbabilisticModel

### LINKED ISSUE #187: Dedicated API for latent model predictions
As a user, I want to be able to access predictions of actual observations for all models, including any noise contributions. In GPflow this captured by `predict_y`. I also want to be able to access predictions of the latent model where such a model exists. In GPflow this is `predict_f`. I want to be able to do this everywhere I can access a model, such as when querying the model myself, and in defining new acquisition functions. This is so that I can define acquisition functions which need to differentiate between the two.

Q: Do we want to just support latent models in the Bayesian sense, or latent models in a more general sense? Is a "latent model" in the general sense of use if we don't know what kind of latent model it is, and how the observation model is derived from the latent model? Indeed, can we always assume the latent model and observation model always have the same relationship? What happens if we can't?

Q: Do we also need the same for sample?

- Separate `ProbabilisticModel` instances for the latent and observation models. I don't know how we'd pass these around (the observation model would always be available, but the latent would be optional).
- A flag for the mode which changes whether the model is latent or observation. Would this support the case where we want both available? Also, changing the meaning of an entire class based on a flag sounds wrong.

uri-granta:


trieste/models/model_interfaces.py
```diff
@@ -71,6 +71,20 @@ def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
...
+        ``query_points`` for the actual observations, including noise contributions.
+
+        Note that this is not supported by all models.
+
+        :param query_points: The points at which to make predictions, of shape [..., D].
+        :return: The mean and variance of the independent marginal distributions at each point in
+            ``query_points``. For a predictive distribution with event shape E, the mean and
+            variance will both have shape [...] + E.
+        """
+        raise NotImplementedError(f"Model {self!r} does not support predicting actual observations")
```
henrymoss: Model {self!r} does not support predicting actual observations, just the latent function.

trieste/acquisition/sampler.py
```diff
@@ -109,7 +109,11 @@ def sample(self, at: TensorType) -> TensorType:
         """
         tf.debugging.assert_shapes([(at, ["N", None])])
 
-        fmean, fvar = self._model.predict(at)
+        try:
+            fmean, fvar = self._model.predict_y(at)
+        except NotImplementedError:
+            fmean, fvar = self._model.predict(at)
```
henrymoss, line 115: LGTM

trieste/models/model_interfaces.py
```diff
@@ -71,6 +71,23 @@ def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
         """
         raise NotImplementedError
 
+    def predict_y(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
+        """
+        Return the mean and variance of the independent marginal distributions at each point in
+        ``query_points`` for the actual observations, including noise contributions.
```
henrymoss: I dont think you need the word actual, either here or line 87. It jars on me for some reason, maybe its just not standard terminology. Observations are observations...

henrymoss (approved):

PR #255, opened 2021-05-14
Compute the posterior covariance between sets of points for the GPR and SGPR models.

johnamcleod:


trieste/models/model_interfaces.py
```diff
@@ -311,6 +311,43 @@ def update(self, dataset: Dataset) -> None:
...
+
+        Kx2 = self.model.kernel(x, query_points_2)
+        Linv_Kx2 = tf.linalg.triangular_solve(L, Kx2)
+
+        K12 = self.model.kernel(query_points_1, query_points_2)
+        cov = K12 - tf.tensordot(tf.transpose(Linv_Kx1), Linv_Kx2, [[-1], [-2]])
+
+        tf.debugging.assert_shapes(
+            [(query_points_1, ["N", "D"]), (query_points_2, ["M", "D"]), (cov, ["N", "M"])]
+        )
```
henrymoss, line 347: Probably not required. The first two bits are already checked.

trieste/models/model_interfaces.py
```diff
@@ -311,6 +311,43 @@ def update(self, dataset: Dataset) -> None:
...
+        self, query_points_1: TensorType, query_points_2: TensorType
+    ) -> TensorType:
+        """
+        Compute the posterior covariance between sets of query points.
+
+        .. math:: \Sigma_{12} = K_{12} - K_{x1}(K_{xx} + \sigma^2 I)^{-1}K_{x2}
+
+        :param query_points_1: Set of query points with shape [N, D]
+        :param query_points_2: Sets of query points with shape [M, D]
+
```
henrymoss, line 324: What happens if we have additional batch dimensions? i.e X_1 is [A, N, D] and X_2 is [B, M, D] ? I guess this is only really defined for A=B?

trieste/models/model_interfaces.py
```diff
@@ -311,6 +311,43 @@ def update(self, dataset: Dataset) -> None:
...
+        :return: Covariance matrix between the sets of query points with shape [N, M]
+        """
+        tf.debugging.assert_shapes([(query_points_1, ["N", "D"]), (query_points_2, ["M", "D"])])
+
+        x, _ = self.model.data
+        num_data = x.shape[0]
+        s = tf.linalg.diag(tf.fill([num_data], self.model.likelihood.variance))
+
```
henrymoss, line 332: No jitter?

trieste/models/model_interfaces.py
johnamcleod: The GPflow GPR model doesn't support leading batch dimensions, so these tensors need to have shapes `[N, D]` and `[M, D]` respectively.

trieste/models/model_interfaces.py
johnamcleod: The GPflow GPR model doesn't use jitter - the likelihood variance has to be sufficiently large to avoid cholesky errors.

henrymoss (approved):

henrymoss:
This is great! Thanks so much.

PR #257, opened 2021-05-17
Add qEHVI acquisition function

TsingQAQ:
A PR for (MC based) Batch expected  hyper-volume improvement. 

### New Features  
- qEHVI based on [1]'s expression

### Reference
- [1] [Daulton, S., Balandat, M., & Bakshy, E. (2020). Differentiable Expected Hypervolume Improvement for Parallel Multi-Objective Bayesian Optimization. arXiv preprint arXiv:2006.05078.](https://arxiv.org/abs/2006.05078)


TsingQAQ:
Am still a bit not catch how could we extend the test for `q>1`. Tried to gather some idea from test of qEI but seems not very easy to do. The difference is for qEI, the MC can be simply done by computing the improvement through the `max(sampled reward)` (i.e., [these lines of code](https://github.com/secondmind-labs/trieste/blob/f2540a08c262ccc40c129d74eeefb3cda7096439/tests/unit/acquisition/test_function.py#L694-L695)), this is yet not possible for EHVI and one have to still compute the hypervolume of MC samples, in this case, there is nothing different from the calculation of original qEHVI acquisition function. 

Any suggestions on improving test are greatly appreciated.

trieste/acquisition/function.py
```diff
@@ -617,6 +617,118 @@ def ehvi_based_on_partitioned_cell(
...
+            ``jitter`` is negative.
+        """
+        tf.debugging.assert_positive(sample_size)
+        tf.debugging.assert_greater_equal(jitter, 0.0)
+
+        super().__init__()
+
+        self._sample_size = sample_size
+        self._jitter = jitter
+        self.q = -1
```
uri-granta: Is this just batch size? If so, maybe call it `self._batch_size` since I don't think we use the q terminology anywhere else? Also it might be clearer to initialise it to None rather than -1.

```
self._batch_size: Optional[int] = None
```

trieste/acquisition/function.py
```diff
@@ -617,6 +617,118 @@ def ehvi_based_on_partitioned_cell(
...
+    def _cache_q_subset_indices(self, q: int) -> None:
+        """
+        Cache indices containing all subsets of `q`.
+
+        :param q: batch size
+        """
+        if q != self.q:  # regenerate subset indices
+            indices = list(range(q))
+            self.q_subset_indices = {
+                f"q_choose_{i}": tf.constant(list(combinations(indices, i)))
```
uri-granta: It's a bit weird to have a string index like this when you could just use `i`? Also, you don't need to listify indices, as combinations accepts an iterable.

```suggestion
                i: tf.constant(list(combinations(range(q), i)))
```

TsingQAQ: Sorry for the late response and thanks for the comments. Just thinking maybe it helps to clarify why we earlier have the `q_choose_{}` here. 

The string is more for interpretability. For instance, when `i=2`, means we are doing inclusion-exclusion with the case we are only choose 2 point from `q` point, ofc it can be removed or replaced by comment if that is a more proper way :)

henrymoss:
What is the situation with this PR @TsingQAQ ? Is it ready for review?

TsingQAQ:
> What is the situation with this PR @TsingQAQ ? Is it ready for review?

i think it can be reviewed, maybe just be mindful that current test only covered for `q=1` (as a verification of it is able to reproduce analytical `EHVI`), as mentioned [above](https://github.com/secondmind-labs/trieste/pull/257#issuecomment-842292836).

henrymoss:
@TsingQAQ Look at lines 265-362 of https://github.com/pytorch/botorch/blob/master/test/acquisition/multi_objective/test_monte_carlo.py  where they are able to test qEHVI for q>1

tests/integration/test_multi_objective_bayesian_optimization.py
```diff
@@ -75,4 +99,4 @@ def build_stacked_independent_objectives_model(data: Dataset) -> ModelStack:
     ideal_pf = tf.cast(VLMOP2().gen_pareto_optimal_points(100), dtype=tf.float64)
     ideal_hv = Pareto(ideal_pf).hypervolume_indicator(ref_point)
 
-    assert tf.math.log(ideal_hv - obs_hv) < -3.5
+    assert tf.math.log(ideal_hv - obs_hv) < performance_threshold
```
henrymoss: Maybe call it convergence_threshold?

tests/unit/acquisition/test_function.py
```diff
@@ -695,6 +698,212 @@ def test_expected_hypervolume_improvement(
...
+    qehvi_acq = qehvi_builder.prepare_acquisition_function(_model_based_tr_dataset, model)
+    ehvi_acq = expected_hv_improvement(model, _model_based_pareto, _reference_pt)
+
+    test_xs = tf.convert_to_tensor(
+        list(itertools.product(*[list(tf.linspace(-1, 1, data_num_seg_per_dim))] * input_dim)),
+        dtype=training_input.dtype,
+    )  # [test_num, input_dim]
+    test_xs = tf.expand_dims(test_xs, -2)  # add Batch dim: q=1
+
+    npt.assert_allclose(ehvi_acq(test_xs), qehvi_acq(test_xs), rtol=1e-2, atol=1e-2)
```
henrymoss, line 854: Why such low high tolerances?

tests/util/sampler.py
```diff
@@ -0,0 +1,41 @@
...
+        :param jitter: placeholder
+        :return: The samples, of shape `[..., S, B, L]`, where `S` is the `sample_size`, `B` the
+            number of points per batch, and `L` the dimension of the model's predictive
+            distribution.
+        """
+
+        tf.debugging.assert_rank_at_least(at, 2)
+        tf.debugging.assert_greater_equal(jitter, 0.0)
+
+        batch_size = at.shape[-2]
```
henrymoss: Why do we have this?

trieste/acquisition/function.py
```diff
@@ -697,6 +697,115 @@ def ehvi_based_on_partitioned_cell(
...
 
 
+class BatchMonteCarloExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder):
+    """
+    Builder for the batch expected hypervolume improvement acquisition function.
+    The implementation of the acquisition function largely
+    follows :cite:`daulton2020differentiable`
+    """
+
+    def __init__(self, sample_size: int = 512, *, jitter: float = DEFAULTS.JITTER):
```
henrymoss, line 756: what does the * in here mean?

tests/unit/acquisition/test_function.py
TsingQAQ: you mean `rtol` and `atol` looks quite big?

henrymoss: yep

tests/util/sampler.py
TsingQAQ: It is aimed for testing, in the [referred code](https://github.com/pytorch/botorch/blob/master/test/acquisition/multi_objective/test_monte_carlo.py), they test the correctness of the utility function through explicitly specified mc samples from objective distribution.

This class is hence given as one may specify mc samples in this sampler, it will just return those fixed samples for acquisition function calculation, which can be used as a verification whether utility & acquisition function returns expected values.

trieste/acquisition/function.py
TsingQAQ: this follows the convention of [`BatchMonteCarloExpectedImprovement`](https://github.com/secondmind-labs/trieste/blob/28243c5fd3ff3bec89ad3621fb2a5a2c0b34c6bd/trieste/acquisition/function.py#L711), where i assume the intention of `*` is used for forcing `jitter` must be specified only in kwargs manner,  (not something like `BatchMonteCarloExpectedHypervolumeImprovement(1000, 1e-10)`), CMIIW.

tests/unit/acquisition/test_function.py
TsingQAQ: This comes from the original difficulty of doing multi-dimensional mc integration. If we only set an `rtol=0.01`  like [`MC-EI`](https://github.com/secondmind-labs/trieste/blob/28243c5fd3ff3bec89ad3621fb2a5a2c0b34c6bd/tests/unit/acquisition/test_function.py#L738), seems we need an extremely large MC sample number (e.g., `1_000_000` for 2 obj case), which is slow and may cause OOM for test with 3 objectives.

Am also thinking perhaps we could later leverage Loka's customer sampler PR to reduce the variance of MC for instance in `BatchReparametrizationSampler`.

henrymoss: Good idea!

tests/util/sampler.py
henrymoss: What do you think of this @uri-granta ?

trieste/acquisition/function.py
```diff
@@ -697,6 +697,115 @@ def ehvi_based_on_partitioned_cell(
...
+            ``jitter`` is negative.
+        """
+        tf.debugging.assert_positive(sample_size)
+        tf.debugging.assert_greater_equal(jitter, 0.0)
+
+        super().__init__()
+
+        self._sample_size = sample_size
+        self._jitter = jitter
+        self._batch_size: Optional[int] = None
```
henrymoss: whats this for?

trieste/acquisition/function.py
```diff
@@ -697,6 +697,115 @@ def ehvi_based_on_partitioned_cell(
...
+    :param reference_point: The reference point for calculating hypervolume
+    :return: The batch expected hypervolume improvement acquisition
+        function for objective minimisation.
+    """
+
+    def acquisition(at: TensorType) -> TensorType:
+        _batch_size = at.shape[-2]  # B
+
+        def gen_q_subset_indices(q: int) -> list:
+            indices = list(range(q))
```
henrymoss: Say what this function does (inline if possible)

trieste/acquisition/function.py
```diff
@@ -697,6 +697,115 @@ def ehvi_based_on_partitioned_cell(
...
+            return [tf.constant(list(combinations(indices, i))) for i in range(1, q + 1)]
+
+        # [..., S, B, num_obj]
+        samples = sampler.sample(at, jitter=sampler_jitter)
```
henrymoss: keep comments in line

trieste/acquisition/function.py
TsingQAQ: Whoops, seems no need now, will remove.

TsingQAQ:
Maybe one thing to make sure at this stage is, currently the implementation, inherits from `SingleModelAcquisitionBuilder`, utilizes the `batchify` as the optimizer to jointly optimize the acq function. While the sequential greedy approach mentioned in reference may lead to better performance when q and dim is high (ref Fig 2.b of reference [1]). 

Do we need to somehow enable this sequential greedy approach for qehvi as well? And if yes, do we need both optimization strategies provided in a single `BatchMonteCarloExpectedHypervolumeImprovement`? (not sure if this is easily doable atm as the two approach have different acq base class )

tests/integration/test_multi_objective_bayesian_optimization.py
```diff
@@ -31,13 +34,34 @@
 
 @random_seed
 @pytest.mark.parametrize(
-    "num_steps, acquisition_rule",
+    "num_steps, acquisition_rule, convergence_threshold",
     [
-        (20, EfficientGlobalOptimization(ExpectedHypervolumeImprovement().using(OBJECTIVE))),
+        pytest.param(
```
henrymoss, line 39: What does this pytest.param do?

tests/util/sampler.py
```diff
@@ -0,0 +1,41 @@
+from abc import ABC
+
+import tensorflow as tf
+
+from trieste.type import TensorType
+from trieste.utils import DEFAULTS
+
+
+class PseudoBatchReparametrizationSampler(ABC):
+    """A Sampler that takes the specified sample as deterministic samples`."""
```
henrymoss: that *returns* the specified sample

tests/util/sampler.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, samples: TensorType):
+        """
+        :param samples `[S, B, L]`, where `S` is the `sample_size`, `B` the
+            number of points per batch, and `L` the dimension of the model's predictive
+            distribution.
+        """
+        tf.debugging.assert_shapes(
+            [(samples, ["S", "B", "L"])],
+            message="This acquisition function only supports batch sizes of one.",
```
henrymoss: This *sampler*

trieste/acquisition/function.py
```diff
@@ -697,6 +697,113 @@ def ehvi_based_on_partitioned_cell(
...
+class BatchMonteCarloExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder):
+    """
+    Builder for the batch expected hypervolume improvement acquisition function.
+    The implementation of the acquisition function largely
+    follows :cite:`daulton2020differentiable`
+    """
+
+    def __init__(self, sample_size: int = 512, *, jitter: float = DEFAULTS.JITTER):
+        """
+        :param sample_size: The number of samples for each batch of points.
```
henrymoss: samples of what? Be more specific here

trieste/acquisition/function.py
```diff
@@ -697,6 +697,113 @@ def ehvi_based_on_partitioned_cell(
...
+        sampler = BatchReparametrizationSampler(self._sample_size, model)
+
+        return batch_ehvi(sampler, self._jitter, _pf, _reference_pt)
+
+
+def batch_ehvi(
+    sampler, sampler_jitter: float, pareto: Pareto, reference_point: TensorType
+) -> AcquisitionFunction:
+    """
+    :param sampler: The posterior sampler to sample the possible observations @ query points `at`
```
henrymoss: say *at* not @

trieste/acquisition/function.py
```diff
@@ -697,6 +697,113 @@ def ehvi_based_on_partitioned_cell(
...
+    :param sampler_jitter: The size of the jitter to use in sampler
+    :param pareto: Pareto class
```
henrymoss: Be more specific

trieste/acquisition/function.py
```diff
@@ -697,6 +697,113 @@ def ehvi_based_on_partitioned_cell(
...
+
+        q_subset_indices = gen_q_subset_indices(_batch_size)
+
+        hv_contrib = 0.0
+        lb_points, ub_points = pareto.hypercell_bounds(
+            tf.constant([-inf] * samples.shape[-1], dtype=at.dtype), reference_point
+        )
+
+        def hv_contrib_on_samples(obj_samples: TensorType) -> TensorType:
+            # [..., S, Cq_j, j, num_obj] -> [..., S, Cq_j, num_obj]
```
henrymoss, line 837: What is Cq_j? Perhpas this should be defined above where it is introduced?

trieste/acquisition/function.py
```diff
@@ -697,6 +697,113 @@ def ehvi_based_on_partitioned_cell(
...
+            overlap_vertices = tf.reduce_max(obj_samples, axis=-2)
```
henrymoss, line 838: Can you explain what an overlap vertices is?

henrymoss (commented):
Nearly there. Thanks for the extra tests. Do you want to add an integration test aswell? i.e. look at tests/integration/test_multi_objective....

tests/integration/test_multi_objective_bayesian_optimization.py
TsingQAQ: am using it more for an id printing purpose, e.g., `ehvi_vlmop2` printed during testing maybe more interpretable as telling what the test is for than `20-acquisition_rule0--3.65` when having too many tests.

trieste/acquisition/function.py
TsingQAQ: Will add some text, it is the length of `j`th indices of return of `gen_q_subset_indices`

TsingQAQ: Added explanation : `# gather all combinations having j points from q batch points (Cq_j)`

trieste/acquisition/function.py
TsingQAQ: Added an explanation: `# calculate samples overlapped area's hvi for obj_samples`, overlap vertices is the vertices of the overlapped area, hope this has increased some interpretability.

TsingQAQ:
> Nearly there. Thanks for the extra tests. Do you want to add an integration test aswell? i.e. look at tests/integration/test_multi_objective....  

Maybe one weird thing to mention: the integrating test's performance for VLMOP2 performance is getting a little worse after update repo (e.g., swich of optimizer and opt model at beginning...). 
Say log HV difference (smaller the better):
-  -3.49 → -3.445 (for q=2)
- -3.27 → -3.2105 (for q=4)

henrymoss (commented):
Did you add an integration test?

TsingQAQ:
> Did you add an integration test?

i assume that means adding BO tests of `qehvi` on VLMOP2?  We had a q=2 and q=4 test cases. Not sure if i miss-understood the question or some extra experiments are expected.

tests/util/sampler.py
```diff
@@ -0,0 +1,42 @@
+from abc import ABC
+
+import tensorflow as tf
+
+from trieste.type import TensorType
+from trieste.utils import DEFAULTS
+
+
+class PseudoBatchReparametrizationSampler(ABC):
```
uri-granta: Shouldn't this inherit from `trieste.acqusition.Sampler`? 

FYI An upcoming PR (https://github.com/secondmind-labs/trieste/pull/302) is moving the toy objective functions in trieste.utils into a separate directory, so that they can depend on the rest of trieste code without circular dependencies. It may be worth moving this to the same directory, as similarly it's a toy utility for experimentation rather than a something used to implement the rest of trieste.

trieste/acquisition/function.py
```diff
@@ -745,6 +745,121 @@ def ehvi_based_on_partitioned_cell(
...
+                (ub_points[tf.newaxis, tf.newaxis, :, tf.newaxis, :] - overlap_vertices), 0.0
+            )  # [..., S, K, Cq_j, num_obj]
+
+            areas_j = tf.reduce_sum(  # sum over all subsets Cq_j -> [..., S, K]
+                tf.reduce_prod(lengths_j, axis=-1), axis=-1  # calc hvi within each K
+            )
+
+            return tf.reduce_sum(areas_j, axis=-1)  # sum over cells -> [..., S]
+
+        for j in range(1, _batch_size + 1):  # Inclusion-Exclusion loop
```
uri-granta: Doesn't make a difference right now, but at some point we'll update the acquisition function so that it isn't retraced every time (I'm working on the infrastructure necessary for this at the moment).

```suggestion
        for j in tf.range(1, _batch_size + 1):  # Inclusion-Exclusion loop
```

trieste/acquisition/function.py
```diff
@@ -745,6 +745,121 @@ def ehvi_based_on_partitioned_cell(
...
+        _pf = Pareto(mean)
+        _reference_pt = get_reference_point(_pf.front)
+
+        sampler = BatchReparametrizationSampler(self._sample_size, model)
+
+        return batch_ehvi(sampler, self._jitter, _pf, _reference_pt)
+
+
+def batch_ehvi(
+    sampler, sampler_jitter: float, pareto: Pareto, reference_point: TensorType
```
```suggestion
    sampler: Sampler, sampler_jitter: float, pareto: Pareto, reference_point: TensorType
```

uri-granta (approved):
Looks good to me!

tests/util/sampler.py
TsingQAQ: >Shouldn't this inherit from `trieste.acqusition.Sampler`  

hmmm, in that case we may need also passing in a not used pseudo `Probabilistic` model for `PseudoBatchReparametrizationSampler` instantiation,  which is a bit cumbersome to me, is it worth doing this?

> It may be worth moving this to the same directory, as similarly it's a toy utility for experimentation rather than a something used to implement the rest of trieste.    

Am a bit hesitate to do such a move like `objectives`, please correct me if am wrong here: as the current `PseudoBatchReparametrizationSampler` is only expected to be used within tests, unless somewhere else expect this functionality as well, it is safe to stay within tests?

henrymoss:
@TsingQAQ bring it up to date, then we can merge :)

TsingQAQ:
@henrymoss @uri-granta  i am a bit puzzled by the test log arguing ([line 225](https://github.com/secondmind-labs/trieste/runs/3161724434#step:5:225)) in `test_variational_gaussian_process_predict`, which is not expected to be touched by this PR at all. Is there anything am not dealing properly?

tests/util/sampler.py
TsingQAQ: Now it has been modified to inherit from the `BatchReparametrizationSampler`

henrymoss (approved):

henrymoss (approved):

PR #258, opened 2021-05-21
Add mechanism to mark slow running tests

### LINKED ISSUE #256: Run additional tests on develop after merging
As trieste grows, and as we improve our currently limited integration testing, it will no longer be feasible to run all the tests before every commit. Relatedly, we don't want to dissuade people from writing new tests for fear they would slow down development.

One solution is to have a suite of tests that are only run after merging.

* Consistently reproducible tests that take no more than a few hours can be run automatically after each merge to develop, allowing us to link any regressions to the offending commit.
* Inconsistent or very long running tests can be run on a schedule, allowing us to observe gradual regressions or intermittent issues.

uri-granta:
After this PR:

- you can now mark slow running tests with `@pytest.mark.slow`.
- these tests will be skipped by default, but can be run with `pytest --runslow yes`, `tox -e alltests` or `task alltests`
- (alternatively you can use `pytest --runslow only` or `task slowtests` to run just the slow tests)
- running tests via `task` now also shows the slowest running tests
- there is a github workflow to run all the tests (including the slow ones) after a merge to develop or release

TODO

- figure out how failures can be reported (email/status/badge/?). This may require merging first.
- figure out what to do for slow running notebooks. This should probably be in a different PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,11 @@ Run tests with
 $ tox -e tests
 ```
 
+To save time, some slower tests are not run by default. Run these with:
+```bash
+$ tox -e alltests
+```
+
 #### Code formatting
 
 We format all Python code, other than the notebooks, with [black](https://black.readthedocs.io/en/stable/), [flake8](https://flake8.pycqa.org/en/latest/), and [isort](https://pycqa.github.io/isort/). You may need to run these before pushing changes, with (in the repository root)
@@ -89,8 +94,10 @@ pip install -r common_build/taskipy/requirements.txt -c common_build/taskipy/con
 
 You can then use `task` to run various common tasks:
 
-- `task tests` to run all the tests (including type checking);
+- `task tests` to run all the normal tests (including type checking);
 - `task quicktests` to run just the unit tests (starting with the last failure and exiting immediately on any error);
+- `task alltests` to run all the tests including the slow ones;
+- `task slowtests` to run just the slow tests;
 - `task mypy` to run just the type checks;
 - `task format` to reformat the code using black, flake8 and isort;
 - `task check_format` to check whether the code is correctly formatted.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,9 +94,9 @@ pip install -r common_build/taskipy/requirements.txt -c common_build/taskipy/con
 
 You can then use `task` to run various common tasks:
 
-- `task tests` to run all the normal tests (including type checking);
+- `task tests` to run all the tests (including type checking, but excluding slow tests);
 - `task quicktests` to run just the unit tests (starting with the last failure and exiting immediately on any error);
-- `task alltests` to run all the tests including the slow ones;
+- `task alltests` to run all the tests including the slow tests;
 - `task slowtests` to run just the slow tests;
 - `task mypy` to run just the type checks;
 - `task format` to reformat the code using black, flake8 and isort;
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,36 @@
+# Copyright 2021 The Trieste Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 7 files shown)


trieste/acquisition/rule.py
```diff
@@ -200,7 +200,7 @@ def acquire(
         points = self._optimizer(search_space, acquisition_function)
 
         if isinstance(self._builder, GreedyAcquisitionFunctionBuilder):
-            for i in range(
+            for _ in range(
                 self._num_query_points - 1
             ):  # greedily allocate remaining batch elements
                 greedy_acquisition_function = self._builder.prepare_acquisition_function(
```
henrymoss, line 206: Good spot

henrymoss:
Could some info about this be added into our contributing doc?

henrymoss (approved):

CONTRIBUTING.md
```diff
@@ -89,8 +94,10 @@ pip install -r common_build/taskipy/requirements.txt -c common_build/taskipy/con
 
 You can then use `task` to run various common tasks:
 
-- `task tests` to run all the tests (including type checking);
+- `task tests` to run all the normal tests (including type checking);
```
johnamcleod: I am no longer sure what to expect here from the description. This command runs all of the unit and integration tests, along with the type checking, but _not_ some tests which are marked as slow or linting? Perhaps this could say something like `run all tests (including type checking), but excluding slow tests`?

CONTRIBUTING.md
```diff
@@ -89,8 +94,10 @@ pip install -r common_build/taskipy/requirements.txt -c common_build/taskipy/con
...
 - `task quicktests` to run just the unit tests (starting with the last failure and exiting immediately on any error);
+- `task alltests` to run all the tests including the slow ones;
```
johnamcleod:
```suggestion
- `task alltests` to run all the tests including the slow ones (including type checking);
```

CONTRIBUTING.md
```diff
@@ -89,8 +94,10 @@ pip install -r common_build/taskipy/requirements.txt -c common_build/taskipy/con
...
+- `task slowtests` to run just the slow tests;
```
johnamcleod, line 100:
```suggestion
- `task slowtests` to run just the slow tests (including type checking);
```

tests/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+    config.addinivalue_line("markers", "slow: mark test as slow to run")
+
+
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--runslow") == "no":
+        skip_slow = pytest.mark.skip(reason="need --runslow option to run")
+        for item in items:
+            if "slow" in item.keywords:
+                item.add_marker(skip_slow)
+    elif config.getoption("--runslow") == "only":
```
johnamcleod, line 37: I cannot remember if there is any validation of these values. Would it make sense to raise an exception here if the user has set it to something other than yes/no/only?

tox.ini
```diff
@@ -35,6 +35,8 @@ commands =
     format: isort --check .
     tests: pip install . -r tests/requirements.txt -c tests/constraints.txt
     tests: pytest
+    alltests: pip install . -r tests/requirements.txt -c tests/constraints.txt
```
johnamcleod, line 38: I don't know why trieste uses both `tox` and `poetry` commands, but should there be commands here to run the "slow tests" and the "quick tests"? It looks like the other `poetry` commands are reproduced here.

johnamcleod (approved):
Looks good to me! I have added a few comments.

tests/conftest.py
uri-granta: This is already validated by the `choices` argument to addoption. The options appear in `pytest --help` and you get the following error message if you mistype:

```
pytest: error: argument --runslow: invalid choice: 'maybe' (choose from 'yes', 'no', 'only')
```

tox.ini
uri-granta: We're not using poetry at all actually, just taskipy. The taskipy commands are merely intended to make it easier to run things locally, flexibly and quickly, while the tox commands are intended to be more reliable (in that they generate the venv each time) and are what's used by the build checks.

PR #259, opened 2021-05-26
Add custom sampler

### LINKED ISSUE #175: More space-filling sampling method than tf.uniform

In the `Box` class the sampling method is `tf.random.uniform`. I found it less space-filling than for example 
`tf.math.sobol_sample`. This will make the metamodel better approximate the function at the initial phase of optimization. it might be good if the user can choose what kind of sampler to use in the `Box` class. 
as a comparison, botorch had a set of samplers we can select ([link](https://botorch.org/api/sampling.html)) and in the GPflowOpt we have `Design` class ([link](https://github.com/GPflow/GPflowOpt/blob/master/gpflowopt/design.py)) . 

**Additional context**
to give some context what's available there: [link](https://como.ceb.cam.ac.uk/media/preprints/c4e-Preprint-182.pdf)

satrialoka:
Add custom sampler functionality for Trieste's Box (WIP) resolve #175
add design.py to implement the sampler. Currently implemented sampler are:

- Random 
- Sobol sequence
- Halton sequence


henrymoss:
Hi @satrialoka  and @icouckuy, this PR is definitely in the right direction.

henrymoss:
@satrialoka  please could you explain why you chose to have a separate design of experiment class. It feels like a lot of unneeded code. Why is this better than just having some extra methods to the Box search space (e.g. .sample_halton) ?

satrialoka:
Yeah seems it's indeed add unneeded code. My reason to add that is to make users easily implement their own DoE/Sampler method by having a base abstract class for it. Do you think adding function in BOX is enough? If so, I'll be happy to change the current implementation.

icouckuy:
Indeed, we were also thinking about lhd and other designs. If we restrict to Sobol/Halton we can indeed simplify things

henrymoss:
I think its worth just adding the functionality into box for now. :)

henrymoss:
Brill. Please can you add you descriptions in the code to say what these functions do, what inputs/outputs they have and what the inputs do, i.e. following the format of other functions in that class.

You could also add some unit tests, like we have for Box.sample(), e.g. check the size of the output is as expected, check that it only works for valid number of samples, check that we get deterministic samples back for fixed choices of  seed/skip

satrialoka:
for the test, I am just copying what they did in test_box_sampling.
please check, thanks!

docs/notebooks/sampler.pct.py
```diff
@@ -0,0 +1,42 @@
...
+initial_query_points = search_space.sample_halton(num_initial_points)
+plt.scatter(initial_query_points[:, 0], initial_query_points[:, 1])
+
+# %% [markdown]
+# Sample 50 initial points using Sobol sequence
+# %%
+search_space = trieste.space.Box([0., 0.], [1., 1.])
+num_initial_points = 100
+initial_query_points = search_space.sample_sobol(num_initial_points)
+plt.scatter(initial_query_points[:, 0], initial_query_points[:, 1])
```
henrymoss: I don't think its worth having a notebook just for this

trieste/space.py
```diff
@@ -241,11 +247,48 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
         return tf.reduce_all(value >= self._lower) and tf.reduce_all(value <= self._upper)
 
     def sample(self, num_samples: int) -> TensorType:
+        """
+        Sampling points randomly
```
henrymoss: Sample randomly from the space.

trieste/space.py
```diff
@@ -241,11 +247,48 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+        """
         dim = tf.shape(self._lower)[-1]
         return tf.random.uniform(
             (num_samples, dim), minval=self._lower, maxval=self._upper, dtype=self._lower.dtype
         )
 
+    def sample_halton(self, num_samples: int, seed: int = 0) -> TensorType:
+        """
+        Sampling points using randomized halton sequence
+
```
henrymoss: Sample from the space using a Halton sequence. The resulting samples are guaranteed to be diverse and are reproducible by using the same choice of `seed`.

trieste/space.py
```diff
@@ -241,11 +247,48 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+            return []
+        dim = tf.shape(self._lower)[-1]
+        return (self._upper - self._lower) * tfp.mcmc.sample_halton_sequence(
+            dim=dim, num_results=num_samples, dtype=self._lower.dtype, seed=seed
+        ) + self._lower
+
+    def sample_sobol(self, num_samples: int, skip: int = 0) -> TensorType:
+        """
+        Arange points using sobol sample
+
```
henrymoss: Some comment similar to Halton. Can  you also comment why you would use one rather than the other?

tests/unit/test_space.py
```diff
@@ -308,6 +308,20 @@ def test_box_sampling(num_samples: int) -> None:
     _assert_correct_number_of_unique_constrained_samples(num_samples, box, samples)
 
 
+@pytest.mark.parametrize("num_samples", [0, 1, 10])
+def test_box_sobol_sampling(num_samples: int) -> None:
```
henrymoss: test_box_sobol_sampling_returns_correct_shape

tests/unit/test_space.py
```diff
@@ -308,6 +308,20 @@ def test_box_sampling(num_samples: int) -> None:
...
+    box = Box(tf.zeros((3,)), tf.ones((3,)))
+    sobol_samples = box.sample_sobol(num_samples)
+    _assert_correct_number_of_unique_constrained_samples(num_samples, box, sobol_samples)
+
+
+@pytest.mark.parametrize("num_samples", [0, 1, 10])
+def test_box_halton_sampling(num_samples: int) -> None:
```
henrymoss: test_box_halton_sampling returns_correct_shape

trieste/space.py
```diff
@@ -241,11 +247,48 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
         return tf.reduce_all(value >= self._lower) and tf.reduce_all(value <= self._upper)
 
     def sample(self, num_samples: int) -> TensorType:
+        """
+        Sampling points randomly
+
+        :param num_samples: The number of points to sample from this search space.
+        :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
+            from this search space.
```
henrymoss: What shape are we expecting?

trieste/space.py
```diff
@@ -241,11 +247,48 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
         )
 
+    def sample_halton(self, num_samples: int, seed: int = 0) -> TensorType:
+        """
+        Sampling points using randomized halton sequence
+
+        :param num_samples: The number of points to sample from this search space.
+        :param seed: Random seed for the halton sequence
+        :return: ``num_samples`` of points, using halton sequence.
+        """
```
henrymoss: shape?

trieste/space.py
```diff
@@ -241,11 +247,48 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+            dim=dim, num_results=num_samples, dtype=self._lower.dtype, seed=seed
+        ) + self._lower
+
+    def sample_sobol(self, num_samples: int, skip: int = 0) -> TensorType:
+        """
+        Arange points using sobol sample
+
+        :param num_samples: The number of points to sample from this search space.
+        :param skip: The number of initial points of the Sobol sequence to skip
+        :return: ``num_samples`` of points, using sobol sequence.
```
henrymoss: shape?

tests/unit/test_space.py
```diff
@@ -308,6 +308,20 @@ def test_box_sampling(num_samples: int) -> None:
...
+    box = Box(tf.zeros((3,)), tf.ones((3,)))
+    halton_samples = box.sample_halton(num_samples)
+    _assert_correct_number_of_unique_constrained_samples(num_samples, box, halton_samples)
+
+
```
henrymoss, line 324: Please also add two tests to check that halton/sobol return the same sequence when given the same random seed / skip. For example, see line 342 in https://github.com/secondmind-labs/trieste/blob/develop/tests/unit/acquisition/test_sampler.py

trieste/space.py
```diff
@@ -241,11 +247,48 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
         return tf.reduce_all(value >= self._lower) and tf.reduce_all(value <= self._upper)
 
     def sample(self, num_samples: int) -> TensorType:
+        """
+        Sampling points randomly
+
+        :param num_samples: The number of points to sample from this search space.
+        :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
+            from this search space.
+        """
```
henrymoss, line 257: Please add in tf.debugging.assert_positive(num_samples) for all the three sample methods along with tests that the methods fail with invalid sample size. For example see line 38 in https://github.com/secondmind-labs/trieste/blob/develop/tests/unit/acquisition/test_sampler.py

trieste/space.py
```diff
@@ -241,11 +247,48 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+    def sample_halton(self, num_samples: int, seed: int = 0) -> TensorType:
+        """
+        Sampling points using randomized halton sequence
+
+        :param num_samples: The number of points to sample from this search space.
+        :param seed: Random seed for the halton sequence
+        :return: ``num_samples`` of points, using halton sequence.
+        """
+        if num_samples == 0:
+            return []
```
henrymoss: why do we have this behaviour?

docs/notebooks/sampler.pct.py
satrialoka: agree, I'll remove it

trieste/space.py
satrialoka: hmm, not sure. the difference between the two is, Halton here is stochastic, while Sobol that we used here is deterministic. we can also add details about two by referring to some paper like https://zancojournals.su.edu.krd/index.php/JPAS/article/view/1031 but not sure if it is needed here.

tests/unit/test_space.py
satrialoka: Do you mean to test if two methods should return different sequences?

trieste/space.py
satrialoka: we can use 0 as the num_samples in tf.random.uniform and it will return empty tensor. this is not the case for Sobol and Halton, because using 0 num_samples will raise an error. to make things have the same behaviors I add this line.

trieste/space.py
satrialoka: I used .assert_non_negative, since we also accept 0 as num_samples

trieste/space.py
henrymoss: Okay. Please can you write a test to check that this behaviour happens?

tests/unit/test_space.py
henrymoss: test that halton returns the same ouptut for a given random seed and different outputs for different seeds. And also check the same for Sobol.

satrialoka: I see okayy 👌🏼

trieste/space.py
satrialoka: will do!

satrialoka: wait, actually, this test is included in the `test_box_sampling_returns_correct_shape` we have 0 as test param there.

tests/unit/test_space.py
satrialoka: for Halton the seed test doesn't pass, then I checked on the python terminal if the problem is in the tfp and got: 
```
>>> tfp.mcmc.sample_halton_sequence(dim=2,num_results=5,seed=1)
<tf.Tensor: shape=(5, 2), dtype=float32, numpy=
array([[0.30109066, 0.61624485],
       [0.55109066, 0.94957817],
       [0.05109064, 0.06068926],
       [0.92609066, 0.3940226 ],
       [0.42609066, 0.72735596]], dtype=float32)>
>>> tfp.mcmc.sample_halton_sequence(dim=2,num_results=5,seed=1)
<tf.Tensor: shape=(5, 2), dtype=float32, numpy=
array([[0.28878212, 0.07074466],
       [0.5387821 , 0.7374114 ],
       [0.0387821 , 0.5151891 ],
       [0.9137821 , 0.18185577],
       [0.41378212, 0.8485224 ]], dtype=float32)>
```
which is strange. I'll investigate it more

henrymoss: Cheers

satrialoka: as you can see above using the same seed the tfp's halton sample returns the different points. tried to set tf random seed still got the same behavior. do you think there is a bug on the tfp.mcmc.sample_halton_sequence?

henrymoss: I'm not sure. Maybe raise an issue or check with the tfp people?

satrialoka: I'll make an issue on tfp github

satrialoka: okay, so we need to set and reset `tf.random.set_seed()` with the same seed in between calls to make it works.
thanks to @TsingQAQ  who pointed this out to me.

tests/unit/test_space.py
```diff
@@ -302,12 +302,65 @@ def test_box_contains_raises_on_point_of_different_shape(
...
+    sobol_samples_2 = box.sample_sobol(num_samples=100, skip=skip)
+    npt.assert_allclose(sobol_samples_1, sobol_samples_2)
+
+
+@pytest.mark.parametrize("seed", [1, 42, 123])
+def test_box_halton_sampling_returns_same_points_for_same_seed(seed: int) -> None:
+    box = Box(tf.zeros((3,)), tf.ones((3,)))
+    tf.random.set_seed(1)
+    halton_samples_1 = box.sample_halton(num_samples=100, seed=seed)
+    tf.random.set_seed(1)
```
henrymoss: We would want this seed setting to happen in the sampler itself, so that it is reproducible. What do you think @uri-granta ?

henrymoss: See discussion above for context

henrymoss: @satrialoka  please add tf.random.set_seed(seed) into the .sample_halton method

henrymoss: We want to be able to call .sample_halton(num_samples=100, seed=seed) and get exactly the same back

satrialoka: okay

trieste/space.py
```diff
@@ -241,11 +247,55 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
         return tf.reduce_all(value >= self._lower) and tf.reduce_all(value <= self._upper)
 
     def sample(self, num_samples: int) -> TensorType:
+        """
+        Sample randomly from the space.
+
+        :param num_samples: The number of points to sample from this search space.
+        :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
+            from this search space with shape ('N', 'D').
```
henrymoss: with shape '[num_samples, D]' , where `D` is the search space dimension.

trieste/space.py
```diff
@@ -241,11 +247,55 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
         )
 
+    def sample_halton(self, num_samples: int, seed: int = 0) -> TensorType:
+        """
+        Sample from the space using a Halton sequence. The resulting samples are guaranteed to be
+        diverse and are reproducible by using the same choice of ``seed``.
+
+        :param num_samples: The number of points to sample from this search space.
+        :param seed: Random seed for the halton sequence
+        :return: ``num_samples`` of points, using halton sequence with shape ('N', 'D').
```
henrymoss: with shape '[num_samples, D]' , where `D` is the search space dimension.

trieste/space.py
```diff
@@ -241,11 +247,55 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+        ) + self._lower
+
+    def sample_sobol(self, num_samples: int, skip: int = 0) -> TensorType:
+        """
+        Arange points using sobol sample. The resulting samples are deterministic and skipped
+        the first ``skip`` points of the sobol sequence
+
+        :param num_samples: The number of points to sample from this search space.
+        :param skip: The number of initial points of the Sobol sequence to skip
+        :return: ``num_samples`` of points, using sobol sequence with shape ('N', 'D').
```
henrymoss: with shape '[num_samples, D]' , where `D` is the search space dimension.

trieste/space.py
```diff
@@ -241,11 +247,55 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+    def sample_halton(self, num_samples: int, seed: int = 0) -> TensorType:
+        """
+        Sample from the space using a Halton sequence. The resulting samples are guaranteed to be
+        diverse and are reproducible by using the same choice of ``seed``.
+
+        :param num_samples: The number of points to sample from this search space.
+        :param seed: Random seed for the halton sequence
+        :return: ``num_samples`` of points, using halton sequence with shape ('N', 'D').
+        """
+        tf.random.set_seed(seed)
```
henrymoss, line 273: tf.random.set_seed(seed) # to ensure reproducibility

trieste/space.py
```diff
@@ -241,11 +247,55 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+            return []
+        dim = tf.shape(self._lower)[-1]
+        return (self._upper - self._lower) * tfp.mcmc.sample_halton_sequence(
+            dim=dim, num_results=num_samples, dtype=self._lower.dtype, seed=seed
+        ) + self._lower
+
+    def sample_sobol(self, num_samples: int, skip: int = 0) -> TensorType:
+        """
+        Arange points using sobol sample. The resulting samples are deterministic and skipped
+        the first ``skip`` points of the sobol sequence
```
henrymoss: Sample a diverse set from the space using a Sobol sequence. If `skip` is specified, the the resulting samples are reproducible.

trieste/space.py
```diff
@@ -241,11 +247,55 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+        tf.random.set_seed(seed)
+        tf.debugging.assert_non_negative(num_samples)
+        if num_samples == 0:
+            return []
+        dim = tf.shape(self._lower)[-1]
+        return (self._upper - self._lower) * tfp.mcmc.sample_halton_sequence(
+            dim=dim, num_results=num_samples, dtype=self._lower.dtype, seed=seed
+        ) + self._lower
+
+    def sample_sobol(self, num_samples: int, skip: int = 0) -> TensorType:
```
henrymoss: skip: Optional[int]=None

henrymoss: We only want deterministic samples if we specify a skip

henrymoss: but random samples if not

trieste/space.py
```diff
@@ -241,11 +247,55 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+
+        :param num_samples: The number of points to sample from this search space.
+        :param skip: The number of initial points of the Sobol sequence to skip
+        :return: ``num_samples`` of points, using sobol sequence with shape ('N', 'D').
+        """
+        tf.debugging.assert_non_negative(num_samples)
+        if num_samples == 0:
+            return []
+        dim = tf.shape(self._lower)[-1]
+        return (self._upper - self._lower) * tf.math.sobol_sample(
```
henrymoss, line 302: Add in 
```
if skip is None: # generate random skip
    skip = tf.random.uniform([1],maxval=2**32,dtype=tf.int64)[0]
```

trieste/space.py
```diff
@@ -241,11 +247,55 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+            from this search space with shape ('N', 'D').
+        """
+        tf.debugging.assert_non_negative(num_samples)
+
         dim = tf.shape(self._lower)[-1]
         return tf.random.uniform(
             (num_samples, dim), minval=self._lower, maxval=self._upper, dtype=self._lower.dtype
         )
 
+    def sample_halton(self, num_samples: int, seed: int = 0) -> TensorType:
```
henrymoss: seed: Optional[int]=None

trieste/space.py
henrymoss:
```
if seed is not None: # ensure reproducibility
     tf.random.set_seed(seed)
```

trieste/space.py
satrialoka: seems sobol sequence doesn't support `tf.int64`, I changed it to` tf.int32`,

trieste/space.py
```diff
@@ -241,11 +247,62 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+            tf.random.set_seed(seed)
+        dim = tf.shape(self._lower)[-1]
+        return (self._upper - self._lower) * tfp.mcmc.sample_halton_sequence(
+            dim=dim, num_results=num_samples, dtype=self._lower.dtype, seed=seed
+        ) + self._lower
+
+    def sample_sobol(self, num_samples: int, skip: Optional[int] = None) -> TensorType:
+        """
+        Sample a diverse set from the space using a Sobol sequence.
+        If ``skip`` is specified, the the resulting samples are reproducible.
```
henrymoss: then the

henrymoss (commented):
One last comment then we can merge. 

Please can you add two tests to check that halton/sobol return different sequences when we dont specific seed/skip

henrymoss (approved):

trieste/space.py
```diff
@@ -241,11 +247,62 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
+
+        :param num_samples: The number of points to sample from this search space.
+        :param seed: Random seed for the halton sequence
+        :return: ``num_samples`` of points, using halton sequence with shape '[num_samples, D]' ,
+            where D is the search space dimension.
+        """
+
+        tf.debugging.assert_non_negative(num_samples)
+        if num_samples == 0:
+            return []
```
uri-granta: Return a tensor? (and ditto for sobol)

```suggestion
            return tf.constant([])
```

uri-granta (approved):
Looks good to me

trieste/space.py
satrialoka: 👌🏼

PR #263, opened 2021-06-02
Hopefully fix docs deployment

uri-granta:
Looks like https://github.com/secondmind-labs/trieste/pull/240 broke the docs deployment script (which is only run at release time) by keeping the autoapi files around when building and (more importantly) adding autoapi to .gitignore to stop them appearing in git status. Unfortunately, the deployment phase tries to add files at that directory when updating the docs.

This change explicitly adds those files with `-f` to override the gitignore. It also allows manually triggering the docs deployment so we don't need to rerelease. Once the docs are fixed I'll remove that ability.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 on:
+  workflow_dispatch:
   push:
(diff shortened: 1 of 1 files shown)


peter-moffatt (approved):
Looks good!

PR #266, opened 2021-06-09
Random Fourier feature sampler

### LINKED ISSUE #84: Add bibliographic reference for Thompson Sampling
As a user, I want the Thompson sampling acquisition rule to include a reference (or references) to the literature, so that I can learn more about how the algorithm works and get general context.

Here's a conversation about which references are good

> joel: what's the best paper to reference for thompson sampling in BO?
> joel: ... I'm looking for a good paper to reference for TS in the toolbox
> joel: the "de facto" reference if you will
> sattar: This one is the first one with theoretical results: https://arxiv.org/abs/1704.00445
> sattar: For general reference I would suggest:
> sattar: http://proceedings.mlr.press/v84/kandasamy18a/kandasamy18a.pdf
> sattar: https://arxiv.org/abs/1706.01825
> sattar: https://arxiv.org/abs/1704.00445

henrymoss:
This is the first go at a RFF sampler for GPR models. 

I have followed the implementation discussed in Appendix A of https://arxiv.org/pdf/1406.2541.pdf

1. @vdutor , please can you check that this makes sense with respect to our use of the RandomFourierFeatures functionality from gpflux.
2. @johnamcleod , please could you have a look at how I have coded up the linalg. I don't use Cholesky, so am worried about how stable the computations are. 
3.  @vpicheny , please can you check that the sampler is in the right place. After this PR, I will do another PR where we add an acquisition function that maximizes the sampled trajectories, allowing EGO to perform (batch) Thompson sampling.

Once you guys are happy, I will add in some tests.


johnamcleod (commented):
It looks like a bunch of temporary files have been committed.

henrymoss:
@johnamcleod, I tried to update the requirements to add Gpflux and everything has gone crazy!  I will sort this out.
Please just look at acquisition/sampler.py .

trieste/models/model_interfaces.py
```diff
@@ -306,6 +321,14 @@ def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
     def predict_y(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
         return self.model.predict_y(query_points)
 
+    def get_kernel(self) -> gpflow.kernels.Kernel:
```
johnamcleod, line 317: Why is this method necessary? `predictor.model.kernel` seems quite concise, and highlights that the kernel is associated to the underlying model.

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,190 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+
+        self._dataset = dataset
+        self._model = model
+
+    def __repr__(self) -> str:
+        """"""
+        return f"{self.__class__.__name__}({self._model!r}, {self._dataset!r})"
+
+    @abstractmethod
+    def sample(self) -> Callable[[TensorType], TensorType]:
```
johnamcleod: This seems to be an unnecessary level of indirection. This class only has one method, so I would expect that this object is passed around to the place where samples are needed.

I would also expect a method called `sample` to return a sample, not a function for sampling.

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,190 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+    def __init__(self, dataset: Dataset, model: GPflowPredictor, num_features: int = 1000):
+        """
+        :param dataset: The data from the observer. Must be populated.
+        :param model: The model to sample from.
+        :param num_features: The number of features used to approximate the kernel.
+        :raise ValueError: If ``dataset`` is empty.
+        """
+        super().__init__(dataset, model)
+        self._num_features = num_features  # [m]
+        self._num_data = len(self._dataset.query_points)  # [n]
```
johnamcleod: This comments are to connect the variable names to the terms of the equations in the class docstring? The square brackets suggest to me that they are shape annotations.

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,190 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+        the [n,m] matrix of feature evaluations across the training data :math:`\Phi`
+        and observation noise variance :math:`\sigma^2`.
+
+        :return: The posterior distribution for theta.
+        """
+
+        phi = self._feature_functions(self._dataset.query_points)  # [n, m]
+        design_matrix = tf.matmul(phi, phi, transpose_a=True)  # [m, m]
+        s = self._noise_variance * tf.eye(self._num_features, dtype=phi.dtype)
+        inv_design_matrix = tf.linalg.inv(design_matrix + s)
```
johnamcleod: I think it would be better to avoid inverting the matrix. If you get the cholesky factorisation of the design matrix you can use the `cholesky_solve` function to compute the mean and covariance.

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,190 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+
+        theta_posterior_mean = tf.matmul(
+            tf.matmul(inv_design_matrix, phi, transpose_b=True), self._dataset.observations
+        )  # [m, 1]
+        theta_posterior_mean = tf.squeeze(theta_posterior_mean)  # [m]
```
johnamcleod: It might be better to avoid `squeeze` here. If you are expecting the shape to be `[m, 1]`, it would be better to index into the tensor.

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,190 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+        matrix of feature evaluations across the training data :math:`\Phi` and
+        observation noise variance :math:`\sigma^2`.
+
+        :return: The posterior distribution for theta.
+        """
+
+        phi = self._feature_functions(self._dataset.query_points)  # [n, m]
+        gram_matrix = tf.matmul(phi, phi, transpose_b=True)  # [n, n]
+        s = self._noise_variance * tf.eye(self._num_data, dtype=phi.dtype)
+        inv_gram_matrix = tf.linalg.inv(gram_matrix + s)
```
johnamcleod: Again, it might be better to avoid the matrix inverse.

trieste/models/model_interfaces.py
```diff
@@ -98,6 +98,21 @@ def get_observation_noise(self) -> TensorType:
         """
         raise NotImplementedError(f"Model {self!r} does not provide scalar observation noise")
 
+    def covariance_between_points(
```
johnamcleod: Why has this method been added to the interface? It doesn't appear to be used in the PR. If it is not needed here, could the change be deferred until it is needed?

johnamcleod (commented):
Is it possible to add some unit tests for this functionality?

trieste/models/model_interfaces.py
henrymoss: My mistake, Crept in from another branch

trieste/acquisition/sampler.py
henrymoss: I don't understand your first point?

trieste/acquisition/sampler.py
henrymoss: Thanks, this is the sort of wisdom I wanted!

trieste/models/model_interfaces.py
henrymoss: I wanted to stress that only gpflow models have this attribute. More general (non-GP ones) will not

henrymoss: Its just to inform people when they build new GPflow wrappers

johnamcleod: `predictor.model` has type `GPModel` for this class, and "not GP" models will not be able to call `.kernel` on the model. This should be highlighted by any editor which is aware of types, and will be caught by the static checks. I am not convinced of the need for another getter object on this class.

trieste/acquisition/sampler.py
johnamcleod: The flow to get samples is:
`s = Sampler(); sampling_fn = s.sample(); samples = sampling_fn(X)`.
I do not understand why the `sampling_fn` step is needed when it could be:
`s = Sampler(); samples = s.sample(X)`.

Building a smaller object and returning it can make sense when the initial class is very large and has many responsibilities, and it would not make sense to pass the whole thing around. This case is different because the class does nothing else. There is only one method to call on this object.

This would also be a more intuitive behaviour given the method name. Method names on classes can be thought of as the "verbs" that that the object can perform. This "sampler" can "sample", but that produces a "sampling function" not samples.

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,205 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+            ):  # if m < n  then calculate posterior in design space (an m*m matrix inversion)
+                self._theta_posterior = self._prepare_theta_posterior_in_design_space()
+            else:  # if n < m  then calculate posterior in gram space (an n*n matrix inversion)
+                self._theta_posterior = self._prepare_theta_posterior_in_gram_space()
+
+            self._pre_calc = True
+
+        def trajectory(x: TensorType) -> TensorType:
+            feature_evaluations = self._feature_functions(x)  # [N, m]
+            theta_sample = self._theta_posterior.sample(1)  # [1, m]
```
TsingQAQ: A stupid question: is it make sense to move `self._theta_posterior` outside `trajectory`? In that sense, the `trajectory` can be a deterministic trajectory sample, which can be optimized for instance.

henrymoss: You are right! That was always my intension.

henrymoss: Good spot

henrymoss: I will fix this and add a test to check that the trajectories are deterministic

trieste/acquisition/sampler.py
```diff
@@ -19,20 +19,23 @@
...
 from scipy.optimize import bisect
 
+from ..data import Dataset
 from ..models import ProbabilisticModel
 from ..type import TensorType
 from ..utils import DEFAULTS
 
 
-class Sampler(ABC):
+class DiscreteSampler(ABC):
```
uri-granta, line 35: In future we'll have to think about the compatibility issues involved with renaming classes. For now I expect it's ok (though if we think anyone might already be using this, we could always add a `Sampler = DiscreteSampler` alias).

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,206 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+
+        self._dataset = dataset
+        self._model = model
+
+    def __repr__(self) -> str:
+        """"""
+        return f"{self.__class__.__name__}({self._model!r}, {self._dataset!r})"
+
+    @abstractmethod
+    def get_trajectory(self) -> Callable[[TensorType], TensorType]:
```
uri-granta: Suggestion: maybe define `TrajectoryFunction = Callable[[TensorType], TensorType]` at the top and use that?

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,206 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+            theta_posterior_mean, theta_posterior_chol_covariance
+        )
+
+    def get_trajectory(self) -> Callable[[TensorType], TensorType]:
+        """
+        Generate an approximate function draw (trajectory) by sampling weights
+        and evaluating the feature functions.
+
+        If this is the first call, then we calculate the posterior distributions for
+        the feautre weights.
```
uri-granta:
```suggestion
        the feature weights.
```

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,206 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+        if not self._pre_calc:
+            self._feature_functions = RandomFourierFeatures(
+                self._kernel, self._num_features, dtype=self._dataset.query_points.dtype
+            )
+
+            if (
+                self._num_features < self._num_data
+            ):  # if m < n  then calculate posterior in design space (an m*m matrix inversion)
+                self._theta_posterior = self._prepare_theta_posterior_in_design_space()
+            else:  # if n < m  then calculate posterior in gram space (an n*n matrix inversion)
```
uri-granta:
```suggestion
            else:  # if n <= m  then calculate posterior in gram space (an n*n matrix inversion)
```

tests/unit/acquisition/test_sampler.py
```diff
@@ -266,16 +268,113 @@ def test_batch_reparametrization_sampler_sample_raises_for_inconsistent_batch_si
...
+def test_discrete_sampler_reprs(discrete_sampler) -> None:
     assert (
-        repr(sampler(20, QuadraticMeanAndRBFKernel()))
-        == f"{sampler.__name__}(20, QuadraticMeanAndRBFKernel())"
+        repr(discrete_sampler(20, QuadraticMeanAndRBFKernel()))
+        == f"{discrete_sampler.__name__}(20, QuadraticMeanAndRBFKernel())"
     )
+
+
+def test_rff_sampler_repr_includes_class_name() -> None:
```
uri-granta, line 286: If you want to you could put all the rff tests in a class for clarity:

```python
class TestRandomFourierFeatureThompsonSampler:
    def test_rff_sampler_repr_includes_class_name(self) -> None:
        ...
    ...
```

uri-granta (approved):
LGTM

trieste/acquisition/sampler.py
henrymoss: None every uses Sampler, just the subclasses. But I get your point!

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,218 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+
+class ContinuousSampler(ABC):
+    r"""
+    An :class:`ContinuousSampler` samples a specific quantity according
+    to an underlying :class:`ProbabilisticModel`.
+    Unlike our :class:`DiscreteSampler`, :class:`ContinuousSampler` returns a
+    queryable function (a trajectory) that provides the sample's value at a query location.
+    """
+
+    def __init__(self, dataset: Dataset, model: ProbabilisticModel):
```
TsingQAQ, line 305: Is it a good idea that we can also have a `sample_size` arg here, so that sample method(i.e., `get_trajectory`) can return a `list[TrajectoryFunction]`  for instance?

henrymoss: I think its fine just to call this multiple times if you want multiple samples. Otherwise we have to deal with varied sized outputs (i.e. sometimes just a function, sometimes a list of functions).

tests/util/model.py
```diff
@@ -92,8 +92,11 @@ def __init__(
...
     ):
-        kernel = tfp.math.psd_kernels.ExponentiatedQuadratic(kernel_amplitude)
-        super().__init__([lambda x: quadratic(x - x_shift)], [kernel], noise_variance)
+        self.kernel = tfp.math.psd_kernels.ExponentiatedQuadratic(kernel_amplitude)
+        super().__init__([lambda x: quadratic(x - x_shift)], [self.kernel], noise_variance)
 
     def __repr__(self) -> str:
         return "QuadraticMeanAndRBFKernel()"
+
+    def get_kernel(self) -> tfp.math.psd_kernels.PositiveSemidefiniteKernel:
```
vdutor, line 101: Q: why not use `@property`?

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,218 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+
+        try:
+            self._noise_variance = model.get_observation_noise()
+            self._kernel = model.get_kernel()
+        except (NotImplementedError, AttributeError):
+            raise ValueError(
+                """
+            Thompson sampling with random Fourier features only currently supports models
+            with likelihood.variance and kernel attributes.
+            """
```
vdutor: minor: formatting looks off.

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,218 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+            )
+
+        self._pre_calc = False
```
vdutor: minor: Looks like an important (hidden) attribute. Maybe add a brief docstring.

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,218 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+        self._num_data = len(self._dataset.query_points)  # n
+
+        try:
+            self._noise_variance = model.get_observation_noise()
+            self._kernel = model.get_kernel()
+        except (NotImplementedError, AttributeError):
+            raise ValueError(
+                """
+            Thompson sampling with random Fourier features only currently supports models
+            with likelihood.variance and kernel attributes.
```
vdutor: i.e. Gaussian likelihoods?

trieste/acquisition/sampler.py
```diff
@@ -276,3 +279,218 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+    inversion of a gram matrix. As well as being more efficient in early BO
+    steps (where :math:`n<m`), this second computation method allows must larger choices
+    of m (as required to approximate very flexible kernels).
+    """
+
+    def __init__(self, dataset: Dataset, model: ProbabilisticModel, num_features: int = 1000):
+        """
+        :param dataset: The data from the observer. Must be populated.
+        :param model: The model to sample from.
+        :param num_features: The number of features used to approximate the kernel.
```
vdutor: Minor: perhaps mention the default value (1000) and in which settings it is expected to work well. Looks a bit like a magic number now.

trieste/acquisition/sampler.py
johnamcleod: The name of the type has changed though, so type annotations would need to be updated. Projects which consume `trieste` may also have added their own custom subclasses which would break.

henrymoss: Understood :) I think more carefully before making these sort of changes in the future!

vdutor (approved):
Looks great to me. Briefly double-checked the maths as well and they look sound too.

PR #268, opened 2021-06-09
Add gpflux dependency

uri-granta:
Add gpflux dependency to trieste (but don't actually use it yet).

Doing this revealed a weakness in the constraint file handling which will be addressed in a future PR.

It also revealed a possible issue with the latest version of black, whereby it no longer excludes .tox files by default. This is being worked around for the moment.


henrymoss:
Do we care about the format changes to some files?

uri-granta:
> Do we care about the format changes to some files?

The docstring changes are due to the new version of black. They're good changes in that they better correspond to https://www.python.org/dev/peps/pep-0257/ so I don't see any reason to prevent the upgrade. The only issue people might have is when running black locally they'll need to upgrade to the latest version via `pip install -r common_build/format/requirements.txt -c common_build/format/constraints.txt` but I can post a message to Slack to tell people that.

henrymoss (approved):

PR #271, opened 2021-06-16
Avoid retracing acquisition functions

uri-granta:
This PR avoids retracing acquisition functions by updating them rather than generating them afresh each optimization loop, and compiling them with tf.function. For simplicity, this is made optional and backwards compatible: users can choose whether to implement the update methods or not (at the cost of performance if they don't).

Note that updating the model wrappers to be compatible with being updated involves adding some tf.Variables that slow down the non-sparse GPR and VGP models. In some cases this slowdown can outweigh the speedup from not recompiling the acquisition function. If this becomes a real issue in the future we can add more architecture to support models that don't allow AF updates.

TIP: when reviewing this, add ?w=1 to the URL to ignore whitespace changes.

**Still left to do (in a separate PR)**

- update all the acquisition functions:
- [x] ExpectedImprovement
- [x] AugmentedExpectedImprovement
- [x] MinValueEntropySearch
- [ ] NegativeLowerConfidenceBound
- [ ] ProbabilityOfFeasibility
- [ ] ExpectedConstrainedImprovement
- [ ] ExpectedHypervolumeImprovement
- [ ] BatchMonteCarloExpectedImprovement.
- [x] LocalPenalizationAcquisitionFunction
- update all the penalizers too:
- [x] soft_local_penalizer
- [ ] hard_local_penalizer


trieste/acquisition/rule.py
```diff
@@ -195,18 +197,25 @@ def acquire(
...
         :return: The single (or batch of) points to query, and `None`.
         """
+        if self._acquisition_function is None:
+            self._acquisition_function = self._builder.prepare_acquisition_function(
+                datasets, models
+            )
+        else:
+            self._acquisition_function = self._builder.update_acquisition_function(
+                self._acquisition_function, datasets, models
+            )
```
henrymoss, line 207: Very nice!

trieste/acquisition/rule.py
```diff
@@ -195,18 +197,25 @@ def acquire(
...
-        points = self._optimizer(search_space, acquisition_function)
+        points = self._optimizer(search_space, self._acquisition_function)
 
         if isinstance(self._builder, GreedyAcquisitionFunctionBuilder):
             for _ in range(
                 self._num_query_points - 1
             ):  # greedily allocate remaining batch elements
-                greedy_acquisition_function = self._builder.prepare_acquisition_function(
-                    datasets, models, pending_points=points
+                self._acquisition_function = self._builder.update_acquisition_function(
```
henrymoss, line 215: Having an acq as an attribute of rule will be helpful for debugging aswell, e.g. plotting acq functions evolving and the like

trieste/acquisition/sampler.py
```diff
@@ -237,7 +237,7 @@ def sample(self, at: TensorType) -> TensorType:
 
         if tf.size(self._eps) == 0:
             self._eps.assign(
-                tf.random.normal([self._sample_size, mean.shape[-1]], dtype=tf.float64)
+                tf.random.normal([self._sample_size, tf.shape(mean)[-1]], dtype=tf.float64)
             )  # [S, L]
```
henrymoss, line 241: Whats this for? Just for the tensroflow retracing?

trieste/models/model_interfaces.py
```diff
@@ -357,7 +368,7 @@ def model(self) -> GPR | SGPR:
         return self._model
 
     def update(self, dataset: Dataset) -> None:
-        x, y = self.model.data
+        x, y = map(lambda var: var.value(), self.model.data)
```
henrymoss: what happens if you keep using these as variables?

trieste/acquisition/sampler.py
uri-granta: Yes. Compiling the graph means the variable `mean` only knows the statically determined shape (which may have None in it). You have to use `tf.shape` to get the dynamic shape at each execution.

trieste/models/model_interfaces.py
uri-granta: What do you mean?

trieste/acquisition/sampler.py
henrymoss: makes sense and sounds familiar

trieste/models/model_interfaces.py
henrymoss: does .value mean that the variable just becomes a constant?

uri-granta: I think that value() is needed to get the shape a few lines further down. Also to for the type checking on the DataSet initialisation to pass. Either way, it's fine to be explicit here.

trieste/bayesian_optimizer.py
```diff
@@ -310,13 +310,11 @@ def optimize(
...
 
         for step in range(num_steps):
-            if track_state:
-                history.append(Record(datasets, models, acquisition_state))
-
             try:
                 if track_state:
-                    models = copy.deepcopy(models)
-                    acquisition_state = copy.deepcopy(acquisition_state)
+                    models_copy = copy.deepcopy(models)
```
johnamcleod, line 323: I assume this change is necessary because the `deepcopy`d compiled models no longer work properly. I presume the `Record` saves the copy of the model so it can be queried later? Does that version of the model work properly?

trieste/models/model_interfaces.py
```diff
@@ -384,7 +396,7 @@ def covariance_between_points(
         """
         tf.debugging.assert_shapes([(query_points_1, ["N", "D"]), (query_points_2, ["M", "D"])])
 
-        x, _ = self.model.data
+        x, _ = map(lambda var: var.value(), self.model.data)
```
johnamcleod: This looks like it is doing an extra conversion which is not necessary: the `y` data. Could this be replaced with `self.model.data[0].value()` or something?

trieste/acquisition/function.py
```diff
@@ -55,18 +55,38 @@
...
 class AcquisitionFunctionBuilder(ABC):
-    """An :class:`AcquisitionFunctionBuilder` builds an acquisition function."""
+    """An :class:`AcquisitionFunctionBuilder` builds and updates an acquisition function."""
 
     @abstractmethod
     def prepare_acquisition_function(
         self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
     ) -> AcquisitionFunction:
         """
+        Build an acquisition function.
```
johnamcleod: Nit: The function name is "prepare" while the docstring say "build". Should these be consistent?

trieste/acquisition/function.py
```diff
@@ -55,18 +55,38 @@
...
         """
 
+    def update_acquisition_function(
+        self,
+        function: AcquisitionFunction,
+        datasets: Mapping[str, Dataset],
+        models: Mapping[str, ProbabilisticModel],
+    ) -> AcquisitionFunction:
+        """
+        Update an acquisition function. To avoid retracing every optimization loop, use this
```
johnamcleod: I wonder if this docstring should be expanded. 

- What is the scope of the update? 
- Would it be worth mentioning that retracing is specific to combining this with `tf.function`?
- It seems here that "use" is meant to be understood as "override in a subclass" rather than "call"?

trieste/acquisition/function.py
```diff
@@ -131,35 +172,58 @@ def prepare_acquisition_function(
...
 
-def expected_improvement(model: ProbabilisticModel, eta: TensorType) -> AcquisitionFunction:
-    r"""
-    Return the Expected Improvement (EI) acquisition function for single-objective global
-    optimization. Improvement is with respect to the current "best" observation ``eta``, where an
-    improvement moves towards the objective function's minimum, and the expectation is calculated
-    with respect to the ``model`` posterior. For model posterior :math:`f`, this is
 
-    .. math:: x \mapsto \mathbb E \left[ \max (\eta - f(x), 0) \right]
+class expected_improvement:
```
johnamcleod: It looks like the `ExpectedImprovement` and `AugmentedExpectedImprovement` could be a single class parameterised by the lowercase versions. Is there a reason to keep them separate?

uri-granta: Agree, but will leave this to a separate PR.

trieste/bayesian_optimizer.py
uri-granta: Yes, but I'll add a test to show this.

trieste/models/model_interfaces.py
```diff
@@ -537,6 +590,26 @@ def __init__(self, model: VGP, optimizer: Optimizer | None = None):
...
         self._model = model
 
+        # GPflow stores num_data as a number. However, since we want to be able to update it
+        # without having to retrace the acquisition functions, put it in a Variable instead.
+        # So that the elbo method doesn't fail we also need to turn it into a property.
+        if not isinstance(self._model, VGPWrapper):
+            self._model._num_data = tf.Variable(
+                model.num_data or 0, trainable=False, dtype=tf.float64
+            )
+            self._model.__class__ = VGPWrapper
```
henrymoss, line 655: Whats this doing?

uri-granta: This turns num_data into a property backed by a tf.variable, since the gpflow elbo method expects to be able to access it as such. Just turning num_data into a variable fails for this reason.

tests/unit/test_bayesian_optimizer.py
```diff
@@ -259,7 +259,7 @@ def __deepcopy__(self, memo: dict[int, object]) -> _UncopyableModel:
     with pytest.raises(_Whoops):
         result.unwrap()
 
-    assert len(history) == 4
+    assert len(history) == 3
```
henrymoss, line 262: Why has this changed?

henrymoss (approved):

tests/unit/test_bayesian_optimizer.py
uri-granta: As discussed: this is because we now store the model copy in the history rather than the original model. This test uses a model that can only be copied 3 times, meaning we now only store it three times.

trieste/models/model_interfaces.py
```diff
@@ -499,6 +511,32 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
         multiple_assign(self.model, current_best_parameters)
 
 
+class NumDataPropertyMixin:
+    """Mixin class for exposing num_data as a property, stored in a tf.Variable. This is to work
```
henrymoss, line 515: What does Mixin mean?

trieste/models/model_interfaces.py
```diff
@@ -499,6 +511,32 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+    @num_data.setter
+    def num_data(self, value: TensorType) -> None:
+        self._num_data.assign(value)
+
+
+class SVGPWrapper(SVGP, NumDataPropertyMixin):
+    """A wrapper around GPFlow's SVGP class that stores num_data in a tf.Variable and exposes
+    it as a property."""
+
+
```
henrymoss, line 539: Do we really need this (and the VGP one below). It just complicated things. Why not always assume its a standard SVGP coming into the SParseVariational Class?

trieste/acquisition/function.py
```diff
@@ -132,35 +174,56 @@ def prepare_acquisition_function(
...
+        return function
+
 
-def expected_improvement(model: ProbabilisticModel, eta: TensorType) -> AcquisitionFunction:
-    r"""
-    Return the Expected Improvement (EI) acquisition function for single-objective global
-    optimization. Improvement is with respect to the current "best" observation ``eta``, where an
-    improvement moves towards the objective function's minimum, and the expectation is calculated
-    with respect to the ``model`` posterior. For model posterior :math:`f`, this is
+class expected_improvement:
```
henrymoss: I do wonder if we should define a base class for these acquisition function bits?

henrymoss: We need documentation saying that these acq functions need an update and a __call__ and what these bits are for

trieste/acquisition/function.py
```diff
@@ -326,47 +410,85 @@ def prepare_acquisition_function(
...
+        gamma = (tf.squeeze(self._samples) - fmean) / fsd
 
         log_minus_cdf = normal.log_cdf(-gamma)
         ratio = tf.math.exp(normal.log_prob(gamma) - log_minus_cdf)
         f_acqu_x = -gamma * ratio / 2 - log_minus_cdf
 
         return tf.math.reduce_mean(f_acqu_x, axis=1, keepdims=True)
 
-    return acquisition
-
```
henrymoss, line 369: How come we dont have a new lower_confidence_bound class (its still a function like before). Just because it doesnt have anything to update doesnt mean we dont want to stop recompiling it?

trieste/acquisition/function.py
```diff
@@ -1116,64 +1299,96 @@ def prepare_acquisition_function(
...
+        """
+        :param function: The acquisition function to update.
+        :param dataset: The data from the observer.
+        :param model: The model over the specified ``dataset``.
+        :param pending_points: Points already chosen to be in the current batch (of shape [M,D]),
+            where M is the number of pending points and D is the search space dimension.
+        :return: The updated acquisition function.
+        """
+        tf.debugging.assert_positive(len(dataset))
+        tf.debugging.Assert(None not in [self._base_acquisition_function], [])
```
henrymoss, line 1334: What does this line do?

trieste/acquisition/function.py
```diff
@@ -1116,64 +1299,96 @@ def prepare_acquisition_function(
...
+        tf.debugging.assert_rank(pending_points, 2)
 
-        penalization = cast(Callable[..., PenalizationFunction], self._lipschitz_penalizer)(
-            model, pending_points, self._lipschitz_constant, self._eta
-        )
+        if self._penalized_acquisition is not None and isinstance(
+            self._penalization, UpdatablePenalizationFunction
+        ):
+            # if possible, just update the penalization function variables
+            self._penalization.update(pending_points, self._lipschitz_constant, self._eta)
```
henrymoss, line 1335: pending points is of variable (growing) size so how does not trigger retracing?

trieste/acquisition/function.py
```diff
@@ -1116,64 +1299,96 @@ def prepare_acquisition_function(
...
+            return self._penalized_acquisition
+        else:
```
henrymoss, line 1348: I dont see a situation where a penalization function isnt updatable! Maybe we can just force them to be updatable

trieste/models/model_interfaces.py
uri-granta: A mixin is "a class that contains methods for use by other classes without having to be the parent class of those other classes". It's useful here as both wrappers want the same behaviour.

trieste/models/model_interfaces.py
uri-granta: Even if we only supported passing in standard SVGPs into SparseVariational, we'd still need this, as we need to turn those standard SVGPs into SVGPs that store num_data in a Variable and expose it as a property.

trieste/acquisition/function.py
uri-granta: I'll add an ABC (though note these acquisition functions don't need an update: whether they have one, and what it looks like, is specific to the implementations).

trieste/acquisition/function.py
uri-granta: Because I haven't updated all the acquisition functions yet. See the text box for the PR.

trieste/acquisition/function.py
uri-granta: Checks that the base acquisition function already been constructed by a previous call to prepare_acquisition_function. The funny syntax is to make tensorflow happy (and is copied from similar checks elsewhere in this file).

trieste/models/model_interfaces.py
henrymoss: I see

trieste/acquisition/function.py
uri-granta: It's still more hassle to write updateable functions than non-updateable ones. And this bit of code is still required for the first time we generate the penalization function, so I don't think we gain anything by removing support for non-updateable functions.

tests/integration/test_bayesian_optimization.py
```diff
@@ -132,3 +133,9 @@ def build_model(data: Dataset) -> GaussianProcessRegression:
     # this is a regression test
     assert tf.reduce_any(tf.reduce_all(relative_minimizer_err < 0.05, axis=-1), axis=0)
     npt.assert_allclose(best_y, SCALED_BRANIN_MINIMUM, rtol=0.005)
+
+    # check that acquisition functions defined as classes aren't being retraced unnecessarily
+    if isinstance(acquisition_rule, EfficientGlobalOptimization):
+        acquisition_function = acquisition_rule._acquisition_function
+        if isinstance(acquisition_function, AcquisitionFunctionClass):
+            assert acquisition_function.__call__._get_tracing_count() == 3  # type: ignore
```
henrymoss, line 141: Lovely!

henrymoss (approved):

PR #272, opened 2021-06-22
decomposition of model_interfaces.py

hstojic:
Hello,

here is a draft for decomposing the model_interfaces (now also renamed to interfaces)

some remarks:
- I didn't move optimizers yet, that would be the next step - it needs some further rethinking and changes in some classes as default gpflow optimizer is spreaded elsewhere
- I haven't touched tests, after agreeing on final form I can do that part as well
- it would be great to eliminate gpflow imports anywhere else from the code, isolate it to models_gpflow
- this PR will make the toolbox incompatible with previous versions, so perhaps this is something to implement for trieste 1.x


trieste/models/config.py
```diff
@@ -13,14 +13,28 @@
...
+types.
+"""
+supported_models: dict[Any, Callable[[Any, Optimizer], TrainableProbabilisticModel]] = {
+    GPR: GaussianProcessRegression,
+    SGPR: GaussianProcessRegression,
+    VGP: VariationalGaussianProcess,
+}
 
 
 def _default_optimizer() -> gpflow.optimizers.Scipy:
```
hstojic: I think it would be good to eliminate gpflow here, but what to use instead, tensorflow? since that is a library used throughout?

trieste/models/interfaces.py
```diff
@@ -0,0 +1,229 @@
...
+        Return the variance of observation noise.
+
+        Note that this is only supported for models with with homoscedastic noise, and so this
+        function returns a scalar.
+
+        :return: The observation noise.
+        """
+        raise NotImplementedError(f"Model {self!r} does not provide scalar observation noise")
+
+    def get_kernel(self) -> gpflow.kernels.Kernel:
```
hstojic, line 95: this is a recent addition? this is another model-specific dependency  that we should probably remove, that is, the output type, not the method - any suggestions?

trieste/models/config.py
henrymoss: do you mean for the optimizer?

trieste/models/interfaces.py
henrymoss: I could just say its a Callable? But I don't envisage us every using a model that has  a kernel that isnt a gpflow one (e.g. even gpflux uses the gpflow kernels)

trieste/models/config.py
hstojic: yes, text refers to `_default_optimizer`, while I marked GPR, SGPR and VGP which is also an import from gpflow

trieste/models/interfaces.py
hstojic: right, that would be a generic thing to do here, though its not that useful any more as a type check

does this really need to be a part of this class? it assumes a nonparametric probabilistic model, no? perhaps it could be a method in ` GPflowPredict` which is sort of an interface for gpflow models? - there it would be fully appropriate

trieste/models/config.py
joelberkeley: i think of this file as gpflow-specific. I'd personally move the whole thing into the models_gpflow package

trieste/models/interfaces.py
henrymoss: You are right that get_kernel should be in the Gpflow class. I will move that and get_observation noise into  Gpflow class

joelberkeley: you may find that acquisition functions don't type check if you do that. Having looked at where `get_kernel` is used, I'm not sure what you can do about that other than raise a `TypeError` when you try to use random fourier.

joelberkeley: but don't let me discourage you

hstojic: @henrymoss I'll wait for the merge of your PR to merge this one then

trieste/models/config.py
hstojic: ok, then probably I misunderstood how this config is used, thanks for the tip, that would resolve these issues effectively

trieste/models/interfaces.py
joelberkeley: @henrymoss while you're doing that, `get_kernel` _returns_ the error where it should raise it

hstojic:
Finally updated all this. Merged in latest `develop`, moved around all the tests, eliminated gpflow from other parts of the code.

I also attempted to fix `config.py` by creaating a parent class and creating model type specific subclass, but there typing turned out to be messy and mypy complains - I would appreciate some help in this department :) (@joelberkeley?) 
 
I haven't touched optimizers yet, not crucial for now as everything can work without that part, so I left it for another PR.

uri-granta (approved):
This looks great. I'll take a look at the typing next.

Note that it'll result in a small merge conflict for https://github.com/secondmind-labs/trieste/pull/320 (cc @sebastianober)

Also probably worth building the docs manually using tox to see that they still look ok.

uri-granta:
Looking in more detail at the typing failures, there are quite a few issues. Happy to discuss.

hstojic:
All fixed now and merged in latest `develop`. 

I returned `get_kernel` method to `ProbabilisticModel` and we'll address that in another PR.

PR #273, opened 2021-06-23
Get the test to run

johnamcleod:


trieste/models/keras/data.py
```diff
@@ -0,0 +1,43 @@
...
+        self._networks = networks
+
+    def __call__(self, data: Dataset) -> TrainingData:
+        inputs = {}
+        outputs = {}
+
+        for index, network in enumerate(self._networks):
+            input_name = network.input_tensor_spec.name + "_" + str(index)
+            if index == 0:
+                output_name = 'dense'
```
johnamcleod, line 35: This isn't going to work in general. We will need to figure out how to get the input and output names from the model.

tests/integration/test_bayesian_optimization.py
```diff
@@ -31,14 +31,17 @@
...
+from trieste.models.keras.data import EnsembleDataTransformer
+from trieste.models.optimizer import TFKerasOptimizer, TrainingData
 from trieste.models.keras.models import NeuralNetworkEnsemble
 from trieste.models.keras.networks import MultilayerFcNetwork
 from trieste.models.keras.utils import get_tensor_spec_from_data
 from trieste.space import Box
 from trieste.utils.objectives import BRANIN_MINIMIZERS, BRANIN_MINIMUM, branin, mk_observer
 
 
+tf.keras.backend.set_floatx('float64')
```
johnamcleod, line 43: This might not be necessary.

hstojic: right, I was wondering about that one as well, was on my to do list to check, lets set it up that way for now as trieste seems to use 64 evereywhere

trieste/models/keras/data.py
hstojic: not sure I understand completely why naming is so relevant, but its definitely a good idea to have a dedicated class for these trieste Dataset to keras data tranformations

hstojic (approved):
looks good to me

PR #274, opened 2021-06-24
Thompson Sampling Overhaul

henrymoss:
I have overhauled our Thompson sampling functionality to support Fourier feature decompositions and have a more consistent sampling interface. 

Significant changes are as follows:

- we no longer have separate discrete and continuous sampler classes. We now have a general Sampler class and a ThompsonSampler subclass. The ThompsonSampler class differs from Sampler as it has an additional input to determine if we return samples of the function's minimal value (as required for entropy search) or its minimiser (as required for Thompson sampling).
- The RandomFourierFeatureThompsonSampler no longer has a prebuild logic as this turned out to be unnecessary.  This class also now has a sample method (over a discrete set of locations) to match the other ThompsonSampler classes. We can still access the full sampled trajectory, which will be required for a future continuous Thompson sampler PR.
- The ThompsonSampling rule is now more appropriately named DiscreteThompsonSampling and now supports exact or approximate (feature-based) Thompson sampling over a discrete candidate set.
- The MinValueEntropySearch acquisition function can now get its min value samples using Gumbel, exact (discrete) Thompson sampling or approximate (feature-based) Thompson sampling.

### DIFF
--- a/trieste/acquisition/function.py
+++ b/trieste/acquisition/function.py
@@ -32,7 +32,7 @@ from ..space import SearchSpace
 from ..type import TensorType
 from ..utils import DEFAULTS
 from ..utils.pareto import Pareto, get_reference_point
-from .sampler import BatchReparametrizationSampler, GumbelSampler
+from .sampler import BatchReparametrizationSampler, GumbelSampler, RandomFourierFeatureThompsonSampler, ExactThompsonSampler
 
 AcquisitionFunction = Callable[[TensorType], TensorType]
 """
@@ -242,17 +242,21 @@ class MinValueEntropySearch(SingleModelAcquisitionBuilder):
     minimisation. :class:`MinValueEntropySearch` estimates the information in the distribution
     of the objective minimum that would be gained by evaluating the objective at a given point.
 
-    This implementation largely follows :cite:`wang2017max` and samples the objective minimum
-    :math:`y^*` via a Gumbel sampler.
+    This implementation largely follows :cite:`wang2017max` and samples the objective's minimum
+    :math:`y^*` via a Gumbel sampler. TODO talk about costs of sampling
     """
 
-    def __init__(self, search_space: SearchSpace, num_samples: int = 10, grid_size: int = 5000):
+    def __init__(self, search_space: SearchSpace, num_samples: int = 10, grid_size: int = 5000, use_thompson: bool = False, num_fourier_features: Optional[int]=None):
         """
         :param search_space: The global search space over which the optimisation is defined.
         :param num_samples: Number of samples to draw from the distribution over the minimum of the
             objective function.
         :param grid_size: Size of the grid with which to fit the Gumbel distribution. We recommend
             scaling this with search space dimension.
+        :param use_thompson: If True then use Thompson sampling to sample the objective's 
+            minimum, else use Gumbel sampling.
+        :param num_fourier_features: Number of fourier features used for approximate Thompson 
+            sampling. If None, then do exact Thompson sampling.
         """
         self._search_space = search_space
 
@@ -263,6 +267,14 @@ class MinValueEntropySearch(SingleModelAcquisitionBuilder):
         if grid_size <= 0:
             raise ValueError(f"grid_size must be positive, got {grid_size}")
         self._grid_size = grid_size
+
+        if type(num_fourier_features) is int:
+            if not use_thompson:
+                raise ValueError(f"Fourier features approximation can only be applied to thompson sampling however `use_thompson` is {use_thompson}.")
+            if num_fourier_features<=0:
+                raise ValueError(f"num_fourier_features must be positive, got {num_fourier_features}")
+        self._use_thompson = use_thompson
+        self._num_fourier_features = num_fourier_features
 
     def prepare_acquisition_function(
         self, dataset: Dataset, model: ProbabilisticModel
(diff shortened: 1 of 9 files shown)


trieste/acquisition/function.py
```diff
@@ -264,6 +284,21 @@ def __init__(self, search_space: SearchSpace, num_samples: int = 10, grid_size:
             raise ValueError(f"grid_size must be positive, got {grid_size}")
         self._grid_size = grid_size
 
+        if type(num_fourier_features) is int:
```
uri-granta: It's more Pythonic to check for optional values like this:

```suggestion
        if num_fourier_features is not None:
```

trieste/acquisition/function.py
```diff
@@ -264,6 +284,21 @@ def __init__(self, search_space: SearchSpace, num_samples: int = 10, grid_size:
...
+            if not use_thompson:
+                raise ValueError(
+                    f"""
+                    Fourier features approximation can only be applied to thompson sampling
```
```suggestion
                    Fourier features approximation can only be applied to Thompson sampling
```

trieste/acquisition/function.py
```diff
@@ -277,14 +312,28 @@ def prepare_acquisition_function(
         if len(dataset.query_points) == 0:
             raise ValueError("Dataset must be populated.")
 
-        gumbel_sampler = GumbelSampler(self._num_samples, model)
+        if not self._use_thompson:  # use Gumbel sampler
+            sampler: ThompsonSampler = GumbelSampler(
+                self._num_samples, model, sample_min_value=True
+            )
+        else:
```
uri-granta: Clearer as a single if/elif/else block:

```suggestion
        elif self._num_fourier_features is not None:
```

trieste/acquisition/sampler.py
```diff
@@ -61,33 +61,77 @@ def sample(self, at: TensorType) -> TensorType:
         """
 
 
-class DiscreteThompsonSampler(DiscreteSampler):
+class ThompsonSampler(Sampler, ABC):
     r"""
-    This sampler provides approximate Thompson samples of the objective function's
+    A :class:`ThompsonSampler` samples either the minimum values or minimisers of a function
+    modelled by an underlying :class:`ProbabilisticModel` across a  discrete set of points.
```
uri-granta: If we're using US spelling :-)

```suggestion
    modeled by an underlying :class:`ProbabilisticModel` across a  discrete set of points.
```

trieste/acquisition/sampler.py
```diff
@@ -61,33 +61,77 @@ def sample(self, at: TensorType) -> TensorType:
...
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""{self.__class__.__name__}(
+        {self._sample_size!r},
+        {self._model!r},
+        {self._sample_min_value})
+        """
+
+    @abstractmethod
```
uri-granta: You don't need to repeat the abstract method here as it's already inherited from `Sampler` (Python will figure out that this must be an abstract base class as it doesn't implement it).

trieste/acquisition/sampler.py
```diff
@@ -61,33 +61,77 @@ def sample(self, at: TensorType) -> TensorType:
...
-        samples = self._model.sample(at, self._sample_size)  # [self._sample_size, len(at), 1]
-        samples_2d = tf.squeeze(samples, -1)  # [self._sample_size, len(at)]
-        indices = tf.math.argmin(samples_2d, axis=1)
-        unique_indices = tf.unique(indices).y
-        thompson_samples = tf.gather(at, unique_indices)
+        samples = self._model.sample(at, self._sample_size)  # [S, N, 1]
+
+        if self._sample_min_value:
+            thompson_samples = tf.reduce_min(samples, axis=1)  # [S, 1]
+
```
uri-granta: (this blank line is a bit confusing)

trieste/acquisition/sampler.py
```diff
@@ -99,8 +143,31 @@ class GumbelSampler(DiscreteSampler):
     Samples are obtained via the Gumbel distribution by sampling :math:`r` uniformly from
     :math:`[0, 1]` and applying the inverse probability integral transform
     :math:`y = \mathcal G^{-1}(r; a, b)`.
+
```
uri-granta, line 139: (Superfluous blank line?)

trieste/acquisition/sampler.py
```diff
@@ -99,8 +143,31 @@ class GumbelSampler(DiscreteSampler):
...
+
+    Note that the :class:`GumbelSampler` can only sample a function's minimal value and not
+    its minimiser.
     """
 
+    def __init__(self, sample_size: int, model: ProbabilisticModel, sample_min_value: bool = False):
```
uri-granta: The init functions don't need to be the same for all the ThompsonSamplers: methods do (since you might not know which type of thompson sampler you're calling them on) but for a constructor you always know the type. Since GumbelSampler doesn't support sampling the minimiser then there's no point in including `sample_min_value` as an argument. Just call `super().__init__(sample_size, model, True)` at the end.

```suggestion
    def __init__(self, sample_size: int, model: ProbabilisticModel):
```

trieste/acquisition/rule.py
```diff
@@ -212,13 +212,29 @@ def acquire(
...
+    def __init__(
+        self,
+        num_search_space_samples: int,
+        num_query_points: int,
+        num_fourier_features: Optional[int] = None,
+    ):
         """
         :param num_search_space_samples: The number of points at which to sample the posterior.
         :param num_query_points: The number of points to acquire.
+        :num_fourier_features: The number of features used to approximate the kernel. We recommend
```
uri-granta: Maybe comment that `None` means exact?

trieste/acquisition/rule.py
```diff
@@ -228,12 +244,21 @@ def __init__(self, num_search_space_samples: int, num_query_points: int):
                 f"Number of query points must be greater than 0, got {num_query_points}"
             )
 
+        if (type(num_fourier_features) is int) and (num_fourier_features <= 0):
```
```suggestion
        if num_fourier_features is not None and num_fourier_features <= 0:
```

uri-granta (approved):
Looks good!

trieste/acquisition/rule.py
```diff
@@ -212,13 +212,31 @@ def acquire(
...
+class DiscreteThompsonSampling(AcquisitionRule[None, SearchSpace]):
+    r"""
+    Implements Thompson sampling for choosing optimal points.
 
-    def __init__(self, num_search_space_samples: int, num_query_points: int):
+    This rule returns the minimizers of functions sampled from our model and evaluated across
+    a discretization of the search space (containing `N` candidate points).
+
+    The model is sampled either exactly (with an :math:`O(N^3)` complexity), or sampled
+    approximately through a random Fourier `M` feature decompisition
```
vpicheny, line 223: do we have the RFF option here?

PR #275, opened 2021-06-25
Gibbon

henrymoss:
My GIBBON acquisition function for cheap batch information-theoretic BO.

 I have added the acquisition function and test. A notebook will follow in another PR

Also changed default of MES/GIBBON to use the Thompson sampler for min value samples, with Gumbel now as optional.


henrymoss:
@uri-granta  I have had to add some `# type: ignore` flags in function.py (lines 1287 and 1336). 

GIBBON only works for GPR and our test dummy model (as it requires access to noise and covariance_between_points). However, I can't specify that model should be a GPR without the tests breaking (as the test dummy model is not a GPR)

trieste/acquisition/function.py
```diff
@@ -1121,3 +1123,238 @@ def penalization_function(x: TensorType) -> TensorType:
...
+                )
+            if num_fourier_features <= 0:
+                raise ValueError(
+                    f"num_fourier_features must be positive, got {num_fourier_features}"
+                )
+        self._use_thompson = use_thompson
+        self._num_fourier_features = num_fourier_features
+        self._big_batch = big_batch
+
+        self._min_value_samples = None
```
uri-granta, line 1195: (surprised you don't need to type annotate this line)

trieste/acquisition/function.py
```diff
@@ -1121,3 +1123,238 @@ def penalization_function(x: TensorType) -> TensorType:
...
+    :return: The GIBBON acquisition function. This function will raise :exc:`ValueError` or
+        :exc:`~tf.errors.InvalidArgumentError` if used with a batch size greater than one.
+    """
+    tf.debugging.assert_rank(samples, 2)
+    if len(samples) == 0:
+        raise ValueError("Min-value samples must be populated.")
+
+    try:
+        noise_variance = model.get_observation_noise()
+        model.covariance_between_points  # type: ignore
```
uri-granta: Is the point of this to check whether covariance_between_points is defined? If so, I would expect this to raise an `AttributeError` on failure, not `NotImplementedError`. Also, it would probably be clearer to be to use `if hasattr(model, "get_observation_noise")` instead.

trieste/acquisition/function.py
```diff
@@ -1121,3 +1123,238 @@ def penalization_function(x: TensorType) -> TensorType:
...
+
+        tf.debugging.assert_shapes(
+            [(x, [..., 1, None])],
+            message="This acquisition function only supports batches through greedy batch design.",
+        )
+
+        fmean, fvar = model.predict(tf.squeeze(x, -2))
+        yvar = fvar + noise_variance  # need predictive variance of observations
+
+        rho_squared = fvar / yvar  # squared correlation between observations and latent function
```
uri-granta, line 1314: (Random aside: sadly ρ² isn't a valid identifier name, though ρ_squared  and ρᵗʷᵒ are!)

tests/unit/acquisition/test_function.py
```diff
@@ -967,3 +975,228 @@ def test_lipschitz_penalizers_raises_for_invalid_pending_points_shape(
...
+
+        def sample(self, query_points: TensorType, num_samples: int) -> None:
+            return None
+
+        def get_observation_noise(self) -> None:
+            return None
+
+    with pytest.raises(ValueError):
+        model_without_likelihood = dummy_model_without_covariance_between_points()
+        gibbon(model_without_likelihood, tf.constant([1.0]))
```
uri-granta: I'm surprised this test works (I thought it would raise an uncaught AttributeError instead as I commented earlier) but apparently it does!

tests/unit/acquisition/test_function.py
```diff
@@ -312,7 +314,7 @@ def test_min_value_entropy_search_builder_builds_min_value_samples_rff(mocked_mv
     query_points = builder._search_space.sample(num_samples=builder._grid_size)
     query_points = tf.concat([dataset.query_points, query_points], 0)
     fmean, _ = model.predict(query_points)
-    assert max(min_value_samples) < min(fmean)
+    assert max(min_value_samples) < min(fmean) + 1e-4
```
uri-granta, line 317: Why is this necessary (here and further down)?

uri-granta (approved):
Looks good!

trieste/acquisition/function.py
henrymoss: Eh?

tests/unit/acquisition/test_function.py
henrymoss: Under non-zero observations noise (with RFF sampling), we could get sample draws that were all larger than min(fmean), just by chance. As i changed the default number of min-value samples, this seemed to happen when running this test (although it didnt before).

PR #277, opened 2021-06-30
GPR kernel initializer

henrymoss:
If any of our GPR kernel's trainable parameters have priors, then we begin model optimization from the best configuration across a random sample from these parameters' priors.

For trainable parameters without priors, we begin optimization from their initial values.

The sample size is max(1000, 100*d) where d is the number of trainable kernel params with priors.

### DIFF
--- a/trieste/models/model_interfaces.py
+++ b/trieste/models/model_interfaces.py
@@ -28,13 +28,6 @@ from ..type import TensorType
 from ..utils import DEFAULTS
 from .optimizer import Optimizer
 
-RANDOM_INIT_SUPPORTED_KERNELS = (
-    gpflow.kernels.SquaredExponential,
-    gpflow.kernels.Matern12,
-    gpflow.kernels.Matern32,
-    gpflow.kernels.Matern52,
-)
-
 
 class ProbabilisticModel(ABC):
     """A probabilistic model."""
--- a/trieste/models/model_interfaces.py
+++ b/trieste/models/model_interfaces.py
@@ -460,12 +460,12 @@ class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel):
         for _ in tf.range(num_prior_samples):
             randomize_model_hyperparameters(self.model)
             try:
-                likelihood = evaluate_likelihood_of_model_parameters()
+                log_likelihood = evaluate_likelihood_of_model_parameters()
             except tf.errors.InvalidArgumentError:  # allow badly specified priors
-                likelihood = -1e100
+                log_likelihood = -1e100
 
-            if likelihood > max_log_likelihood:  # only keep best kernel params
-                max_log_likelihood = likelihood
+            if log_likelihood > max_log_likelihood:  # only keep best kernel params
+                max_log_likelihood = log_likelihood
                 current_best_parameters = read_values(self.model)
 
         multiple_assign(self.model, current_best_parameters)
--- a/trieste/models/model_interfaces.py
+++ b/trieste/models/model_interfaces.py
@@ -432,10 +432,10 @@ class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel):
         """
 
         num_trainable_params_with_priors = tf.reduce_sum(
-            [tf.size(param) for param in self.model.kernel.trainable_parameters]
+            [tf.size(param) for param in self.model.kernel.trainable_parameters if param.prior]
         )
 
-        if num_trainable_params_with_priors > 1:  # Find a promising kernel initialization
+        if num_trainable_params_with_priors >= 1:  # Find a promising kernel initialization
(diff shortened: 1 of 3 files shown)


trieste/models/model_interfaces.py
```diff
@@ -419,6 +420,56 @@ def get_observation_noise(self):
...
+
+        if num_trainable_params_with_priors >= 1:  # Find a promising kernel initialization
+            num_random_samples = tf.minimum(1000, 100 * num_trainable_params_with_priors)
+            self.find_best_kernel_initialization(num_random_samples)
+
+        self.optimizer.optimize(self.model, dataset)
+
+    def find_best_kernel_initialization(self, num_kernel_samples: int = 100) -> None:
+        """
+        Test `num_kernel_samples` models with kernel parameters sampled from their
```
frgsimpson: I'd consider renaming num_kernel_samples here to something like num_prior_samples because kernel samples has a different meaning.

trieste/models/model_interfaces.py
```diff
@@ -419,6 +420,56 @@ def get_observation_noise(self):
...
+
+        :param dataset: The data with which to optimize the `model`.
+        """
+
+        num_trainable_params_with_priors = tf.reduce_sum(
+            [tf.size(param) for param in self.model.kernel.trainable_parameters if param.prior]
+        )
+
+        if num_trainable_params_with_priors >= 1:  # Find a promising kernel initialization
+            num_random_samples = tf.minimum(1000, 100 * num_trainable_params_with_priors)
```
frgsimpson: I can see the motivation for increasing the n_inits, but it does risk becoming prohibitively slow if there are many params.

trieste/models/model_interfaces.py
```diff
@@ -419,6 +420,60 @@ def get_observation_noise(self):
...
+
+        @tf.function
+        def evaluate_likelihood_of_kernel_parameters() -> tf.Tensor:
+            return self.model.maximum_log_likelihood_objective()
+
+        current_best_parameters = read_values(self.model.kernel)
+        max_log_likelihood = evaluate_likelihood_of_kernel_parameters()
+
+        for _ in tf.range(num_kernel_samples):
+            randomize_kernel_hyperparameters(self.model.kernel)
```
frgsimpson: Would it be faster if this lives inside the tf.function method?

frgsimpson: Also note here we are passing self.model.kernel, we could pass self.model and also randomise any parameters in the likelihood.

frgsimpson (commented):
Looks good! I'd maybe suggest adding the option of applying default priors to the parameters without explicit priors, such as

https://github.com/Prowler-io/kitt/blob/9ec3808d6c8afe2ddb3197f4cb83d26203326c82/kitt/data/priors.py#L15

trieste/models/model_interfaces.py
```diff
@@ -419,6 +420,60 @@ def get_observation_noise(self):
...
+            ]
+        )
+
+        if num_trainable_params_with_priors >= 1:  # Find a promising kernel initialization
+            num_random_samples = tf.minimum(1000, 100 * num_trainable_params_with_priors)
+            self.find_best_kernel_initialization(num_random_samples)
+
+        self.optimizer.optimize(self.model, dataset)
+
+    def find_best_kernel_initialization(self, num_kernel_samples: int = 100) -> None:
```
uri-granta: Does this need a default? (it's a bit odd that the default here is a fraction of the minimum we'd call it with in `optimize`)

trieste/models/model_interfaces.py
```diff
@@ -419,6 +420,60 @@ def get_observation_noise(self):
...
+
+        current_best_parameters = read_values(self.model.kernel)
+        max_log_likelihood = evaluate_likelihood_of_kernel_parameters()
+
+        for _ in tf.range(num_kernel_samples):
+            randomize_kernel_hyperparameters(self.model.kernel)
+            try:
+                likelihood = evaluate_likelihood_of_kernel_parameters()
+            except tf.errors.InvalidArgumentError:  # allow badly specified priors
+                likelihood = -1e100
```
uri-granta: Why not 0?

uri-granta (approved):
Looks good!

trieste/models/model_interfaces.py
frgsimpson: Ah because this is really the log_likelihood

trieste/models/model_interfaces.py
henrymoss: Done

trieste/models/model_interfaces.py
henrymoss: agreed

trieste/models/model_interfaces.py
henrymoss: made this clear

PR #280, opened 2021-07-01
Showcase recent PRs

henrymoss:
I have added some of our recently added functionality into our PRs and integration tests.

- Added kernel priors in EI and batch EI notebooks and explained how they are used by Trieste
- Added demo of our new space-filling design in EI notebook
- Added lables on some notebook plots
- Added kernel priors/space-filling design to integration tests to reduce the number of steps required to converge (i.e fast tests).
- Slightly changed the convergence criterion of the integration test to better test BO, i.e. a harsher regret criterion but less harsh relative difference criterion (which didnt actually test convergence properly due to the rescaling)
- Added a rescaled branin function to use with kernel priors (used in EI and batch EI notebooks and the integration tests)

### DIFF
--- a/trieste/utils/objectives.py
+++ b/trieste/utils/objectives.py
@@ -29,7 +29,7 @@ from ..observer import MultiObserver, Observer, SingleObserver
 from ..type import TensorType
 
 
-def branin_internals(x: TensorType, scale: TensorType, translate: TensorType) -> TensorType:
+def _branin_internals(x: TensorType, scale: TensorType, translate: TensorType) -> TensorType:
     x0 = x[..., :1] * 15.0 - 5.0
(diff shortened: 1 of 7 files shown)


trieste/utils/objectives.py
```diff
@@ -29,6 +29,19 @@
 from ..type import TensorType
 
 
+def branin_internals(x: TensorType, scale: TensorType, translate: TensorType) -> TensorType:
```
uri-granta: Mark as internal? Otherwise, it should have a docstring.

```suggestion
def _branin_internals(x: TensorType, scale: TensorType, translate: TensorType) -> TensorType:
```

uri-granta (approved):
Looks good!

PR #283, opened 2021-07-08
BayesianOptimizer can fit initial model

henrymoss:
At the moment, we assume that the model given to BayesianOptimizer is already optimized to the initial_data and so it is not optimized in our BO loop until after choosing the first BO point.

I have changed it so that we optimize the model before choosing the first point, with an optional choice to revert back to our current behaviour


henrymoss:
I am having some confusing type issues to do with overloading. Please help me @uri-granta

henrymoss:
I have

uri-granta (approved):
Looks good!

PR #284, opened 2021-07-08
mask infeasible points before prediction in eci

joelberkeley:
In expected constrained improvement, all the query points are used for predictions, then they're masked out with the constraint mask. This is inefficient as it's making predictions then potentially throwing some away. I've changed it so it throws the query points away first, then makes predictions

### DIFF
--- a/trieste/acquisition/function.py
+++ b/trieste/acquisition/function.py
@@ -607,9 +607,9 @@ class ExpectedConstrainedImprovement(AcquisitionFunctionBuilder):
         if not tf.reduce_any(is_feasible):
             return constraint_fn
 
-        feasible_query_points = tf.boolean_mask(objective_dataset.query_points, axis=0)
+        feasible_query_points = tf.boolean_mask(objective_dataset.query_points)
         feasible_mean, _ = objective_model.predict(feasible_query_points)
-        eta = tf.reduce_min(feasible_mean)
+        eta = tf.reduce_min(feasible_mean, axis=0)
 
         return lambda at: expected_improvement(objective_model, eta)(at) * constraint_fn(at)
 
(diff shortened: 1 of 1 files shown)


trieste/acquisition/function.py
```diff
@@ -600,13 +600,14 @@ def prepare_acquisition_function(
 
         constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)
         pof = constraint_fn(objective_dataset.query_points[:, None, ...])
-        is_feasible = pof >= self._min_feasibility_probability
+        is_feasible = tf.squeeze(pof >= self._min_feasibility_probability, axis=-1)
```
joelberkeley, line 603: i'll be honest - i'm a bit vague on this. It makes sense to me and it passes the tests, but I'm not in the place to think it through carefully atm

henrymoss, line 605: Why must this change?

joelberkeley: tbh i'm not sure. maybe cos we were masking over the target space [N, 1] before and now we're masking over the input space [N] + feature_shape

joelberkeley: it kinda makes sense anyway, cos `constraint_fn` outputs [..., 1] and we don't need that trailing dim for the mask (indeed my experience has been that `tf.boolean_mask` doesn't like non-vector masks

joelberkeley: from the [docs](https://www.tensorflow.org/api_docs/python/tf/boolean_mask)

> In general, 0 < dim(mask) = K <= dim(tensor), and mask's shape must match the first K dimensions of tensor's shape.

i.e. pof >=min is [N, 1] and tensor is [N] + feature_shape so it'll only match if we a) squeeze it or b) the feature_shape happens to start with a 1

joelberkeley: you may need to check that, I'm tired and not thinking clearly

joelberkeley: ah yes, and the mean would have been [N, 1] so there was no need for squeeze

henrymoss:
LGTM. Thanks for this @joelberkeley

henrymoss (approved):

PR #285, opened 2021-07-09
Victor/constrained moo

vpicheny:
This PR implements a new acquisition function for the multiobjective constrained case. It basically merges the ExpectedConstrainedImprovent and the ExpectedHypervolumeImprovement functions. Comes with a new notebook, which combines the two corresponding notebooks as well.


vpicheny:
I haven't written the tests yet - I'm a bit unsure about what's needed

trieste/acquisition/function.py
```diff
@@ -745,6 +745,56 @@ def ehvi_based_on_partitioned_cell(
...
+        constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)
+        pof = constraint_fn(objective_dataset.query_points[:, None, ...])
+        is_feasible = pof >= self._min_feasibility_probability
+
+        if not tf.reduce_any(is_feasible):
+            return constraint_fn
+
+        objective_mean, _ = objective_model.predict(objective_dataset.query_points)
+
+        _pf = Pareto(tf.boolean_mask(objective_mean, is_feasible[:, 0]))
```
joelberkeley: see my PR #284 . With that, you can a) make this faster, and b) merge this with `ExpectedConstrainedImprovement` if you so wish

docs/notebooks/multiobjective_constrained.py
```diff
@@ -0,0 +1,222 @@
+# -*- coding: utf-8 -*-
```
joelberkeley: since moo constrained could be in either a moo notebook or a constrained BO notebook, you could split up the notebooks into smaller chunks and then mention other relevant notebooks in each notebook

joelberkeley: that might also make it easier to browse the notebooks from the website

trieste/acquisition/function.py
```diff
@@ -745,6 +745,56 @@ def ehvi_based_on_partitioned_cell(
...
+        _reference_pt = get_reference_point(_pf.front)
+        ehvi = expected_hv_improvement(objective_model, _pf, _reference_pt)
```
TsingQAQ, line 799: Maybe am a bit concern whether the current functionality, the `get_reference_point` and `hypercell_bounds` etc. support extreme constraints: when there are no feasible points at all and we had an empty front.

joelberkeley: there's a check above for the presence of feasible points
```python
        if not tf.reduce_any(is_feasible):
            return constraint_fn
```
is one feasible point sufficient?

TsingQAQ: Ah, yes, missed that one.

trieste/acquisition/function.py
henrymoss: I've merged @joelberkeley's PR #284 . So maybe you can do as he suggests @vpicheny ?

vpicheny: sure, will do!

vpicheny: @joelberkeley what do you have in mind with merging? 

Something like 
`constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)` 
but for the objective?

trieste/acquisition/function.py
```diff
@@ -746,6 +746,57 @@ def ehvi_based_on_partitioned_cell(
...
+        pof = constraint_fn(objective_dataset.query_points[:, None, ...])
+        is_feasible = tf.squeeze(pof >= self._min_feasibility_probability, axis=-1)
+
+        if not tf.reduce_any(is_feasible):
+            return constraint_fn
+
+        feasible_query_points = tf.boolean_mask(objective_dataset.query_points, is_feasible)
+        feasible_mean, _ = objective_model.predict(feasible_query_points)
+
+        _pf = Pareto(tf.boolean_mask(feasible_mean))
```
```suggestion
        _pf = Pareto(feasible_mean)
```
???

trieste/acquisition/function.py
```diff
@@ -746,6 +746,57 @@ def ehvi_based_on_partitioned_cell(
...
+                "Expected hypervolume improvement is defined with respect to existing points in the objective"
+                " data, but the objective data is empty."
+            )
+
+        constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)
+        pof = constraint_fn(objective_dataset.query_points[:, None, ...])
+        is_feasible = tf.squeeze(pof >= self._min_feasibility_probability, axis=-1)
+
+        if not tf.reduce_any(is_feasible):
+            return constraint_fn
```
joelberkeley, line 789: that's a lot of repetition from `ExpectedConstrainedImprovement`

trieste/acquisition/function.py
```diff
@@ -746,6 +746,57 @@ def ehvi_based_on_partitioned_cell(
...
+
+        feasible_query_points = tf.boolean_mask(objective_dataset.query_points, is_feasible)
+        feasible_mean, _ = objective_model.predict(feasible_query_points)
```
joelberkeley, line 792: that's a lot of repetition from `ExpectedConstrainedImprovement`

trieste/acquisition/function.py
```diff
@@ -746,6 +746,57 @@ def ehvi_based_on_partitioned_cell(
...
+    def __repr__(self) -> str:
+        """"""
+        return (
+            f"ExpectedConstrainedHypervolumeImprovement({self._objective_tag!r}, {self._constraint_builder!r},"
+            f" {self._min_feasibility_probability!r})"
+        )
+
+    def prepare_acquisition_function(
+        self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
+    ) -> AcquisitionFunction:
```
joelberkeley, line 764: this function is a combination of eci and ehvi, and should be possible to nicely abstract so that you can switch out ei vs ehvi easily

trieste/acquisition/function.py
joelberkeley: sorry, i missed this. I've looked into this more and I don't know how it's possible. You may just need two versions, though you could pull the first half of the method into a module-private function which both methods call if it's possible to write that in a way that's worth it

docs/notebooks/multi_objective_ehvi.pct.py
```diff
@@ -196,6 +202,142 @@ def log_hv(observations):
 ax.set_ylabel("log HV difference")
 plt.show()
 
+# %% [markdown]
+# ## Multi-objective optimization with constraints
+#
+# EHVI can be adapted to the case of constraints, as we show below. We start by defining a problem with the same objectives as above, but with an inequality constraint, and we define the corresponding `Observer`.
+
+# %%
+class Sim:
```
henrymoss, line 211: Its not clear that this means simulation

henrymoss (approved):

PR #286, opened 2021-07-10
extract TrustRegion from rules

joelberkeley:
We change trust region such that it's no longer an acquisition rule. I think this enables the following:
* All acquisition rules are now fully compatible with trust region
* there is less duplication in the rules
* EGO and thompson sampling no longer need a notion of state

TODO
- [x] update optimizer tests
- [x] handle docstring for trust region
- [x] confirm `y_min` default
- [x] elaborate re state in "recovering ..." notebook
- [x] decide which integration tests should use trust region


docs/notebooks/batch_optimization.pct.py
```diff
@@ -81,7 +81,7 @@ def build_model(data):
 batch_ei_acq = BatchMonteCarloExpectedImprovement(sample_size=1000, jitter=1e-5)
 batch_ei_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=batch_ei_acq)
-points_chosen_by_batch_ei, _ = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
+points_chosen_by_batch_ei = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
```
henrymoss, line 84: It would certainly be nice to get rid of this. It was quite unnatural

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -41,14 +41,13 @@ def __call__(self, x):
 
 # %% [markdown]
 # ## Set up the problem
-# We'll use the same set up as before, except for the acquisition rule, where we'll use `TrustRegion`, which (with non-trivial state) will better illustrate how to recover.
+# We'll use the same set up as before, except we'll also use a `TrustRegion`, which (with non-trivial state) will better illustrate how to recover.
```
henrymoss: Maybe elaborate what non-trivial state means here? Not obvious to those not ordained in trust regions/Trieste

trieste/acquisition/function.py
```diff
@@ -56,7 +56,25 @@
...
+    ) -> T:
+        """
+        :param datasets: The data from the observer.
+        :param models: The models over each dataset in ``datasets``.
+        :return: A value of type `T`.
+        """
+
+
+class AcquisitionFunctionBuilder(Empiric[AcquisitionFunction], ABC):
     """An :class:`AcquisitionFunctionBuilder` builds an acquisition function."""
```
henrymoss: Whats going on here?

trieste/acquisition/rule.py
```diff
@@ -21,48 +21,44 @@
...
 
     @abstractmethod
     def acquire(
         self,
         search_space: SP_contra,
         datasets: Mapping[str, Dataset],
         models: Mapping[str, ProbabilisticModel],
-        state: S | None = None,
-    ) -> tuple[TensorType, S]:
+    ) -> TensorType:
```
henrymoss, line 60: Yep, definitely keen to have this sort of behaviour.

henrymoss (commented):
I'm keen for this sort of change. @uri-granta  what do you reckon?

Thanks for doing this @joelberkeley !

trieste/acquisition/function.py
joelberkeley: i need to discuss this with you. I've abstracted the notion of an acquisition builder, and `prepare_acquisition_function` isn't an appropriate name when we're not building an acquisition function. I didn't want to change `prepare_acquisition_function` everywhere until I'd got your opinion on it

joelberkeley: really, I envisage `Empiric` being a type alias so there's nothing to name (I didn't do that here cos it's out of scope), but we'll need a name at least for now

trieste/bayesian_optimizer.py
```diff
@@ -172,7 +172,7 @@ def optimize(
         model_specs: Mapping[str, ModelSpec],
         *,
         track_state: bool = True,
-        fit_intial_model: bool = True,
+        fit_initial_model: bool = True,
```
joelberkeley, line 176: this is a typo, not intentional, right?

trieste/bayesian_optimizer.py
```diff
@@ -312,29 +343,33 @@ def optimize(
...
                     for tag, model in models.items():
                         dataset = datasets[tag]
                         model.update(dataset)
                         model.optimize(dataset)
 
-                query_points, acquisition_state = acquisition_rule.acquire(
-                    self._search_space, datasets, models, acquisition_state
-                )
+                if trust_region is None:
+                    acquisition_space = self._search_space
```
joelberkeley, line 367: we could make a "unit" empiric and use that, but I felt that was unnecessary for this one use case

trieste/type.py
```diff
@@ -12,10 +12,19 @@
...
 """Type alias for tensor-like types."""
+
+S = TypeVar("S")
+""" Unbound type variable. """
+
+T = TypeVar("T")
+""" Unbound type variable. """
+
+State = Callable[[Optional[S]], Tuple[S, T]]
+""" A `State` represents a stateful function, with state of type `S` and output of type `T`. """
```
joelberkeley, line 30: tbh i'm not sure it's right to allow `Optional[S]`. I think `S` is more appropriate. I may try to figure this out

trieste/acquisition/function.py
joelberkeley: i realise I don't need to do this in this PR. I do recommend it, but it's not necessary here

trieste/acquisition/rule.py
```diff
@@ -444,17 +406,67 @@ def acquire(
...
+
+    :param beta: The inverse of the trust region contraction factor.
+    :param kappa: Scales the threshold for the minimal improvement required for a step to be
+        considered a success.
+    :return: todo
+    """
+    return lambda box: _ContinuousTrustRegion(box, beta, kappa)
+
+
+"""
```
joelberkeley, line 439: not sure what to do with this. Do you need it?

trieste/acquisition/rule.py
```diff
@@ -311,124 +302,95 @@ def acquire(
...
+        self._kappa = kappa
 
-        **Note:** The acquisition search space will never extend beyond the boundary of the
-        ``search_space``. For a local search, the actual search space will be the
-        intersection of the trust region and ``search_space``.
+    @property
+    def default_state(self) -> ContinuousTrustRegionState:
+        space = self._global_search_space
+        eps = 0.5 * (space.upper - space.lower) / (5.0 ** (1.0 / space.lower.shape[-1]))
+        # todo is y_min correct?
```
joelberkeley, line 380: not sure if setting `y_min` to infinity results in the same algorithm

trieste/acquisition/rule.py
```diff
@@ -311,124 +302,95 @@ def acquire(
...
+        return ContinuousTrustRegionState(space, eps, math.inf, True)
```
joelberkeley: not sure if setting `y_min` to infinity results in the same algorithm

tests/integration/test_bayesian_optimization.py
```diff
@@ -83,13 +87,15 @@
                 num_query_points=2,
             ),
         ),
-        (15, TrustRegion()),
         (10, DiscreteThompsonSampling(500, 3)),
         (10, DiscreteThompsonSampling(500, 3, num_fourier_features=1000)),
     ],
 )
+@pytest.mark.parametrize("trust_region", [continuous_trust_region(), None])
```
joelberkeley: this doubles the number of integration tests in this file, which is expensive given these are slow tests. Any preference of which tests to apply trust region to?

docs/notebooks/recovering_from_errors.pct.py
joelberkeley: done. is that what you had in mind?

trieste/bayesian_optimizer.py
henrymoss: typo

tests/integration/test_bayesian_optimization.py
henrymoss: maybe one EGO based and one TS ?

joelberkeley: done

trieste/acquisition/rule.py
```diff
@@ -311,124 +302,95 @@ def acquire(
...
+        self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
+    ) -> State[ContinuousTrustRegionState, Box]:
         dataset = datasets[OBJECTIVE]
 
-        global_lower = search_space.lower
-        global_upper = search_space.upper
+        global_lower = self._global_search_space.lower
+        global_upper = self._global_search_space.upper
 
         y_min = tf.reduce_min(dataset.observations, axis=0)
```
joelberkeley, line 383: @vpicheny I chose the starting `y_min` to be infinite. If there are any finite observations in `dataset.observations` this works correctly, but if they're infinite or there aren't any, this implementation is different to how it was before. What should happen if `dataset.observations` is empty or is all infinite? Should trust region error?

vpicheny: I think this is fine. If you start with y_min=inf, your first step will always be a success, so you'll start by doing at least two global steps, which is reasonable in absence of previous data.

joelberkeley: i made it error if `y_min` is not finite

joelberkeley: ah, i didn't see your message. If `state.y_min` is infinite, and `y_min` is also infinite, you'll go straight to a local search space because `y_min < state.y_min - self._kappa * tr_volume` and `not state.is_global` are both false so `is_global = step_is_success or not state.is_global` is false

trieste/bayesian_optimizer.py
```diff
@@ -182,11 +182,27 @@ def optimize(
...
-        acquisition_state: S | None = None,
+        acquisition_rule: AcquisitionRule[SP],
         *,
+        trust_region: None = None,
+        trust_region_state: None = None,
         track_state: bool = True,
-        fit_intial_model: bool = True,
+        fit_initial_model: bool = True,
+    ) -> OptimizationResult[None]:
+        ...
```
uri-granta, line 193: Wondering if it's worth adding docstrings for the different overloads, as it's getting tricky to distinguish them?

1. Tagged datasets and models, no acquisition rule
2. Tagged datasets and models, explicit acquisition rule, no trust region
3. Tagged datasets and models, explicit acquisition rule, trust regions
4. Single dataset and model, no acquisition rule
5. Single dataset and model, acquisition rule, no trust region
6. Single dataset and model, acquisition rule, trust region

trieste/acquisition/rule.py
```diff
@@ -18,51 +18,46 @@
...
-S = TypeVar("S")
-""" Unbound type variable. """
-
 SP_contra = TypeVar("SP_contra", bound=SearchSpace, contravariant=True)
 """ Contravariant type variable bound to :class:`~trieste.space.SearchSpace`. """
 
 
-class AcquisitionRule(ABC, Generic[S, SP_contra]):
-    """The central component of the acquisition API."""
+class AcquisitionRule(Generic[SP_contra], ABC):
```
uri-granta: I understand removing the S, but why switch the inheritance order?

trieste/type.py
```diff
@@ -12,10 +12,19 @@
...
 """Type alias for tensor-like types."""
+
+S = TypeVar("S")
+"""Unbound type variable."""
+
+T = TypeVar("T")
+"""Unbound type variable."""
+
+State = Callable[[S], Tuple[S, T]]
+"""A `State` represents a stateful value of type `T`, with state of type `S`."""
```
uri-granta: I find this description confusing.

trieste/acquisition/rule.py
```diff
@@ -311,128 +302,102 @@ def acquire(
...
+"""Unbound type variable."""
 
-    @dataclass(frozen=True)
-    class State:
-        """The acquisition state for the :class:`TrustRegion` acquisition rule."""
 
-        acquisition_space: Box
-        """ The search space. """
+class EmpiricStateful(ABC, Generic[S, T]):
+    """A :class:`EmpiricStateful` is an empirical stateful value with a default state."""
```
uri-granta: I don't understand the name. What exactly does empirical mean? What is a stateful value (it seems to be a function that expects a state s and returns a value t and updated state s')? Does an EmpiricalStateful instance represent the value or is it a factor for stateful values using the acquire method?

trieste/acquisition/rule.py
```diff
@@ -311,128 +302,102 @@ def acquire(
...
 
-        eps: TensorType
+    @abstractmethod
+    def acquire(
+        self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
+    ) -> State[S, T]:
         """
-        The (maximum) vector from the current best point to each bound of the acquisition space.
```
uri-granta, line 329: A summary description before the arglist would be useful.

trieste/bayesian_optimizer.py
```diff
@@ -264,12 +295,12 @@ def optimize(
             arguments. Note that if the default is used, this implies the tags must be
             `OBJECTIVE`, the search space can be any :class:`~trieste.space.SearchSpace`, and the
             acquisition state returned in the :class:`OptimizationResult` will be `None`.
-        :param acquisition_state: The acquisition state to use on the first optimization step.
-            This argument allows the caller to restore the optimization process from an existing
-            :class:`Record`.
+        :param trust_region: Defines how to construct the acquisition space from the global space on
+            each step. Defaults to using the global space.
+        :param trust_region_state: Initial state for ``trust_region``.
```
uri-granta: Mention that this defaults to the default state of the trust region?

trieste/bayesian_optimizer.py
```diff
@@ -312,29 +343,40 @@ def optimize(
...
 
-                query_points, acquisition_state = acquisition_rule.acquire(
-                    self._search_space, datasets, models, acquisition_state
-                )
+                if trust_region is None:
+                    acquisition_space = self._search_space
+                else:
+                    trust_region_state, acquisition_space = trust_region_applied.acquire(
+                        datasets, models
+                    )(cast(S, trust_region_state))
```
uri-granta: Is the cast just to avoid the None? If not, why not move the defaulting down here? Probably clearer that way anyway.

```suggestion
                    )(trust_region_state if trust_region_state is not None else trust_region_applied.default_state)
```

trieste/acquisition/rule.py
```diff
@@ -444,17 +409,59 @@ def acquire(
...
-        point = self._optimizer(acquisition_space, acquisition_function)
-        state_ = TrustRegion.State(acquisition_space, eps, y_min, is_global)
+            new_state = ContinuousTrustRegionState(acquisition_space, eps, y_min, is_global)
+            return new_state, acquisition_space
+
+        return go
+
+
+def continuous_trust_region(
+    beta: float = 0.7, kappa: float = 1e-4
```
uri-granta, line 421: I know it's a bad idea but it still makes me sad that we don't write

```suggestion
    β: float = 0.7, κ: float = 1e-4
```

trieste/acquisition/rule.py
```diff
@@ -444,17 +409,59 @@ def acquire(
...
+    """
+    Implements the *trust region* algorithm for constructing a local acquisition space from a global
+    continuous search space. The algorithm is as follows:
+
+    The local acquisition space is the space constructed on each step by this `TrustRegion`,
+    and can be either the global search space, or a trust region which is typically smaller than
+    the global search space and changes on each step.
 
-        return point, state_
+    If no state is specified (it is `None`), the :attr:`default_state` is used.
```
uri-granta: Does this line belong here? By the time we call the continuous trust region we've already defaulted the state, no?

uri-granta (commented):
Overall this looks great. A few questions mostly regarding terminology.

trieste/acquisition/rule.py
joelberkeley: sometimes python seems to complain about invalid method resolution or sommat. This was one of those cases. I have no idea what the difference is in this case

trieste/acquisition/rule.py
joelberkeley: good catch

trieste/bayesian_optimizer.py
joelberkeley: could do. I reckon the method's just doing too much. Others weren't keen on splitting up the method last time I asked

trieste/bayesian_optimizer.py
joelberkeley: will do

trieste/type.py
joelberkeley: so do I

trieste/acquisition/rule.py
joelberkeley: i asked this kind of question on a forum for another language recently. They gave a list of reasons why to not do this, but one of them was that unicode is mangled on iOS. Found that mildly interesting

trieste/acquisition/rule.py
joelberkeley: i'll see if I can make this clearer

trieste/bayesian_optimizer.py
joelberkeley: will do

trieste/bayesian_optimizer.py
joelberkeley: fyi changing it like this means the stored state will be `None` for the first run (if the arg is `None`). Still works fine, just slightly differently

trieste/acquisition/rule.py
joelberkeley: i changed it back and it's fine with it, so error came from elsewhere

trieste/type.py
joelberkeley: updated. is that clearer?

trieste/acquisition/rule.py
joelberkeley: I could make this type notably simpler if I didn't have to add ` default_state` attribute, but I kind of gave up after the fourth or so iteration of the API

trieste/bayesian_optimizer.py
joelberkeley: docstrings don't appear in the docs when I add them to each overload

trieste/acquisition/rule.py
joelberkeley: i've expanded the description for `return`. I tend to prefer doing that when the summary would just re-iterate what's in the arglist and type signature. Do you think the new description is sufficient?

trieste/acquisition/rule.py
joelberkeley: note the structure is like a subclass of the acquisition function builders. My idea was to use a common interface for both, called `Empiric` ~ `Data -> Models -> T` but that would be too big a change for this PR. It's an `EmpiricStateful` because it's a `Data -> Models -> State[S, T]`. I'd actually write it like that if `default_state` wasn't involved

joelberkeley: I guess it would be possible in a compiled language, because we could restrict `S` to have a `default` value, and then `S.default`, but `S` is erased

joelberkeley: @uri-granta , i've changed the design a little on my machine and now we don't have `EmpiricStateful`. Shall I push that or will it confuse?

joelberkeley: here's the change on my fork https://github.com/joelberkeley/trieste/pull/1/files

trieste/type.py
uri-granta: Yes, that's much better thanks.

trieste/acquisition/rule.py
uri-granta: Yes, I think that's much clearer! (though remember to update the docstring references to `Empiric.acquire` and `TrustRegion.default_state`)

uri-granta: Once that's merged, happy to give this a tick.

joelberkeley: done

PR #287, opened 2021-07-12
Specify Model-specific optimization parameters

henrymoss:
Can now specify model-wrapper specific parameters when build our model wrappers. This allows us to control behaviour of of model wrappers (and reduce magic numbers), typically though the specification of model-specific parameters. 

The idea is to have all model specific optimization functionality withim the model wrappers, leaving a nice small set of base-optimizers (e.g. L-BFGS and Adam e.t.c, see models/optimizer.py) that can be used for any of our models.

Three examples of this behaviour included in this PR are

1.  We can control the number of random kernel samples used to init model optimization for GPR (this was previously a magic number with no way to change)
2. We can tell VGP to interleave its chosen optimizer (scipy or tf-based with/without mini-batching) with natgrad steps. Natgrads are specific to VGP (and SVGP, but this will come in a later PR) and so it makes sense to have this contained within the VGP wrapper rather than its own new model optimizer. 
3. Similarly, we can specify the natgrad gamma value when building the VGP wrapper.

Here is an example of the new behaviour, when specifying a VGP model that uses an Adam optimizer for 50 steps interleaved with natgrad steps:

`models = { OBJECTIVE: {"model": vgp_model, "model_args": { "use_natgrads": True }, "optimizer":tf.optimizers.Adam(1e-3), "optimizer_args": { "max_iter": 50 } }}`

Specifying a GPR model that test 10 random kernel param configs before beginning model optimization with a scipy optimizer

`models = { OBJECTIVE: {"model": gpr_model, "model_args": { "num_prior_samples": 10 }, "optimizer": gpflow.optimizers.Scipy() }}`


henrymoss:
This PR also makes a couple of other small related changes:

1. Updated notebooks (failure_ego and expected_improvement) to show off new behaviour.
2. Removed GP specific attriubtes (i.e. kernel) from probabilistic model class.
3. Added proper description and reference for VGP

henrymoss:
I still need to write some tests for changes in config. But I want to know if this PR is a good idea first

trieste/acquisition/sampler.py
```diff
@@ -412,7 +412,7 @@ def __init__(
 
         try:
             self._noise_variance = model.get_observation_noise()
-            self._kernel = model.get_kernel()
+            self._kernel = model.get_kernel()  # type: ignore
```
uri-granta, line 415: Might be better to restrict the models the sampler accepts to GPflowPredictor models instead, as other models with a get_kernel method aren't necessarily going to return a gpflow.kernels.Kernel?

docs/notebooks/expected_improvement.pct.py
```diff
@@ -66,6 +66,9 @@ def build_model(data):
 
     return {
             "model": gpr,
+            "model_args": {
```
uri-granta, line 69: Do we want to mention setting num_prior_samples in the text above (where we talk about packaging it with the gpflow scypy optimizer?

uri-granta (approved):
Looks great!

trieste/acquisition/sampler.py
henrymoss: The issue here is that our dummy test model is not of GPFlowPredictor type, so the type checking gets unhappy in the unit tests.

trieste/models/model_interfaces.py
```diff
@@ -557,6 +607,48 @@ def update(self, dataset: Dataset, *, jitter: float = DEFAULTS.JITTER) -> None:
...
+        If using natural gradients, our optimizer inherits the mini-batch behavior and number
+        of optimization steps as the base optimizer specified when initializing
+        the :class:`VariationalGaussianProcess`.
+        """
+        model = self.model
+
+        if self._use_natgrads:  # optimize variational params with natgrad optimizer
+
+            natgrad_optimizer = gpflow.optimizers.NaturalGradient(gamma=self._natgrad_gamma)
+            base_optimizer: TFOptimizer = self.optimizer
```
uri-granta: mypy can't 'remember' that this was checked previously. Either use `cast` here or assign it to a different variable `self.tfoptimizer` during initialization.

trieste/models/config.py
```diff
@@ -76,7 +79,7 @@ def create_model_interface(self) -> TrainableProbabilisticModel:
 
         for model_type, model_interface in supported_models.items():
             if isinstance(self.model, model_type):
-                return model_interface(self.model, optimizer)
+                return model_interface(self.model, optimizer, **self.model_args)
```
uri-granta: (If this requires a type ignore then so be it.)

trieste/models/model_interfaces.py
henrymoss: @uri-granta what do you mean by cast?

trieste/acquisition/function.py
```diff
@@ -1287,7 +1287,8 @@ def gibbon(
     except NotImplementedError:
```
hstojic, line 1288: so only homoscedastic GP models will have `get_observation_noise` method? should the model be something more specific than `ProbabilisticModel`?

henrymoss: I need to think about this, as quite a few of our acq functions are not really defined properly for *any* prob model. This will be another PR though

trieste/models/model_interfaces.py
```diff
@@ -88,24 +88,15 @@ def predict_y(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
             f"Model {self!r} does not support predicting observations, just the latent function"
         )
 
-    def get_observation_noise(self) -> TensorType:
+    def get_observation_noise(self):
```
hstojic: why removing the output type?

hstojic: also, why not moving the method to the GPpredictor class? we should list here methods that we would expect all the models to have, no?

trieste/models/model_interfaces.py
```diff
@@ -505,19 +509,65 @@ def update(self, dataset: Dataset) -> None:
 
 
 class VariationalGaussianProcess(GPflowPredictor, TrainableProbabilisticModel):
-    """A :class:`TrainableProbabilisticModel` wrapper for a GPflow :class:`~gpflow.models.VGP`."""
+    r"""
+    A :class:`TrainableProbabilisticModel` wrapper for a GPflow :class:`~gpflow.models.VGP`.
 
-    def __init__(self, model: VGP, optimizer: Optimizer | None = None):
+    A Variational Gaussian Process (VGP) approximates the posterior of a GP
+    using the multivariate Gaussian closest to the posterior of the GP by minimzing the
```
hstojic: minimizing instead of minimzing

tests/unit/models/test_model_interfaces.py
```diff
@@ -357,6 +357,15 @@ def test_gaussian_process_regression_pairwise_covariance(gpr_interface_factory)
     np.testing.assert_allclose(expected_covariance, actual_covariance, atol=1e-5)
 
 
+def test_gpr_raises_for_invalid_init() -> None:
```
hstojic: we can be more specific here, using `num_prior_samples` instead of `init` in the title

tests/unit/models/test_model_interfaces.py
```diff
@@ -651,14 +673,70 @@ def test_sparse_variational_optimize(batcher, compile: bool) -> None:
...
+        dict(max_iter=10, batch_size=10, dataset_builder=batcher, compile=compile),
     )
     model = SparseVariational(_svgp(x_observed[:10]), dataset, optimizer=optimizer)
     loss = model.model.training_loss(data)
     model.optimize(dataset)
     assert model.model.training_loss(data) < loss
 
 
+@pytest.mark.parametrize("use_natgrads", [True, False])
+@pytest.mark.parametrize("compile", [True, False])
```
hstojic, line 800: I would move `compile` argument to `conftest.py` file as a fixture, then it doesn't need to be parametrized above the testing function - it seems like an argument that would be tested in several places at least

henrymoss: There is no conftest.py

trieste/models/model_interfaces.py
```diff
@@ -323,6 +314,19 @@ def get_kernel(self) -> gpflow.kernels.Kernel:
         """
         return self.model.kernel
 
+    def get_observation_noise(self):
```
hstojic, line 318: is there a test for this exception?

trieste/models/model_interfaces.py
```diff
@@ -340,15 +344,25 @@ class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel):
...
         :param model: The GPflow model to wrap.
         :param optimizer: The optimizer with which to train the model. Defaults to
             :class:`~trieste.models.optimizer.Optimizer` with :class:`~gpflow.optimizers.Scipy`.
+        :param num_prior_samples: Number of randomly sampled kernels to evaluate before beginning
+            model optimization.
         """
         super().__init__(optimizer)
         self._model = model
 
+        if num_prior_samples <= 0:
```
hstojic: I think `assert` would be more appropriate here than if-else statement

hstojic (approved):
overall looks good, I have left couple of questions and some minor comments

tests/unit/models/test_model_interfaces.py
hstojic: there is one at `tests/conftest.py`, thats the main one, although you can make it subfolder specific as well, by simply creating another one at the subfolder level

trieste/acquisition/function.py
hstojic: hm, true, more of a design question, how to do these checks in general - here I'm just wondering whether checking for existence of this method is the best way to ensure its a homoscedastic GP, perhaps its best left as an assumption that is stressed in the docstring and leave it at that

also, scratch my comment about more specific than `ProbabilisticModel` class, I'm supposedly trying to avoid importing gpflow-related stuff elsewhere

hstojic: above it should be an `assert` statement rather than if-else:
```
if len(samples) == 0:
        raise ValueError("Min-value samples must be populated.")
```

hstojic: scratch the last comment, after a discussion on recent Joel's PR, we should probably use `tf.debugging` assertions everywhere

trieste/models/model_interfaces.py
henrymoss: Will add

docs/notebooks/failure_ego.pct.py
```diff
@@ -110,23 +110,11 @@ def create_classification_model(data):
...
-        adam_opt = tf.optimizers.Adam(1e-3)
-        natgrad_opt = gpflow.optimizers.NaturalGradient(gamma=0.1)
-
-        for step in range(50):
-            loss = self.model.training_loss
-            natgrad_opt.minimize(loss, variational_params)
-            adam_opt.minimize(loss, self.model.trainable_variables)
+# We now specify how Trieste will use our GPflow models within the BO loop.
+#
+# For our `GPR` model, we will use a standard L-BFGS optimizer from Scipy, whereas we will optimze our `VGP` model using altenate Adam steps (to optimize kernel parameter) and NatGrad steps (to optimize variational parameters).
```
vpicheny: alternate

docs/notebooks/failure_ego.pct.py
```diff
@@ -110,23 +110,11 @@ def create_classification_model(data):
 classification_model = create_classification_model(initial_data[FAILURE])
 
 # %% [markdown]
-# ## Create a custom optimize method
+# ## Build Trieste models
```
vpicheny, line 113: it's too bad we won't show any more how to create a custom optimize method. OK for this PR, maybe just create a note somewhere?

trieste/models/model_interfaces.py
```diff
@@ -341,15 +345,26 @@ class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel):
...
+    def __init__(
+        self, model: GPR | SGPR, optimizer: Optimizer | None = None, num_kernel_samples: int = 10
+    ):
         """
         :param model: The GPflow model to wrap.
         :param optimizer: The optimizer with which to train the model. Defaults to
             :class:`~trieste.models.optimizer.Optimizer` with :class:`~gpflow.optimizers.Scipy`.
+        :param num_kernel_samples: Number of randomly sampled kernels (for each kernel parameter) to
+            evaluate before beginning model optimization. Therefore, for a kernel with `p`
+            parameters, we evaluate `p * num_kernel_samples` kernels.
```
vpicheny: maybe this is slightly confusing? change to "for a kernel with p (vector-valued) parameters"

vpicheny (approved):
LGTM

PR #288, opened 2021-07-12
Adding a two-constraint case to the notebooks

vpicheny:
This PR adds another example at the end of the inequality notebook. It re-uses the previous single-constraint problem to make it more lightweight. We do not have a notebook that shows how to use a reducer, and this is an important practical case. It makes the notebook a bit longer.


vpicheny:
This is a very first draft, probably needs quite a bit of tidying up.

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -1,3 +1,18 @@
...
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.9.1
+#   kernelspec:
+#     display_name: trieste
+#     language: python
+#     name: trieste
+# ---
+
```
henrymoss, line 15: Whats this?

henrymoss: Is this new?

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -262,6 +277,93 @@ def batch_efi(at):
     mask_fail=batch_mask_fail.flatten()
 )
 
+# %% [markdown]
+# ## Multiple constraints
```
henrymoss: Constrained optimization with more than one cosntraint

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -262,6 +277,93 @@ def batch_efi(at):
...
+
+
+class Sim2(Sim):
+    threshold2 = 0.5
+
+    @staticmethod
+    def constraint2(input_data):
+        x, y = input_data[:, -2], input_data[:, -1]
+        z = tf.sin(x) * tf.cos(y) - tf.cos(x) * tf.sin(y)
+        return z[:, None]
```
henrymoss, line 278: Maybe write out the whole of this class rather than inherit the objective and constrain1 defined above. That might confuse the basic user?

docs/notebooks/inequality_constraints.pct.py
```diff
@@ -262,6 +277,93 @@ def batch_efi(at):
...
+        CONSTRAINT2: Dataset(query_points, Sim2.constraint2(query_points)),
+    }
+
+
+num_initial_points = 10
+initial_data = observer_two_constraints(search_space.sample(num_initial_points))
+initial_models = trieste.utils.map_values(create_bo_model, initial_data)
+
+# %% [markdown]
+# Now, the probability that the two constraints are feasible is the product of the two feasibilities. Hence, we combine the two `ProbabilityOfFeasibility` into one quantity by using a `Reducer`:
```
henrymoss: using a `Product` `Reducer`.

henrymoss (approved):

PR #289, opened 2021-07-13
Fix cts acquisition optimizer

henrymoss:
We now raise error (rather than crash) if we try to optimize our acq function with more restarts than initial samples.

I have also added a unit test to check this.

### DIFF
--- a/tests/unit/acquisition/test_optimizer.py
+++ b/tests/unit/acquisition/test_optimizer.py
@@ -132,6 +132,7 @@ def test_generate_continuous_optimizer_raises_with_invalid_init_params() -> None
         generate_continuous_optimizer(num_restarts=3),
         generate_continuous_optimizer(sigmoid=True),
         generate_continuous_optimizer(sigmoid=True, num_restarts=3),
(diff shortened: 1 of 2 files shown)


henrymoss:
Also, allowed it to work with num_init_samples=1 by removing a rogue squeeze.

tests/unit/acquisition/test_optimizer.py
```diff
@@ -166,7 +169,6 @@ def test_optimize_batch(
     points = batch_optimizer(search_space, acquisition)
     assert points.shape == [batch_size] + search_space.lower.shape
     for point in points:
-        print(point)
```
uri-granta, line 169: good catch!

trieste/acquisition/optimizer.py
```diff
@@ -123,6 +123,14 @@ def generate_continuous_optimizer(
     if num_restarts <= 0:
         raise ValueError(f"num_parallel must be positive, got {num_restarts}")
```
```suggestion
        raise ValueError(f"num_restarts must be positive, got {num_restarts}")
```

uri-granta (approved):
LGTM

PR #290, opened 2021-07-13
including prior density in kernel param init 

henrymoss:
Including kernel prior term in the likelihood when choosing kernel params


trieste/models/model_interfaces.py
```diff
@@ -453,7 +453,7 @@ def find_best_model_initialization(self, num_prior_samples) -> None:
         @tf.function
         def evaluate_likelihood_of_model_parameters() -> tf.Tensor:
             randomize_model_hyperparameters(self.model)
-            return self.model.maximum_log_likelihood_objective()
+            return self.model.maximum_log_likelihood_objective() + self.model.log_prior_density()
```
vpicheny, line 456: `-self.model._training_loss()`? or clean everything with loss, and min instead of max

PR #292, opened 2021-07-13
Get hartman dataset to pass

hstojic:


hstojic (commented):
all good, neat simplifciation of the code in networks.py :)

PR #294, opened 2021-07-14
Predictive variance acquisition

satrialoka:
Add predictive variance acquisition function for Active Learning
a continuation of #185

Feature Implemented:
PredictiveVariance acquisition function

Follow up:
After this PR, the bayesian active learning by disagreement (BALD) will be implemented

- [x] update notebooks
- [x] add ref

this will close #185


trieste/acquisition/function.py
```diff
@@ -1365,3 +1365,29 @@ def repulsion_term(
...
+        :param dataset: Unused.
+        :param model: The model over the specified ``dataset``.
+        :return: The determinant of the predictive function.
+        """
+        return lambda at: self.acquisition(model, at)
+
+    @staticmethod
+    def acquisition(model: ProbabilisticModel, at: TensorType) -> TensorType:
+        _, covariance = model.predict_joint(at)
+        return tf.linalg.det(covariance)
```
joelberkeley: any specific need for this public method? this could be a function in `prepare_acquisition_function`

satrialoka: no actually, yeah that'll be neat to make it a function inside prepare_acqusition_function

satrialoka: correct me if I am wrong, but making this public method is needed for test code no?
for example in [negative lcb](https://github.com/secondmind-labs/trieste/blob/4ef8d1317defca77b8f0dceed9ccefee894daf8f/tests/unit/acquisition/test_function.py#L375) test. so it should be a function **outside** of PredVar class. WDYT?

trieste/acquisition/function.py
```diff
@@ -1365,3 +1365,29 @@ def repulsion_term(
...
+        :param dataset: Unused.
+        :param model: The model over the specified ``dataset``.
+        :return: The determinant of the predictive function.
+        """
+
+        def determinantcovariance(model: ProbabilisticModel, at: TensorType) -> TensorType:
+            _, covariance = model.predict_joint(at)
+            return tf.linalg.det(covariance)
+
+        return lambda at: determinantcovariance(model, at)
```
joelberkeley: you can use `model` from the larger scope (indeed you're already doing this in the `lambda`)
```suggestion
        def determinantcovariance(at: TensorType) -> TensorType:
            _, covariance = model.predict_joint(at)
            return tf.linalg.det(covariance)

        return determinantcovariance
```

trieste/acquisition/function.py
joelberkeley: when you call `prepare...` you essentially get `determinantcovariance` back immediately, so you can test via `prepare...`, with  a stub model with custom `predict_joint`
```
class Model(ProbabilisticModel):
    def predict_joint(x):
        mean = <some function of x that helps you know this isn't being used>
        cov = <some function of x that makes it easy to test>
        return mean, cov
```
and use any old data (you could use some data that gives confidence it's not being used, but I'm not sure that's necessary).

i'm a big fan of keeping as much out of the public API as possible. It helps keep the intention of the API clear. In some cases, we exposed the underlying acquisition function as well as the `AcquisitionFunctionBuilder`. That might be because users might not want to do the work the builder does in the way the builder does it (EI can use different "best" points, for example), or it could be for ease of use (people are perhaps more used to thinking about LCB than negative LCB). In some cases, it may have just been a mistake. I hope we never did it to help with testing. I don't think it's good practice to make things public solely for testing purposes

satrialoka: Ah yeah, that makes sense. yeah, I think that'll be the best approach here. Thanks for the suggestion!

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,177 @@
+# %% [markdown]
+# # Active Learning
+
+# %% [markdown]
+# Sometimes rather than optimizing a black-box function, we want to learn the function better by reducing our model uncertainty. The one way to do this is by doing sampling in unknown inputs regions (active learning). This notebook demonstrates to perform Bayesian active learning using `trieste`.
```
henrymoss: Sometimes, we may just want to learn a black-box function, rather than optimizing it.  This goal is known as active learning and corresponds to choosing query points that reduce our model uncertainty. This notebook demonstrates to perform Bayesian active learning using `trieste`.

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,177 @@
...
+import tensorflow as tf
+import pandas as pd
+
+np.random.seed(1793)
+tf.random.set_seed(1793)
+
+# %% [markdown]
+# ## Describe the problem
+#
+# In this example, we will consider doing active learning for the log branin function.
```
henrymoss: In this example, we will perform active learning for the log Branin function

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+search_space = Box([0, 0], [1, 1])
+
+fig = plot_function_plotly(log_branin, search_space.lower, search_space.upper, grid_density=20)
+fig.update_layout(height=400, width=400)
+fig.show()
+
+# %% [markdown]
+# We begin our Bayesian active learning steps after evaluating two initial points sampled by Halton sequence.
```
henrymoss: We begin our Bayesian active learning from a two-point initial design built from a space-filling Halton sequence.

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,177 @@
...
+        },
+    }
+
+
+model = build_model(initial_data)
+
+# %% [markdown]
+# ## Active learning using predictive variance
+#
+# To do our first active learning example, we will use `PredictiveVariance`, one of the simplest active learning acquisition function. the `PredictiveVariance` sample a point on highest determinant of the predictive covariance of our model. Here, we will consider single query point using `PredictiveVariance`, the batch case will be shown later. We can utilize trieste's `BayesianOptimizer` to do the active learning steps.
```
henrymoss: For our first active learning example, we will use a simple acquisition function known as `PredictiveVariance` which chooses points for which we are highly uncertain (i.e. the predictive posterior covariance matrix at these points has large determinant).

We will now demonstrate how to choose individual query points using `PredictiveVariance` before moving onto batch active learning. For both cases, we can utilize trieste's `BayesianOptimizer` to do the active learning steps.

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,177 @@
...
+        "optimizer_args": {
+            "minimize_args": {"options": dict(maxiter=100)},
+        },
+    }
+
+
+model = build_model(initial_data)
+
+# %% [markdown]
+# ## Active learning using predictive variance
```
henrymoss, line 68: Can you add a ref here?

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,177 @@
...
+# %% [markdown]
+# Then we can retrieve our final dataset from the active learning steps
+
+# %%
+dataset = result.try_get_final_dataset()
+query_points = dataset.query_points.numpy()
+observations = dataset.observations.numpy()
+
+# %% [markdown]
+# Finally, we can check the performance of our `PredictiveVariance` active learning acquisition function by plotting the predictive variance landscape of our model. We can see how it samples on regions with high variance value.
```
henrymoss: Finally, we can check the performance of our `PredictiveVariance` active learning acquisition function by plotting the predictive variance landscape of our model. We can see how it samples regions for which our model is highly uncertain.

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,177 @@
...
+        xlabel="$X_1$",
+        ylabel="$X_2$",
+    )
+    plot_bo_points(query_points[: num_initial_points + i], ax[0, 0], num_initial_points)
+
+
+# %% [markdown]
+# ## Batch active learning using predictive variance
+#
+# For batch active learning using `PredictiveVariance`, we can set the `num_query_points` with the size of batch we want on the `EfficientGLobalOptimization` rule. The rest step is the same with the single query point of `PredictiveVariance` acquisition function.
```
henrymoss: For batch active learning, we must pass a `num_query_points` input to our `EfficientGLobalOptimization` rule.

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,177 @@
...
+result = bo.optimize(bo_iter, initial_data, model, rule, track_state=True)
+dataset = result.try_get_final_dataset()
+
+query_points = dataset.query_points.numpy()
+observations = dataset.observations.numpy()
+
+from util.plotting import plot_bo_points, plot_function_2d
+
+for i in range(bo_iter):
+
```
henrymoss: Need to describe this plot

trieste/acquisition/function.py
```diff
@@ -1365,3 +1365,29 @@ def repulsion_term(
...
+        """
+        :param dataset: Unused.
+        :param model: The model over the specified ``dataset``.
+        :return: The determinant of the predictive function.
+        """
+
+        def determinantcovariance(at: TensorType) -> TensorType:
+            _, covariance = model.predict_joint(at)
+            return tf.linalg.det(covariance)
+
```
henrymoss: I wonder if its worth being able to specify an optional jitter param. Also, might be worth returning log(det), as this can be calculated much more numerically stably in tensorflow (https://www.tensorflow.org/api_docs/python/tf/linalg/logdet) .

henrymoss (commented):
Have you thought about supporting greedy batch design for this acq function?

trieste/acquisition/function.py
satrialoka: yeah that'll be nice I think. I'll add it.

satrialoka:
> 
> 
> Have you thought about supporting greedy batch design for this acq function?

haven't thought about it, do you think it is necessary?

trieste/acquisition/function.py
satrialoka: got cholesky decomposition error when using `tf.linalg.logdet` for batch case. I'll investigate why this happens

henrymoss: When does the cholesky error pop up? Do you remember to exp afterwards?

henrymoss:
Don't bother with that for now

trieste/acquisition/function.py
satrialoka: ah yeah seems i forgot to exp

docs/notebooks/active_learning.pct.py
satrialoka: Couldnt find any paper that focuses on predictive variance
one candidate is this paper https://arxiv.org/pdf/0805.4359.pdf on chapter 2.2.1, they mention this a litle bit.
WDYT?

satrialoka: @henrymoss  any comment on this?

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,188 @@
+# %% [markdown]
+# # Active Learning
+
+# %% [markdown]
+# Sometimes, we may just want to learn a black-box function, rather than optimizing it. This goal is known as active learning and corresponds to choosing query points that reduce our model uncertainty. This notebook demonstrates to perform Bayesian active learning using `trieste`.
```
vpicheny: how to perform

docs/notebooks/active_learning.pct.py
vpicheny: Possibly it's simple enough that we don't need a ref. In the case of batch > 1, I don't think that using the determinant of the covmatrix is standard and we could use a reference.

vpicheny: maybe use 
MacKay, D. J. (1992). Information-based objective functions for active data selection. Neural computation, 4(4), 590-604.
instead?

vpicheny (approved):

trieste/acquisition/function.py
```diff
@@ -1963,3 +1963,30 @@ def _get_min_value_samples(
...
+        """"""
+        return "PredictiveVariance()"
+
+    def prepare_acquisition_function(
+        self, dataset: Dataset, model: ProbabilisticModel, jitter: float = DEFAULTS.JITTER
+    ) -> AcquisitionFunction:
+        """
+        :param dataset: Unused.
+        :param model: The model over the specified ``dataset``.
+        :return: The determinant of the predictive function.
```
apaleyes: `jitter` is not in the docstring

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,188 @@
...
+
+    _, ax = plot_function_2d(
+        pred_var,
+        search_space.lower - 0.01,
+        search_space.upper + 0.01,
+        grid_density=20,
+        contour=True,
+        colorbar=True,
+        figsize=(10, 6),
+        title=["Variance contour with queried points at iter:" + str(i + 1)],
```
apaleyes:
```suggestion
        title=["Variance contour with queried points at iter: " + str(i + 1)],
```

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,188 @@
...
+        xlabel="$X_1$",
+        ylabel="$X_2$",
+    )
+    plot_bo_points(query_points[: num_initial_points + i], ax[0, 0], num_initial_points)
+
+
+# %% [markdown]
+# ## Batch active learning using predictive variance
+#
+# For batch active learning, we must pass a num_query_points input to our `EfficientGLobalOptimization` rule.
```
apaleyes:
```suggestion
# For batch active learning, we must pass a num_query_points input to our `EfficientGlobalOptimization` rule.
```

docs/notebooks/active_learning.pct.py
hstojic: I don't see the MacKay reference here in the text? can you please add it here as well, not just in `function.py`?

trieste/acquisition/function.py
```diff
@@ -1963,3 +1963,32 @@ def _get_min_value_samples(
...
+    ) -> AcquisitionFunction:
+        """
+        :param dataset: Unused.
+        :param model: The model over the specified ``dataset``.
+        :param jitter: The size of the jitter to use when stabilising the Cholesky decomposition of
+            the covariance matrix.
+        :return: The determinant of the predictive function.
+        """
+
+        def determinantcovariance(at: TensorType) -> TensorType:
```
hstojic: ideally, here you would use the `@tf.function` decorator, right @uri-granta?

for consistency with the rest of the acq functions codebase, could you please move out this function outside of the class, similar to say `lower_confidence_bound` function - this would also facilitate testing, no need to use a builder, so you would test separately builder and the function itself

trieste/acquisition/function.py
```diff
@@ -1963,3 +1963,32 @@ def _get_min_value_samples(
...
+            _, covariance = model.predict_joint(at)
+            return tf.exp(tf.linalg.logdet(covariance + jitter))
+
+        return determinantcovariance
```
hstojic: I think it might be good to define `update_acquisition_function` and return `function` to avoid `prepare_acquisition_function` call

trieste/acquisition/function.py
```diff
@@ -1963,3 +1963,32 @@ def _get_min_value_samples(
...
+        self, dataset: Dataset, model: ProbabilisticModel, jitter: float = DEFAULTS.JITTER
+    ) -> AcquisitionFunction:
+        """
+        :param dataset: Unused.
+        :param model: The model over the specified ``dataset``.
+        :param jitter: The size of the jitter to use when stabilising the Cholesky decomposition of
+            the covariance matrix.
+        :return: The determinant of the predictive function.
+        """
+
```
hstojic, line 2128: it would be good to assert that the model has `predict_joint` method and return a user-friendly message, as only some models will have this method implemented

trieste/acquisition/function.py
```diff
@@ -1963,3 +1963,32 @@ def _get_min_value_samples(
         sampler = ExactThompsonSampler(num_samples, model, sample_min_value=True)
 
     return sampler.sample(sampled_points)
+
+
+class PredictiveVariance(SingleModelAcquisitionBuilder):
+    """
+    Builder for the determinant of the predictive covariance matrix over the batch points.
```
hstojic, line 2104: please add a note that the model needs to be able to supply covariance of the joint marginal distribution (which is also quite expensive to compute)

trieste/acquisition/function.py
uri-granta: Making the acquisition functions support in pace updates is not always trivial (and I don't think we should insist on it for every PR, as we can always add it later). However, in this case there isn't anything that needs updating, so you can have a trivial update_acquisition_function like in NegativeLowerConfidenceBound.

That said, you should probably move the jitter parameter to the initialiser (like in BatchMonteCarloExpectedImprovement) rather than in prepare_acqusition_function, as it doesn't belong in that function signature.

Hope that makes sense!

tests/unit/acquisition/test_function.py
```diff
@@ -1726,3 +1727,33 @@ def acquisition(x: TensorType) -> TensorType:
...
+def test_predictive_variance_builder_builds_predictive_variance() -> None:
+    model = QuadraticMeanAndRBFKernel()
+    acq_fn = PredictiveVariance().prepare_acquisition_function(
+        Dataset(tf.zeros([0, 1]), tf.zeros([0, 1])), model
+    )
+    query_at = tf.linspace([[-10]], [[10]], 100)
+    _, covariance = model.predict_joint(query_at)
+    expected = tf.linalg.det(covariance)
+    npt.assert_array_almost_equal(acq_fn(query_at), expected)
+
```
hstojic, line 1774: please add a test for raising an exception if the model doesn't implement `predict_joint`, e.g. GPflux models do not implement it, even though there is a method (required by the `ProbabilisticModel`

hstojic: also please add `PredictiveVariance` to the `test_single_model_acquisition_function_builder_reprs`

hstojic: next, we would need a test that it gives correct values, not that it just builds an acquisition function, see e.g. `test_expected_improvement` or `test_lower_confidence_bound`

trieste/acquisition/function.py
hstojic: +1 for moving out jitter argument

tests/unit/acquisition/test_function.py
hstojic: finally, if you are willing to delve with `tf.function`, we would need smth like `test_negative_lower_confidence_bound_builder_updates_without_retracing`, I think copying this with minimal modifications should do the trick

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,188 @@
...
+from trieste.models.gpflow import GPflowModelConfig
+
+
+def build_model(data):
+    variance = tf.math.reduce_variance(data.observations)
+    kernel = gpflow.kernels.RBF(variance=variance, lengthscales=[2, 2])
+    gpr = gpflow.models.GPR(data.astuple(), kernel, noise_variance=1e-5)
+    gpflow.set_trainable(gpr.likelihood, False)
+
+    return GPflowModelConfig(
```
hstojic: Lets simplify this and use `GaussianProcessRegression` wrapper directly instead of `GPflowModelConfig`

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,188 @@
...
+        }
+    )
+
+
+model = build_model(initial_data)
+
+# %% [markdown]
+# ## Active learning using predictive variance
+#
+# For our first active learning example, we will use a simple acquisition function known as `PredictiveVariance` which chooses points for which we are highly uncertain (i.e. the predictive posterior covariance matrix at these points has large determinant).
```
hstojic: can you please add a note that a model needs to have `predict_joint` method, i.e. be able to return full covariance - not all models will have that and likely to be expensive

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,188 @@
...
+query_points = dataset.query_points.numpy()
+observations = dataset.observations.numpy()
+
+# %% [markdown]
+# Finally, we can check the performance of our `PredictiveVariance` active learning acquisition function by plotting the predictive variance landscape of our model. We can see how it samples regions for which our model is highly uncertain.
+
+# %%
+from util.plotting import plot_bo_points, plot_function_2d
+
+for i in range(bo_iter):
```
hstojic: you could make this a function, since you use exactly the same code below to plot batch results

docs/notebooks/active_learning.pct.py
```diff
@@ -0,0 +1,188 @@
...
+# After that, we can retrieve our final dataset.
+
+# %%
+dataset = result.try_get_final_dataset()
+query_points = dataset.query_points.numpy()
+observations = dataset.observations.numpy()
+
+
+# %% [markdown]
+# Now we can visualize the batch predictive variance using our plotting function.
```
hstojic, line 165: perhaps you could mention what kind of differences you could expect with using the batch method, say if you use same number of points above and total points here

hstojic (commented):
I have left some suggestions for improvements, let me know if you have any questions

trieste/acquisition/function.py
```diff
@@ -2052,26 +2052,62 @@ class PredictiveVariance(SingleModelAcquisitionBuilder):
...
+                """
+                PredictiveVariance only supports models with a predict_joint method.
+                """
+            )
 
-        def determinantcovariance(at: TensorType) -> TensorType:
-            _, covariance = model.predict_joint(at)
-            return tf.exp(tf.linalg.logdet(covariance + jitter))
+        pv = determinantcovariance(model, self._jitter)
+        return tf.function(lambda at: pv(at))
```
hstojic: you don't need another `tf.function` here I think, right @uri-granta?

uri-granta: That's right: you can just return `determinantcovariance(model, self._jitter)` directly: no need to wrap it up in a lambda (unlike NegativeLowerConfidenceBound, which is negating the lcb).

tests/unit/acquisition/test_function.py
satrialoka: > please add a test for raising an exception if the model doesn't implement `predict_joint`, e.g. GPflux models do not implement it, even though there is a method required by the `ProbabilisticModel`. 

I was using this:

       if not hasattr(model, "predict_joint"):
            raise AttributeError(
                """
                PredictiveVariance only supports models with a predict_joint method.
                """
            )

but it doesn't pass the raise test since the `predicte_joint ` indeed is there in the base class. another option is to try-catch inside the function(not the builder), but it seems to do this inside the builder is much preferable, what are our options here?

docs/notebooks/active_learning.pct.py
satrialoka: okay 👌🏼

trieste/acquisition/function.py
satrialoka: I see, okay 👌🏼

tests/unit/acquisition/test_function.py
satrialoka: maybe @uri-granta  have any suggestion?

hstojic: yes, `ProbabilisticModel` will always have the method, but those models that do not have it should keep the `raise NotImplementedError` so try checking that

satrialoka: This is exactly i don't know how to do, 
some alternatives are checking the method's code by using  `from inspect import getsourcelines` to check whether there is a `NotImplementedError` inside it or not, but importing new lib like this seems not a neat way to do it.

hstojic: Ok, I can chip in here with this test - can you please resolve all other comments in the meantime?

satrialoka: okay 👌🏼

hstojic: great, let me know when you are done!

hstojic: ok, done

hstojic:
I think this should be ready, if tests pass I'll be happy to approve and merge

trieste/acquisition/function.py
```diff
@@ -2125,8 +2125,11 @@ def prepare_acquisition_function(
 
         :return: The determinant of the predictive function.
         """
-        if not hasattr(model, "predict_joint"):
-            raise AttributeError(
+        
+        try:
+            model.predict_joint(dataset.query_points[0:])
+        except NotImplementedError:
+            raise ValueError(
```
satrialoka: Ah yeah we can try-catch inside the builder, was thinking doing this will be expensive compared to doing try-catch inside the function. my bad.

satrialoka:
I am happy with the current changes (I add a correction on the notebook text). I think we're ready to merge

trieste/acquisition/function.py
hstojic: I'm worried about that as well, it could be unnecessarily expensive, even for a single point if its a big dataset - hm, probably we should move it to the acq function itself, it will raise it only on starting to acquire points but its worth the savings - can you please do that?

satrialoka: Sure!

trieste/acquisition/function.py
```diff
@@ -2097,3 +2097,71 @@ def _get_min_value_samples(
...
+
+    :param model: The model of the objective function.
+    :param jitter: The size of the jitter to use when stabilising the Cholesky decomposition of
+            the covariance matrix.
+    """
+
+    @tf.function
+    def acquisition(x: TensorType) -> TensorType:
+
+        try:
```
hstojic, line 2157: @uri-granta is try-except ok inside tf.function?

hstojic: checking this in a builder seems unnecessarily costly

hstojic (approved):
ok, lets merge this in

PR #295, opened 2021-07-14
ordinal space 

satrialoka:
ordinal space implementation

TODO:

- [x] implement ordinal space
- [x] add test


henrymoss:
How are the tests coming along?

satrialoka:
> 
> 
> How are the tests coming along?

working on it, I am thinking to add rounding checks beside the Box like tests

trieste/space.py
```diff
@@ -339,3 +339,117 @@ def __mul__(self, other: Box) -> Box:
 
     def __deepcopy__(self, memo: dict[int, object]) -> Box:
         return self
+
+
+class OrdinalSearchSpace(Box):
+    r"""
+    Ordinal search space representing a :math:`D`-dimensional box in
```
henrymoss: a *discrete* D-dimensional ...

henrymoss: Say something like: Generates evenly-spaced values in an interval for each dimension

trieste/space.py
```diff
@@ -339,3 +339,117 @@ def __mul__(self, other: Box) -> Box:
...
+        :raise ValueError (or InvalidArgumentError): If any of the following are true:
+
+            - ``lower``, ``upper`` and ``stepsizes`` have invalid shapes.
+            - ``lower``, ``upper`` and ``stepsizes`` do not have the same floating point type.
+            - ``upper`` is not greater than ``lower`` across all dimensions.
+        """
+
+        super().__init__(lower, upper)
+        tf.debugging.assert_shapes([(stepsizes, ["D"])])
+        tf.assert_rank(lower, 1)
```
henrymoss: this assert already done in __init__

trieste/space.py
```diff
@@ -339,3 +339,117 @@ def __mul__(self, other: Box) -> Box:
...
+
+    def __repr__(self) -> str:
+        """"""
+        return f"OrdinalSearchSpace({self._lower!r}, {self._upper!r}, {self._stepsizes!r})"
+
+    def sample(self, num_samples: int) -> TensorType:
+        """
+        :param num_samples: The number of points to sample from this search space.
+        :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
+            from this search space. Rounded by the ``stepsize``.
```
henrymoss: Get rid of "rounded by ..."

trieste/space.py
```diff
@@ -339,3 +339,117 @@ def __mul__(self, other: Box) -> Box:
...
+        return samples
+
+    def sample_sobol(self, num_samples: int, skip: Optional[int] = None) -> TensorType:
+        """
+        :param num_samples: The number of points to sample from this search space.
+        :return: ``num_samples`` of points, using sobol sequence with shape '[num_samples, D]' ,
+            where D is the search space dimension. Rounded by the ``stepsize``.
+        """
+        samples = super().sample_sobol(num_samples, skip)
+        samples = tf.round(samples / self._stepsizes) * self._stepsizes
```
henrymoss: I'm not sure if we should have these. Maybe ask Ivo what he thinks. This rounding means that they are not really sobol or halton anymore.

trieste/space.py
henrymoss: Give an example like "OrdinalSearchSpace([-2.0, -2.0], [2.0, 2.0], [0.2, 0.3])" provide the space with elements ....

trieste/space.py
```diff
@@ -339,3 +339,117 @@ def __mul__(self, other: Box) -> Box:
...
+            dtypes to those of the other :class:`OrdinalSearchSpace`.
+        """
+        if self.lower.dtype is not other.lower.dtype:
+            return NotImplemented
+
+        product_lower_bound = tf.concat([self._lower, other.lower], axis=-1)
+        product_upper_bound = tf.concat([self._upper, other.upper], axis=-1)
+        product_stepsizes = tf.concat([self._stepsizes, other.stepsizes], axis=-1)
+
+        return OrdinalSearchSpace(product_lower_bound, product_upper_bound, product_stepsizes)
```
henrymoss, line 482: You need to overwrite the __contains  method that you have inherited from box as it doesnt work atm

tests/unit/test_space.py
```diff
@@ -423,3 +423,71 @@ def test_box_deepcopy() -> None:
...
+        pytest.param([0.0, 1.0, 2.0], [1.0, 2.0, 3.0], [0.1, 0.2, 0.3]),
+        pytest.param([1.0, 10.0], [10.0, 100.0], [2, 10]),
+    ],
+)
+def test_ordinalsearchspace_halton_sampling_return_correct_rounded_points(lower, upper, stepsizes):
+    ordinalsp = OrdinalSearchSpace(lower, upper, stepsizes)
+    box = Box(lower, upper)
+    ordinalsample = ordinalsp.sample_halton(5, seed=0)
+    boxsample = box.sample_halton(5, seed=0)
+    npt.assert_allclose(ordinalsample, tf.round(boxsample / stepsizes) * stepsizes)
```
henrymoss: You are missing lots of tests (recreate the ones we have for box)

1.  test_ordinal_raises_if_stepsizes_have_invalid_shape
3. test_ordinal_bounds_and_stepsize_attributes()
4. test_ordinal_contains_point
5. test_ordinal_does_not_contain_point
6. test_ordinal___mul___bounds_are_the_concatenation_of_original_bounds()
7. test_ordinal___mul___raises_if_bounds_have_different_types

satrialoka: I'll write these, thanks for listing the test functions!

trieste/space.py
satrialoka: hmm it's quite tricky to write this membership function, I am not sure how to check if the points match with the interval since modulo is not working on floating point. any idea?

henrymoss: Not sure, you will have to think about it. But it is important to have this functionality. Maybe ask Ivo

satrialoka: okay 👌🏼

henrymoss:
Whats left on this PR?

satrialoka:
> 
> 
> Whats left on this PR?

The __contains method and its test, I am thinking of a "hacky" way to do this using `math.isclose`, I'll try it tomorrow.

henrymoss:
Is this done?

satrialoka:
> Is this done?

yes I suppose, but better to review it especially the \_contains\_ method

trieste/space.py
```diff
@@ -331,3 +332,138 @@ def __mul__(self, other: Box) -> Box:
...
+            from the search space.
+        """
+        if not shapes_equal(value, self._lower):
+            raise ValueError(
+                f"value must have same dimensionality as search space: {self._lower.shape},"
+                f" got shape {value.shape}"
+            )
+
+        isclose_array = np.vectorize(np.isclose)
+
```
henrymoss: Can this not be done in tf?

satrialoka: they have `tf.experimental.numpy.isclose ` in tensorflow. and this function vectorization might be possible with `tf.map_fn`. what do you think?

henrymoss: yep give it a go. We dont want numpy

satrialoka: 👌🏼

tests/unit/test_space.py
```diff
@@ -423,3 +423,121 @@ def test_box_deepcopy() -> None:
...
+
+
+@pytest.mark.parametrize(
+    "lower, upper, stepsizes",
+    [
+        pytest.param([0.0, 1.0], [1.0, 2.0], [0.1, 0.2]),
+        pytest.param([0.0, 1.0, 2.0], [1.0, 2.0, 3.0], [0.1, 0.2, 0.3]),
+        pytest.param([1.0, 10.0], [10.0, 100.0], [2, 10]),
+    ],
+)
```
henrymoss, line 487: Does this work for step size that is a fraction?

trieste/space.py
```diff
@@ -331,3 +331,140 @@ def __mul__(self, other: Box) -> Box:
...
+    @property
+    def stepsizes(self) -> TensorType:
+        """the step sizes of the ordinal space"""
+        return self._stepsizes
+
+    def __contains__(self, value: TensorType) -> bool | TensorType:
+        """
+        Return `True` if ``value`` is a member of this search space, else `False`. A point is a
+        member if all of its coordinates lie in the closed intervals bounded by the lower and upper
+        bounds and spaced by some step sizes.
```
henrymoss: lie on the discretization specified by the given lower and upper bounds and step size.

trieste/space.py
```diff
@@ -331,3 +331,140 @@ def __mul__(self, other: Box) -> Box:
...
+                f" got shape {value.shape}"
+            )
+
+        return (
+            tf.reduce_all(value >= self._lower)
+            and tf.reduce_all(value <= self._upper)
+            and tf.reduce_all(
+                tf.map_fn(
+                    lambda x: tf.experimental.numpy.isclose(x[0], x[1]),
+                    (value / self._stepsizes, tf.round(value / self._stepsizes)),
```
henrymoss, line 448: What is this actually doing?

henrymoss (commented):
Nearly there! Thanks for all your work on this!

trieste/space.py
satrialoka: A point on the ordinal space should be divisible by its step size, `value % stepsizes == 0`. But, modulo doesn't work well on float. As you know, there some limitation on  [floating point arithmetic](https://docs.python.org/3/tutorial/floatingpoint.html). So I just take `isclose()` between the result of the division with its nearest integer, the floating point error should be a very small numerical difference between them if the value is divisible by stepsizes. 
Not sure yet what will goes wrong with this approach.

tests/unit/test_space.py
satrialoka: not sure it will, using fraction is quite risky since there will be a chance that a floating point error present on the result, hence decimal is preferred.

henrymoss: Sorry, I meant decimal. Doesn't seem to be tested? so have a step size of 0.1 for example

satrialoka: ah as you see there the couple first test points are indeed using decimal

henrymoss: 1.0, 10.0 and 100.0 are still integers. I mean a step size of fraction (expressed as a decimal), like 0.1 or 0,2. i.e. we might want a space of [0,0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9,1]

satrialoka: yes it's on line 449 and 450. the stepsizes are the third list of the pytest.param
```
        pytest.param([0.0, 1.0], [1.0, 2.0], [0.1, 0.2]),
        pytest.param([0.0, 1.0, 2.0], [1.0, 2.0, 3.0], [0.1, 0.2, 0.3]),
```
is this what you mean?

henrymoss: OFC. Sorry was being stupid

satrialoka: No problem

trieste/space.py
```diff
@@ -331,3 +331,140 @@ def __mul__(self, other: Box) -> Box:
 
     def __deepcopy__(self, memo: dict[int, object]) -> Box:
         return self
+
+
+class OrdinalSearchSpace(Box):
+    r"""
+    Ordinal search space representing a discrete :math:`D`-dimensional box in
+    :math:`\mathbb{R}^D`. This class inherit :class:`Box` and generates evenly-spaced value
```
```suggestion
    :math:`\mathbb{R}^D`. This class inherit :class:`Box` and generates evenly-spaced values
```

trieste/space.py
```diff
@@ -331,3 +331,140 @@ def __mul__(self, other: Box) -> Box:
...
+        :raise ValueError (or InvalidArgumentError): If any of the following are true:
+
+            - ``lower``, ``upper`` and ``stepsizes`` have invalid shapes.
+            - ``lower``, ``upper`` and ``stepsizes`` do not have the same floating point type.
+            - ``upper`` is not greater than ``lower`` across all dimensions.
+        """
+
+        super().__init__(lower, upper)
+        tf.debugging.assert_shapes([(lower, ["D"]), (stepsizes, ["D"])])
+        tf.assert_rank(stepsizes, 1)
```
uri-granta, line 384: Should we check that the stepsize values aren't 0 or negative?

trieste/space.py
```diff
@@ -331,3 +331,140 @@ def __mul__(self, other: Box) -> Box:
...
+        return f"OrdinalSearchSpace({self._lower!r}, {self._upper!r}, {self._stepsizes!r})"
+
+    def sample(self, num_samples: int) -> TensorType:
+        """
+        :param num_samples: The number of points to sample from this search space.
+        :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
+            from this search space.
+        """
+        samples = super().sample(num_samples)
+        samples = tf.round(samples / self._stepsizes) * self._stepsizes + 0
```
uri-granta, line 414: Mightn't this produce values outside the box? E.g. wouldn't `OrdinalSearchSpace([0.0], [1.0], [0.6])` round a 0.95 sample to 1.2?

(If that's a problem then do we need to make sure that the solution doesn't skew the sampling distribution? E.g. if we just round 0.95 down to 0.6 then that'd make 0.6 more likely than 0.0. Perhaps we should initialise lower and upper to be the smallest/biggest multiples of stepsize in the box, and document that in the docstring?)

uri-granta (commented):
Looks good (and good testing too). Just a couple of questions.

trieste/space.py
satrialoka: Make sense, I'll add it.

trieste/space.py
satrialoka: Yeah it produce values outside the box, my bad.

> Perhaps we should initialise lower and upper to be the smallest/biggest multiples of stepsize in the box, and document that in the docstring?)

do you mean we should replace the user predefined bound to that smallest/biggest multiples? 
what if we just filter out of bound point and replace it with lower/upper bound? this will make the box not "evenly spaced" though.

uri-granta: Yes, my suggestion was changing the bounds so that they're multiples. @henrymoss Do you have an opinion on this?

henrymoss: Yep . Lets fix stepsizes to be multiples!

satrialoka: added, please check. however, it seems need more test points inside `test_ordinalsearchspace_bounds_changes_to_nearest_multiples` , since there might be some floating-point problems.

trieste/space.py
```diff
@@ -331,3 +331,149 @@ def __mul__(self, other: Box) -> Box:
...
+        self,
+        lower: Sequence[float] | TensorType,
+        upper: Sequence[float] | TensorType,
+        stepsizes: Sequence[float] | TensorType,
+    ):
+
+        r"""
+        If ``lower``, ``upper`` and ``stepsizes`` are `Sequence`\ s of floats
+        (such as lists or tuples), they will be converted to tensors of dtype `tf.float64`.
+        if ``lower`` and ``upper`` are not multiple of ``stepsizes`` it will be replaced by
```
```suggestion
        if ``lower`` and ``upper`` are not multiple of ``stepsizes``, they will be replaced by
```

trieste/space.py
```diff
@@ -331,3 +331,149 @@ def __mul__(self, other: Box) -> Box:
...
+        :param stepsizes: The step size of each dimension of the Ordinal Space. Must have
+            shape [D] for positive D, and if a tensor, must have float type.
+        :raise ValueError (or InvalidArgumentError): If any of the following are true:
+
+            - ``lower``, ``upper`` and ``stepsizes`` have invalid shapes.
+            - ``lower``, ``upper`` and ``stepsizes`` do not have the same floating point type.
+            - ``upper`` is not greater than ``lower`` across all dimensions.
+        """
+
+        super().__init__(lower, upper)
```
uri-granta, line 381: This doesn't make sense here if we're potentially changing the lower/upper bounds later on. Maybe move it to the end of the function and pass in the updated values of lower/upper instead? (you can use normal local variables for those too, as it's `super()`'s responsibility to set the private variables)

This will also catch the case where lower is rounded up and upper is rounded down, leading to lower > upper. You should probably test this by checking that e.g. the following raises an error:

```python
OrdinalSearchSpace([0.1], [0.9], [1.0])
```

trieste/space.py
```diff
@@ -331,3 +331,149 @@ def __mul__(self, other: Box) -> Box:
...
+        tf.debugging.assert_positive(len(stepsizes), message="stepsizes cannot be empty")
+
+        if isinstance(stepsizes, Sequence):
+            self._stepsizes = tf.constant(stepsizes, dtype=tf.float64)
+        else:
+            self._stepsizes = tf.convert_to_tensor(stepsizes)
+
+            tf.debugging.assert_same_float_dtype([self._lower, self._stepsizes])
+
+        self._lower = tf.math.ceil(tf.divide(self._lower, self._stepsizes)) * self._stepsizes
```
uri-granta: FYI I think this correctly handles floating point errors: `tf.math.ceil` rounds 5.0000001 to 5, not 6, as it's very close to 5. Therefore, while `math.ceil((5/3)/(1/3))` return `6`, `tf.math.ceil(tf.divide(5/3, 1/3))` returns 5.

Maybe update `test_ordinalsearchspace_bounds_changes_to_nearest_multiples` so it checks `OrdinalSearchSpace([5/3],[2.0],[1/3])` too.

uri-granta (commented):
Just a couple more comments. Almost good to check in! (also sorry for the slow response, was away on holiday)

trieste/space.py
satrialoka: tried to do that, just realized that `super()` also responsible to change the dtype of _upper and _lower, thus if I put it at the end of the __init__ the dtype equality tests fails. Should we add Box's upper and lower dtype conversion here? or we can re do lower>upper assert here since it less redundant. what do you think? 
I am not quite sure how to make sure that we always have same dtype.

uri-granta: Another possibility then is to leave super() where it is, but then call it _again_ at the end of the function with the rounded bounds?

trieste/space.py
satrialoka: checked on python terminal yes it does work, but strangely that test fail (maybe because of dtype?). 
fixed by subtracting the term inside ceil with small jitter though:
 `self._lower = tf.math.ceil(tf.divide(self._lower, self._stepsizes) - 1e-6) * self._stepsizes`

trieste/space.py
satrialoka: sorry wasn't aware of your last comment. I'll try it.

satrialoka: Done!

hstojic:
this PR is now obsolete I think with the new support for product spaces
@satrialoka are you fine with closing this one?

satrialoka:
> this PR is now obsolete I think with the new support for product spaces @satrialoka are you fine with closing this one?

Sure!

PR #297, opened 2021-07-15
Sample from constrained kernel params before model fitting

henrymoss:
This PR allows our kernel initialization routine for GPR models to sample kernel parameters with (Sigmoid) constraints. Previously, this routine only sampled parameters with hyper-priors.

Our sampling strategy follows the following hierarchy

1. If a param is constrained with a sigmoid bijector then we sample from its constrained domain (ignoring its priors if it has them).
2. If a param is not constrained but has a prior then we sample from the prior
3. otherwise we do not sample the param and fix its values to its initialization.

This new change will be detailed in my imminent notebook overhaul.

### DIFF
--- a/tests/unit/models/test_model_interfaces.py
+++ b/tests/unit/models/test_model_interfaces.py
@@ -754,3 +754,25 @@ def test_randomize_model_hyperparameters_randomizes_kernel_parameters_with_const
     npt.assert_raises(AssertionError, npt.assert_allclose, 1.0, kernel.variance)
     npt.assert_array_equal(dim, kernel.lengthscales.shape)
     npt.assert_raises(AssertionError, npt.assert_allclose, [0.2] * dim, kernel.lengthscales)
+
+
+@random_seed
+@pytest.mark.parametrize("dim", [1, 10])
+def test_randomize_model_hyperparameters_samples_from_constraints_when_given_prior_and_constraint(
+    dim: int,
+) -> None:
+    kernel = gpflow.kernels.RBF(variance=1.0, lengthscales=[0.2] * dim)
+    upper = tf.cast([0.5] * dim, dtype=tf.float64)
+
+    lower = upper / 100
+    kernel.lengthscales = gpflow.Parameter(
+        kernel.lengthscales, transform=tfp.bijectors.Sigmoid(low=lower, high=upper)
+    )
+    kernel.lengthscales.prior = tfp.distributions.Uniform(low=10.0, high=100.0)
+
+    kernel.variance.prior = tfp.distributions.LogNormal(loc=np.float64(-2.0), scale=np.float64(1.0))
+
+    randomize_model_hyperparameters(kernel)
+
+    npt.assert_array_less(kernel.lengthscales, [0.5] * dim)
+    npt.assert_raises(AssertionError, npt.assert_allclose, [0.2] * dim, kernel.lengthscales)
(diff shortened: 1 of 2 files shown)


henrymoss:
To use this in practice, do something like
 `upper = (search_space.upper-search_space.lower)*tf.math.sqrt(tf.cast(len(search_space.upper),dtype=tf.float64))`
`lower = upper / 100`
```
model.kernel.lengthscales = gpflow.Parameter(
        model.kernel.lengthscales, transform=tfp.bijectors.Sigmoid(low=lower, high=upper)
    )
```

when defining your model

trieste/models/model_interfaces.py
```diff
@@ -424,30 +425,47 @@ def optimize(self, dataset: Dataset) -> None:
...
+        the kernel. The best configuration is used as the starting point for model optimization.
+
+        For trainable parameters constrained to lie in a finite interval (through a sigmoid
+        bijector), we begin model optimization from the best of a random sample from these
+        parameters' acceptable domains.
+
+        For trainable parameters without constraints but with priors, we begin model optimization
+        from the best of a random sample from these parameters' priors.
+
+        For trainable parameters with neither priors or constraints, we begin optimization from
```
uri-granta:
```suggestion
        For trainable parameters with neither priors nor constraints, we begin optimization from
```

tests/unit/models/test_model_interfaces.py
```diff
@@ -406,27 +417,62 @@ def test_find_best_model_initialization_only_changes_params_with_priors(
...
+    x = tf.constant(np.arange(1, 5).reshape(-1, 1), dtype=gpflow.default_float())  # shape: [4, 1]
+    model = gpr_interface_factory(x, _3x_plus_10(x))
+    model.model.kernel = gpflow.kernels.RBF(lengthscales=[0.2] * dim)
+
+    if isinstance(model, (VariationalGaussianProcess, SparseVariational)):
+        pytest.skip("find_best_model_initialization is only implemented for the GPR models.")
+
+    upper = tf.cast([10.0] * dim, dtype=tf.float64)
+    lower = upper / 100
+    model.model.kernel.lengthscales = gpflow.Parameter(
```
uri-granta, line 439: Have you tested parameters with both constraints _and_ priors?

uri-granta (approved):

tests/unit/models/test_model_interfaces.py
vpicheny: is this something gpflow would allow? if it is I agree this is an important test, we need to make sure it either returns an error or if follows the expected hierarchy.

trieste/models/model_interfaces.py
```diff
@@ -589,10 +607,19 @@ def _assert_data_is_compatible(new_data: Dataset, existing_data: Dataset) -> Non
...
+    Sets hyperparameters to random samples from their constrained domains or (if not constraints
+    are available) their prior distributions.
 
     :param model: Any GPModel from gpflow.
     """
-    for parameter in model.trainable_parameters:
-        if parameter.prior is not None:
-            parameter.assign(parameter.prior.sample())
+    for param in model.trainable_parameters:
+        if isinstance(param.bijector, tfp.bijectors.Sigmoid):
```
vpicheny, line 616: do we have other possible transformations? or is Sigmoid the only case to consider?

tests/unit/models/test_model_interfaces.py
henrymoss: Gpflow allows it. I will add the test.

PR #298, opened 2021-07-15
Standardise data

johnamcleod:


setup.py
```diff
@@ -43,5 +43,6 @@
         # tensorflow!=2.2.0,!=2.3.0 because of https://github.com/advisories/GHSA-8fxw-76px-3rxv
         "tensorflow>=2.1,!=2.2.0,!=2.3.0",
         "tensorflow-probability>=0.9",
+        "sklearn"
```
hstojic, line 46: huh, I'm a bit reluctant to add this dependency just for this standardization thingie :|

tests/integration/models/keras/test_predictions.py
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pytest
 import numpy as np
 import tensorflow as tf
+from sklearn.preprocessing import StandardScaler
```
hstojic, line 17: I think this one assumes Gaussian distribution of the variables? min-max normalization would probably work better?

PR #299, opened 2021-07-15
use tf.debugging for search space assertions

joelberkeley:
Use `tf.debugging` for assertions in space.py, so that error handling works in `tf.function`.

Also changed an implementation while I was at it

### DIFF
--- a/trieste/space.py
+++ b/trieste/space.py
@@ -114,7 +114,7 @@ class DiscreteSearchSpace(SearchSpace):
         :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
             from this search space.
         """
-        tf.debugging.assert_less_equal(num_samples, self._points.shape[0])
+        tf.debugging.assert_less_equal(num_samples, len(self._points))
         return tf.random.shuffle(self._points)[:num_samples, :]
 
     def __mul__(self, other: DiscreteSearchSpace) -> DiscreteSearchSpace:
--- a/trieste/space.py
+++ b/trieste/space.py
@@ -117,7 +117,7 @@ class DiscreteSearchSpace(SearchSpace):
         tf.debugging.assert_less_equal(
             num_samples,
             len(self._points),
-            message="Number of samples cannot be greater than the number of points in search space"
+            message="Number of samples cannot be greater than the number of points in search space",
         )
         return tf.random.shuffle(self._points)[:num_samples, :]
 
--- a/trieste/space.py
+++ b/trieste/space.py
@@ -114,7 +114,11 @@ class DiscreteSearchSpace(SearchSpace):
         :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
             from this search space.
         """
-        tf.debugging.assert_less_equal(num_samples, len(self._points))
+        tf.debugging.assert_less_equal(
+            num_samples,
+            len(self._points),
+            message="Number of samples cannot be greater than the number of points"
+        )
         return tf.random.shuffle(self._points)[:num_samples, :]
 
     def __mul__(self, other: DiscreteSearchSpace) -> DiscreteSearchSpace:
--- a/trieste/space.py
+++ b/trieste/space.py
@@ -14,7 +14,9 @@
 """ This module contains implementations of various types of search space. """
 from __future__ import annotations
 
+import operator
 from abc import ABC, abstractmethod
+from functools import reduce
 from typing import Optional, Sequence, TypeVar, overload
 
 import tensorflow as tf
@@ -45,8 +47,8 @@ class SearchSpace(ABC):
         :param value: A point to check for membership of this :class:`SearchSpace`.
         :return: `True` if ``value`` is a member of this search space, else `False`. May return a
             scalar boolean `TensorType` instead of the `bool` itself.
-        :raise ValueError (or InvalidArgumentError): If ``value`` has a different dimensionality
-            from this :class:`SearchSpace`.
+        :raise ValueError (or tf.errors.InvalidArgumentError): If ``value`` has a different
+            dimensionality from this :class:`SearchSpace`.
         """
 
     @abstractmethod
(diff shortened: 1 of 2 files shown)


trieste/space.py
```diff
@@ -65,17 +67,10 @@ def __pow__(self: SP, other: int) -> SP:
...
-            raise ValueError(f"Exponent must be strictly positive, got {other}")
-
-        space = self
-
-        for _ in range(other - 1):
-            space *= self
-
-        return space
+        tf.debugging.assert_positive(other, message="Exponent must be strictly positive")
+        return reduce(operator.mul, [self] * other)
```
joelberkeley, line 73: felt like doing this, can undo if you'd prefer

trieste/space.py
```diff
@@ -235,14 +223,14 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
-        if not shapes_equal(value, self._lower):
-            raise ValueError(
-                f"value must have same dimensionality as search space: {self._lower.shape},"
-                f" got shape {value.shape}"
-            )
+        tf.debugging.assert_equal(
+            shapes_equal(value, self._lower),
+            True,
+            message="value must have same dimensionality as search space"
+        )
```
joelberkeley: this is a bit of a hack, to get around the fact that `tf.debugging.assert_shapes` considers 0 and [0] to have the same shape

hstojic (commented):
so we should use tf.debugging asserts throughout the code, rather than plain asserts (and ifelse statements with raising errors/warnings/etc)?

joelberkeley:
> so we should use tf.debugging asserts throughout the code, rather than plain asserts (and ifelse statements with raising errors/warnings/etc)?

`tf.debugging` works as expected with `tf.function`. `raise` doesn't (it will only raise when the function is compiled). `assert` is only used in source code for internal correctness checks. I don't believe trieste uses `assert` outside tests (`assert` is fine in tests, if you're not using `tf.function`)

trieste/space.py
hstojic: definitely more elegant! 
am I missing something or `assert_positive` means >0 while here we want >=1?

trieste/space.py
```diff
@@ -119,13 +114,7 @@ def sample(self, num_samples: int) -> TensorType:
...
             from this search space.
         """
-        num_points = self._points.shape[0]
-        if num_samples > num_points:
-            raise ValueError(
-                "Number of samples cannot be greater than the number of points"
-                f" {num_points} in discrete search space, got {num_samples}"
-            )
-
+        tf.debugging.assert_less_equal(num_samples, len(self._points))
```
hstojic: why not keeping the message? is it self explanatory enough?

trieste/space.py
joelberkeley: ~you're right. maybe i though positive meant strictly positive~

joelberkeley: scrap that ... > 0 is the same as >=1 for integers, and that's also what `assert_positive` does

trieste/space.py
joelberkeley: will add message

trieste/space.py
hstojic: > `tf.debugging` works as expected with `tf.function`. `raise` doesn't (it will only raise when the function is compiled). `assert` is only used in source code for internal correctness checks. I don't believe trieste uses `assert` outside tests (`assert` is fine in tests, if you're not using `tf.function`)

I see, I didn't know about the difference with respect to compilation. 
So `assert` has the same issue if tf.function is used? 
Actually, I have noticed we are dominantly using if-else accompanied with raising an error, while to me it usually seems like it should be an assert statement?

Regarding `tf.function`, at the moment it is used only in model optimization, is that correct?

hstojic:
> > so we should use tf.debugging asserts throughout the code, rather than plain asserts (and ifelse statements with raising errors/warnings/etc)?
> 
> `tf.debugging` works as expected with `tf.function`. `raise` doesn't (it will only raise when the function is compiled). `assert` is only used in source code for internal correctness checks. I don't believe trieste uses `assert` outside tests (`assert` is fine in tests, if you're not using `tf.function`)

I see, I didn't know about the difference with respect to compilation.
So assert has the same issue if tf.function is used?
Actually, I have noticed we are dominantly using if-else accompanied with raising an error, while to me it usually seems like it should be an assert statement?

Regarding tf.function, at the moment it is used only in model optimization, is that correct?

joelberkeley:
> I see, I didn't know about the difference with respect to compilation.
> So assert has the same issue if tf.function is used?

yes

> Actually, I have noticed we are dominantly using if-else accompanied with raising an error, while to me it usually seems like it should be an assert statement?

I don't think `assert` is meant to be used for argument validation

> Regarding tf.function, at the moment it is used only in model optimization, is that correct?

internally, yes, but users can wrap anything they like in `tf.function`

hstojic:
> > Actually, I have noticed we are dominantly using if-else accompanied with raising an error, while to me it usually seems like it should be an assert statement?
> 
> I don't think `assert` is meant to be used for argument validation

Why not? It seems to me that plenty of people use it that way, e.g. https://wiki.python.org/moin/UsingAssertionsEffectively

> > Regarding tf.function, at the moment it is used only in model optimization, is that correct?
> 
> internally, yes, but users can wrap anything they like in `tf.function`

true... but then that would be an argument for using `tf.debugging` throughout the code

hstojic (approved):
regardless of the discussion above, content wise this is good to go in my opinion

hstojic:
a thought on `tf.function` and users passing any piece of our code through it - it would be good to test that (all of) our code *can* be used within `tf.function` without errors. Some python code can cause errors within `tf.function`, we had that issue in Bellman.

joelberkeley:
> a thought on `tf.function` and users passing any piece of our code through it - it would be good to test that (all of) our code _can_ be used within `tf.function` without errors. Some python code can cause errors within `tf.function`, we had that issue in Bellman.

this crossed my mind before. I guess I never thought it high priority. Not sure why, since trieste has already had problems with `tf.function`

re asserts and exceptions, this [Q&A](https://softwareengineering.stackexchange.com/q/15515/271736) might be of interest

joelberkeley:
btw i don't have merge permissions. feel free to merge

PR #300, opened 2021-07-15
enable slicing on Dataset

joelberkeley:
Add slicing to Dataset

TODO
- [ ] confirm constructor error checking
- [ ] check tests

### DIFF
--- a/trieste/data.py
+++ b/trieste/data.py
@@ -36,7 +36,7 @@ class Dataset:
 
     def __post_init__(self) -> None:
         """
-        :raise ValueError (or InvalidArgumentError): If ``query_points`` or ``observations`` have
+        :raise tf.errors.InvalidArgumentError: If ``query_points`` or ``observations`` have
             rank less than two, or they have unequal shape in any but their first dimension.
         """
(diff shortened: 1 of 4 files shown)


trieste/data.py
```diff
@@ -36,23 +36,19 @@ class Dataset:
...
         tf.debugging.assert_rank_at_least(self.observations, 2)
 
-        if 0 in (self.query_points.shape[-1], self.observations.shape[-1]):
-            raise ValueError(
-                f"query_points and observations cannot have dimension 0, got shapes"
-                f" {self.query_points.shape} and {self.observations.shape}."
+        for x, name in ((self.query_points, "query_points"), (self.observations, "observations")):
+            tf.debugging.assert_positive(
+                x.shape[1:],
+                message=f"{name} cannot have empty trailing dimensions, got shape {x.shape}",
```
joelberkeley, line 48: it appears this was interpreting all leading dimensions (all except the last) as batch dims, but I think it's meant to interpret them all as dimensions except the first, which is the number of points

joelberkeley: THIS DEFINITELY NEEDS CHECKING

PR #301, opened 2021-07-16
fixed Slack invitation links and made wording around it more precise

hstojic:


vpicheny (approved):
I haven't tested if the link works, I just assume that you have!

PR #302, opened 2021-07-17
adding search space definitions for objectives

### LINKED ISSUE #296: Searhc space for implemented objective functions
It would be convenient to have search space defined in the toolbox for the implemented objective functions (Branin, Hartmann etc in `trieste/utils/objectives.py`), similar to minimizers and minima

For example, for branin it would be
```
BRANIN_SEARCH_SPACE = Box([0,1], [0,1])
```

hstojic:
This closes #296


trieste/utils/objectives.py
```diff
@@ -89,6 +90,10 @@ def scaled_branin(x: TensorType) -> TensorType:
 """ The global minimum of the :func:`branin` function, with shape [1] and dtype float64. """
 
 
+BRANIN_SEARCH_SPACE = Box([0.0]*2, [1.0]*2)
+""" The search space for the :func:`branin` function. """
```
joelberkeley: fyi you can write `Box([0.0], [1.0]) ** 2`

hstojic: thanks for the tip!

trieste/utils/objectives.py
```diff
@@ -27,6 +27,7 @@
 from ..data import Dataset
 from ..observer import MultiObserver, Observer, SingleObserver
 from ..type import TensorType
+from ..space import Box
```
hstojic, line 27: heh, this seems to have created a circular import issue : | I hate these... 

it is logical to have these search space definitions in this file, alongside the objectives and minimizers, but then it will be tricky to get rid of this issue - suggestions anyone? @joelberkeley perhaps (since you took a look already...)?

uri-granta: One possible fix would be to stop utils importing objectives in its init.py file, since space only depends on shapes_equal from utils.misc. Objectives would then need to be explicitly imported as utils.objectives (just like multi_objectives already has to).

uri-granta: Though this does point to the utils directory being used for two different things: general utility functions used in the actual code, and toy utility function for experimentation and writing tests.

joelberkeley: > Though this does point to the utils directory being used for two different things: general utility functions used in the actual code, and toy utility function for experimentation and writing tests.

how about moving the objectives to a new root-level package?

hstojic:
I think this is probably the best solution, smth like `trieste/example_objectives/`?

Sent from ProtonMail mobile

trieste/utils/objectives.py
hstojic: I think this is probably the best solution, smth like `trieste/example_objectives/` or simply `trieste/objectives/`

trieste/objectives/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This package contains examples of popular objective functions used in (Bayesian) optimization.
+"""
+
+from . import multi_objectives, observer, single_objectives
```
uri-granta: `observer.py` should probably be a top level trieste module, like space.py, rather than part of the objectives package.

Also wondering whether it might make sense to `import * from single_objectives` here, so you can just write `from trieste.objectives import branin` (but still need `from trieste.objectives.multi_objectives import VLMOP2` for the multi objective functions).

uri-granta (commented):
Looks good but

trieste/objectives/__init__.py
uri-granta: Also make sure you note the renaming in the git commit message as this will need to go into the release notes for the next release.

hstojic: here `observer.py` contains util functions for making an observer from the objective function, so I think it should be here as these functions have always been used together with these example objective functions. Do you think they would be useful generally? if so, then we should move them to `observer.py` at the root.  Otherwise, we can name the file differently.

I think I would agree with the second suggestion, but not a strong preference

uri-granta (approved):

PR #303, opened 2021-07-19
Victor/fixkernelconstraints

vpicheny:
This PR fixes 3 things:

- in `find_best_model_initialization` we now use the training loss so that it also works if we don't use any prior
- we "squeeze" the hyperparameters by an epsilon before searching for new ones. This avoids the following bug: say at the previous BO iteration, the optimal lengthscale reached its upper bound; if `find_best_model_initialization` doesn't find any better set of hyperparameters, you would re-assign the upper bound to the hyperparameters, then the untransformed value is +Inf and you get an error message.
- the previous sampling would return a single value for all lengthscales. They are now all different.

### DIFF
--- a/trieste/models/model_interfaces.py
+++ b/trieste/models/model_interfaces.py
@@ -469,30 +469,26 @@ class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel):
         """
 
         @tf.function
-        def evaluate_likelihood_of_model_parameters() -> tf.Tensor:
+        def evaluate_loss_of_model_parameters() -> tf.Tensor:
             randomize_model_hyperparameters(self.model)
-            return -self.model.training_loss()
-            # return self.model.maximum_log_likelihood_objective() + self.model.log_prior_density()
+            return self.model.training_loss()
 
         current_best_parameters = read_values(self.model)
-        max_log_likelihood = -self.model.training_loss()
-        # max_log_likelihood = (
-        #     self.model.maximum_log_likelihood_objective() + self.model.log_prior_density()
-        # )
+        min_loss = self.model.training_loss()
 
         found_better = False
         for _ in tf.range(num_kernel_samples):
             try:
-                log_likelihood = evaluate_likelihood_of_model_parameters()
+                train_loss = evaluate_loss_of_model_parameters()
             except tf.errors.InvalidArgumentError:  # allow badly specified kernel params
-                log_likelihood = -1e100
+                train_loss = 1e100
 
-            if log_likelihood > max_log_likelihood:  # only keep best kernel params
-                max_log_likelihood = log_likelihood
+            if train_loss < min_loss:  # only keep best kernel params
+                min_loss = train_loss
                 current_best_parameters = read_values(self.model)
                 found_better = True
 
-        if found_better:
+        if found_better:  # trick to avoid issues with Sigmoid
             multiple_assign(self.model, current_best_parameters)
 
 
--- a/trieste/models/model_interfaces.py
+++ b/trieste/models/model_interfaces.py
@@ -476,7 +476,6 @@ class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel):
         current_best_parameters = read_values(self.model)
         min_loss = self.model.training_loss()
 
-        found_better = False
         for _ in tf.range(num_kernel_samples):
             try:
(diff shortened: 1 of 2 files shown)


trieste/models/model_interfaces.py
```diff
@@ -615,11 +614,19 @@ def randomize_model_hyperparameters(model: gpflow.models.GPModel) -> None:
...
         elif param.prior is not None:
             param.assign(param.prior.sample())
+
+
+def squeeze_hyperparameters(model: gpflow.models.GPModel, epsilon: float = 1e-8) -> None:
+    for param in model.trainable_parameters:
+        if isinstance(param.bijector, tfp.bijectors.Sigmoid):
+            squeezed_param = tf.math.minimum(param, param.bijector.high - epsilon)
+            squeezed_param = tf.math.maximum(squeezed_param, param.bijector.low + epsilon)
+            param.assign(squeezed_param)
```
henrymoss: Perhaps add a little comment saying what this does/why it is needed? e.g. just use what you wrote in the PR description

trieste/models/model_interfaces.py
```diff
@@ -615,11 +614,19 @@ def randomize_model_hyperparameters(model: gpflow.models.GPModel) -> None:
     for param in model.trainable_parameters:
         if isinstance(param.bijector, tfp.bijectors.Sigmoid):
             sample = tf.random.uniform(
-                [1],
+                param.bijector.low.shape,
```
henrymoss, line 617: Good spot

trieste/models/model_interfaces.py
henrymoss: This might  also be unit-testable

henrymoss: For example

```
@random_seed
def test_squeeze_hyperparameters_when_param_at_edge_of_bounds() -> None:
    kernel = gpflow.kernels.RBF(variance=1.0, lengthscales=[0.1 + 1e-5,0.5 - 1e-5])
    upper = tf.cast([0.5, 0.5], dtype=tf.float64)
    lower = upper / 5
    
    kernel.lengthscales = gpflow.Parameter(
        kernel.lengthscales, transform=tfp.bijectors.Sigmoid(low=lower, high=upper)
    )

    squeeze_hyperparameters(model,epsilon=1e-2)

    npt.assert_array_less(kernel.lengthscales, [0.1,0.5] - 1e-2)
    npt.assert_array_less(-kernel.lengthscales, -[0.1,0.5] + 1e-2)
```

trieste/models/model_interfaces.py
henrymoss: Also, could do with a unit test.

henrymoss: Something like

```
@random_seed
def test_randomize_model_hyperparameters_samples_different_values_for_multi_dimensional_params(
) -> None:
    kernel = gpflow.kernels.RBF(variance=1.0, lengthscales=[0.2, 0.2])
    upper = tf.cast([10.0] * dim, dtype=tf.float64)
    lower = upper / 100
    kernel.lengthscales = gpflow.Parameter(
        kernel.lengthscales, transform=tfp.bijectors.Sigmoid(low=lower, high=upper)
    )

    randomize_model_hyperparameters(kernel)

    npt.assert_raises(AssertionError, npt.assert_allclose, kernel.lengthscales[0], kernel.lengthscales[1])


```

trieste/models/model_interfaces.py
uri-granta: (Yes, a docstring like for randomize_model_hyperparameters would be good.)

uri-granta (approved):
Looks good! Regarding tests, apart from @henrymoss's suggestion of testing the single-value-return bug fix and the squeeze parameters implementation, it might also be good to have explicit tests for the other two points of the ticket description:

1. check that `find_best_model_initialization` now handles params with no priors (see `test_find_best_model_initialization_improves_likelihood` and `test_find_best_model_initialization_only_changes_params_with_priors`)
2. check that `find_best_model_initialization` doesn't result in an +inf error if the optimal lengthscales has already reached its upper bound.

trieste/models/model_interfaces.py
vpicheny: thanks! done

trieste/models/model_interfaces.py
vpicheny: thanks! done

PR #305, opened 2021-07-21
Victor/squeezesoftplus

vpicheny:
Adds a "squeeze" functionality for softplus transformed variables.


vpicheny:
I decided to remove a test where I was trying to trick gpflow into pushing the hyperparameters to their boundary. It turns out that this is very difficult to achieve, and it's also very difficult to test that your model will return an error with +Inf before squeezing it.

vpicheny:
I also don't understand what's causing the issue with the docs

trieste/models/model_interfaces.py
```diff
@@ -624,22 +624,37 @@ def randomize_hyperparameters(object: gpflow.Module) -> None:
...
-    Squeezes the parameters to be strictly inside their range defined by the Sigmoid.
+    Squeezes the parameters to be strictly inside their range defined by the Sigmoid,
+    or strictly greater than the limit defined by the Softplus.
     This avoids having Inf unconstrained values when the parameters are exactly at the boundary.
 
     :param object: Any gpflow Module.
-    :param alpha: the proportion of the range with which to squeeze
-    :raise ValueError: If ``alpha`` is not in (0,1).
+    :param alpha: the proportion of the range with which to squeeze (or the absolute value
+    with which to translate the parameter for Softplus without lower bound)
```
henrymoss: you need a indent here. This might be the doc issue

trieste/models/model_interfaces.py
```diff
@@ -624,22 +624,37 @@ def randomize_hyperparameters(object: gpflow.Module) -> None:
             param.assign(param.prior.sample())
 
 
-def squeeze_hyperparameters(object: gpflow.Module, alpha: float = 1e-2) -> None:
+def squeeze_hyperparameters(
+    object: gpflow.Module, alpha: float = 1e-2, epsilon: float = 1e-7
```
henrymoss, line 717: this seems like quite a high default for alpha?

vpicheny: possibly, it's only 1% of the range though. Should I change to 1e-3? 1e-4? more?

henrymoss: I meant that this could be a bit too big if used to translate (when using softplus without a lower bound)

tests/unit/models/test_model_interfaces.py
```diff
@@ -511,27 +511,6 @@ def test_find_best_model_initialization_improves_likelihood(
...
-    upper = tf.cast([0.5], dtype=tf.float64)
-    lower = upper / 5.0
-    model.model.kernel.lengthscales = gpflow.Parameter(
-        model.model.kernel.lengthscales, transform=tfp.bijectors.Sigmoid(low=lower, high=upper)
-    )
-    gpflow.set_trainable(model.model.kernel.variance, False)
-    gpflow.set_trainable(model.model.likelihood.variance, False)
-
-    model.optimize(Dataset(x, tf.zeros_like(x)))
-    model.find_best_model_initialization(2)
```
henrymoss, line 555: How do you know that it will always go to 0.5?

tests/unit/models/test_model_interfaces.py
```diff
@@ -855,5 +834,9 @@ def test_squeeze_hyperparameters_when_param_at_edge_of_bounds() -> None:
     kernel.lengthscales = gpflow.Parameter(
         kernel.lengthscales, transform=tfp.bijectors.Sigmoid(low=lower, high=upper)
     )
-    squeeze_hyperparameters(kernel, 0.1)
+    squeeze_hyperparameters(kernel, alpha=0.1)
     npt.assert_array_almost_equal(kernel.lengthscales, [0.1 + 4e-2, 0.5 - 4e-2])
```
henrymoss, line 930: Im not sure this test is doing anything. Do we not need the lengthscales to be closer to their bounds for the squeeze to do anything (as epsilon is 1e-7)?

trieste/models/model_interfaces.py
```diff
@@ -624,22 +624,37 @@ def randomize_hyperparameters(object: gpflow.Module) -> None:
...
     for param in object.trainable_parameters:
         if isinstance(param.bijector, tfp.bijectors.Sigmoid):
-            epsilon = (param.bijector.high - param.bijector.low) * alpha
-            squeezed_param = tf.math.minimum(param, param.bijector.high - epsilon)
-            squeezed_param = tf.math.maximum(squeezed_param, param.bijector.low + epsilon)
+            delta = (param.bijector.high - param.bijector.low) * alpha
+            squeezed_param = tf.math.minimum(param, param.bijector.high - delta)
+            squeezed_param = tf.math.maximum(squeezed_param, param.bijector.low + delta)
             param.assign(squeezed_param)
+        elif isinstance(param.bijector, tfp.bijectors.Chain):
```
uri-granta: Could combine into one if test? Also should probably check bijector length.

```suggestion
        elif (
            isinstance(param.bijector, tfp.bijectors.Chain) and
            len(param.bijector.bijectors) == 2 and
            isinstance(param.bijector.bijectors[0], tfp.bijectors.Shift) and
            isinstance(param.bijector.bijectors[1], tfp.bijectors.Softplus)
        ):
```

trieste/models/model_interfaces.py
```diff
@@ -624,22 +624,37 @@ def randomize_hyperparameters(object: gpflow.Module) -> None:
...
+    Squeezes the parameters to be strictly inside their range defined by the Sigmoid,
+    or strictly greater than the limit defined by the Softplus.
     This avoids having Inf unconstrained values when the parameters are exactly at the boundary.
 
     :param object: Any gpflow Module.
-    :param alpha: the proportion of the range with which to squeeze
-    :raise ValueError: If ``alpha`` is not in (0,1).
+    :param alpha: the proportion of the range with which to squeeze (or the absolute value
+    with which to translate the parameter for Softplus without lower bound)
+    :param epsilon: the value with which to offset the parameter
```
uri-granta: Mention the use case? Also a full stop.

```suggestion
    :param epsilon: the value with which to offset the Softplus limit.
```

trieste/models/model_interfaces.py
```diff
@@ -624,22 +624,37 @@ def randomize_hyperparameters(object: gpflow.Module) -> None:
...
 
 
-def squeeze_hyperparameters(object: gpflow.Module, alpha: float = 1e-2) -> None:
+def squeeze_hyperparameters(
+    object: gpflow.Module, alpha: float = 1e-2, epsilon: float = 1e-7
+) -> None:
     """
-    Squeezes the parameters to be strictly inside their range defined by the Sigmoid.
+    Squeezes the parameters to be strictly inside their range defined by the Sigmoid,
+    or strictly greater than the limit defined by the Softplus.
```
uri-granta: Do you need to mention that it's a shifted Softplus or is that implicit?

tests/unit/models/test_model_interfaces.py
```diff
@@ -855,5 +834,9 @@ def test_squeeze_hyperparameters_when_param_at_edge_of_bounds() -> None:
...
+
+    lik = gpflow.likelihoods.Gaussian(variance=1.01e-6)
```
uri-granta, line 935: I would put this in a separate test. Something like `test_squeeze_hyperparameters_softplus` and rename the other test something like `test_squeeze_hyperparameters_sigmoid`.

Also would be good to add a simple test that confirms that passing in a 0 or negative epsilon results in a ValueError.

trieste/models/model_interfaces.py
uri-granta: Is the code actually handling the case of a Softplus without a lower bound?

uri-granta (commented):
A few suggestions and one question.

trieste/models/model_interfaces.py
vpicheny: I use alpha for relative change (sigmoid) and epsilon for absolute change (softplus). It's extra complexity but I don't think we can find a single default value for both that would be satisfying.

tests/unit/models/test_model_interfaces.py
vpicheny: We apply the squeeze even if the parameters are within bounds.

trieste/models/model_interfaces.py
vpicheny: will that work? if the length isn't > 1 `param.bijector.bijectors[1]` won't exist and this will return an error? or is the `and` taken sequentially?

trieste/models/model_interfaces.py
vpicheny: no, the transform has to be shift + softplus, I only coded that case because it's the `positive` transform used in gpflow. But you have cases where the shift is zero.

henrymoss (approved):
LGTM

PR #306, opened 2021-07-26
trust region refactor

joelberkeley:
This is a much simpler alternative to #286 . While that PR special-cases trust region itself, this PR special-cases only stateful acquisition rules

It's kind of two PRs, and I can split them up if you prefer. The first curries the `acquire` method on `TrustRegion`, moving the state parameter to the return type. This means we can then parametrize over the return type, now `T_co`. In the second, we let users choose which rule to use with `TrustRegion`.

### DIFF
--- a/docs/notebooks/batch_optimization.pct.py
+++ b/docs/notebooks/batch_optimization.pct.py
@@ -87,7 +87,7 @@ from trieste.acquisition.rule import EfficientGlobalOptimization
 batch_ei_acq = BatchMonteCarloExpectedImprovement(sample_size=1000, jitter=1e-5)
 batch_ei_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=batch_ei_acq)
-points_chosen_by_batch_ei, _ = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
+points_chosen_by_batch_ei = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
 
 # %% [markdown]
 # then we do the same with `LocalPenalizationAcquisitionFunction` ...
@@ -98,7 +98,7 @@ from trieste.acquisition import LocalPenalizationAcquisitionFunction
 local_penalization_acq = LocalPenalizationAcquisitionFunction(search_space, num_samples=2000)
 local_penalization_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=local_penalization_acq)
-points_chosen_by_local_penalization, _ = local_penalization_acq_rule.acquire_single(
+points_chosen_by_local_penalization = local_penalization_acq_rule.acquire_single(
     search_space, initial_data, model)
 
 # %% [markdown]
@@ -110,7 +110,7 @@ from trieste.acquisition import GIBBON
 gibbon_acq = GIBBON(search_space, grid_size = 2000)
 gibbon_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=gibbon_acq)
-points_chosen_by_gibbon, _ = gibbon_acq_rule.acquire_single(
+points_chosen_by_gibbon = gibbon_acq_rule.acquire_single(
     search_space, initial_data, model)
 
 # %% [markdown]
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 from __future__ import annotations
 
+from typing import List, Union, cast
+
 import gpflow
 import numpy.testing as npt
 import pytest
@@ -46,55 +48,65 @@ from trieste.objectives import (
 from trieste.objectives.utils import mk_observer
 from trieste.observer import OBJECTIVE
 from trieste.space import Box, SearchSpace
+from trieste.types import State, TensorType
 
 
 @random_seed
 @pytest.mark.parametrize(
     "num_steps, acquisition_rule",
-    [
-        (20, EfficientGlobalOptimization()),
-        (25, EfficientGlobalOptimization(AugmentedExpectedImprovement().using(OBJECTIVE))),
-        (
-            15,
-            EfficientGlobalOptimization(
-                MinValueEntropySearch(BRANIN_SEARCH_SPACE, num_fourier_features=1000).using(
-                    OBJECTIVE
-                )
+    cast(
+        List[
+            Union[
+                AcquisitionRule[TensorType, Box],
+                AcquisitionRule[State[TensorType, TrustRegion.State], Box],
+            ]
+        ],
+        [
+            (20, EfficientGlobalOptimization()),
+            (25, EfficientGlobalOptimization(AugmentedExpectedImprovement().using(OBJECTIVE))),
+            (
+                15,
+                EfficientGlobalOptimization(
+                    MinValueEntropySearch(BRANIN_SEARCH_SPACE, num_fourier_features=1000).using(
+                        OBJECTIVE
+                    )
+                ),
             ),
-        ),
-        (
-            10,
-            EfficientGlobalOptimization(
-                BatchMonteCarloExpectedImprovement(sample_size=500).using(OBJECTIVE),
-                num_query_points=3,
+            (
+                10,
+                EfficientGlobalOptimization(
+                    BatchMonteCarloExpectedImprovement(sample_size=500).using(OBJECTIVE),
+                    num_query_points=3,
+                ),
             ),
-        ),
-        (
-            10,
-            EfficientGlobalOptimization(
-                LocalPenalizationAcquisitionFunction(
-                    BRANIN_SEARCH_SPACE,
-                ).using(OBJECTIVE),
-                num_query_points=3,
+            (
+                10,
+                EfficientGlobalOptimization(
+                    LocalPenalizationAcquisitionFunction(
+                        BRANIN_SEARCH_SPACE,
+                    ).using(OBJECTIVE),
+                    num_query_points=3,
+                ),
             ),
-        ),
-        (
-            10,
-            EfficientGlobalOptimization(
-                GIBBON(
-                    BRANIN_SEARCH_SPACE,
-                ).using(OBJECTIVE),
-                num_query_points=2,
+            (
+                10,
+                EfficientGlobalOptimization(
+                    GIBBON(
+                        BRANIN_SEARCH_SPACE,
+                    ).using(OBJECTIVE),
+                    num_query_points=2,
+                ),
             ),
-        ),
-        (15, TrustRegion()),
-        (10, DiscreteThompsonSampling(500, 3)),
-        (10, DiscreteThompsonSampling(500, 3, num_fourier_features=1000)),
-    ],
+            (15, TrustRegion()),
+            (10, DiscreteThompsonSampling(500, 3)),
+            (10, DiscreteThompsonSampling(500, 3, num_fourier_features=1000)),
+        ],
+    ),
 )
 def test_optimizer_finds_minima_of_the_scaled_branin_function(
     num_steps: int,
-    acquisition_rule: AcquisitionRule[None, SearchSpace] | AcquisitionRule[TrustRegion.State, Box],
+    acquisition_rule: AcquisitionRule[TensorType, SearchSpace]
+    | AcquisitionRule[State[TensorType, TrustRegion.State], Box],
 ) -> None:
     search_space = BRANIN_SEARCH_SPACE
 
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -98,6 +98,16 @@ from trieste.types import State, TensorType
                 ),
             ),
             (15, TrustRegion()),
+            (
+                15,
+                TrustRegion(
+                    EfficientGlobalOptimization(
+                        MinValueEntropySearch(BRANIN_SEARCH_SPACE, num_fourier_features=1000).using(
+                            OBJECTIVE
+                        )
+                    )
+                ),
+            ),
             (10, DiscreteThompsonSampling(500, 3)),
             (10, DiscreteThompsonSampling(500, 3, num_fourier_features=1000)),
         ],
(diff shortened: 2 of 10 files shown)


trieste/bayesian_optimizer.py
```diff
@@ -182,7 +183,20 @@ def optimize(
         num_steps: int,
         datasets: Mapping[str, Dataset],
         model_specs: Mapping[str, ModelSpec],
-        acquisition_rule: AcquisitionRule[S, SP],
+        acquisition_rule: AcquisitionRule[TensorType, SP],
```
joelberkeley, line 186: unfortunately this clashes with the other overload. I believe this is because `TensorType` is untyped. I'm not sure what to do about that

trieste/bayesian_optimizer.py
```diff
@@ -182,7 +183,20 @@ def optimize(
...
+    ) -> OptimizationResult[None]:
+        ...
+
+    @overload
+    def optimize(
+        self,
+        num_steps: int,
+        datasets: Mapping[str, Dataset],
+        model_specs: Mapping[str, ModelSpec],
+        acquisition_rule: AcquisitionRule[State[S | None, TensorType], SP],
```
joelberkeley, line 202: this is a slightly odd way to do this, allowing `None` as the new state, but it works, is a limitation rather than a hack, and means fewer changes. You may want to change it in the future if you want to allow `AcquisitionRule`s that don't have a "special" (e.g. starting) state

trieste/acquisition/rule.py
```diff
@@ -425,36 +398,39 @@ def acquire(
...
-            step_is_success = y_min < state.y_min - self._kappa * tr_volume
-
-            eps = (
-                state.eps
-                if state.is_global
-                else state.eps / self._beta
-                if step_is_success
-                else state.eps * self._beta
-            )
+        def go(state: TrustRegion.State | None) -> tuple[TrustRegion.State | None, TensorType]:
```
joelberkeley, line 401: the type checking on these is horrible. Because `tf.Tensor` is untyped both as an object and a class, if you get the return values the wrong way round, mypy doesn't notice, even if your state is typed

trieste/bayesian_optimizer.py
```diff
@@ -182,7 +183,20 @@ def optimize(
         num_steps: int,
         datasets: Mapping[str, Dataset],
         model_specs: Mapping[str, ModelSpec],
-        acquisition_rule: AcquisitionRule[S, SP],
+        acquisition_rule: AcquisitionRule[TensorType, SP],
+        *,
+        track_state: bool = True,
+        fit_intial_model: bool = True,
+    ) -> OptimizationResult[object]:
```
joelberkeley, line 193: I've used `object` here instead of `None` because mypy can't differentiate between `TensorType` and `State[...]`, so the return types clash if it's `None`. I think this should be OK cos `None` and `object` have almost the same interface.

tests/integration/test_bayesian_optimization.py
```diff
@@ -83,7 +83,7 @@
                 num_query_points=2,
             ),
         ),
-        (15, TrustRegion()),
+        (15, TrustRegion()),  # type: ignore
```
henrymoss: Why this?

trieste/acquisition/rule.py
```diff
@@ -37,86 +38,72 @@
...
 
 
-class AcquisitionRule(ABC, Generic[S, SP_contra]):
-    """The central component of the acquisition API."""
+class AcquisitionRule(ABC, Generic[T_co, SP_contra]):
+    """
+    The central component of the acquisition API.
+
+    An :class:`AcquisitionRule` can produce any value from the search space for this step, and the
+    historic data and models. This value is typically a set of query points, either as is (see e.g.
```
henrymoss: "as is", what does this mean?

trieste/acquisition/rule.py
```diff
@@ -37,86 +38,72 @@
...
-    batchify,
-    generate_continuous_optimizer,
-)
+from .optimizer import AcquisitionOptimizer, automatic_optimizer_selector, batchify
 from .sampler import ExactThompsonSampler, RandomFourierFeatureThompsonSampler, ThompsonSampler
 
-S = TypeVar("S")
-""" Unbound type variable. """
+T_co = TypeVar("T_co", covariant=True)
+""" Unbound covariant type variable. """
```
henrymoss, line 46: What does this do?

trieste/acquisition/rule.py
```diff
@@ -425,36 +391,39 @@ def acquire(
...
-        acquisition_function = self._builder.prepare_acquisition_function(datasets, models)
-        point = self._optimizer(acquisition_space, acquisition_function)
-        state_ = TrustRegion.State(acquisition_space, eps, y_min, is_global)
+            points = self._rule.acquire(acquisition_space, datasets, models)
+            state_ = TrustRegion.State(acquisition_space, eps, y_min, is_global)
+
+            return state_, points
 
-        return point, state_
+        return go
```
henrymoss, line 445: Why does acquire now return a function?

trieste/acquisition/rule.py
joelberkeley: if `Foo` is parametrized by a **covariant** type variable (as `F[T_co]`), then `Foo[Child]` is a subclass of `Foo[Parent]`
if `Foo` is parametrized by a **contravariant** type variable (as `F[T_contra]`), then `Foo[Parent]` is a subclass of `Foo[Child]`
if `Foo` is parametrized by an **invariant** type variable (as `F[T]`), then neither of `Foo[Child]` or `Foo[Parent]` is a subclass of the other

In python, it's standard to suffix co- and contravariant type variables by `_co` and `_contra`

joelberkeley: this means that an `AcquisitionRule[int]` is a subclass of `AcquisitionRule[float]` because `int` is (at type checking at least), a subclass of `float`

trieste/acquisition/rule.py
joelberkeley: it's how i've made the `state` part of the return type rather than a built-in part of acquisition rules.

an alternative approach could be to simply make `TrustRegion` have some mutable state. You may choose to do this in future, but there are two reasons I didn't. One is historical: trieste was like this when I joined the team. The other is because it's good form to avoid mutable state, and I didn't want to consider the consequences of using mutable state in this PR

trieste/acquisition/rule.py
joelberkeley: as a `TensorType` rather wrapped in some context, like a state transition: `State[S, TensorType]` or error: `Result[TensorType]` (of course we don't have that in trieste)

joelberkeley: i agree it's not very clear. I'll see if I can clarify

trieste/acquisition/rule.py
joelberkeley: specifically, that I can use an `AcquisitionRule[int]` when an `AcquisitionRule[float]` is required. Or perhaps more relevantly, I can use an `AcquisitionRule[tf.Tensor]` where an `AcquisitionRule[TensorType]` is required because a `tf.Tensor` is a `tf.Tensor | np.ndarray`.

trieste/acquisition/rule.py
joelberkeley: it's perhaps better to think of it as `TrustRegion` returns a `State`ful set of points. That is, the value of the points depends on some state, and computing them can update that state. That it's a function is merely a choice of representation

tests/integration/test_bayesian_optimization.py
joelberkeley: mypy is too eager and infers the element type of this list as `tuple[int, AcquisitionRule[TensorType, Box]]`

joelberkeley: I could cast the whole list but I thought that'd be more messy than it's worth

trieste/acquisition/rule.py
joelberkeley: i've expanded on it. how is it now?

trieste/acquisition/rule.py
joelberkeley: note it's not needed for this PR, but I think it's good form

docs/notebooks/batch_optimization.pct.py
```diff
@@ -87,7 +87,7 @@ def build_model(data):
 batch_ei_acq = BatchMonteCarloExpectedImprovement(sample_size=1000, jitter=1e-5)
 batch_ei_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=batch_ei_acq)
-points_chosen_by_batch_ei, _ = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
+points_chosen_by_batch_ei = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
```
uri-granta, line 90: Would be good to also have an example of using TrustRegion in one of the notebooks (maybe in ecovering_from_errors like you did before)?

tests/unit/acquisition/test_rule.py
```diff
@@ -280,7 +283,7 @@ def test_trust_region_for_unsuccessful_global_to_local_trust_region_unchanged()
 
 
 def test_trust_region_for_successful_local_to_global_trust_region_increased() -> None:
-    tr = TrustRegion(NegativeLowerConfidenceBound(0).using(OBJECTIVE))
+    tr = TrustRegion(EfficientGlobalOptimization(NegativeLowerConfidenceBound(0)))
```
uri-granta: Maybe add some tests for `TrustRegion(DiscreteThompsonSampling(...))` as that would probe the usefulness of this change?

trieste/acquisition/rule.py
```diff
@@ -37,86 +38,73 @@
...
 """ Contravariant type variable bound to :class:`~trieste.space.SearchSpace`. """
 
 
-class AcquisitionRule(ABC, Generic[S, SP_contra]):
-    """The central component of the acquisition API."""
+class AcquisitionRule(ABC, Generic[T_co, SP_contra]):
+    """
+    The central component of the acquisition API.
+
+    An :class:`AcquisitionRule` can produce any value from the search space for this step, and the
```
uri-granta, line 56: The main thing I'm not certain about in this PR is that this definition suggests that `T_co` can be anything, but in practice it has to be either `TensorType` or `State[Optional[State], TensorType]` if we want to be able to use it in the BO loop. Should we state this explicitly in the definition of `T_co`?

trieste/bayesian_optimizer.py
uri-granta: Ugh! Maybe put a comment?

trieste/bayesian_optimizer.py
```diff
@@ -332,9 +360,12 @@ def optimize(
                         model.update(dataset)
                         model.optimize(dataset)
 
-                query_points, acquisition_state = acquisition_rule.acquire(
-                    self._search_space, datasets, models, acquisition_state
-                )
+                points_or_stateful = acquisition_rule.acquire(self._search_space, datasets, models)
+
+                if callable(points_or_stateful):
```
uri-granta, line 366: Nice

uri-granta (commented):
Looks good! Just a couple of comments.

docs/notebooks/batch_optimization.pct.py
joelberkeley: ? I haven't changed that notebook

tests/unit/acquisition/test_rule.py
joelberkeley: hmmm ... fair point. I'll have to think how much I can be bothered. That's quite a non-trivial thing to do

trieste/acquisition/rule.py
joelberkeley: I don't know of any way we could type it, since a `TypeVar` can't contain another `TypeVar` and `State` has to be invariant in its state type. Did you have something in mind?

on the subject, not everyone uses the BO loop

trieste/bayesian_optimizer.py
joelberkeley: done

tests/unit/acquisition/test_rule.py
joelberkeley: i've used a stub acquisition rule so I don't need to understand thompson sampling

trieste/acquisition/rule.py
uri-granta: Ok. Then maybe just change this comment to be a bit more assertive? Something like: `To use this in the Bayesian Optimization loop, this value needs to be a set of query points, either on its own...`

uri-granta (approved):
Happy for this to be merged as soon as @henrymoss is ok with it.

trieste/acquisition/rule.py
joelberkeley: how about appending
```
    Indeed, to use an :class:`AcquisitionRule` in the main
    :class:`~trieste.bayesian_optimizer.BayesianOptimizer` Bayesian optimization loop, the rule
    must return either a `TensorType` or `State`-ful `TensorType`.
```

tests/integration/test_bayesian_optimization.py
joelberkeley: haha lol ok so mypy's decided to fail with "unused `# type: ignore`" if i have it but a type error if I don't. `cast` it is

tests/integration/test_bayesian_optimization.py
```diff
@@ -46,55 +48,65 @@
...
-        ),
-        (15, TrustRegion()),
-        (10, DiscreteThompsonSampling(500, 3)),
-        (10, DiscreteThompsonSampling(500, 3, num_fourier_features=1000)),
-    ],
+            (15, TrustRegion()),
+            (10, DiscreteThompsonSampling(500, 3)),
+            (10, DiscreteThompsonSampling(500, 3, num_fourier_features=1000)),
+        ],
+    ),
```
henrymoss: Could we have an integration test of Trust Region combined with a more complicated acq? e.g MES?

trieste/acquisition/rule.py
henrymoss: Thanks for explaining!

tests/integration/test_bayesian_optimization.py
joelberkeley: in addition to the one for trust region with ego or instead of?

henrymoss: as well

joelberkeley: added mes

PR #307, opened 2021-07-27
stricter type checking

joelberkeley:
We configure mypy to error on a number of things esp missing types, and move the mypy config to pyproject.toml


tests/unit/models/test_model_interfaces.py
```diff
@@ -288,15 +298,13 @@ def _vgp_matern(x: tf.Tensor, y: tf.Tensor) -> VGP:
         (VariationalGaussianProcess, _vgp),
     ],
 )
-def _gpr_interface_factory(
-    request,
-) -> Callable[[TensorType, TensorType, Optimizer | None], GaussianProcessRegression]:
+def _gpr_interface_factory(request: Any) -> _ModelFactoryType:
```
joelberkeley, line 301: this looks like a [`FixtureRequest`](https://docs.pytest.org/en/latest/reference/reference.html#request) but that doesn't type check. This may be explained by
> has an optional param attribute

tests/util/misc.py
```diff
@@ -34,7 +34,7 @@
 )
 """ Error types thrown by TensorFlow's debugging functionality for tensor shapes. """
 
-C = TypeVar("C", bound=Callable)
+C = TypeVar("C", bound=Callable[..., object])
```
joelberkeley, line 37: is this right?

pyproject.toml
```diff
@@ -18,6 +18,18 @@ requires = [
...
+files = "trieste,tests,notebooks/**/*.py"
+ignore_missing_imports = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+strict_equality = true
+
+[[tool.mypy.overrides]]
+module = ["trieste.*", "tests.*"]
+disallow_any_generics = true
+disallow_untyped_defs = true
```
joelberkeley, line 31: these would mean contributors can't merge if they haven't fully specified the types

pyproject.toml
```diff
@@ -18,6 +18,18 @@ requires = [
...
 
+[tool.mypy]
+files = "trieste,tests,notebooks/**/*.py"
+ignore_missing_imports = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+strict_equality = true
+
+[[tool.mypy.overrides]]
+module = ["trieste.*", "tests.*"]
```
hstojic: does this mean we need to specify types in tests? that seems unnecessary...

pyproject.toml
```diff
@@ -18,6 +18,18 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[tool.mypy]
```
hstojic, line 21: its great that mypy options can be specified here, thanks!

pyproject.toml
joelberkeley: it would mean that, yes. I personally don't agree that it's unnecessary, but of course it's up to you and the team what to do

joelberkeley: i think this is generally the case with this PR. It does change the approach to contributions, so I expect everyone will want to know of this change

pyproject.toml
hstojic: I support this, it does increase the barrier on contributors, but it doesn't make sense we do typing on users' PR's

pyproject.toml
hstojic: I have raised a discussion point on this, I have mixed feelings about it, lets see

joelberkeley: I could omit `"tests.*"` for now. It would be very easy to add again later

hstojic: ok, lets do that for now and raise a separate PR later on if after a discussion we decide to go with that

joelberkeley: done

hstojic (approved):
ok, looks good now - ready to merge?

joelberkeley:
ready to merge

PR #309, opened 2021-07-27
use tf.debugging for acquisition functions

joelberkeley:


trieste/acquisition/function.py
```diff
@@ -557,16 +548,18 @@ def __init__(
...
-                f"Minimum feasibility probability must be between 0 and 1 inclusive,"
-                f" got {min_feasibility_probability}"
-            )
+        if isinstance(min_feasibility_probability, (int, float)):
+            tf.debugging.assert_greater_equal(float(min_feasibility_probability), 0.0)
+            tf.debugging.assert_less_equal(float(min_feasibility_probability), 1.0)
+        else:
+            dtype = min_feasibility_probability.dtype
+            tf.debugging.assert_greater_equal(min_feasibility_probability, tf.cast(0.0, dtype))
+            tf.debugging.assert_less_equal(min_feasibility_probability, tf.cast(1.0, dtype))
```
joelberkeley, line 562: i'm aware this is very verbose. The difficulty is that `0` is a `float` statically but not at runtime so a user can pass 0 and it pass mypy but be interpreted as an integer at runtime

trieste/acquisition/function.py
```diff
@@ -557,16 +548,18 @@ def __init__(
...
-                f"Minimum feasibility probability must be between 0 and 1 inclusive,"
-                f" got {min_feasibility_probability}"
-            )
+        if isinstance(min_feasibility_probability, (int, float)):
+            tf.debugging.assert_greater_equal(float(min_feasibility_probability), 0.0)
+            tf.debugging.assert_less_equal(float(min_feasibility_probability), 1.0)
+        else:
+            dtype = min_feasibility_probability.dtype
+            tf.debugging.assert_greater_equal(min_feasibility_probability, tf.cast(0, dtype))
+            tf.debugging.assert_less_equal(min_feasibility_probability, tf.cast(1, dtype))
```
joelberkeley, line 562: i'm aware this is very verbose. The difficulty is that `0` is a `float` statically but not at runtime so a user can pass 0 and it passes mypy but be interpreted as an integer at runtime

tests/thirdparty/test_tensorflow.py
```diff
@@ -0,0 +1,39 @@
...
+def test_raise_is_incompatible_with_tf_function() -> None:
+    @tf.function
+    def f(a: tf.Tensor) -> tf.Tensor:
+        if a <= tf.constant(0):
+            raise ValueError
+
+        return a
+
+    with pytest.raises(ValueError):
+        f(tf.constant(1))  # note that 1 should *not* trigger the error branch, but does
```
joelberkeley, line 29: i'm surprised by this, so it would be good if someone checked i've got this right

trieste/acquisition/function.py
```diff
@@ -275,29 +273,23 @@ def __init__(
...
         """
-        self._search_space = search_space
+        tf.debugging.assert_positive(num_samples)
+        tf.debugging.assert_positive(grid_size)
 
-        if num_samples <= 0:
-            raise ValueError(f"num_samples must be positive, got {num_samples}")
-        self._num_samples = num_samples
+        if num_fourier_features is not None:
+            tf.debugging.Assert(use_thompson, [])
```
joelberkeley, line 286: i've dropped error messages where I've used `Assert` because `Assert` doesn't have a message argument

joelberkeley: i could perhaps put them back in as separate `tf.print` statements

tests/thirdparty/test_tensorflow.py
hstojic: this seems more line the problem is in the if-else statement rather than raising exceptions?

hstojic: why "thirdparty"?

joelberkeley: > this seems more line the problem is in the if-else statement rather than raising exceptions?

the if-else is fine. `tf.function` follows every branch of a function when tracing, but it apparently doesn't compile exceptions into the graph, it just raises them.

> why "thirdparty"?

cos we're testing tensorflow rather than trieste

trieste/acquisition/function.py
hstojic: strange it doesn't have message argument, but `tf.print` probably isnt a good solution either, lets leave it as is

hstojic (commented):
overall looks good, just some minor questions to address first
branch also needs to be updated :)

hstojic (approved):
looks good!

PR #310, opened 2021-07-28
MES/GIBBON upgrade

henrymoss:
I have added a couple of tweaks to improve our information-theoretic acquisition functions.

1. I have improved the stability of the calculation of the inverse mills ratio required for MES/GIBBON by using tensorflows log_prob and log_cdf methods to calculate the ratio in log space. This greatly improves calculating stability and means we no longer need to clip the resulting ratio.
2. I have changed the optional big batch rescaling behaviour of GIBBON to be the default behaviour (and renamed it) to follow the new (and final) version of the GIBBON paper.
3. I have added a demonstration of GIBBON into our batch BO notebook


uri-granta (approved):
LGTM!

PR #313, opened 2021-07-29
Better error handling for continous acquisition optimizer

henrymoss:
If the scipy optimizer fails (e.g. decides that it hasnt optimized the acquisition function properly) then we currently recommend a point of nans for the next objective function evaluation.

Now, if optimization fails then we repeat from random starting locations (up to five times) and if scipy has still not found a maximizer then we report a useful error and stop Trieste. 

Previously, we would not know about this error until we tried to query our objective function at a nan location.


trieste/acquisition/optimizer.py
```diff
@@ -162,16 +162,31 @@ def _objective() -> TensorType:
             return -target_func(bijector.forward(variable[:, None, :]))  # [1]
 
         chosen_points = tf.ones([0, len(space.lower)], dtype=trial_search_space.dtype)  # [0, D]
-        for i in tf.range(num_restarts):  # restart optimization
+        for i in tf.range(num_restarts):  # perfom optimization from chosen starting points
```
vpicheny: possible confusion: if restart=1 you would run 2 optimisations? change to num_bfgs_runs for clarity?

trieste/acquisition/optimizer.py
```diff
@@ -162,16 +162,31 @@ def _objective() -> TensorType:
...
             variable.assign(bijector.inverse(initial_points[i : i + 1]))  # [1, D]
             gpflow.optimizers.Scipy().minimize(_objective, (variable,), **opt_kwargs)
             chosen_points = tf.concat(
                 [chosen_points, bijector.forward(variable)], axis=0
             )  # [i+1, D]
 
         chosen_func_values = target_func(chosen_points[:, None, :])  # [num_restarts, 1]
+        chosen_point = tf.gather(chosen_points, tf.argmax(chosen_func_values))  # [1, D]
```
vpicheny: is this NaN-proof?

trieste/acquisition/optimizer.py
```diff
@@ -162,16 +162,31 @@ def _objective() -> TensorType:
...
 
-        return tf.gather(chosen_points, tf.argmax(chosen_func_values))  # [1, D]
+        fail_count = 0
+        while tf.reduce_any(tf.math.is_nan(chosen_point)):  # repeat optimization if failed
+            variable.assign(bijector.inverse(space.sample(1)[i : i + 1]))  # random start
+            gpflow.optimizers.Scipy().minimize(_objective, (variable,), **opt_kwargs)
+            chosen_point = bijector.forward(variable)
+            fail_count += 1
+            if fail_count == 5:
```
vpicheny: I would to not hard-code this and pass it as a parameter. Maybe change name to `num_recovery_bfgs_run`?

tests/unit/acquisition/test_optimizer.py
```diff
@@ -40,6 +40,10 @@ def test_generate_random_search_optimizer_raises_with_invalid_sample_size() -> N
         generate_random_search_optimizer(num_samples=-5)
 
 
+def _jerky_function() -> AcquisitionFunction:
```
vpicheny: does this guarantees a failure? why not a test function that returns NaN?

tests/unit/acquisition/test_optimizer.py
```diff
@@ -141,11 +145,29 @@ def test_continuous_optimizer(
...
+        generate_continuous_optimizer(sigmoid=True, num_restarts=3),
+        generate_continuous_optimizer(sigmoid=True, num_restarts=1, num_initial_samples=1),
+    ],
+)
+def test_optimize_continuous_raises_for_impossible_optimization(
+    optimizer: AcquisitionOptimizer,
+) -> None:
+    search_space = Box([-1], [2])
+    with pytest.raises(ValueError):
+        optimizer(search_space, _jerky_function())
```
vpicheny: this doesn't assess that the optimiser does the right number of restarts, do we have any way of testing that?

trieste/acquisition/optimizer.py
henrymoss: weirdly it is. tf.argmax behaves like np.nanargmax

trieste/acquisition/optimizer.py
```diff
@@ -32,6 +33,12 @@
 """ Type variable bound to :class:`~trieste.space.SearchSpace`. """
 
 
+class FailedOptimizationError(Exception):
+    """Raised when an acquisition optimizer fails to optimize"""
+
```
uri-granta, line 38: Not sure you need the blank line, or even the pass (since you've got a docstring)

trieste/acquisition/optimizer.py
```diff
@@ -105,32 +115,40 @@ def generate_continuous_optimizer(
...
+    optimizations, each starting from one of the top  `num_optimization_runs` initial query points.
+
+    If all `num_optimization_runs` optimizations fail to converge then we run up to
+    `num_recovery_runs` starting from random locations.
 
-    The default behaviour of this method is to return a L-BFGS-B optimizer that perfoms
-    a single optimization.
+    The default behavior of this method is to return a L-BFGS-B optimizer that performs
+    a single optimization from the best of 1000 initial locations. If this optimization fails then
+    we run up to five recovery runs starting from random locations.
```
```suggestion
    we run up to `num_recovery_runs` recovery runs starting from random locations.
```

trieste/acquisition/optimizer.py
```diff
@@ -105,32 +115,40 @@ def generate_continuous_optimizer(
...
 
     :param num_initial_samples: The size of the random sample used to find the starting point(s) of
         the optimization.
     :param sigmoid: If True then use L-BFGS, otherwise use L-BFGS-B.
-    :param num_restarts: The number of separate optimizations to run.
+    :param num_optimization_runs: The number of separate optimizations to run.
+    :param num_recovery_runs: The number of backup optimization runs.
```
```suggestion
    :param num_recovery_runs: The maximum number of recovery optimization runs in case of failure.
```

trieste/acquisition/optimizer.py
```diff
@@ -161,17 +179,46 @@ def optimize_continuous(space: Box, target_func: AcquisitionFunction) -> TensorT
...
-            )  # [i+1, D]
-
-        chosen_func_values = target_func(chosen_points[:, None, :])  # [num_restarts, 1]
-
-        return tf.gather(chosen_points, tf.argmax(chosen_func_values))  # [1, D]
+        def _perform_optimization(starting_point: TensorType) -> OptimizeResult:
+            variable.assign(bijector.inverse(starting_point))  # [1, D]
+            return gpflow.optimizers.Scipy().minimize(_objective, (variable,), **opt_kwargs)
+
+        pass_count = 0
```
uri-granta: You're really using this like a boolean, so it would probably be clearer to have something like:

```suggestion
        successful_optimization = False
```

trieste/acquisition/optimizer.py
```diff
@@ -161,17 +179,46 @@ def optimize_continuous(space: Box, target_func: AcquisitionFunction) -> TensorT
...
+
+                new_point = bijector.forward(variable)  # [1, D]
+                new_point_score = target_func(new_point[:, None, :])  # [1, 1]
+
+                if new_point_score > chosen_point_score:  # if found a better point then keep
+                    chosen_point = new_point  # [1, D]
+                    chosen_point_score = new_point_score  # [1, 1]
+
+        fail_count = 0
+        while pass_count == 0:  # if all optimizations failed then try from random start
```
uri-granta: This currently fails spectacularly with num_recovery_runs is 0! Clearer as:

```suggestion
        if not successful_optimization:
           for _ in range(num_recovery_runs):
               opt_result = _perform_optimization(space.sample(1))
               if opt_result.success:
                   chosen_point = bijector.forward(variable)  # [1, D]
                   break
            else:  # if we didn't break out of the loop
                raise FailedOptimizationError(...)
```

tests/unit/acquisition/test_optimizer.py
```diff
@@ -141,11 +150,30 @@ def test_continuous_optimizer(
     expected_maximizer: list[list[float]],
     optimizer: AcquisitionOptimizer,
 ) -> None:
-
     maximizer = optimizer(search_space, _quadratic_sum(shift))
     npt.assert_allclose(maximizer, expected_maximizer, rtol=1e-3)
 
 
+@pytest.mark.parametrize("num_optimization_runs", [1, 10])
```
uri-granta, line 160: You could probably do more in depth testing by mocking gpflow.optimizers.Scipy().minimize so that it behaves like you want: e.g. fail three times then succeed. If it's not too difficult then I think that's definitely worth doing.

uri-granta (commented):
Looks good, but a few comments.

tests/unit/acquisition/test_optimizer.py
henrymoss: I tried this but couldnt get it to work. Do you have any examples of mocking something like that?

uri-granta: The following should work (though you'll need to fix the num_recovery_runs = 0 bug mentioned below first):

```python
@pytest.mark.parametrize("num_failed_runs", range(4))
@pytest.mark.parametrize("num_recovery_runs", range(4))
def test_optimize_continuous_recovery_runs(num_failed_runs, num_recovery_runs) -> None:

    scipy_minimize = Scipy.minimize
    failed_runs = 0

    def mock_minimize(self, *args, **kwargs):
        nonlocal failed_runs
        result = scipy_minimize(self, *args, **kwargs)
        if failed_runs < num_failed_runs:
            failed_runs += 1
            result.success = False
        return result

    with unittest.mock.patch(
            'trieste.acquisition.optimizer.gpflow.optimizers.Scipy.minimize',
            mock_minimize
    ):
        optimizer = generate_continuous_optimizer(
            num_optimization_runs=1, num_recovery_runs=num_recovery_runs
        )
        if num_failed_runs > num_recovery_runs:
            with pytest.raises(FailedOptimizationError) as e:
                optimizer(Box([-1], [1]), _quadratic_sum([0.5]))
        else:
            optimizer(Box([-1], [1]), _quadratic_sum([0.5]))
```

uri-granta (approved):
LGTM

PR #320, opened 2021-08-05
Fixed SVGP interface

sebastianober:
Fixed SVGP interface to be more consistent with how the other GPflow interfaces work:

- Added SVGP in `supported_models`
- Fixed implementation so SVGP can be used the same way GPR, SGPR, VGP, etc are
- Renamed 'SparseVariational' to 'SparseVariationalGaussianProcess' to be more consistent


trieste/models/model_interfaces.py
```diff
@@ -504,29 +504,47 @@ class SparseVariational(GPflowPredictor, TrainableProbabilisticModel):
...
+        model: SVGP,
+        optimizer: Optimizer | None = None
+    ):
         """
         :param model: The underlying GPflow sparse variational model.
-        :param data: The initial training data.
         :param optimizer: The optimizer with which to train the model. Defaults to
-            :class:`~trieste.models.optimizer.Optimizer` with :class:`~gpflow.optimizers.Scipy`.
+            :class:`~trieste.models.optimizer.Optimizer` with :class:`~tf.optimizers.Adam`.
+        :param max_num_inducing_points: The maximum number of inducing points allowed.
```
hstojic: `max_num_inducing_points` is not present as an argument in `__init__` above?

trieste/models/model_interfaces.py
```diff
@@ -504,29 +504,47 @@ class SparseVariational(GPflowPredictor, TrainableProbabilisticModel):
...
         """
+
+        if optimizer is None:
+            optimizer = Optimizer(tf.optimizers.Adam())
```
hstojic: TFOptimizer might be more appropriate?

trieste/models/model_interfaces.py
```diff
@@ -678,6 +696,7 @@ def perfom_optimization_step() -> None:  # alternate with natgrad optimizations
     GPR: GaussianProcessRegression,
     SGPR: GaussianProcessRegression,
     VGP: VariationalGaussianProcess,
+    SVGP: SparseVariational
```
hstojic: you mentioned that you would rename `SparseVariational` into something else for consistency?

trieste/models/model_interfaces.py
```diff
@@ -504,29 +504,47 @@ class SparseVariational(GPflowPredictor, TrainableProbabilisticModel):
...
-        return f"SparseVariational({self._model!r}, {self._data!r}, {self.optimizer!r})"
+        return f"SparseVariational({self._model!r}, {self.optimizer!r})"
 
     @property
     def model(self) -> SVGP:
         return self._model
 
     def update(self, dataset: Dataset) -> None:
-        _assert_data_is_compatible(dataset, self._data)
+        # Hard-code asserts from _assert_data_is_compatible because model doesn't store dataset
```
hstojic, line 578: this stuff looks like it might be useful to write a test to catch the issue that happened with the previous implementation

trieste/models/model_interfaces.py
sebastianober: Thanks! I initially tried to add this as an option but decided to just get a basic interface working instead and add that later, but forgot to remove the relevant line.

trieste/models/model_interfaces.py
sebastianober: I did originally but that caused some tests to fail so I changed it back (commit e2003f9). What would be the best way to handle this sort of thing in future?

hstojic: feel free to suggest a better naming, you will just need to adjust the tests as well wherever they use the old name
dont hesitate to change the tests, especially for these kind of non-substantive changes

sebastianober: ok sure, I think for the moment I'm happy with SparseVariational, as Henry pointed out SparseVariationalGaussianProcess gets quite long

hstojic:
I'm not sure whats going on here, if I look at the files changed tab, it's as if changes are compared to some old version of `develop`?

sebastianober:
> I'm not sure whats going on here, if I look at the files changed tab, it's as if changes are compared to some old version of `develop`?

I'm not particularly sure how this happened...

hstojic:
> > I'm not sure whats going on here, if I look at the files changed tab, it's as if changes are compared to some old version of `develop`?
> 
> I'm not particularly sure how this happened...

OK, solved, it happens from time to time it seems: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

I have just changed the comparison branch to something else and then returned to `develop`

sebastianober:
> > > I'm not sure whats going on here, if I look at the files changed tab, it's as if changes are compared to some old version of `develop`?
> > 
> > 
> > I'm not particularly sure how this happened...
> 
> OK, solved, it happens from time to time it seems: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch
> 
> I have just changed the comparison branch to something else and then returned to `develop`

Oh that's great, thanks so much!

trieste/models/model_interfaces.py
```diff
@@ -419,6 +419,9 @@ def covariance_between_points(
 
         :return: Covariance matrix between the sets of query points with shape [N, M]
         """
+        if isinstance(self.model, SGPR):
```
hstojic, line 422: it would be good to raise an Issue on GIthub regarding the covariance for SGPRs, so we have it on our to do list - probably it would entail a subclass of `GaussianProcessRegression` where covariance method is additionally implemented

trieste/models/model_interfaces.py
```diff
@@ -542,39 +545,49 @@ class SparseVariational(GPflowPredictor, TrainableProbabilisticModel):
...
     """
 
-    def __init__(self, model: SVGP, data: Dataset, optimizer: Optimizer | None = None):
+    def __init__(self, model: SVGP, optimizer: Optimizer | None = None):
         """
         :param model: The underlying GPflow sparse variational model.
-        :param data: The initial training data.
         :param optimizer: The optimizer with which to train the model. Defaults to
-            :class:`~trieste.models.optimizer.Optimizer` with :class:`~gpflow.optimizers.Scipy`.
+            :class:`~trieste.models.optimizer.TFOptimizer` with :class:`~tf.optimizers.Adam`.
```
hstojic: perhaps mention `batch_size` as well?

tests/unit/models/test_model_interfaces.py
```diff
@@ -255,39 +255,37 @@ def _mock_data() -> tuple[tf.Tensor, tf.Tensor]:
...
-    variational_variables = [m.q_mu.unconstrained_variable, m.q_sqrt.unconstrained_variable]
-    gpflow.optimizers.Scipy().minimize(m.training_loss_closure(), variational_variables)
     return m
 
 
 class _ModelFactoryType(Protocol):
     def __call__(
         self, x: TensorType, y: TensorType, optimizer: Optimizer | None = None
-    ) -> GaussianProcessRegression:
+    ) -> tuple[GaussianProcessRegression, Callable]:
```
hstojic: I'm not sure this would be a `callable`, perhaps some parent class of `GPR`, `SVGP` etc - `GPModel` perhaps?

tests/unit/models/test_model_interfaces.py
```diff
@@ -297,15 +295,17 @@ def __call__(
...
         (GaussianProcessRegression, _sgpr),
         (VariationalGaussianProcess, _vgp),
+        (SparseVariational, _svgp),
     ],
 )
 def _gpr_interface_factory(request: Any) -> _ModelFactoryType:
     def model_interface_factory(
         x: TensorType, y: TensorType, optimizer: Optimizer | None = None
-    ) -> GaussianProcessRegression:
+    ) -> tuple[GaussianProcessRegression, Callable]:
```
hstojic: same as above, is `Callable` the right thing here?

tests/unit/models/test_model_interfaces.py
```diff
@@ -297,15 +295,17 @@ def __call__(
...
         model_interface: type[GaussianProcessRegression] = request.param[0]
         base_model: GaussianProcessRegression = request.param[1](x, y)
-        return model_interface(base_model, optimizer=optimizer)
+        _reference_model: Callable = request.param[1]
```
hstojic: why not initializing with (x,y) immediately here as well?

hstojic:
looks fairly good for now, I have some minor comments - after fixing issues with tests and addressing these few comments this should be good to go

trieste/models/model_interfaces.py
sebastianober: Done! See #329

tests/unit/models/test_model_interfaces.py
sebastianober: If it is not an instance of the class yet (since it hasn't been initialized yet) but is the constructor is it a `Callable` or a `GPModel`?

tests/unit/models/test_model_interfaces.py
sebastianober: For tests such as `test_gaussian_process_regression_update` we want to compare the interface on updating from data `(x, y)` to `new_data = (x_new, y_new)` with the `update` interface method. To do this it seems to make most sense to initialize the interface with `(x, y)`, then use `model.update(new_data)` and initialize the GPflow reference with `new_data` instead of trying to update the GPflow reference from `(x, y)` to `new_data`. With this setup I need to pass the constructor, instead of an instance. What are your thoughts on this?

trieste/models/model_interfaces.py
sebastianober: Done, thanks!

tests/unit/models/test_model_interfaces.py
hstojic: right, I've noticed that, I think it depends which case you end up using more often - you can still do this even if its an instance, just call the gpr_factory again with new data  - feel free to choose the variant you think is more useful

tests/unit/models/test_model_interfaces.py
hstojic: huh, not sure, I'm not an expert on typing and mypy - @uri-granta or @joelberkeley?

tests/unit/models/test_model_interfaces.py
sebastianober: I feel like this way is a bit of a cleaner way of doing it, but if you prefer I can change it to what you suggest, or maybe there's even a way of separating things out so we have a reference model maker that takes the model interface and a dataset and returns the reference model we want?

tests/unit/models/test_model_interfaces.py
joelberkeley: it's a `Callable` *and* a `type[<whatever_model_type>]`, but I'd be generally wary of using the latter - there's no guarantee `__init__`'s signature is preserved in subclasses

that said, can you say more about why you've chosen to use SVGP to compare to SparseVariational etc? I would have thought they're all trying to approximate a GPR and that comparing to GPR would be sufficient, and more importantly, simpler

tests/unit/models/test_model_interfaces.py
hstojic: I dont have a strong preference really

tests/unit/models/test_model_interfaces.py
joelberkeley: also, `Callable` takes arguments e.g. `Callable[[Dataset], GaussianProcessRegression]` for input `Dataset` and output `GaussianProcessRegression`. Do use those if possible, whatever the args would be here

tests/unit/models/test_model_interfaces.py
sebastianober: Ok in that case I'll just leave it

tests/unit/models/test_model_interfaces.py
sebastianober: Just to be sure, in the latter case you're saying the correct typing would then be `type[<whatever model type>]`? I'll have a look at adding the args as you suggest, thanks!

For your second question, there's really no guarantee that SVGP will be close to GPR or any of the other model types, particularly as SVGP can be quite finicky to train. So after discussing with Henry and Hrvoje, we decided it would be best to change things so that we compare the behavior of the interfaces to the corresponding reference models. It felt like previously the tests were relying on the simplicity of the dataset and model to pass when compared to GPR, instead of actually testing whether the internal models were acting as a reference model would. So basically I changed things to 1) use more complicated data and kernel, and 2) compare the internal interface models to the equivalent reference models.

joelberkeley: re first qu - thinking

re second qu, ok, thanks. It's interesting to learn that. I'm a beginner with sparse GPs and VI and keen to learn where I can, but I was of the understanding that they're approximations to vanilla GPs at least in some sense. Have I got that right? If so, perhaps the approximation doesn't imply similar predictions/lml, but maybe there are other metrics that demonstrate similarity, in which case we could test for them instead. Even if this is accurate, the change is likely outside the scope of this work

joelberkeley: re first qu, sorry I rushed my first answer, I'm kind of not following this PR in detail. I've looked at the code more carefully, and I see that actually `type[<whatever model type>]` isn't right at all. Re `Callable`, it's not possible in Python to specify the optional arg `num_inducing: int = 2` in `_svgp` etc. with `Callable`, but it looks like you're not using that anyway - is that right? If you are, you can use a `Protocol` (just like the one here) with an appropriate `__call__` method.

regarding the return type, perhaps they're `Internal/External...Mixins` from gpflow, can't recall the name. That said, I'd ask if it's better to have each test doing one thing rather than each of them testing all the models. If a test requires an `External...` model type like `SVGP` you could have a separate set of tests for those. If there is commonality between that test and those for `Internal...`, there may be other abstractions that can reduce duplication.

i'm somewhat throwing ideas out there, don't take them too seriously

tests/unit/models/test_model_interfaces.py
```diff
@@ -989,7 +1105,8 @@ def test_gaussian_process_deep_copyable(gpr_interface_factory) -> None:
 
     # check that updating the original doesn't break or change the deepcopy
     x_new = tf.concat([x, tf.constant([[10.0], [11.0]], dtype=gpflow.default_float())], 0)
-    new_data = Dataset(x_new, _3x_plus_10(x_new))
+    new_data = Dataset(x_new, _2sin_x_over_3(x_new))
+    print(new_data)
```
hstojic: print is not needed here, right?

tests/unit/models/test_model_interfaces.py
sebastianober: For the typing: I just had a thought and that's that I don't think we really ever set `num_inducing` to anything other than the default in these tests, and if I'm honest I don't really see a need to set them to anything else. So maybe we can just hard code it and get rid of the arg? 

That's an interesting suggestion with the `Internal` vs `External` models, and that might solve some of the problems with having to check the model type and then decide whether or not you need to pass data to train. I'll have a think about it for sure!

One of the main motivations for a bunch of these changes was that we're planning on adding an interface to GPflux for DGPs, which won't act at all like GPR models. So that's partly why we made the changes with the reference models. But maybe then it would also make much more sense to have separate tests for different model types as you suggest.

joelberkeley: > So maybe we can just hard code it and get rid of the arg?

that would make things easier for sure

> we're planning on adding an interface to GPflux for DGPs

ah good to hear that's in the works. BTW I haven't been in discussions re this PR with hrvoje and henry, and I'm also out of the loop wrt project direction, so they'll know things I don't

tests/unit/models/test_model_interfaces.py
sebastianober: Oh yep you're right, I put that in while debugging and forgot to remove it!

tests/unit/models/test_model_interfaces.py
```diff
@@ -255,39 +255,37 @@ def _mock_data() -> tuple[tf.Tensor, tf.Tensor]:
...
+        gpflow.kernels.Matern32(), gpflow.likelihoods.Gaussian(), x[:num_inducing], num_data=len(x)
+    )
 
 
 def _vgp(x: tf.Tensor, y: tf.Tensor) -> VGP:
     likelihood = gpflow.likelihoods.Gaussian()
-    kernel = gpflow.kernels.Linear()
+    kernel = gpflow.kernels.Matern32()
     m = VGP((x, y), kernel, likelihood)
-    variational_variables = [m.q_mu.unconstrained_variable, m.q_sqrt.unconstrained_variable]
```
hstojic, line 273: how come this part can be removed?

tests/unit/models/test_model_interfaces.py
```diff
@@ -586,6 +648,38 @@ def test_vgp_update_updates_num_data() -> None:
...
+    data = Dataset(x, _3x_plus_10(x))
+    m = VariationalGaussianProcess(_vgp(data.query_points, data.observations))
+
+    x_new = tf.concat([x, tf.constant([[10.0], [11.0]], dtype=gpflow.default_float())], 0)
+    new_data = Dataset(x_new, _3x_plus_10(x_new))
+
+    reference_model = _vgp(data.query_points, data.observations)
+    reference_model_new = _vgp(new_data.query_points, new_data.observations)
+
+    internal_model = m.model
```
hstojic: a bit nitpicky, but its too verbose with all these variable assignments - everything can be called from within allclose function in relatively succinct fashion

tests/unit/models/test_model_interfaces.py
sebastianober: Before, when we were comparing VGP to GPR, we had to train the variational parameters before testing how similar it was. This is no longer necessary since we're comparing VGP to VGP.

tests/unit/models/test_model_interfaces.py
sebastianober: Hopefully better now!

hstojic (approved):
Looks good now

PR #323, opened 2021-08-08
rename module type to types

joelberkeley:
`import type` clashes with the inbuilt `type`. This PR renames the module to `types`


hstojic (commented):
I understand its not a good practice to have the same name as a builtin function, but to me it seems there shouldn't be any clash, usually we are having statements like
`from trieste.type import TensorType` - here there is no confusion, right?
did you ever encounter an issue with this?

joelberkeley:
in the trust region PR, I had a case where I couldn't import `State` from `type` as is because there was already a `State` class in rule.py. I ended up going with `import type as types` ... `types.State`. It's an edge case, no doubt, just thought it would be good to do

hstojic:
ok, fair enough, better to be on the safe side and do these things early in the development

hstojic (approved):
good to go, let me know when you are ready to merge

joelberkeley:
ready to merge

PR #324, opened 2021-08-08
rearrange utils package

joelberkeley:
This PR refactors the utils package. We
* move pareto.py out of util because it's not really the right place for it. utli is meant to be generic, but pareto is very specific.
* move pareto.py to acquisition, since nothing else uses it. I seem to recall @satrialoka being against this when it was introduced. Am I right? What other options are available?
* rename utils to util everywhere, since it's shorter and conveys as much meaning. This may be better as a separate PR
* merge utils/\_\_init\_\_.py and utils/misc.py into a single module in the src root

### DIFF
--- a/docs/notebooks/batch_optimization.pct.py
+++ b/docs/notebooks/batch_optimization.pct.py
@@ -87,7 +87,7 @@ from trieste.acquisition.rule import EfficientGlobalOptimization
 batch_ei_acq = BatchMonteCarloExpectedImprovement(sample_size=1000, jitter=1e-5)
 batch_ei_acq_rule = EfficientGlobalOptimization(  # type: ignore
     num_query_points=10, builder=batch_ei_acq)
-points_chosen_by_batch_ei, _ = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
+points_chosen_by_batch_ei = batch_ei_acq_rule.acquire_single(search_space, initial_data, model)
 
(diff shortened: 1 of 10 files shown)


satrialoka:
Yeah, I was considering putting it there similar to what botorch did , but I think it is ok to move it to acquisition. Another option is to keep putting it inside utils under multiobjective folder same as what botorch [did](https://github.com/pytorch/botorch/tree/master/botorch/utils/multi_objective)

joelberkeley:
you know much more about how people will "think of" pareto - where they'll expect to find it etc. All I really feel is that it shouldn't be in util. It's possible there isn't really a great place for it with the directory structure we have

TsingQAQ:
Hi Joel and Loka, there is a [draft PR](https://github.com/secondmind-labs/trieste/pull/326) on modifying of Pareto: it takes your suggestions and the `Pareto.py` have been moved to acquisition. Please let me know if this makes sense as you expected/there are better ways of doing so.

joelberkeley:
ok. Unless you actually planned on this change for your PR anyway, it would be easier for me at least if it was only in one PR

hstojic:
I am not sure I get the final message, @joelberkeley do you want to proceed with the PR or not?

joelberkeley:
@hstojic on reflection, I may simplify this PR, and wait for jixiang to complete his related PR first

PR #325, opened 2021-08-08
hide private functionality from docs

joelberkeley:
we hide some private functionality from the docs by commenting out docstrings

### DIFF
--- a/trieste/acquisition/sampler.py
+++ b/trieste/acquisition/sampler.py
@@ -432,6 +432,8 @@ class RandomFourierFeatureThompsonSampler(ThompsonSampler):
         else:  # if n <= m  then calculate posterior in gram space (an n*n matrix inversion)
             self._theta_posterior = self._prepare_theta_posterior_in_gram_space()
 
+        self._pre_calc = False  # Flag so we only calculate the posterior for the weights once.
+
     def __repr__(self) -> str:
         """"""
         return f"""{self.__class__.__name__}(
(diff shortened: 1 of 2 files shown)


trieste/utils/pareto.py
```diff
@@ -47,23 +48,19 @@ def non_dominated(observations: TensorType) -> tuple[TensorType, TensorType]:
     return tf.boolean_mask(observations, dominance == 0), dominance
 
 
+@dataclass(frozen=True)
```
joelberkeley, line 52: this was just begging to be a dataclass

trieste/acquisition/sampler.py
```diff
@@ -432,30 +432,24 @@ def __init__(
         else:  # if n <= m  then calculate posterior in gram space (an n*n matrix inversion)
             self._theta_posterior = self._prepare_theta_posterior_in_gram_space()
 
-        self._pre_calc = False  # Flag so we only calculate the posterior for the weights once.
```
joelberkeley: i noticed nothing is using this. Should it be used somewhere or deleted? @henrymoss

hstojic (commented):
looks good to me, I'm not sure about `self._pre_calc` though - I'll let Henry address that and then its good to go

joelberkeley:
> looks good to me, I'm not sure about `self._pre_calc` though - I'll let Henry address that and then its good to go

i've reverted that so we can merge. Merge when ready

hstojic:
ok, merging if checks are ok

hstojic (approved):
all good

PR #330, opened 2021-08-13
Reduce RAM usage in integration tests

henrymoss:
Our MO integration tests previously had a massive matrix multiplication that required >10GB of RAM. I've now changed some of the parameters in the tests (e.g. num MC samples) to avoid such a large calculation.

The resulting tests will now run on any machine.

We also now force the user to specify the number of MC samples when defining a BatchMonteCarloExpectedHypervolumeImprovement acquisition function, just like we do for BatchMonteCarloExpectedImprovement. This will avoid future similar accidental memory explosions.


hstojic (approved):
looks good!

PR #336, opened 2021-08-18
Minor fixes to error messages

sebastianober:
Two minor fixes to the error messages in the model interfaces.


hstojic (approved):
looks good

PR #339, opened 2021-08-19
`module_deepcopy` not needed any more

### LINKED ISSUE #335: change how we make deep copies of models
given that we are moving to supporting other model types than GPflow, we'll probably need to rethink how we make deep copies of the models in each iteration, perhaps the top interface (e.g. `ProbabilisticModel`) should require that each model type should implement some kind of `save_model` method that `BayesianOptimizer` would call, since different model types might do it differently (e.g. for Keras networks data does not need to be saved, only the architecture and weights) - then interface of each model type (e.g. `GPflowPredictor`) would actually define the save method.

any thoughts?
@joelberkeley perhap? you have written the current `module_deepcopy` function, right?

hstojic:
GPflow models couldn't be properly deep copied in TF Probability versions < 0.12, but now since we use 0.13 `module_deepcopy` function that provided deep copy functionality is not needed any more - tests all pass and it seems like we should be fine without it

there is a related discussion in issue #335


uri-granta (approved):
Looks good! Note that we officially also support tf-probability 0.12, not just 0.13 (required for tensorflow 2.4). However there's no reason to suspect that it'll break that, and if we'll test it manually when we do the next release (assuming we still need that support, which hopefully we won't).

PR #340, opened 2021-08-20
Update BatchMonteCarloExpectedImprovement

uri-granta:
Update BatchMonteCarloExpectedImprovement acquisition function builder to avoid retracing each time.


tests/integration/test_bayesian_optimization.py
```diff
@@ -73,7 +73,7 @@
                 ),
             ),
             (
-                10,
+                12,
```
uri-granta, line 76: This is purely due to the BatchReparametrizationSampler picking different _eps values to before.

trieste/acquisition/sampler.py
```diff
@@ -269,6 +269,10 @@ def __init__(self, sample_size: int, model: ProbabilisticModel):
             tf.ones([0, 0, sample_size], dtype=tf.float64), shape=[None, None, sample_size]
         )  # [0, 0, S]
 
+        # for some reason graph compilation is resulting in self._eps reporting the wrong shape
+        # we therefore use an extra boolean variable to keep track of whether it's initialised
+        self._initialized = tf.Variable(False)
```
henrymoss, line 274: Why does this have to be a tf Variable?

henrymoss (commented):
LGTM

trieste/acquisition/sampler.py
uri-granta: For the same reason self._eps needs to be: so it can be updated in the compiled graph.

uri-granta:
> LGTM

You ok to tick then?

henrymoss (approved):

PR #343, opened 2021-08-20
Decomposition of model-specific optimizer functions

sebastianober:
Following #272 it would be good to decompose model-specific optimizer functions into their own files corresponding to each model.

### DIFF
--- a/trieste/models/optimizer.py
+++ b/trieste/models/optimizer.py
@@ -162,7 +162,7 @@ def _create_tf_optimizer(
 
 @singledispatch
 def create_loss_function(
-    model: gpflow.models.GPModel, dataset: TrainingData, compile: bool = False
+    model: tf.Module, dataset: TrainingData, compile: bool = False
 ) -> LossClosure:
     """
     Generic function for building a loss function for a specified `model` and `dataset`.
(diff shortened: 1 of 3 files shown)


trieste/models/optimizer.py
```diff
@@ -161,14 +160,6 @@ def _create_tf_optimizer(
...
-def _create_scipy_optimizer(
-    optimizer: gpflow.optimizers.Scipy,
-    optimizer_args: Dict[str, Any],
-) -> Optimizer:
-    return Optimizer(optimizer, **optimizer_args)
-
-
 @singledispatch
 def create_loss_function(
     model: gpflow.models.GPModel, dataset: TrainingData, compile: bool = False
```
hstojic: additional gpflow leftovers...

sebastianober: Should be fixed now

sebastianober:
TODO: decide what to do about Keras vs TF optimizers - do we want a `KerasOptimizer` class? Perhaps this might be something that would be better for a new issue?

sebastianober:
@hstojic Any ideas on what you want to do with this? Should we just leave it with my GPflux PR?

hstojic:
> @hstojic Any ideas on what you want to do with this? Should we just leave it with my GPflux PR?

I'll pick this up and sort out the PR, this week or the next

hstojic:
new PR #372 is replacing this one

Issue #215, opened 2021-03-30
Add ask-tell optimizer


joelberkeley:
**Describe the feature you'd like**
As a user, I'd like an ask-tell optimizer so that ...

joelberkeley:
This could use coroutines (they seem to be a python core feature designed for this use case). For example
```python
def optimize(
    search_space: SP,
    datasets: Mapping[str, Dataset],
    models: Mapping[str, TrainableProbabilisticModel],
    acquisition_rule: AcquisitionRule[S, SP],
    acquisition_state: Optional[S] = None,
) -> Generator[Result[TensorType], Mapping[str, Dataset], None]:
    while True:
        try:
            query_points, acquisition_state = acquisition_rule.acquire(
                search_space, datasets, models, acquisition_state
            )

            observer_output = yield Ok(query_points)

            datasets = {tag: datasets[tag] + observer_output[tag] for tag in observer_output}

            for tag, model in models.items():
                dataset = datasets[tag]
                model.update(dataset)
                model.optimize(dataset)
        except Exception as e:
            yield Err(e)  # todo is this right?
```
used like
```python
ask_tell = optimize(search_space, initial_data, models, EfficientGlobalOptimization())
new_point = next(ask_tell)

if new_point.is_ok:
    new_data = observer(new_point.unwrap())
    new_point = ask_tell.send(new_data)
```

satrialoka:
I have tried above example for non-async case, and it was pretty nice. 
regarding the function naming, Isn't it more natural for using the ask/tell name itself rather than the default python's coroutine function name? for example `next(ask_tell)` could be modified as `ask_tell.ask()`, `ask_tell.send(new_data)` could be `ask_tell.tell(new_data)`.

joelberkeley:
@satrialoka re `.ask()` and `.tell()`, possibly, but you'd have to ask if it's worth wrapping it for the different names. Wrapping it may lose out on certain aspects of coroutines, and it would of course imply a little extra maintenance. I don't know the answer.

satrialoka:
ah yea, you're right.  maybe @uri-granta has a comment on this?

uri-granta:
Keeping the optimizers as bare coroutines (as opposed to a wrapped class) is attractive to me too, but there's no reason we can't also define standalone `ask` and `tell` functions right?

```python
AskTellOptimizer = Generator[Result[TensorType], Mapping[str, Dataset], None]

[...]

PR #346, opened 2021-09-13
Ask tell API

apaleyes:
Related issue: #215 

This PR adds high level ask/tell interface. It is essentially a stripped down `optimize` function with the loop, all history and state tracking removed.

Contrary to the discussion in the issue, I've went with simpler approach and no coroutines. This allows for a nice way of saving and loading state, which can be important in complex scenarios where the optimization has to be stopped for a considerable time to perform the experiment. There are also a few one line fixes in here.

Once we agree on the API, the next PR with docs and example notebook will follow.


trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+""" Unbound type variable. """
+
+SP = TypeVar("SP", bound=SearchSpace)
+""" Type variable bound to :class:`SearchSpace`. """
+
+
+class AskTellOptimizer(Generic[S, SP]):
+    """
+    This class provides Ask/Tell optimization interface. It is designed for those use cases
+    when control of the optimization loop by Trieste is impossible.
```
vpicheny: impossible or not desirable

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+    That could happen because objective function cannot be implemented/called in Python, or if
```
vpicheny: example seems over-specific, remove this sentence?

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+        search_space: SP,
+        datasets: Mapping[str, Dataset] | Dataset,
+        model_specs: Mapping[str, ModelSpec] | ModelSpec,
+        fit_model: bool,
+        acquisition_rule: AcquisitionRule[S, SP],
+        acquisition_state: Optional[S] = None,
+    ):
+        """
+        :param search_space: The space over which to search.
+        :param datasets: The known observer query points and observations for each tag.
```
vpicheny: observer query points?

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+        :param model_specs: The model to use for each :class:`~trieste.data.Dataset` in
+            ``datasets``.
+        :param fit_model: If `True`, models passed in will be optimized on the given data.
+            If `False`, the models are assumed to be optimized already.
+        :param acquisition_rule: The acquisition rule, which defines how to search for a new point
```
vpicheny, line 137: do we want to default to `EfficientGlobalOptimization` as in `bayesian_optimizer`?

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+
+        self._datasets = datasets
+        self._models = map_values(create_model, model_specs)
+
+        if fit_model:
+            # there is no default value for `fit_model` parameter
+            # to force the user to think about it
+            # before invoking ask/tell API
+            # there is no sensible default here:
+            # `False` may result in suboptimal performance
```
vpicheny: I think this is overspecific - we may just want to say that if model training is carried out outside this interface then the user should set it to False.

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+        """Suggests a point (or points in batch mode) to observe by optimizing the acquisition function.
+        If the acquisition is stateful, its state is saved.
+
+        :return: A :class:`TensorType` instance representing suggested point(s).
+        """
+        points_or_stateful = self._acquisition_rule.acquire(
+            self._search_space, self._datasets, self._models
+        )
+
+        if callable(points_or_stateful):
```
vpicheny, line 267: I don't understand what this does

apaleyes:
After discussion with @vpicheny , I am going to add a notebook to this PR, to help reviewers see the API in action.

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+class AskTellOptimizer(Generic[S, SP]):
+    """
+    This class provides Ask/Tell optimization interface. It is designed for those use cases
+    when control of the optimization loop by Trieste is impossible.
+    That could happen because objective function cannot be implemented/called in Python, or if
+    optimization state cannot be maintained in memory.
+    For more details about the Bayesian Optimization routine, refer to :class:`BayesianOptimizer`.
+    """
+
+    def __init__(
```
uri-granta, line 121: Maybe model this more closely on the overloaded `BayesianOptimizer.optimize` method? E.g. make the rule optional (defaulting to `EfficientGlobalOptimization()`), ensure datasets and model_specs match up type-wise (either both are dicts or neither is), etc? Happy to discuss if you like.

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+            # there is no sensible default here:
+            # `False` may result in suboptimal performance
+            # `True` may result in two model training calls per optimization step
+            for tag, model in self._models.items():
+                dataset = datasets[tag]
+                model.update(dataset)
+                model.optimize(dataset)
+
+    @classmethod
+    def from_record(
```
uri-granta, line 200: Might be nice if this and `get_state` had obviously symmetric names? e.g. `from_record`/`to_record`? Also maybe place them one after the other?

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+            # there is no default value for `fit_model` parameter
+            # to force the user to think about it
+            # before invoking ask/tell API
+            # there is no sensible default here:
+            # `False` may result in suboptimal performance
+            # `True` may result in two model training calls per optimization step
+            for tag, model in self._models.items():
+                dataset = datasets[tag]
+                model.update(dataset)
+                model.optimize(dataset)
```
uri-granta, line 191: You should add a  `def __repr__(self) -> str` for easier debugging (see e.g. BayesianOptimizer)

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+
+    @classmethod
+    def from_record(
+        cls, search_space: SP, acquisition_rule: AcquisitionRule[S, SP], record: Record[S]
```
uri-granta: Maybe put record first and make acquisition_rule optional (assuming you end up doing that for init)? Also it might make sense to expose search_space and acquisition_rule as @properties. (There'd probably be no harm doing that for `BayesianOptimizer` too, though it's more obviously useful here.)

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+        :param search_space: The space over which to search.
+        :param acquisition_rule: The acquisition rule, which defines how to search for a new point
+            on each optimization step.
+        :param record: Optimization state record.
+        """
+        # here we are recovering previously saved optimization state
+        # there the model was already trained
+        # thus there is no need to train it again
+        fit_model = False
+        return cls(
```
uri-granta, line 223: With this many arguments, it's probably clearer to use named arguments either for all of them or all but the important first three.
```suggestion
        return cls(
            search_space,
            record.datasets,
            record.models,
            fit_model=False,  # model is already trained
            acquisition_rule=acquisition_rule,
            acquisition_state=record.acquisition_state,
        )        
```

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+            model.optimize(dataset)
+
+    def get_state(self) -> Record[S]:
+        """Collects the current state of the optimization, which includes datasets,
+        models and acquisition state (if applicable).
+
+        :return: An optimization state record.
+        """
+        return Record(
+            datasets=self._datasets, models=self._models, acquisition_state=self._acquisition_state
```
uri-granta: Do we need to deepcopy the models and acquisition_state to avoid the record being modified by future ask/tell commands? 

If you think it would be useful to expose _both_ a deecopied record that can be used safely and a current state that is more efficient then maybe call the first `to_record()` and make the second a `current_state` `@property`?

trieste/ask_tell_optimization.py
uri-granta: This is a Joelism for handling acquisition rules that support state (first branch) together with ones that don't (second branch). Both branches calculate the new query points; the first branch also uses and updates the state.

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,181 @@
...
+        points_or_stateful = self._acquisition_rule.acquire(
+            self._search_space, self._datasets, self._models
+        )
+
+        if callable(points_or_stateful):
+            self._acquisition_state, query_points = points_or_stateful(self._acquisition_state)
+        else:
+            query_points = points_or_stateful
+
+        return query_points
```
uri-granta, line 272: Q: Do we want to protect against calling ask twice in a row without an intermediate tell? Especially with stateful rules? Not sure.

tests/integration/test_ask_tell_optimization.py
```diff
@@ -0,0 +1,142 @@
...
+        # in first the same optimizer object is always used
+        # in second new optimizer is created at each step (e.g. from serialized state)
+        new_point = ask_tell.ask()
+
+        if reload_state:
+            state = ask_tell.get_state()
+
+        new_data = observer(new_point)
+
+        if reload_state:
```
uri-granta, line 122: Not part of this PR, but would be nice to test actually serialising records to disk.

tests/integration/test_ask_tell_optimization.py
```diff
@@ -0,0 +1,142 @@
...
+            ask_tell = AskTellOptimizer.from_record(search_space, acquisition_rule, state)
+
+        ask_tell.tell(new_data)
+
+    final_state = ask_tell.get_state()
+    dataset = final_state.dataset
+
+    print(dataset)
```
uri-granta: Stray print?

tests/unit/test_ask_tell_optimization.py
```diff
@@ -0,0 +1,255 @@
...
+) -> None:
+    dataset_with_key_1 = {"1": init_dataset}
+    model_with_key_1 = {"1": model}
+    new_data_with_key_2 = {"2": mk_dataset([[1.0]], [[1.0]])}
+
+    ask_tell = AskTellOptimizer(
+        search_space, dataset_with_key_1, model_with_key_1, True, acquisition_rule
+    )
+    with pytest.raises(ValueError):
+        ask_tell.tell(new_data_with_key_2)
```
uri-granta, line 281: Good unit testing

uri-granta (commented):
Looks good, and reassuringly not over-engineered! Various comments.

trieste/ask_tell_optimization.py
apaleyes: added

trieste/ask_tell_optimization.py
apaleyes: removed the sentence

trieste/ask_tell_optimization.py
apaleyes: rephrased, don't know how that made sense to me before

trieste/ask_tell_optimization.py
apaleyes: Uri also suggested the same, makes sense to keep API similar. agreed!

trieste/ask_tell_optimization.py
apaleyes: yeah, by the time I finished this PR doubt crept in. made it default to True

trieste/ask_tell_optimization.py
apaleyes: thanks Uri! Importantly, when the state is used, acquire returns a Callable (that when called gives you state and points) instead of just points

apaleyes: i'll add a comment in the code explaining this

trieste/ask_tell_optimization.py
apaleyes: great idea about `to_record`, will do!

trieste/ask_tell_optimization.py
apaleyes: added

trieste/ask_tell_optimization.py
apaleyes: agreed with the default rule idea (Victor said as much above), but probably will pass on overloads for type matching. we have some validation in place to control this

trieste/ask_tell_optimization.py
apaleyes: yup, now that the rule can be optional this needs some reordering indeed

trieste/ask_tell_optimization.py
apaleyes: agreed

trieste/ask_tell_optimization.py
apaleyes: i think we actually might want several asks in a row sometimes. i'll next work on async batch BO, where this is likely to be the case

tests/integration/test_ask_tell_optimization.py
apaleyes: i understand not all types of models can be serialized at the moment, so it may be a bit tricky

tests/integration/test_ask_tell_optimization.py
apaleyes: it is, removed

tests/unit/test_ask_tell_optimization.py
apaleyes: thx

trieste/ask_tell_optimization.py
apaleyes: what would properties be good for?

trieste/ask_tell_optimization.py
apaleyes: That's a good question, haven't really thought about it. It may be also relevant in the context of retracing, so i won't change anything in that regard just yet.

trieste/ask_tell_optimization.py
apaleyes: i ended up adding some overloads, but mostly for acquisitions

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,251 @@
...
+                f"new_data keys {new_data.keys()} doesn't "
+                f"match dataset keys {self._datasets.keys()}"
+            )
+
+        self._datasets = {tag: self._datasets[tag] + new_data[tag] for tag in new_data}
+
+        for tag, model in self._models.items():
+            dataset = self._datasets[tag]
+            model.update(dataset)
+            model.optimize(dataset)
```
henrymoss, line 294: I wonder if you always want to optimize the model after every tell?

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,154 @@
...
+num_initial_points = 5
+initial_query_points = search_space.sample(num_initial_points)
+observer = mk_observer(scaled_branin)
+initial_data = observer(initial_query_points)
+model = build_model(initial_data)
+
+# %% [markdown]
+# ## Timing acquisition function: simple use case for Ask-Tell
+#
+# Let's say we are very concerned with the performance of the acqusition function, and want a simple way of measuring its performance over the course of the optimization. At the time of writing these lines, regular Trieste's optimizer does not provide such coustomization functionality, and this is where Ask-Tell comes in handy.
```
henrymoss, line 57: I think I would prefer to have a more compelling example (perhaps as well as this one). e.g something about going away to make physical experiments

trieste/ask_tell_optimization.py
apaleyes: It might be a useful idea, although I am struggling to come up with a concrete example. It is easy to add it later if needed, so let's leave it like this for now, and see if relevant use case comes up. Sounds good?

henrymoss: If the model fitting is slow, you might want to do a few ask-tells before refitting. But yeah, lets leave it fo rnow

docs/tutorials.rst
```diff
@@ -39,11 +39,13 @@ The following tutorials (or sections thereof) explain how to use and extend spec
 * :doc:`How do I set up a batch Bayesian optimization routine?<notebooks/batch_optimization>`
 * :ref:`How do I make a custom acquisition function?<notebooks/failure_ego:Create a custom acquisition function>`
 * :doc:`How do I recover a failed optimization loop?<notebooks/recovering_from_errors>`
+* :doc:`Does Trieste support Ask-Tell optimization?<notebooks/ask_tell_optimization>`
```
hstojic: point of this interface is user-friendly access to BO for users without much experience with BO, right? so we might want to use something more descriptive here instead of ask-tell?

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,251 @@
...
+from .models import ModelSpec, create_model
+from .observer import OBJECTIVE
+from .space import SearchSpace
+from .types import State, TensorType
+from .utils import Ok, map_values
+
+S = TypeVar("S")
+""" Unbound type variable. """
+
+SP = TypeVar("SP", bound=SearchSpace)
```
hstojic, line 38: this can be imported from `space`, right?

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,251 @@
...
+from .acquisition.rule import AcquisitionRule, EfficientGlobalOptimization
+from .bayesian_optimizer import OptimizationResult, Record
+from .data import Dataset
+from .models import ModelSpec, create_model
+from .observer import OBJECTIVE
+from .space import SearchSpace
+from .types import State, TensorType
+from .utils import Ok, map_values
+
+S = TypeVar("S")
```
hstojic, line 35: this can be imported from `types`, right?

docs/tutorials.rst
apaleyes: i don't think that's the goal. if anything Ask/Tell is more complex for users to use, because they need to handle the optimization loop themselves.

that said, i am happy to update the description to make it more user-friendly. How about `Does Trieste have interface for external control of the optimization loop?`

trieste/ask_tell_optimization.py
apaleyes: It's the same one line plus few raised eyebrows, so we are hardly saving while harming readability. I suspect if anything, it may only confuse mypy. The purpose of these type variables, as i understand it, is to serve as placeholders for actual types, and making use of the same placeholder in two different places doesn't sound right. @uri-granta could probably comment on that with more knowledge

docs/notebooks/ask_tell_optimization.pct.py
hstojic: coustomization -> customization

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,154 @@
...
+    })
+
+num_initial_points = 5
+initial_query_points = search_space.sample(num_initial_points)
+observer = mk_observer(scaled_branin)
+initial_data = observer(initial_query_points)
+model = build_model(initial_data)
+
+# %% [markdown]
+# ## Timing acquisition function: simple use case for Ask-Tell
```
hstojic, line 51: this is nice but less important than the other two use cases, I would put this one at the end

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,154 @@
...
+
+    new_data_point = observer(new_point)
+
+    dataset = dataset + new_data_point
+
+plot_ask_tell_regret(ask_tell.to_result())
+
+
+# %% [markdown]
+# ## External experiment: manipulating optimization state
```
hstojic: probably the most important use case, I would move it to the top

another example is optimising hyperpars of a big model that takes couple of days - might be closer for many users

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,154 @@
...
+#
+# Now let's suppose you are optimizing a real life process, e.g. a lab experiment. This time you cannot even express the objective function in Python code. Instead you would like to ask Trieste what configuration to run next, go to the lab, perform the experiment, collect data, feed it back to Trieste and ask for the next configuration, and so on. In such case you want to save the optimization state and then load it later.
+#
+# Of course we cannot perform a real physical experiment within this notebook, so we will just mimick it.
+
+# %%
+for step in range(n_steps):
+    print(f"Ask Trieste for configuration #{step}")
+    new_config = ask_tell.ask()
```
hstojic, line 142: reusing the `ask_tell` object from the example above in the last iteration?

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,154 @@
...
+    ax[0].set_ylabel("Regret")
+    ax[0].set_ylim(0.001, 100)
+    ax[0].set_xlabel("# evaluations")
+
+plot_ask_tell_regret(ask_tell.to_result())
+
+# %% [markdown]
+# ## Model switching: using only Ask part
+#
+# A slightly more complex use case. Let's suppose we want to switch between two models depending on some criteria dynamically during the optimization loop. Why would we do it? [Who knows!](https://bayesopt.github.io/papers/2014/paper13.pdf). In that case we can only use Ask part of the Ask-Tell interface.
```
hstojic: we are making it stranger than it is, model selection is a common issue that a lot of people would naturally wonder about - now that we would support more model types it would be even easier to do

docs/tutorials.rst
hstojic: in some ways its more complex, in others easier... but yes, description is definitely more user-friendly :)

trieste/ask_tell_optimization.py
hstojic: I agree, we are not saving much here and it seems to be harmless duplication. In any case, I'm still learning about typing, so you are probably right.

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: ok, will remove this weak attempt at humor

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: Good catch! This is now being rewritten though

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: good point, will mention this in the description

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: the order here is from the simplest to more advanced, as is befitting the tutorial

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: fixed!

docs/tutorials.rst
apaleyes: ok, changed

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
...
+    loaded_state = pickle.loads(saved_state)
+    ask_tell = AskTellOptimizer.from_record(loaded_state, search_space)
+    ask_tell.tell(Dataset(new_config, new_datapoint))
+
+
+plot_ask_tell_regret(ask_tell.to_result())
+
+
+# %% [markdown]
+# A word of warning. Serialization is not guaranteed to work smoothly with every GPflow model, so apply caution when using similar technique in your problems.
```
hstojic: perhaps remove GPflow, since we will have multiple model classes?

apaleyes: But we can't claim anything about other types of models yet, can we? Although perhaps we can say that this simple serialization technique isn't guaranteed to work with tensorflow in general. Would that sound better?

apaleyes: Updated to this more general formulation

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
+# %% [markdown]
+# # Ask-Tell Optimization Interface
+
+# %% [markdown]
+# In this notebook we will illustrate the use of Ask-Tell interface in Trieste. It might be useful in the cases when you want to have greater control of the optimization loop, or if letting Trieste manage this loop is impossible.
+#
+# First, some commmon code we will be using in the notebook.
```
vpicheny: maybe cross-ref to the intro notebook for more details on how to set up the problem?

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
...
+
+plot_ask_tell_regret(ask_tell.to_result())
+
+
+# %% [markdown]
+# ## External experiment: storing optimizer state
+#
+# Now let's suppose you are optimizing a process that takes hours or even days to complete, e.g. a lab experiment or a hyperparameter optimization of a big machine learning model. This time you cannot even express the objective function in Python code. Instead you would like to ask Trieste what configuration to run next, go to the lab, perform the experiment, collect data, feed it back to Trieste and ask for the next configuration, and so on. It would be very convenient to be able to store intermediate optimization state to disk or database or other storage, so that your machine can be switched off while you are waiting for observation results. 
+# 
+# In this section we'll show how you could do it with Ask-Tell in Trieste. Of course we cannot perform a real physical experiment within this notebook, so we will just mimick it ny using pickle to write optimization state and read it back.
```
vpicheny: typo: "it ny using"

apaleyes: fixed

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: Good point, added a reference

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,254 @@
...
+class AskTellOptimizer(Generic[SP]):
+    """
+    This class provides Ask/Tell optimization interface. It is designed for those use cases
+    when control of the optimization loop by Trieste is impossible or not desirable.
+    For more details about the Bayesian Optimization routine, refer to :class:`BayesianOptimizer`.
+    """
+
+    @overload
+    def __init__(
+        self,
```
vpicheny, line 51: I think that we are missing an overload for the single model / single dataset case, which is quite user-friendly.

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,254 @@
...
+        datasets: Mapping[str, Dataset] | Dataset,
+        model_specs: Mapping[str, ModelSpec] | ModelSpec,
+        acquisition_rule: AcquisitionRule[TensorType | State[S | None, TensorType], SP]
+        | None = None,
+        acquisition_state: Optional[S] = None,
+        *,
+        fit_model: bool = True,
+    ):
+        """
+        :param search_space: The space over which to search.
```
vpicheny: ....to search for the next query point?

apaleyes: sure, why not

trieste/ask_tell_optimization.py
apaleyes: Ok, help me understand:

        datasets: Mapping[str, Dataset] | Dataset,
        model_specs: Mapping[str, ModelSpec] | ModelSpec,

That already allows single instance for each of the params. Why have separate overload?

vpicheny: in bayesian_optimization we have:     
```
@overload
    def optimize(
        self,
        num_steps: int,
        datasets: Dataset,
        model_specs: ModelSpec,
        *,
        track_state: bool = True,
        fit_initial_model: bool = True,
    ) -> OptimizationResult[None]:
```

so you don't have to bother with a tag if you have a single model

apaleyes: Sure, same is allowed with ask/tell. The method signature allows it, and code handles it too, type checks pass. All notebook examples are with single dataset/model instance pair. So I am just not sure what overload is for.

vpicheny: wait - why do we have it for `BayesianOptimizer` then?

vpicheny: @uri-granta ??

apaleyes: Checked [PEP](https://www.python.org/dev/peps/pep-0484/#function-method-overloading), and these overloads make more sense now. They exist to reinforce the fact that datasets and models can either both be a single instance, or both a mapping. Not (dataset, mapping) or (mapping, model) allowed. It seems reasonable and a good typecheck, so I've added similar overloads to AskTellOptimizer

vpicheny (approved):

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
...
+    """kernel_func should be a function that takes variance as a single input parameter"""
+    variance = tf.math.reduce_variance(data.observations)
+    if kernel_func is None:
+        kernel = gpflow.kernels.Matern52(variance=variance)
+    else:
+        kernel = kernel_func(variance)
+    gpr = gpflow.models.GPR(data.astuple(), kernel, noise_variance=1e-5)
+    gpflow.set_trainable(gpr.likelihood, False)
+
+    return GPflowModelConfig(**{
```
hstojic: since you are using a default optimizer, you can simplify things here a bit by returning a `GaussianProcessRegression` object instead, then below you don't need to use `create_model` function

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,290 @@
...
+        """Creates new :class:`~AskTellOptimizer` instance from provided optimization state.
+
+        :param record: Optimization state record.
+        :param search_space: The space over which to search for the next query point.
+        :param acquisition_rule: The acquisition rule, which defines how to search for a new point
+            on each optimization step. Defaults to
+            :class:`~trieste.acquisition.rule.EfficientGlobalOptimization` with default
+            arguments.
+        """
+        # we are recovering previously saved optimization state
```
hstojic, line 217: perhaps move this comment to the docstring?

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,290 @@
...
+            fit_model=False,
+        )
+
+    def to_record(self) -> Record[S]:
+        """Collects the current state of the optimization, which includes datasets,
+        models and acquisition state (if applicable).
+
+        :return: An optimization state record.
+        """
+        models_copy = copy.deepcopy(self._models)
```
hstojic, line 238: should we check here if deepcopy method is available and give a user-friendly message that the model cannot be saved? though perhaps a better place for this check might be in  models interfaces...
(Keras models don't have deepcopy I think, and would produce a less user-friendly message)

docstring could perhaps include a warning to users related to this, along the lines what you have in the notebook

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,290 @@
...
+        acquisition_state_copy = copy.deepcopy(self._acquisition_state)
+        return Record(
+            datasets=self._datasets, models=models_copy, acquisition_state=acquisition_state_copy
+        )
+
+    def to_result(self) -> OptimizationResult[S]:
+        """Converts current state of the optimization
+        into an :class:`~trieste.data.OptimizationResult` object."""
```
hstojic: an -> a

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,290 @@
...
+        )
+
+    def to_record(self) -> Record[S]:
+        """Collects the current state of the optimization, which includes datasets,
+        models and acquisition state (if applicable).
+
+        :return: An optimization state record.
+        """
+        models_copy = copy.deepcopy(self._models)
+        acquisition_state_copy = copy.deepcopy(self._acquisition_state)
```
hstojic, line 235: is this meaningful only when acquisition function is stateful? even if this passes for stateless acquisition functions, would it mean that you couldn't continue optimization from a `Record`?
seems like it might be worth having a warning about this in the docstring

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,290 @@
...
+            raise ValueError(
+                f"new_data keys {new_data.keys()} doesn't "
+                f"match dataset keys {self._datasets.keys()}"
+            )
+
+        self._datasets = {tag: self._datasets[tag] + new_data[tag] for tag in new_data}
+
+        for tag, model in self._models.items():
+            dataset = self._datasets[tag]
+            model.update(dataset)
```
hstojic, line 293: just to make sure, does this update models in the `self._models` object?

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
+# %% [markdown]
+# # Ask-Tell Optimization Interface
+
+# %% [markdown]
+# In this notebook we will illustrate the use of Ask-Tell interface in Trieste. It might be useful in the cases when you want to have greater control of the optimization loop, or if letting Trieste manage this loop is impossible.
```
henrymoss: In this notebook we will illustrate the use of **an** Ask-Tell interface in Trieste. **It is** useful for cases where you want to have greater control of the optimization loop, or **when letting** Trieste manage this loop is impossible.

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
...
+#
+# First, some code to setup the problem we will be using throughout the notebook. If you would like more details about this problem setup, please refer to [introductory EI notebook](expected_improvement.ipynb).
```
henrymoss: First, some code to **set up** the problem we will be using throughout the notebook. If you would like more details about this problem setup, please refer to [introductory EI notebook](expected_improvement.ipynb).

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
...
+num_initial_points = 5
+initial_query_points = search_space.sample(num_initial_points)
+observer = mk_observer(scaled_branin)
+initial_data = observer(initial_query_points)
+model = build_model(initial_data)
+
+# %% [markdown]
+# ## Timing acquisition function: simple use case for Ask-Tell
+#
+# Let's say we are very concerned with the performance of the acqusition function, and want a simple way of measuring its performance over the course of the optimization. At the time of writing these lines, regular Trieste's optimizer does not provide such customization functionality, and this is where Ask-Tell comes in handy.
```
henrymoss, line 53: I think this is the least exciting of the examples? Is there a reason why this is the first one in the notebook?

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
...
+    stop = timeit.default_timer()
+
+    print(f"Time at step {step + 1}: {stop - start}")
+
+    new_data = observer(new_point)
+    ask_tell.tell(new_data)
+
+
+# %% [markdown]
+# Once ask-tell optimization is over, you can get an optimization result object from it and perform whatever analysis you need. Just like with regular Trieste optimization interface. For instance here we will plot regret for each optimization step.
```
henrymoss: Once ask-tell optimization is over, you can extract an optimization result object and perform whatever analysis you need, just like with regular Trieste optimization interface. For instance, here we will plot regret for each optimization step.

docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -0,0 +1,166 @@
...
+    ax.set_ylabel("Regret")
+    ax.set_ylim(0.001, 100)
+    ax.set_xlabel("# evaluations")
+
+plot_ask_tell_regret(ask_tell.to_result())
+
+# %% [markdown]
+# ## Model selection: using only Ask part
+#
+# A slightly more complex use case. Let's suppose we want to switch between two models depending on some criteria dynamically during the optimization loop. So we want greater control of the model such as training it outside of Trieste. In that case we can only use Ask part of the Ask-Tell interface.
```
henrymoss: **We now turn to a** slightly more complex use case. Let's suppose we want to switch between two models depending on some criteria dynamically during the optimization loo**p, e.g. we want greater to be able to train a model outside of Trieste**. In **this** case we can only use Ask part of the Ask-Tell interface.

tests/integration/test_bayesian_optimization.py
```diff
@@ -56,9 +56,12 @@
...
         List[
-            Union[
-                AcquisitionRule[TensorType, Box],
-                AcquisitionRule[State[TensorType, TrustRegion.State], Box],
+            Tuple[
+                int,
+                Union[
+                    AcquisitionRule[TensorType, Box],
+                    AcquisitionRule[State[TensorType, TrustRegion.State], Box],
+                ],
```
henrymoss, line 64: WHy has this changed?

tests/unit/test_ask_tell_optimization.py
```diff
@@ -0,0 +1,294 @@
...
+    new_data_with_key_2 = {"2": mk_dataset([[1.0]], [[1.0]])}
+
+    ask_tell = AskTellOptimizer(
+        search_space, dataset_with_key_1, model_with_key_1, acquisition_rule
+    )
+    with pytest.raises(ValueError):
+        ask_tell.tell(new_data_with_key_2)
+
+
+def test_ask_tell_optimizer_default_acqusition_requires_objective_tag(
```
hstojic: acqusition -> acquisition

hstojic (approved):
left a few minor comments, but overall looks great!

trieste/models/gpflow/models.py
```diff
@@ -57,8 +57,27 @@ def __init__(
...
+        return self._model
+
+    def _ensure_variable_model_data(self) -> None:
         # GPflow stores the data in Tensors. However, since we want to be able to update the data
         # without having to retrace the acquisition functions, put it in Variables instead.
+        # To tell is the data is stored properly, we look at its first dimension size
+        # If it is a number, TF won't allow it to change, and we reset it to None
+
+        if self._model.data[0].shape[0] is None and self._model.data[1].shape[0] is None:
+            # both query points and observations are in right shape
```
henrymoss: I dont get what this is doing?

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: fair enough, changed it

trieste/ask_tell_optimization.py
apaleyes: Left it where it is, but added some short explanation of this behavior to the docstring too

trieste/ask_tell_optimization.py
apaleyes: fixed

trieste/ask_tell_optimization.py
apaleyes: How can this be checked? Just because an object doesn't override `__deepcopy__` doesn't mean it cannot be deepcopied. What would happen if one tries to deepcopy a Keras model?

trieste/ask_tell_optimization.py
apaleyes: Yup!

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: changed

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: fixed

docs/notebooks/ask_tell_optimization.pct.py
apaleyes: Because it's the simplest I could think of, to get users familiar with the interface without any kind of complex mechanics. Besides, that's the only one in the notebook that looks real without pretending.

tests/integration/test_bayesian_optimization.py
apaleyes: The inputs are actually tuples of int and acq function, but only type of the acquisition function was declared. Why mypy hasn't caught it is the real question, to which I have no answer

tests/unit/test_ask_tell_optimization.py
apaleyes: fixed!

trieste/models/gpflow/models.py
apaleyes: clarified "properly stored" bit

trieste/models/gpflow/models.py
```diff
@@ -57,8 +57,28 @@ def __init__(
...
+        return f"GaussianProcessRegression({self._model!r}, {self.optimizer!r})"
+
+    @property
+    def model(self) -> GPR | SGPR:
+        return self._model
+
+    def _ensure_variable_model_data(self) -> None:
         # GPflow stores the data in Tensors. However, since we want to be able to update the data
         # without having to retrace the acquisition functions, put it in Variables instead.
+        # To tell if the data is stored in a way it can be updated without retracing,
```
uri-granta: I don't think this comment quite explains why we need this. Maybe something like:

> Following seriasliation-deserialisation, the dynamic Variable shape can get overwritten (preventing  subsequent changes to the data) so we also ensure it's correct before every update.

trieste/models/gpflow/models.py
```diff
@@ -57,8 +57,28 @@ def __init__(
...
+        # we look at its first dimension size.
+        # If it is a number, TF won't allow it to change, and we reset it to None
+
+        if self._model.data[0].shape[0] is None and self._model.data[1].shape[0] is None:
```
uri-granta: Maybe as follows (slightly cleaner and more robust):

```suggestion
        if all(isinstance(x, tf.Variable) and x.shape[0] is None for x in self._model.data):
```

trieste/ask_tell_optimization.py
uri-granta: Yes, the overoads are intended to catch accidental errors such as passing in state for a rule that doesn't expect it, or passing in a dataset mapping with a single model (which isn't necessarily an error, but is unlikely to be on purpose).

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,291 @@
...
+    ):
+        ...
+
+    @overload
+    def __init__(
+        self,
+        search_space: SP,
+        datasets: Mapping[str, Dataset],
+        model_specs: Mapping[str, ModelSpec],
+        acquisition_rule: AcquisitionRule[TensorType, SP] | None = None,
```
uri-granta: The rule shouldn't be optional in this overload. Not having a rule was already covered by a previous overload.

```suggestion
        acquisition_rule: AcquisitionRule[TensorType, SP],
```

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,291 @@
...
+    ):
+        ...
+
+    @overload
+    def __init__(
+        self,
+        search_space: SP,
+        datasets: Mapping[str, Dataset],
+        model_specs: Mapping[str, ModelSpec],
+        acquisition_rule: AcquisitionRule[State[S | None, TensorType], SP] | None = None,
```
uri-granta: Ditto.

```suggestion
        acquisition_rule: AcquisitionRule[State[S | None, TensorType], SP],
```

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,291 @@
...
+    ):
+        ...
+
+    @overload
+    def __init__(
+        self,
+        search_space: SP,
+        datasets: Dataset,
+        model_specs: ModelSpec,
+        acquisition_rule: AcquisitionRule[TensorType, SP] | None = None,
```
uri-granta: same comment as above

trieste/ask_tell_optimization.py
```diff
@@ -0,0 +1,291 @@
...
+    ):
+        ...
+
+    @overload
+    def __init__(
+        self,
+        search_space: SP,
+        datasets: Dataset,
+        model_specs: ModelSpec,
+        acquisition_rule: AcquisitionRule[State[S | None, TensorType], SP] | None = None,
```
uri-granta: and again

trieste/ask_tell_optimization.py
uri-granta: We could catch any exception and reraise with an additional error message, but I don't think that'll really add much value. Maybe mention in the docstring that this will fail if the models or acquisition state are not deep copyable?

tests/integration/test_ask_tell_optimization.py
```diff
@@ -0,0 +1,150 @@
...
+)
+from trieste.objectives.utils import mk_observer
+from trieste.observer import OBJECTIVE
+from trieste.space import Box, SearchSpace
+from trieste.types import State, TensorType
+
+
+@random_seed
+@pytest.mark.parametrize(
+    "num_steps, reload_state, acquisition_rule_fn",
```
uri-granta, line 46: If the test isn't too slow then it might make sense to just parametrize `reload_state` separately, so each test case gets run once with reload_state and the other without. If it is slow, then you should consider whether we really want to run all the permutations.

uri-granta (approved):
A few comments but overall LGTM!

trieste/models/gpflow/models.py
apaleyes: It's not only about serialization though. I've merged the two description into a single comment

trieste/models/gpflow/models.py
apaleyes: looks good!

tests/integration/test_ask_tell_optimization.py
apaleyes: that's a good point. i guess we can remove one permutation indeed

trieste/ask_tell_optimization.py
apaleyes: fixed

trieste/ask_tell_optimization.py
hstojic: Joelism :)

PR #348, opened 2021-09-13
Use extend-exclude for flake8 and black

apaleyes:
This PR makes a bit more extensive use of `extend-exclude` option, pun intended.

I've hit a couple of issues running all format checks because in my dev setup `.venv` lives in the project root. So both tools would validated files inside this `.venv`.

Black by default can respect .gitignore file, which covers .venv, .tox and many more. But if you use `exclude` option, this default behavior is overriden. So for black we simply use `extend-exclude` instead of `exclude`, and then clean up in other places.

flake8 doesn't respect .gitignore at all, but it has sensbile defaults that cover .tox. Thus for flake8 we only need to add `.venv` to the list of excludes, with the same `extend-exclude` option.

While this is wasn't an issue in Trieste development so far, I argue it would very contributor-friendly to support .venv in project root,  as this setup is [quite](https://github.com/python-poetry/poetry/issues/108) [popular](https://github.com/pypa/pipenv/issues/259).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ testpaths = [
 line-length = 100
 target-version = ['py37']
 include = '\.py$'
-exclude = '\.pct.py$'
+extend-exclude = '\.pct.py$'
 
 [tool.isort]
 profile = "black"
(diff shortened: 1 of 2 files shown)


hstojic (approved):
looks reasonable to me, thanks for spotting this!

hstojic:
ready for merging in?

apaleyes:
Thanks, @hstojic ! I can't merge it in myself though

PR #349, opened 2021-09-14
Refactor GIBBON to avoid retracing

uri-granta:


trieste/acquisition/function.py
```diff
@@ -477,6 +464,8 @@ def __init__(self, model: ProbabilisticModel, samples: TensorType):
 
     def update(self, samples: TensorType) -> None:
         """Update the acquisition function with new samples."""
+        tf.debugging.assert_rank(samples, 2)
+        tf.debugging.assert_positive(len(samples))
         self._samples.assign(samples)
```
henrymoss, line 469: I've added in a few of these extra asserts. Was there a reason not to have these before? I guess they will need unit testing now

trieste/acquisition/function.py
```diff
@@ -1685,156 +1677,281 @@ def prepare_acquisition_function(
...
+        self,
+        model: ProbabilisticModel,
+        pending_points: TensorType,
+        rescaled_repulsion: bool = True,
+    ):
+        r"""
+        GIBBON's repulsion term encourages diversity within the batch
+        (achieving high values for points with low predictive correlation).
+
+        The term :math:`r=\log |C|` is given by the log determinant of the predictive
```
henrymoss: the *replusion* term

trieste/acquisition/function.py
```diff
@@ -1685,156 +1677,281 @@ def prepare_acquisition_function(
...
+            num_samples,
+            model,
+            dataset,
+            sample_min_value=True,
+            num_features=num_fourier_features,
+        )
+    else:  # use exact Thompson sampler
+        sampler = ExactThompsonSampler(num_samples, model, sample_min_value=True)
+
+    return sampler.sample(sampled_points)
```
henrymoss, line 1959: This needs unit testing

uri-granta: These is a private function so wouldn't normally be unit tested directly. The code paths are tested instead via MinValueEntropySearch and GIBBON.

henrymoss (approved):
LGTM!

PR #350, opened 2021-09-15
Added Michalewicz function

sebastianober:


tests/unit/objectives/test_single_objectives.py
```diff
@@ -81,7 +92,7 @@ def test_objective_maps_minimizers_to_minimum(
 ) -> None:
     objective_values_at_minimizers = objective(minimizers)
     tf.debugging.assert_shapes([(objective_values_at_minimizers, [len(minimizers), 1])])
-    npt.assert_allclose(objective_values_at_minimizers, tf.squeeze(minimum), atol=1e-4)
+    npt.assert_allclose(objective_values_at_minimizers, tf.squeeze(minimum), atol=2e-4)
```
henrymoss: How many samples do you need to get within the existing tolerance?

tests/unit/objectives/test_single_objectives.py
```diff
@@ -37,6 +37,13 @@
     LOGARITHMIC_GOLDSTEIN_PRICE_MINIMIZER,
     LOGARITHMIC_GOLDSTEIN_PRICE_MINIMUM,
     LOGARITHMIC_GOLDSTEIN_PRICE_SEARCH_SPACE,
+    MICHALEWICZ_2_MINIMIZER,
+    MICHALEWICZ_2_MINIMUM,
+    MICHALEWICZ_2_SEARCH_SPACE,
+    MICHALEWICZ_5_MINIMUM,
+    MICHALEWICZ_5_SEARCH_SPACE,
+    MICHALEWICZ_10_MINIMUM,
+    MICHALEWICZ_10_SEARCH_SPACE,
```
henrymoss: Where are the minimizers for 2 and 5?

sebastianober: I don't know them, they're not listed on the [https://www.sfu.ca/~ssurjano/michal.html](url) site...

tests/unit/objectives/test_single_objectives.py
sebastianober: There aren't any samples here?

tests/unit/objectives/test_single_objectives.py
henrymoss: Work them out? Probably useful to have them?

tests/unit/objectives/test_single_objectives.py
henrymoss: My bad. So how come we are outside tolerance then???

henrymoss (approved):

tests/unit/objectives/test_single_objectives.py
```diff
@@ -50,6 +59,9 @@
     hartmann_3,
     hartmann_6,
     logarithmic_goldstein_price,
+    michalewicz_2d,
```
hstojic: we should probably be consistent with naming of other functions and remove `d` part at the end in these 2d, 5d and 10d

hstojic (approved):
looks good, we would need to change the names slightly, see my comment above

tests/unit/objectives/test_single_objectives.py
sebastianober: Good point, I've fixed this now. Anything else?

PR #352, opened 2021-09-21
Update more acquisition function builders

uri-granta:
Update the following acquisition function builders to support in place updates and avoid retracing:

* ExpectedConstrainedImprovement
* ExpectedHypervolumeImprovement
* ExpectedConstrainedHypervolumeImprovement

Still to do (in a separate PR):

* BatchMonteCarloExpectedHypervolumeImprovement


trieste/acquisition/function.py
```diff
@@ -706,6 +706,9 @@ def __init__(
         self._objective_tag = objective_tag
         self._constraint_builder = constraint_builder
         self._min_feasibility_probability = min_feasibility_probability
+        self._constraint_fn: Optional[AcquisitionFunction] = None
+        self._improvement_fn: Optional[AcquisitionFunction] = None
+        self._constrained_fn: Optional[AcquisitionFunction] = None
```
apaleyes: Can these have a bit more distinctive names? It is quite hard to figure out which of two `_constrain...` is being used while reading the code.

trieste/acquisition/function.py
```diff
@@ -735,7 +738,52 @@ def prepare_acquisition_function(
             " objective data, but the objective data is empty.",
         )
 
-        constraint_fn = self._constraint_builder.prepare_acquisition_function(datasets, models)
+        self._constraint_fn = self._constraint_builder.prepare_acquisition_function(
+            datasets, models
+        )
+        pof = self._constraint_fn(objective_dataset.query_points[:, None, ...])
```
apaleyes, line 744: perhaps it won't hurt to spell out `pof` here?

trieste/acquisition/function.py
```diff
@@ -745,9 +793,20 @@ def prepare_acquisition_function(
         feasible_query_points = tf.boolean_mask(objective_dataset.query_points, is_feasible)
         feasible_mean, _ = objective_model.predict(feasible_query_points)
         eta = tf.reduce_min(feasible_mean, axis=0)
-        function = expected_improvement(objective_model, eta)
+        tf.debugging.Assert(isinstance(self._improvement_fn, expected_improvement), [])
+        self._improvement_fn.update(eta)  # type: ignore
```
apaleyes: Is the expectation that if lines 747-748 are entered (early return in `prepare`), execution never reaches this line here? If it does, this assertion will fail

trieste/acquisition/function.py
```diff
@@ -735,7 +738,52 @@ def prepare_acquisition_function(
...
+        eta = tf.reduce_min(feasible_mean, axis=0)
+        self._improvement_fn = expected_improvement(objective_model, eta)
+
+        @tf.function
+        def constrained_function(x: TensorType) -> TensorType:
+            return cast(AcquisitionFunction, self._improvement_fn)(x) * cast(
+                AcquisitionFunction, self._constraint_fn
+            )(x)
+
+        self._constrained_fn = cast(AcquisitionFunction, constrained_function)
```
apaleyes: Why does this cast work?

trieste/acquisition/function.py
```diff
@@ -745,9 +793,20 @@ def prepare_acquisition_function(
...
+            return self._constrained_fn
+
+        @tf.function
+        def constrained_function(x: TensorType) -> TensorType:
+            return cast(AcquisitionFunction, self._improvement_fn)(x) * cast(
+                AcquisitionFunction, self._constraint_fn
+            )(x)
 
-        return lambda at: function(at) * constraint_fn(at)
+        self._constrained_fn = constrained_function
```
apaleyes: There is a cast in similar situation in `prepare`

trieste/acquisition/function.py
```diff
@@ -735,7 +738,52 @@ def prepare_acquisition_function(
...
+        tf.debugging.assert_positive(
+            len(objective_dataset),
+            message="Expected improvement is defined with respect to existing points in the"
+            " objective data, but the objective data is empty.",
+        )
+        tf.debugging.Assert(None not in [self._constraint_fn], [])
+
+        constraint_fn = cast(AcquisitionFunction, self._constraint_fn)
+        self._constraint_builder.update_acquisition_function(constraint_fn, datasets, models)
         pof = constraint_fn(objective_dataset.query_points[:, None, ...])
```
apaleyes, line 786: ditto about `pof`

trieste/acquisition/function.py
uri-granta: :+1: 

FYI this PR isn't actually ready yet (I should have put it in draft, sorry). Still missing new tests, and I'm panning to convert another few functions as part of the same PR.

trieste/acquisition/function.py
uri-granta: Good catch. This does need to be handled (and I'll make sure it's covered in a unit test).

trieste/acquisition/function.py
uri-granta: This cast shouldn't actually be necessary, as mypy should know that constrained_function is an AcquisitionFunction and not None. For some reason though it immediately forgets this on the next line. Though it'd probably be clearer to remove the cast and just return constrained_function instead.

trieste/acquisition/function.py
uri-granta: See comment above. The reason a cast isn't necessary here is that mypy infers that self._constrained_fn can't be None due to the earlier if clause. (Though as I said, I'll remove the cast from further up too.)

trieste/acquisition/function.py
apaleyes: ok, noted, will hold off further reviewing for a bit

trieste/acquisition/function.py
```diff
@@ -744,10 +791,36 @@ def prepare_acquisition_function(
...
+        """
         eta = tf.reduce_min(feasible_mean, axis=0)
-        function = expected_improvement(objective_model, eta)
 
-        return lambda at: function(at) * constraint_fn(at)
+        if self._expected_improvement_fn is None:
+            self._expected_improvement_fn = expected_improvement(objective_model, eta)
+        else:
+            tf.debugging.Assert(isinstance(self._expected_improvement_fn, expected_improvement), [])
+            self._expected_improvement_fn.update(eta)  # type: ignore
```
henrymoss, line 823: It there any reason why this is its own function?

uri-granta: Yes, it allows ExpectedConstrainedHypervolumeImprovement to inherit from ExpectedConstrainedImprovement without having to duplicate lots of code. Also, it's now used both when preparing and updating the functions.

trieste/acquisition/function.py
```diff
@@ -779,37 +852,70 @@ def prepare_acquisition_function(
...
+            partition_bounds[1], trainable=False, shape=[None, partition_bounds[1].shape[-1]]
+        )
+        self._cross_index = tf.constant(
+            list(product(*[[0, 1]] * self._lb_points.shape[-1]))
+        )  # [2^d, indices_at_dim]
+
+    def update(self, partition_bounds: tuple[TensorType, TensorType]) -> None:
+        """Update the acquisition function with new partition bounds."""
+        self._lb_points.assign(partition_bounds[0])
+        self._ub_points.assign(partition_bounds[1])
```
apaleyes, line 915: May we ever want to serialize an acquisition object? If there is a chance, e.g. pickle the whole BayesianOptimizer object for some reason, it's probably worth ensuring dynamic shapes here, as we did in models recently. If serializing acquisition object makes no sense, don't bother

trieste/acquisition/function.py
```diff
@@ -779,37 +852,70 @@ def prepare_acquisition_function(
...
 
+    def update_acquisition_function(
+        self, function: AcquisitionFunction, dataset: Dataset, model: ProbabilisticModel
+    ) -> AcquisitionFunction:
+        """
+        :param function: The acquisition function to update.
+        :param dataset: The data from the observer.
+        :param model: The model over the specified ``dataset``.
+        """
+        tf.debugging.assert_positive(len(dataset), message="Dataset must be populated.")
```
apaleyes, line 863: For education more than anything else: why use TF's assert instead of just plain if...raise?

trieste/acquisition/function.py
```diff
@@ -735,7 +738,51 @@ def prepare_acquisition_function(
...
+        """
+        objective_model = models[self._objective_tag]
+        objective_dataset = datasets[self._objective_tag]
+
+        tf.debugging.assert_positive(
+            len(objective_dataset),
+            message="Expected improvement is defined with respect to existing points in the"
+            " objective data, but the objective data is empty.",
+        )
+        tf.debugging.Assert(None not in [self._constraint_fn], [])
```
apaleyes: Is that just `self._constraint_fn is not None`?

tests/unit/acquisition/test_function.py
```diff
@@ -1692,18 +1738,29 @@ def prepare_acquisition_function(
         ) -> AcquisitionFunction:
             return lambda x: tf.ones_like(tf.squeeze(x, -2))
 
-    data = {"foo": Dataset(train_x, model_pred_observation)}
+    data = {"foo": Dataset(train_x[:5], model_pred_observation[:5])}
```
apaleyes, line 1741: Why are these sliced now?

tests/unit/acquisition/test_function.py
```diff
@@ -1692,18 +1738,29 @@ def prepare_acquisition_function(
...
     npt.assert_allclose(echvi(at), ehvi(at))
 
+    new_data = {"foo": Dataset(train_x, model_pred_observation)}
+    up_echvi = builder.update_acquisition_function(echvi, new_data, models_)
+    assert up_echvi == echvi
+    up_ehvi = (
+        ExpectedHypervolumeImprovement()
+        .using("foo")
+        .prepare_acquisition_function(new_data, models_)
+    )
```
apaleyes, line 1759: Why is that necessary, if basically the same happened two lines above?

trieste/acquisition/function.py
```diff
@@ -779,37 +852,70 @@ def prepare_acquisition_function(
...
+        """
+        self._model = model
+        self._lb_points = tf.Variable(
+            partition_bounds[0], trainable=False, shape=[None, partition_bounds[0].shape[-1]]
+        )
+        self._ub_points = tf.Variable(
+            partition_bounds[1], trainable=False, shape=[None, partition_bounds[1].shape[-1]]
+        )
+        self._cross_index = tf.constant(
+            list(product(*[[0, 1]] * self._lb_points.shape[-1]))
```
apaleyes, line 909: Not sure I understand what cross_index is for, probably worth explaining in a comment. Also:

```suggestion
            list(product([0, 1] * self._lb_points.shape[-1]))
```

apaleyes (commented):
Nothing major, just some nitpicking and clarifying questions

trieste/acquisition/function.py
uri-granta: If ... raise doesn't work well with functions wrapped with tf.function.

trieste/acquisition/function.py
uri-granta: I'm pretty sure there was some obscure tensorflow reason, but changing it doesn't seem to make the unit tests fail! I'll update this (and a few other similar cases) and see if anything breaks.

tests/unit/acquisition/test_function.py
uri-granta: The testnow  prepares the function with some of the data, then later updates it with all of the data (to test that updating works without retracing).

tests/unit/acquisition/test_function.py
uri-granta: This is creating an updated ehvi function based on the new data to compare the updated echvi function (it's actually creating a brand new function but that's ok as we're not testing updating ehvi here, only echvi).

trieste/acquisition/function.py
uri-granta: Not quite. Product takes a list of arguments, not a single argument that's a list: `product([0,1], [0,1], [0,1])` not `product([[0,1],[0,1],[0,1]])` or `product([0,1,0,1,0,1])`.

apaleyes: i stand corrected, thanks

tests/unit/acquisition/test_function.py
apaleyes: ok, i see , thanks

tests/unit/acquisition/test_function.py
apaleyes: got it

trieste/acquisition/function.py
uri-granta: I can't immediately see why we'd do this, so will leave it for now.

apaleyes: ok!

apaleyes (approved):

henrymoss (approved):

Issue #353, opened 2021-09-22
Allow specifying reference point upfront in multi-objective optimization acquisition function 


TsingQAQ:
**Describe the feature you'd like**
As a user, i'd like to specify the reference point upfront (now has been dynamically updated in Trieste), this may due to several reasons like:
1. i don't want the acquisition function to be too greedy (especially when there is only 1 pf points).
2. i have expert knowledge where the Pareto Frontier (PF) might be, specify reference point may leads to faster locating of the expected PF.

**Additional context**
In case we are specifying reference point upfront, there might be a situation that (in the beginning of BO) below the reference point there are no observations at all (e.g., 1. the problem is difficult, 2. there is no feasible point exists), in this scenario, there are actually 2 options: 
o1: We may want the partition method just return us the non-dominated region that has been formulated by anti-reference point and the specified reference point. 
o2: We can also just raise an error as [here](https://github.com/pytorch/botorch/blob/70143a73fec92d0df9ed6707baedf9215b8ac44d/botorch/acquisition/multi_objective/analytic.py#L136-L140).

PR #356, opened 2021-09-27
Allow empty obs in non-dominated space partition

TsingQAQ:
This is a PR for allowing empty/None observation specified for non-dominated partition. In some cases, there maybe no observations below a specified reference point. In this scenario, this functionality will just return a hyper-cell (bounds) that has been constructed by anti_reference and reference point, driving greedy acquisition function search within this region. 

**Main Modifications**
- change `observations` to optional arg
- change the order of args of `prepare_default_non_dominated_partition_bounds`
- Add return case when `observations` is not specified (or empty)

This PR is the 1st step of tackling #353 (option 1).

**Note**: we leave the responsibility of screening observations below reference point within `prepare_acquisition_function`.

### DIFF
--- a/trieste/acquisition/multi_objective/partition.py
+++ b/trieste/acquisition/multi_objective/partition.py
@@ -63,7 +63,7 @@ def prepare_default_non_dominated_partition_bounds(
         tf.debugging.assert_greater_equal(
             ref,
             anti_ref,
-            message="reference point containing dimensionality below default "
+            message=f"reference point: {ref} containing at least one value below default "
             "anti-reference point ([-1e10, ..., -1e10]), try specify a lower "
             "anti-reference point.",
         )
@@ -71,23 +71,24 @@ def prepare_default_non_dominated_partition_bounds(
             tf.debugging.assert_greater_equal(
                 obs,
                 anti_ref,
-                message="observations containing points below default "
+                message=f"observations: {obs} containing at least one value below default "
                 "anti-reference point ([-1e10, ..., -1e10]), try specify a lower "
                 "anti-reference point.",
             )
         return anti_ref
 
     tf.debugging.assert_shapes([(reference, ["D"])])
-    if (  # prepare default anti_reference point if not specified
-        anti_reference is None
-    ):  # if anti_reference point is not specified, use a -1e10 as default (act as -inf)
+    if anti_reference is None:
+        # if anti_reference point is not specified, use a -1e10 as default (act as -inf)
         anti_reference = specify_default_anti_reference_point(reference, observations)
-    else:  # anti_reference point is specified
+    else:
+        # anti_reference point is specified
         tf.debugging.assert_shapes([(anti_reference, ["D"])])
 
     if is_empty_obs(observations):  # if no valid observations
         assert tf.reduce_all(tf.less_equal(anti_reference, reference)), ValueError(
-            "anti_reference point contains at least one value larger than reference point"
+            f"anti_reference point: {anti_reference} contains at least one value larger "
+            f"than reference point: {reference}"
         )
         return tf.expand_dims(anti_reference, 0), tf.expand_dims(reference, 0)
     elif tf.shape(observations)[-1] > 2:
--- a/trieste/acquisition/multi_objective/partition.py
+++ b/trieste/acquisition/multi_objective/partition.py
@@ -38,7 +38,7 @@ def prepare_default_non_dominated_partition_bounds(
 
     :param observations: The observations for all objectives, with shape [N, D], if not specified
         or is an empty Tensor, a single non-dominated partition bounds constructed by reference
-        and anti_reference point.
+        and anti_reference point will be returned.
     :param anti_reference: a worst point to use with shape [D].
         Defines the lower bound of the hypercell. If not specified, will use a default value:
         -[1e10] * D.
@@ -53,10 +53,12 @@ def prepare_default_non_dominated_partition_bounds(
         shape.
     """
 
-    def not_valid_obs(obs: TensorType) -> bool:
+    def not_valid_obs(obs: Optional[TensorType]) -> bool:
         return obs is None or tf.equal(tf.size(observations), 0)
 
-    def specify_default_anti_reference_point(ref: TensorType, obs: TensorType) -> TensorType:
+    def specify_default_anti_reference_point(
+        ref: TensorType, obs: Optional[TensorType]
+    ) -> TensorType:
         anti_ref = -1e10 * tf.ones(shape=(tf.shape(reference)), dtype=reference.dtype)
         tf.debugging.assert_greater_equal(
             ref,
@@ -65,7 +67,7 @@ def prepare_default_non_dominated_partition_bounds(
             "anti-reference point ([-1e10, ..., -1e10]), try specify a lower "
             "anti-reference point.",
         )
-        if not not_valid_obs(obs):
+        if not not_valid_obs(obs):  # make sure given (valid) observations are larger than -1e10
             tf.debugging.assert_greater_equal(
                 obs,
                 anti_ref,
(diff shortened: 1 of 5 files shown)


trieste/acquisition/multi_objective/partition.py
```diff
@@ -45,35 +49,48 @@ def prepare_default_non_dominated_partition_bounds(
...
+    ):  # if anti_reference point is not specified, use a -1e10 as default (act as -inf)
+        anti_reference = specify_default_anti_reference_point(reference, observations)
+    else:  # anti_reference point is specified
+        tf.debugging.assert_shapes([(anti_reference, ["D"])])
+
+    if not_valid_obs(observations):  # if no valid observations
+        assert tf.reduce_all(tf.less_equal(anti_reference, reference)), ValueError(
+            "anti_reference points contains dimension bigger than reference point"
+        )
+        return tf.expand_dims(anti_reference, 0), tf.expand_dims(reference, 0)
```
vpicheny, line 93: is it OK that this doesn't return a `_BoundIndexPartition`?

trieste/acquisition/multi_objective/partition.py
```diff
@@ -45,35 +49,48 @@ def prepare_default_non_dominated_partition_bounds(
...
+    tf.debugging.assert_shapes([(reference, ["D"])])
+    if (  # prepare default anti_reference point if not specified
+        anti_reference is None
+    ):  # if anti_reference point is not specified, use a -1e10 as default (act as -inf)
+        anti_reference = specify_default_anti_reference_point(reference, observations)
+    else:  # anti_reference point is specified
+        tf.debugging.assert_shapes([(anti_reference, ["D"])])
+
+    if not_valid_obs(observations):  # if no valid observations
+        assert tf.reduce_all(tf.less_equal(anti_reference, reference)), ValueError(
```
vpicheny, line 89: isn't this something we want to test all the time? (i.e. whether or not observations are valid or not)

trieste/acquisition/multi_objective/partition.py
```diff
@@ -45,35 +49,48 @@ def prepare_default_non_dominated_partition_bounds(
...
         tf.debugging.assert_greater_equal(
-            reference,
-            anti_reference,
+            ref,
+            anti_ref,
             message="reference point containing dimensionality below default "
             "anti-reference point ([-1e10, ..., -1e10]), try specify a lower "
             "anti-reference point.",
         )
+        if not not_valid_obs(obs):  # make sure given (valid) observations are larger than -1e10
```
vpicheny: `not_valid` seems a bit misleading. Change to `is_empty_obs`?

trieste/acquisition/multi_objective/partition.py
vpicheny: also change message to "anti_reference point contains at least one value larger than reference point" ?

vpicheny (commented):
This strategy seems like a good default when the reference dominates the observations

trieste/acquisition/multi_objective/partition.py
TsingQAQ: Was originally expecting calling `prepare_default_non_dominated_partition_bounds` is the same as calling a `any_partition().partition_bounds()`. The latter is expected to return just a tuple of _static_ `lb_points` and `ub_points`.    
So as long as a _static_ `lb_points` & `ub_points` is acceptable for acq, not return a `_BoundIndexPartition_instance`/`any_partition_instance` is ok?

trieste/acquisition/multi_objective/partition.py
TsingQAQ: > isn't this something we want to test all the time? (i.e. whether or not observations are valid or not)

Not sure if got the point (please correct if misinterpreted), thought we only need to test when user specified a reference point (as otherwise the current automatic reference selection would always result valid `observation`).

trieste/acquisition/multi_objective/partition.py
vpicheny: my bad, I didn't realise we were only using `partition_bounds()`
This is fine then.

trieste/acquisition/multi_objective/partition.py
vpicheny: My point was just that we could test `anti_reference << reference` all the time, not only if `not_valid_obs(observations)`.

Anyway this is not very important.

trieste/acquisition/multi_objective/partition.py
```diff
@@ -45,35 +49,48 @@ def prepare_default_non_dominated_partition_bounds(
...
+        return anti_ref
 
-    if tf.shape(observations)[-1] > 2:
+    tf.debugging.assert_shapes([(reference, ["D"])])
+    if (  # prepare default anti_reference point if not specified
+        anti_reference is None
+    ):  # if anti_reference point is not specified, use a -1e10 as default (act as -inf)
+        anti_reference = specify_default_anti_reference_point(reference, observations)
+    else:  # anti_reference point is specified
+        tf.debugging.assert_shapes([(anti_reference, ["D"])])
```
apaleyes: I find this formatting a bit hard to follow. How about this?

```suggestion
    if (anti_reference is None):
        # if anti_reference point is not specified, use a -1e10 as default (act as -inf)
        anti_reference = specify_default_anti_reference_point(reference, observations)
    else:
        # anti_reference point is specified
        tf.debugging.assert_shapes([(anti_reference, ["D"])])
```

trieste/acquisition/multi_objective/partition.py
```diff
@@ -45,35 +49,48 @@ def prepare_default_non_dominated_partition_bounds(
...
+
+    if is_empty_obs(observations):  # if no valid observations
+        assert tf.reduce_all(tf.less_equal(anti_reference, reference)), ValueError(
+            "anti_reference point contains at least one value larger than reference point"
```
apaleyes:
```suggestion
            f"anti_reference point {anti_reference} contains at least one value larger than reference point {reference}"
```

apaleyes (commented):
That looks good. Just one minor suggestion. There is quite a bit of validation going on (good thing!), but almost all error messages thrown contain no runtime information, e.g. variable values. Having those values in would tremendously help debugging, so if you could augment some error messages with data that caused them, it would be awesome. One idea is below.

trieste/acquisition/multi_objective/partition.py
TsingQAQ: Yes, we only provide bounds here. Not immediately aware of a case we need `partition_instance`, unless we need somehow dynamically change partition bounds on the fly during one iteration (if having this use case please indicate), otherwise maybe we can just provide _static_ `partition_bounds` as a "default" choice.

trieste/acquisition/multi_objective/partition.py
TsingQAQ: ah i see, was thinking as `prepare_default_non_dominated_partition_bounds` only add a "new" partition strategy for empty observations, hence we only do `anti_reference << reference` in this new partition. As in other cases, this function will invoke other partition method and its these partition methods responsibility to check if `anti_reference << reference` (i.e., [this test](https://github.com/secondmind-labs/trieste/blob/45f99cab9d0bdc5a174fdf5d742f3f52a20ae9ea/tests/unit/acquisition/multi_objective/test_partition.py#L118-L127) and [this test](https://github.com/secondmind-labs/trieste/blob/45f99cab9d0bdc5a174fdf5d742f3f52a20ae9ea/tests/unit/acquisition/multi_objective/test_partition.py#L218-L232))

trieste/acquisition/multi_objective/partition.py
TsingQAQ: Yep, thanks for the suggestions!

vpicheny (approved):

PR #357, opened 2021-09-28
GPflux interface with models

sebastianober:


tests/integration/models/gpflux/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+tf.keras.backend.set_floatx("float64")
+
+
+@pytest.fixture(name="depth", params=[2, 3])
+def _depth_fixture(request: Any) -> int:
+    return request.param
+
+
+@pytest.fixture(name="hartmann_6_dataset_function", scope="session")
```
hstojic, line 31: I think this can be moved one level up, I will use it in neural nets as well

tests/integration/models/gpflux/conftest.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2021 The Trieste Contributors
```
hstojic: lets move the fixtures one level up, to a conftest in a parent folder, now I'm thinking fixtures will be too scattered around, other developers will have hard time finding them

setting the dtype should stay here on the other hand, valid for gpflux only

tests/integration/test_bayesian_optimization.py
```diff
@@ -164,3 +172,67 @@ def build_model(data: Dataset) -> GaussianProcessRegression:
...
+    "num_steps, acquisition_rule",
+    [
+        (5, DiscreteThompsonSampling(1000, 50)),
+    ],
+)
+def test_two_layer_dgp_optimizer_finds_minima_of_michalewicz_function(
+    num_steps: int, acquisition_rule: AcquisitionRule[TensorType, SearchSpace]
+) -> None:
+
+    search_space = Box([0, 0], [pi, pi])
```
hstojic: there should be michalewicz_search_space variable in objectives, right?

tests/integration/test_bayesian_optimization.py
```diff
@@ -164,3 +172,67 @@ def build_model(data: Dataset) -> GaussianProcessRegression:
...
+
+        optimizer = tf.optimizers.Adam(0.01)
+        fit_args = {
+            "batch_size": batch_size,
+            "epochs": epochs,
+            "verbose": 0,
+            "callbacks": tf.keras.callbacks.LearningRateScheduler(scheduler),
+        }
+
+        config = GPfluxModelConfig(
```
hstojic: perhaps its more direct to simply use `DeepGaussianProcess` wrapper rather than the Config contraption?

tests/unit/models/gpflux/conftest.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2021 The Trieste Contributors
```
hstojic: same as for integration test, for usability lets move fixtures to a conftest in models directory

trieste/models/gpflux/__init__.py
```diff
@@ -16,7 +16,12 @@
 This package contains the primary interface for deep Gaussian process models. It also contains a
 number of :class:`TrainableProbabilisticModel` wrappers for GPflux-based models. Note that currently
 copying/saving models is not supported, so in a Bayes Opt loop `track_state` should be set False.
+<<<<<<< HEAD
```
hstojic: merge leftover

trieste/models/gpflux/architectures.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    # Input data to model must be np.ndarray for k-means algorithm
+    if isinstance(query_points, tf.Tensor):
+        query_points = query_points.numpy()
+
+    num_data = len(query_points)
+
+    # Pad query_points with additional random values to provide enough inducing points
+    if num_inducing > len(query_points):
```
hstojic, line 62: why does it make sense to have more inducing points than query points?

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+from ...data import Dataset
+from ...types import TensorType
+from ..interfaces import TrainableProbabilisticModel
+from .interface import GPfluxPredictor
+from .utils import sample_dgp
+
+
+class DeepGaussianProcess(GPfluxPredictor, TrainableProbabilisticModel):
+    """
+    A :class:`TrainableProbabilisticModel` wrapper for a GPflux :class:`~gpflux.models.DeepGP` with
```
hstojic, line 33: perhaps additional note about architectures that they are convenient for creating a DeepGP?

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+        new_num_data = inputs.shape[0]
+        self.model_gpflux.num_data = new_num_data
+
+        # Update num_data for each layer, as well as make sure dataset shapes are ok
+        for i, layer in enumerate(self.model_gpflux.f_layers):
+            layer.num_data = new_num_data
+
+            if isinstance(layer, LatentVariableLayer):
+                inputs = layer(inputs)
+                continue
```
hstojic, line 132: why do you need `continue` here?

hstojic (commented):
I went through the code already few times, its difficult to find improvements - some fresh eyes are needed :)

perhaps you could ping Vincent D to check out this one (perhaps focusing on sampling, update and architecture)?

tests/integration/models/gpflux/test_predictions.py
```diff
@@ -0,0 +1,120 @@
...
+) -> None:
+    dataset_size = 1000
+    num_inducing = 200
+    batch_size = 200
+    epochs = 500
+
+    example_data = hartmann_6_dataset_function(dataset_size)
+
+    # Trieste implementation
+    dgp = build_vanilla_deep_gp(
```
hstojic: same code snippet for building a model as above, right? perhaps wrap in an internal function and use it in both places?

tests/integration/models/gpflux/test_predictions.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # GPflux implementation
+    config = Config(
+        num_inducing=num_inducing, inner_layer_qsqrt_factor=1e-5, likelihood_noise_variance=1e-2
+    )
+
+    gpflux_model = build_constant_input_dim_deep_gp(example_data.query_points, depth, config)
+
+    keras_model = gpflux_model.as_training_model()
+    keras_model.compile(tf.optimizers.Adam(0.01))
```
hstojic: why not using the object `optimizer` defined above?

tests/integration/models/gpflux/test_predictions.py
```diff
@@ -0,0 +1,120 @@
...
+    keras_model.fit(
+        {"inputs": example_data.query_points, "targets": example_data.observations},
+        batch_size=batch_size,
```
hstojic: you can use `fit_args` defined above as well?

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+        optimizer: tf.optimizers.Optimizer | None = None,
+        fit_args: Dict[str, Any] | None = None,
+    ):
+        """
+        :param model: The underlying GPflux deep Gaussian process model.
+        :param optimizer: The optimizer with which to train the model. Defaults to
+            :class:`~trieste.models.optimizer.TFOptimizer` with :class:`~tf.optimizers.Adam`. Only
+            the optimizer itself is used; other args relevant for fitting should be passed as part
+            of `fit_args`.
+        :param fit_args: A dictionary of arguments to be used in the Keras `fit` method. Default to
```
apaleyes: This needs a link to Keras docs that lists possible args.

```suggestion
        :param fit_args: A dictionary of arguments to be used in the Keras `fit` method. Defaults to
```

tests/unit/models/gpflux/test_models.py
```diff
@@ -0,0 +1,246 @@
...
+    elbo = model.model_gpflux.elbo(data)
+    model.optimize(dataset)
+    assert model.model_gpflux.elbo(data) > elbo
+
+
+def test_dgp_loss(two_layer_model: Callable[[TensorType], DeepGP]) -> None:
+    x = tf.constant(np.arange(5).reshape(-1, 1), dtype=gpflow.default_float())
+    y = fnc_3x_plus_10(x)
+
+    reference_model = two_layer_model(x)
```
hstojic, line 163: not sure we should use `two_layer_model` which is built using Trieste's architecture function, if we want a GPflux reference model? not sure if relying on architectures code being properly tested is ok...

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+
+        super().__init__()
+
+        if optimizer is None:
+            self._optimizer = tf.optimizers.Adam()
+        else:
+            self._optimizer = optimizer
+
+        if not isinstance(self._optimizer, tf.keras.optimizers.Optimizer):
+            raise ValueError("Must use a Keras/TF optimizer for DGPs, not wrapped in TFOptimizer")
```
apaleyes: That's at odds with `tf.optimizers.Optimizer` type declared above. Also, how about this for an error message?

```suggestion
        if not isinstance(self._optimizer, tf.keras.optimizers.Optimizer):
            raise ValueError(f"Optimizer for DeepGPs must be from Keras/TF, received {type(optimizer)}")
```

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+
+        self.original_lr = self._optimizer.lr.numpy()
```
apaleyes, line 73: Remembering our conversation in the previous PR, just a sanity check: was this `numpy()` call intended?

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+                }
+            )
+        else:
+            self.fit_args = fit_args
+
+        if not all([isinstance(layer, (GPLayer, LatentVariableLayer)) for layer in model.f_layers]):
+            raise ValueError(
+                "`DeepGaussianProcess` can only be built out of `GPLayer` or"
+                "`LatentVariableLayer`"
+            )
```
apaleyes: Ideally that would specify what wrong layer was received

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+        return self._optimizer
+
+    def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
+        def get_samples(query_points: TensorType, num_samples: int) -> TensorType:
+            samples = []
+            for _ in range(num_samples):
+                samples.append(sample_dgp(self.model_gpflux)(query_points))
+            return tf.stack(samples)
+
+        return get_samples(query_points, num_samples)
```
apaleyes: Why define a function and immediately call it?

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+                inducing_variable = layer.inducing_variable.inducing_variable
+
+            if inputs.shape[-1] != inducing_variable.Z.shape[-1]:
+                raise ValueError(
+                    f"Shape {inputs.shape} of input to layer {layer} is incompatible with shape"
+                    f" {inducing_variable.Z.shape} of that layer. Trailing dimensions must match."
+                )
+
+            if i == len(self.model_gpflux.f_layers) - 1:
+                if dataset.observations.shape[-1] != layer.q_mu.shape[-1]:
```
apaleyes: Very minor: if ... if ... -> if ... and ...

trieste/models/gpflux/models.py
```diff
@@ -0,0 +1,158 @@
...
+    def optimize(self, dataset: Dataset) -> None:
+        """
+        Optimize the model with the specified `dataset`.
+        :param dataset: The data with which to optimize the `model`.
+        """
+        self.model_keras.fit(
+            {"inputs": dataset.query_points, "targets": dataset.observations}, **self.fit_args
+        )
+
+        # Reset lr in case there was an lr schedule
```
apaleyes: Can you elaborate a bit more in the comment? This seems to be talking about some obscure tf/keras mechanics

trieste/models/gpflux/architectures.py
```diff
@@ -0,0 +1,87 @@
...
+        if search_space is not None:
+            if not isinstance(search_space, Box):
+                raise ValueError("Currently only `Box` instances are supported for `search_space`.")
```
apaleyes:
```suggestion
                raise ValueError(f"Currently only `Box` instances are supported for `search_space`, got {type(search_space)}.")
```

trieste/models/gpflux/architectures.py
```diff
@@ -0,0 +1,87 @@
...
+        num_inducing=num_inducing,
+        inner_layer_qsqrt_factor=inner_layer_sqrt_factor,
+        likelihood_noise_variance=likelihood_noise_variance,
+        whiten=True,  # whiten = False not supported yet in GPflux for this model
+    )
+
+    model = build_constant_input_dim_deep_gp(query_points, num_layers, config)
+
+    # If num_inducing is larger than the number of provided query points, the initialization for
+    # num_data will be wrong.
```
apaleyes: Is this comment supposed to warn, and if so, about what? `num_inducing > len(query_points)` seems to be handled already

trieste/models/gpflux/architectures.py
```diff
@@ -0,0 +1,87 @@
...
+import numpy as np
+import tensorflow as tf
+from gpflux.architectures import Config, build_constant_input_dim_deep_gp
+from gpflux.models import DeepGP
+
+from ...space import Box
+from ...types import TensorType
+
+
+def build_vanilla_deep_gp(
```
apaleyes, line 34: Would be great to understand why this function has to live in Trieste, and not in GPflux. It doesn't take in or return a Trieste-typed object (except for optional space)

tests/unit/models/gpflux/test_models.py
```diff
@@ -0,0 +1,246 @@
...
+# limitations under the License.
+
+"""
+In this module, we test the *behaviour* of trieste models against reference GPflux models (thus
+implicitly assuming the latter are correct).
+*NOTE:* Where GPflux models are used as the underlying model in an trieste model, we should
+*not* test that the underlying model is used in any particular way. To do so would break
+encapsulation. For example, we should *not* test that methods on the GPflux models are called
+(except in the rare case that such behaviour is an explicitly documented behaviour of the
+trieste model).
```
apaleyes: Would be nice to show some respect to **T**rieste

tests/unit/models/gpflux/test_models.py
```diff
@@ -0,0 +1,246 @@
...
+    dataset = Dataset(*data)
+
+    optimizer = tf.optimizers.Adam()
+
+    fit_args = {"batch_size": batch_size, "epochs": 10, "verbose": 0}
+
+    model = DeepGaussianProcess(two_layer_model(x_observed), optimizer, fit_args)
+    elbo = model.model_gpflux.elbo(data)
+    model.optimize(dataset)
+    assert model.model_gpflux.elbo(data) > elbo
```
apaleyes, line 156: Tests like that make me increasingly uneasy about exposing GPflux in Trieste wrapper. As an optimization toolkit, Trieste shouldn't be verifying internal particularities of GPflux model

tests/integration/models/gpflux/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+from typing import Any, Callable
+
+import pytest
+import tensorflow as tf
+
+from tests.util.trieste.utils.objectives import hartmann_6_dataset
+from trieste.data import Dataset
+
+tf.keras.backend.set_floatx("float64")
```
apaleyes: that shall probably be done in test files, not in fixtures

apaleyes (commented):
Getting there! Very nice to see another good-sized PR

trieste/models/gpflux/architectures.py
sebastianober: I think there are two reasons: 1) In Bayes opt you might start out with relatively few query points, but eventually end up with more query points than inducing points, 2) in DGPs it's possible that having more inducing points than query points could still improve the inference since there is no result that I'm aware of that states the approximate posterior becomes exact (or even optimal in some sense) when `num_inducing == num_data`

trieste/models/gpflux/models.py
sebastianober: `LatentVariableLayer`s don't have inducing points, so we just propagate the inputs and skip.

tests/integration/models/gpflux/conftest.py
sebastianober: I have moved them. For the `floatx` stuff I have discussed with Andrei and we have decided the best course of action is to have a fixture for setting the `floatx` and then resetting it to the original value, so that this can be done on a test-by-test basis. This uses the pytest yield fixture [https://docs.pytest.org/en/latest/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization](url)

tests/integration/test_bayesian_optimization.py
sebastianober: Yep, done

tests/integration/test_bayesian_optimization.py
sebastianober: Done

trieste/models/gpflux/models.py
sebastianober: Done

tests/integration/models/gpflux/test_predictions.py
sebastianober: I'm a bit wary of doing this, I think it might then share the optimizers between the two models, and since the callback is meant to modify the optimizer I want to make sure that the implementations are independent of each other.

tests/integration/models/gpflux/conftest.py
sebastianober: Should be addressed now.

tests/integration/models/gpflux/test_predictions.py
sebastianober: I've had a go at this.

tests/unit/models/gpflux/test_models.py
sebastianober: Thanks, fixed.

trieste/models/gpflux/models.py
sebastianober: `tf.keras.optimizers.Optimizer` is equivalent to `tf.optimizers.Optimizer`, so this shouldn't really matter. I've changed it however, as well as the error message.

trieste/models/gpflux/models.py
sebastianober: Yes in this case it's what I want.

trieste/models/gpflux/architectures.py
sebastianober: It's not meant to warn, it's just to explain why I went through and changed `num_data` manually. If `num_inducing > len(query_points)`, `num_data` won't be equal to the number of query points originally received, since the number of query points has been artificially expanded. This is basically a quirk of how `build_constant_input_dim_deep_gp` from GPflux works.

trieste/models/gpflux/architectures.py
sebastianober: We wanted to provide the user with some ready-made architectures they can straightforwardly plug in. Deep GPs aren't the easiest models to create, even with a package like GPflux, so we provided this. This folder will probably be more useful if/when we expand it with more complicated DGP models.

trieste/models/gpflux/models.py
sebastianober: Done

tests/unit/models/gpflux/test_models.py
sebastianober: This is more testing that the `optimize` method (which is a Trieste method) works (to an extent), i.e. that the elbo of the model actually increases as we train. We have similar tests I believe in the GPflow folder. I'm not sure how to otherwise test that the underlying model is actually being changed when we call `optimize`

apaleyes: ok, i guess question is a bit out of scope of this PR anyway

trieste/models/gpflux/architectures.py
apaleyes: ah, i see. thanks for explanation, that's fine then

tests/integration/models/gpflux/conftest.py
hstojic: ah, thats a neat solution!

trieste/models/gpflux/architectures.py
hstojic: ok, thanks for the clarification

tests/unit/models/gpflux/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
hstojic: this file should be deleted now, right?

tests/integration/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+import tensorflow as tf
+
+
+# Teardown fixture to set keras floatx to float64 then return it to previous value at test finish
```
apaleyes:
```suggestion
# Setup and teardown fixture to set keras floatx to float64 before test starts
# then return it to previous value at test finish
```

tests/integration/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+# pytest uses yield in a funny way, so we use type ignore
+@pytest.fixture(name="keras_float")  # type: ignore
+def _keras_float() -> None:
+    curr_float = tf.keras.backend.floatx()
```
apaleyes: There is really no point in such shortenings

```suggestion
    current_float = tf.keras.backend.floatx()
```

tests/unit/models/gpflux/test_architectures.py
```diff
@@ -0,0 +1,109 @@
...
+    inner_layer_sqrt_factor = 1e-5
+    likelihood_noise_variance = 1e-2
+
+    vanilla_deep_gp = build_vanilla_deep_gp(
+        x,
+        num_layers=num_layers,
+        num_inducing=num_inducing,
+        inner_layer_sqrt_factor=inner_layer_sqrt_factor,
+        likelihood_noise_variance=likelihood_noise_variance,
+    )
```
apaleyes: Since all variable names are exactly the same as the parameter names, there is probably no point in duplicating them

tests/unit/models/gpflux/test_models.py
```diff
@@ -0,0 +1,252 @@
...
+
+    ref_samples = get_samples(test_x, num_samples)
+
+    ref_mean = tf.reduce_mean(ref_samples, axis=0)
+    ref_variance = tf.reduce_mean((ref_samples - ref_mean) ** 2)
+
+    error = 1 / tf.sqrt(tf.cast(num_samples, tf.float32))
+    npt.assert_allclose(sample_mean, ref_mean, atol=2 * error)
+    npt.assert_allclose(sample_mean, 0, atol=error)
+    npt.assert_allclose(sample_variance, ref_variance, atol=4 * error)
```
apaleyes, line 217: Aren't these all equivalent?

apaleyes (approved):
I don't have any major comments here, just random nitpicking. Big props for floatx fixtures!

tests/unit/models/gpflux/conftest.py
sebastianober: Yes, thanks!

tests/unit/models/gpflux/test_models.py
sebastianober: I don't think so?

apaleyes: i was confused, you are right

vdutor (approved):
LGTM - the float fixtures are really neat.

hstojic (approved):
Looks good, ready for merging

PR #359, opened 2021-09-29
Bump version to 0.7.0

uri-granta:


henrymoss (approved):

hstojic (approved):

hstojic:
what do you think about adding a RELEASE.md file with a main record of release notes? one doesn't need to go GH to find when was something changed etc

PR #360, opened 2021-09-30
Update BatchMonteCarloExpectedHypervolumeImprovement to avoid retracing

uri-granta:


trieste/acquisition/function.py
```diff
@@ -1215,7 +1247,8 @@ class batch_monte_carlo_expected_improvement(AcquisitionFunctionClass):
     def __init__(self, sample_size: int, model: ProbabilisticModel, eta: TensorType, jitter: float):
         """
 
-        :param sampler:  BatchReparametrizationSampler.
+        :param sample_size:  The number of samples for each batch of points.
+        :param model: The model over the specified ``dataset``.
```
apaleyes, line 1251: There is no dataset specified here, right?

TsingQAQ:
Many thanks for these! Maybe can i mention `BatchMonteCarloExpectedHypervolumeImprovement` may haven't reach its fully functionality (i.e., allow greedy batch, which is claimed having better performance in case of big batch sizes, as mentioned [here](https://github.com/secondmind-labs/trieste/pull/257#issuecomment-865086793)). I was thinking to update this class a bit (say refactoring to inherit from `GreedyAcquisitionFunctionBuilder` and do some changes). As it seems is in-dev in this PR to avoid retracing atm, would you in favor of doing this functionality update first (and you may pitch in to suggest preventing those that may trigger tracing stuffs), or we can do this functionality update later?

trieste/acquisition/function.py
```diff
@@ -1027,43 +1027,77 @@ def prepare_acquisition_function(
...
+        )
+        self._ub_points = tf.Variable(
+            partition_bounds[1], trainable=False, shape=[None, partition_bounds[1].shape[-1]]
+        )
+
+    def update(self, partition_bounds: tuple[TensorType, TensorType]) -> None:
+        """Update the acquisition function with new partition bounds."""
+        self._lb_points.assign(partition_bounds[0])
+        self._ub_points.assign(partition_bounds[1])
+        self._sampler._initialized.assign(False)
```
apaleyes, line 1085: This might be problematic. Not only this is not a part of the sampler interface, in `BatchReparametrizationSampler` this variable itself described as a workaround. Is it possible to create a new sampler object here? Or at least expose this initialization business with some sort of public property?

apaleyes (commented):
Looks good, just one slight concern below

trieste/acquisition/function.py
uri-granta: We want to avoid creating a new sampler, but making this public (e.g. by exposing a reinitiaize method) definitely makes sense.

uri-granta:
Feel free to make the changes first, I'm happy to wait. 

Also, at the moment the `test_batch_monte_carlo_expected_hypervolume_improvement_can_reproduce_ehvi` test is failing for the variance_scale=2 case but passing everywhere else. Not sure why this is: it may be a real issue with something not being updated due to the refactoring, or it may be just the test being fragile and failing due to a different random number. I'll take a look.

TsingQAQ:
> Feel free to make the changes first, I'm happy to wait.
> 
> Also, at the moment the `test_batch_monte_carlo_expected_hypervolume_improvement_can_reproduce_ehvi` test is failing for the variance_scale=2 case but passing everywhere else. Not sure why this is: it may be a real issue with something not being updated due to the refactoring, or it may be just the test being fragile and failing due to a different random number. I'll take a look.

Perfect, will prepare for this recently, and thanks for the test info and effort, will inspect as well from my side.

TsingQAQ:
> Also, at the moment the `test_batch_monte_carlo_expected_hypervolume_improvement_can_reproduce_ehvi` test is failing for the variance_scale=2 case but passing everywhere else. Not sure why this is: it may be a real issue with something not being updated due to the refactoring, or it may be just the test being fragile and failing due to a different random number. I'll take a look.  

i am more thinking is from the fragile of the test itself,  as it seems a bit hard to do MC in higher obj dim (maybe larger variance are there) to get converge to the analytical ehvi.  It seems a larger MC sample size (e.g., `70_000` instead of `50_000`) can get it pass for variance_scale=2.

PR #361, opened 2021-09-30
Increase iteration count because of reports of local failures 

apaleyes:
MES integ test was reported to fail on a number of local machines (@sebastianober , @uri-granta ) while mysteriously passing GitHub Actions build. This commit shall unblock local development, while we shall also create an issue to investigate this further. Note: on my machine it converges after 22 iterations, I've also heard about 21 iterations elsewhere, so the count is increased a bit excessively to allow for local fluctuations.

Also added local .vscode folder to .gitignore.

### DIFF
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -68,7 +68,7 @@ from trieste.types import State, TensorType
             (20, EfficientGlobalOptimization()),
             (25, EfficientGlobalOptimization(AugmentedExpectedImprovement().using(OBJECTIVE))),
             (
-                25,
+                22,
                 EfficientGlobalOptimization(
                     MinValueEntropySearch(BRANIN_SEARCH_SPACE, num_fourier_features=1000).using(
                         OBJECTIVE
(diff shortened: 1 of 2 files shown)


tests/integration/test_bayesian_optimization.py
```diff
@@ -68,7 +68,7 @@
             (20, EfficientGlobalOptimization()),
             (25, EfficientGlobalOptimization(AugmentedExpectedImprovement().using(OBJECTIVE))),
             (
-                15,
+                25,
```
uri-granta: I think it also works with this:

```suggestion
                21,
```

uri-granta (approved):
I took a quick look, and it looks like this is a random number generator issue revealed by the GIBBON refactoring (https://github.com/secondmind-labs/trieste/pull/349) which changed the order of some initialisation. Not sure why the RNG is behaving differently on GH, but since the integration tests should ideally pass regardless of the RNG seed I think this fix is probably best.

tests/integration/test_bayesian_optimization.py
apaleyes: It only works with 22 on my machine.

apaleyes: updated to 22 as the min known to fix all local failures

PR #362, opened 2021-10-01
fixing reference point etc.

vpicheny:
In CEHVI, the reference point is now based on the feasible set of points rather than the optimal ones only.

In the plotting code, a bug is fixed so that the Pareto front excludes infeasible points according to the constraint.

### DIFF
--- a/trieste/acquisition/function.py
+++ b/trieste/acquisition/function.py
@@ -1060,7 +1060,7 @@ class ExpectedConstrainedHypervolumeImprovement(ExpectedConstrainedImprovement):
         feasible_mean, _ = objective_model.predict(feasible_query_points)
 
         _pf = Pareto(feasible_mean)
-        _reference_pt = get_reference_point(_pf.front)
+        _reference_pt = get_reference_point(feasible_mean)
         ehvi = expected_hv_improvement(objective_model, _pf, _reference_pt)
         return lambda at: ehvi(at) * constraint_fn(at)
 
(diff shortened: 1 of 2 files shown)


trieste/bayesian_optimizer.py
```diff
@@ -352,6 +352,7 @@ def optimize(
         history: list[Record[S]] = []
 
         for step in range(num_steps):
+            print(step)
```
hstojic, line 355: monitoring? :)

trieste/acquisition/function.py
```diff
@@ -1060,7 +1060,7 @@ def prepare_acquisition_function(
         feasible_mean, _ = objective_model.predict(feasible_query_points)
 
         _pf = Pareto(feasible_mean)
-        _reference_pt = get_reference_point(_pf.front)
+        _reference_pt = get_reference_point(feasible_mean)
```
TsingQAQ, line 1063: Just for learning curious: as `get_reference_point` is currently only intend to extract the reference point from a 'front' (as conveyed from its input arg's name, though we never check if input is really a 'front'), is 1063 more like a tmp fix, or some ref doc could be mentioned here as well to align with this specific adaptive choice of ref pt in CMOO? And if there is only 1 feasible point, we'd assume this behaves the same?

vpicheny: This is a temporary fix. It's possibly a very bad choice in general. On my current problem it works well because the constraints are quite restrictive. We may want to discuss what we want to do with reference point management in general. It feels that we can't stick to a single solution.

hstojic:
@vpicheny what do you want to do with this PR?

PR #363, opened 2021-10-01
Actually using pending points in prepare

apaleyes:
This PR is an early prep for async BO.

At the moment pending points are assumed in "prepare", but are not used and in fact validated to be None. Since in async scenarios we would like to pass pending points from outside, this PR ensures pending points are propagated both from "prepare" and "update" functions.

### DIFF
--- a/tests/unit/acquisition/test_function.py
+++ b/tests/unit/acquisition/test_function.py
@@ -91,7 +91,7 @@ class _ArbitrarySingleBuilder(SingleModelAcquisitionBuilder):
 
 
 class _ArbitraryGreedySingleBuilder(SingleModelGreedyAcquisitionBuilder):
-    def prepare_acquisition_function(
+    def update_acquisition_function(
         self, dataset: Dataset, model: ProbabilisticModel, pending_points: TensorType = None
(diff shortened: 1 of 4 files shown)


uri-granta:
I'm confused by what this is trying to do. At the moment, prepare is called just once to prepare the function, and from then on only update is called, both to add pending points and to change the dataset every loop. What prevents you from calling update when passing in new pending points from outside?

apaleyes:
Thanks for having a look! That's a reasonable question, I should have clarified it initially.

The problem with current update is that it needs to also "init" the batch, for batch acquisition functions that is. And it uses pending points as an indicator: if we are passed any, that means we are in the middle of the batch, if not, we are getting started, update the batch. Here is the line from LP


        if pending_points is None:
            # update penalization params and base acquisition once per optimization step
            return self._update_base_acquisition_function(dataset, model)

Now we want to be able to pass pending points from outside, and this indication ability is gone - we don't know if we are in the middle of the batch or not. But we still need some way to "init" either the whole function or the batch. `prepare` seems like a natural way of doing it.

Does that make sense?

apaleyes:
closed in favour of #366

PR #366, opened 2021-10-05
Async bo rule and notebook

apaleyes:
This PR replaces and expands #363.

Here we:
* Make changes in greedy acquisition function related to batch initialization and handling of pending points
* Introduce new AsyncEGO rule.
* Provide a notebook that illustrates how one could do async BO with Trieste

Unit tests, docs and other trimmings to come at a later point, I am mostly looking for feedback at this point. One note about new rule is that, unlike TrustRegion, this one works not with other rules but with function builders directly. Happy to hear suggestions of how to change that.

**Update:** all the trimmings are now done. The notebook is decorated with many-many comments to make it as accessible as possible, very happy to improve it further if people still find it unclear what's all this multiprocessing stuff is about.


trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+                    new_optimization_step=False,
                 )
                 chosen_point = self._optimizer(search_space, self._acquisition_function)
                 points = tf.concat([points, chosen_point], axis=0)
 
         return points
 
 
+class AsyncEGO(AcquisitionRule[State[Optional["AsyncEGO.State"], TensorType], SP_contra]):
+    """AsyncEGO rule, as name suggests, is designed for async BO scenarios.
```
```suggestion
    """AsyncEGO rule, as name suggests, is designed for asynchronous BO scenarios.
```

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+    By async BO we understand a use case when multiple objective function can be launched in parallel
+    and are expected to arrive at different times. Instead of waiting for the rest of observations to return,
+    we want to immediately use acquisition function to launch a new observation and avoid wasting computational resources.
+
+    To make the best decision about next point to observe, acquisition function needs to be aware of currently running observations.
+    We call such points "pending". AsyncEGO provides a ``AsyncEGO.State`` object that keeps track of pending points
```
uri-granta: I think we should differentiate between "pending" points used by greedy acquisition functions (points that have been selected but haven't been executed yet) and "outstanding" points, which have been executed but haven't returned.

```suggestion
    We call such points "outstanding". AsyncEGO provides a ``AsyncEGO.State`` object that keeps track of outstanding points
```

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+
+    def __init__(
+        self,
+        builder: GreedyAcquisitionFunctionBuilder | SingleModelGreedyAcquisitionBuilder,
+        optimizer: AcquisitionOptimizer[SP_contra] | None = None,
+        num_query_points: int = 1,
+    ):
+        """
+        :param builder: Acquisition function builder. Only greedy batch approaches are supported,
+            because they are the only ones currently supporting the pending points concept.
```
```suggestion
            because they can be told what points are outstanding.
```

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+            :func:`~trieste.acquisition.optimizer.automatic_optimizer_selector`.
+        :param num_query_points: The number of points to acquire.
+        """
+        if num_query_points <= 0:
+            raise ValueError(
+                f"Number of query points must be greater than 0, got {num_query_points}"
+            )
+
+        if builder is None:
+            raise ValueError("Please specify an acquisition builder")
```
uri-granta, line 249: Might be nice to default to `LocalPenalizationAcquisitionFunction(search_space)` but this requires knowing the global search space. @henrymoss: opinion?

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+                return tf.reduce_all(tf.reduce_any(tf.not_equal(query_points, x), axis=1))
+
+            if state is None:
+                pending_points = None
+            else:
+                pending_points = state.pending_points
+                mask = tf.map_fn(is_not_in_dataset, pending_points, fn_output_signature=tf.bool)
+                pending_points = tf.boolean_mask(pending_points, mask)
+
+            acquisition_function = self._builder.prepare_acquisition_function(
```
uri-granta: Like EGO, we should store the acquisition function in `self._acquisition_function` and update it if it already exists to avoid retracing unnecessarily.

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+        """
+        Constructs a function that, given ``AsyncEGO.State``,
+        returns a new state object and points to evaluate.
+        The state object contains currently known pending points,
+        that is points that were requested for evaluation,
+        but observation for which was not received yet.
+        To keep them up to date, pending points are compared against the given dataset,
+        and whatever points are in the dataset are deleted.
+        Then the current batch is generated by calling the acquisition function,
+        and all points in the batch are added to the known pending points.
```
uri-granta, line 290: Nice clear docstring, thanks!

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+
+        :param search_space: The local acquisition search space for *this step*.
+        :param datasets: The known observer query points and observations.
+        :param models: The models of the specified ``datasets``.
+        :return: A function that constructs the next acquisition state and the recommended query
+            points from the previous acquisition state.
+        """
+
+        def state_func(state: AsyncEGO.State | None) -> tuple[AsyncEGO.State | None, TensorType]:
```
uri-granta: We normally call the state functions `go` for some reason (e.g. see TrustRegion).

```suggestion
        def go(state: AsyncEGO.State | None) -> tuple[AsyncEGO.State | None, TensorType]:
```

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+            @tf.function
+            def is_not_in_dataset(x):
```
```suggestion
            def is_not_in_dataset(x : TensorType) -> TensorType:
```

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+    """AsyncEGO rule, as name suggests, is designed for async BO scenarios.
+    By async BO we understand a use case when multiple objective function can be launched in parallel
+    and are expected to arrive at different times. Instead of waiting for the rest of observations to return,
+    we want to immediately use acquisition function to launch a new observation and avoid wasting computational resources.
+
+    To make the best decision about next point to observe, acquisition function needs to be aware of currently running observations.
+    We call such points "pending". AsyncEGO provides a ``AsyncEGO.State`` object that keeps track of pending points
+    """
+    @dataclass(frozen=True)
+    class State:
```
uri-granta, line 231: What's the reasoning behind using a frozen dataclass for the state, rather than just the pending_points/outstanding_points?

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,144 @@ def acquire(
...
+                pending_points = tf.boolean_mask(pending_points, mask)
+
+            acquisition_function = self._builder.prepare_acquisition_function(
+                datasets, models, pending_points
+            )
+
+            points = self._optimizer(search_space, acquisition_function)
+            pending_points = points
+
+            for _ in range(self._num_query_points - 1):  # greedily allocate batch elements
```
```suggestion
            for _ in range(self._num_query_points - 1):  # greedily allocate additional batch elements
```

trieste/acquisition/function.py
```diff
@@ -1289,6 +1290,9 @@ def update_acquisition_function(
         :param models: The models over each dataset in ``datasets``.
         :param pending_points: Points already chosen to be in the current batch (of shape [M,D]),
             where M is the number of pending points and D is the search space dimension.
+        :param new_optimization_step: Indicates if this call to update_acquisition_function
```
```suggestion
        :param new_optimization_step: Indicates whether this call to update_acquisition_function
```

trieste/acquisition/function.py
```diff
@@ -1289,6 +1290,9 @@ def update_acquisition_function(
...
+            is to start of a new optimization step, of to continue collecting batch of points
```
uri-granta, line 1294:
```suggestion
            is to start of a new optimization step, or to continue collecting batch of points
```

trieste/acquisition/function.py
```diff
@@ -1443,37 +1461,54 @@ def prepare_acquisition_function(
...
-            or ``dataset`` is empty.
+        :raise tf.errors.InvalidArgumentError: If the ``dataset`` is empty.
         """
         tf.debugging.assert_positive(len(dataset))
-        tf.debugging.Assert(pending_points is None, [])
 
-        # no penalization required if no pending_points.
-        return self._update_base_acquisition_function(dataset, model)
+        acq = self._update_base_acquisition_function(dataset, model)
+        if pending_points is not None:
```
uri-granta: Not sure whether this change is actually necessary, as the very first time pending_points should always be None right? Still, if we do make the change then we should probably also handle passing in empty pending_points.

```suggestion
        if pending_points is not None or len(pending_points) == 0:
```

uri-granta (commented):
Overall approach looks good! Various comments. Also you'll need to add unit and integration tests for AsyncEGO (and maybe a couple of unit tests for the function changes).

trieste/acquisition/rule.py
apaleyes: Can you elaborate a bit what's the difference between pending and outstanding? I don't think I understand.

Once the point was chosen by the acquisition function, there is no difference any more, as I understand it. It could have been chosen 5 calls ago or just now, it is still treated exactly the same - as a point which was selected for observations, but result for which hasn't arrived yet.

trieste/acquisition/rule.py
apaleyes: I am a bit vary of closures here updating the `self._acquisition_function` inside the state callable. But will play around and see if that works

trieste/acquisition/rule.py
apaleyes: Since there is only one such instance so far, I dare challenging it! `go` is rather cryptic, I find `state_func` clearer. If anything, probably we shall update TrustRegion?

trieste/acquisition/rule.py
apaleyes: No good reason. I anticipated more stuff in the state, but ended up with only one field.

trieste/acquisition/function.py
apaleyes: I guess that's an attempt at being general. Consider some complex ask/tell scenarios where optimizer instance is created anew for each ask. I agree that's rather hypothetical, but it seems easy to support, hence this change. Will cover the empty tensor case, thanks.

docs/notebooks/async_bo_demo.py
```diff
@@ -0,0 +1,326 @@
...
+sync_bo = AskTellOptimizer(search_space, initial_data, model, local_penalization_rule)
+
+
+# thread-safe queues
+m = Manager()
+points_q = m.Queue()
+observations_q = m.Queue()
+observer_processes = []
+pid = os.getpid()
+
```
henrymoss: I dont really understand any of the the multi processing stuff. This will need to be better documented in the final version!

docs/notebooks/async_bo_demo.py
```diff
@@ -0,0 +1,326 @@
...
+        observer_p.start()
+
+        observer_processes.append(observer_p)
+
+    # BO loop starts here
+    for i in range(num_iterations):
+        print(f"Process {pid}: Main     : iteration {i} starts", flush=True)
+
+        # get a batch of points, send them to queue
+        # each worker picks up a point and processes it
```
henrymoss, line 193: does this give a batch or a single point?

docs/notebooks/async_bo_demo.py
```diff
@@ -0,0 +1,326 @@
...
+            continue
+        
+        new_data = Dataset(query_points=tf.constant(new_data[0], dtype=tf.float64),
+                           observations=tf.constant(new_data[1], dtype=tf.float64))
+
+        bo.tell(new_data)
+        point = bo.ask()
+        print(f"Process {pid}: Main     : acquired point {point}", flush=True)
+        pq.put(np.atleast_2d(point))
+        points_observed += 1
```
henrymoss: This looks like the right logic :)

trieste/acquisition/function.py
henrymoss: I think this is a good change. Certainly better than the implicit behaviour we had before

trieste/acquisition/rule.py
henrymoss: Give it a proper name! AsynchronousEfficientGlobalOptimization

trieste/acquisition/rule.py
henrymoss: Its nice to have a default for sure, but probably too difficult in this case (without passing search space). Once we have a greedy Thompson sampling method, perhaps that could be the default?

trieste/acquisition/rule.py
henrymoss: If we make this change, lets make it consistent across states

henrymoss (commented):
LGTM

trieste/acquisition/rule.py
apaleyes: Actually thinking about it a bit more we shall probably stick to a class for a state, even if it has just one property. It's a bit more extensible and looks more reasonable in types. Or a namedtuple.

trieste/acquisition/rule.py
apaleyes: It had full name initially, but I've cut it down because it just felt too long and hard to read. Do you think I shall unfold it back?

apaleyes: eventually settled on AsyncEfficientGlobalOptimization

trieste/acquisition/rule.py
apaleyes: fixed, except for outstanding/pending

trieste/acquisition/rule.py
apaleyes: no default for now

trieste/acquisition/rule.py
apaleyes: tried it, and it seems to work fine

trieste/acquisition/rule.py
apaleyes: cheers!

trieste/acquisition/rule.py
apaleyes: i made the change in TrustRegion too

trieste/acquisition/rule.py
apaleyes: fixed

trieste/acquisition/rule.py
apaleyes: fixed

trieste/acquisition/function.py
apaleyes: "explicit is better than implicit" The Zen of Python

trieste/acquisition/function.py
apaleyes: fixed

docs/notebooks/async_bo_demo.py
apaleyes: did my best to explain what's going on, let me know if you feel it needs more of anything

trieste/acquisition/rule.py
```diff
@@ -200,14 +202,165 @@ def acquire(
...
+            builder = builder.using(OBJECTIVE)
+
+        self._builder: GreedyAcquisitionFunctionBuilder = builder
+        self._optimizer = optimizer
+        self._num_query_points = num_query_points
+        self._acquisition_function: Optional[AcquisitionFunction] = None
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""AsyncEGO(
```
hstojic: perhaps AsyncEfficientGlobalOptimization rather than AsyncEGO?

trieste/acquisition/function.py
hstojic: I don't see its fixed?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+#       jupytext_version: 1.12.0
+#   kernelspec:
+#     display_name: 'Python 3.7.5 64-bit (''.venv'': venv)'
+#     name: python3
+# ---
+
+# %% [markdown]
+# # Asynchronous Bayesian optimization with Trieste
+#
+# In this notebook we show how Bayesian optimization can be done asynchronuosly. That is pertinent in scenarios when the objective function we are aiming to optimize can be run for several points in parallel, and observations might return back at different times. To avoid wasting resources we immediatelly request the next point asynchronuosly, taking into account points that still being evaluated.
```
hstojic: that still being evaluated -> that are still being evaluated

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+tf.get_logger().setLevel("ERROR")
+import numpy as np
+import math
+
+import time
+import timeit
+
+
+# %% [markdown]
+# First, let's define a simple objective that will emulate workload of variable time. We will be using class BO benchmark function [Branin](https://www.sfu.ca/~ssurjano/branin.html), and insert sleep call in the middle of the calculation to emulate delay. Our sleep delay is a scaled sum of all input values to make sure delays are uneven.
```
hstojic: class BO -> classic Bayesian optimisation?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
hstojic: if you want to add an example here: when optimising hyperparameters of ML models its realistic that for some configurations training time will be much longer

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+# %%
+def objective(points, sleep=True):
+    if points.shape[1] != 2:
+        raise ValueError(f"Incorrect input shape, expected (*, 2), got {points.shape}")
+
+    def scaled_branin(x):
```
hstojic: it should be possible to wrap the existing function that we have in `objectives` here rather than repeating the code, no?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+from trieste.models.gpflow.config import GPflowModelConfig
+
+
+def build_model(data):
+    variance = tf.math.reduce_variance(data.observations)
+    kernel = gpflow.kernels.RBF(variance=variance)
+    gpr = gpflow.models.GPR(data.astuple(), kernel, noise_variance=1e-5)
+    gpflow.set_trainable(gpr.likelihood, False)
+
+    return GPflowModelConfig(
```
hstojic: perhaps we could simplify here and use `GaussianPreocessRegression` wrapper instead?

trieste/acquisition/rule.py
uri-granta: It's like the difference between things in your Amazon basket, and things that you've ordered but haven't arrived yet.

* Greedy acquistiion functions choose points one at a time. The first points are pending until we've selected the entire batch.
* Asyn BO choose some points (greedily) and then ask for them to be evaluated (like normal BO), but then starts another loop before all of them have finished executing. The ones that haven't finished yet are outstanding.

trieste/acquisition/rule.py
```diff
@@ -45,6 +45,9 @@
 T_co = TypeVar("T_co", covariant=True)
 """ Unbound covariant type variable. """
 
+S = TypeVar("S")
```
uri-granta: I'm probably being blind, but where is this used?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+
+# %% [markdown]
+# We now define several common functions to implement the described setup. First we define a worker function that will be running a single observation in a separate process. Worker takes both queues as an input, reads next point from the points queue, makes an observation, and inserts observed data into the observations queue.
+
+# %%
+
+def observer_proc(points_queue, observations_queue):
+    pid = os.getpid()
+
+    while True:
```
hstojic, line 122: I'm not sure I understand why is loop necessary here - this function is used for a single observation no? I see why it would be needed if you are running it until there are no more points, but then description above says single observation, so I'm not sure...

trieste/acquisition/rule.py
uri-granta: Yes, this should match the class name.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+        prc.terminate()
+        prc.join()
+        prc.close()
+
+# %% [markdown]
+# Finally we set some common parameters. See comments below for explanation of what each one means.
+# %%
+# Number of worker processes to run simultaneously
+# Setting this to 1 will turn both setups into non-batch sequential optimization
+num_workers = 3
```
hstojic, line 162: perhaps set to 2, enough to illustrate the point and will reduce computation time?

uri-granta (approved):
Looks good (though my preference would still be to distinguish in name between the greedy pending points and async pending points)

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+# ## Asynchronous optimization
+# This section runs the asynchronous optimization routine. We first setup the [ask/tell optimizer](ask_tell_optimization.ipynb). Notice that we use **AsyncEfficientGlobalOptimization** rule specifically designed for asynchronous scenarios. Next we create thread-safe queues for points and observations, and run the optimization loop.
+#
+# Crucially, even though we are using batch acquisition function Local Penalization, we specify batch size of 1. This is because we don't really want a batch. Since the amout of workers we have is fixed, whenever we see a new observation we only need one point back. However this process can only be done with acquisition funcitons that implement greedy batch collection strategies, because they are able to take into account points that are currently being observed (in Trieste we call them "pending"). Trieste currently provides two such functions: Local Penalization and GIBBON.
+
+# %%
+
+# setup Ask Tell BO
+model_spec = build_model(initial_data)
+model = create_model(model_spec)
```
hstojic: unnecessary if you use the model wrapper directly as suggested above

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+num_observations = num_workers * num_iterations
+# Set this flag to False to disable sleep delays in case you want the notebook to execute quickly
+enable_sleep_delays = True
+
+# %% [markdown]
+# ## Asynchronous optimization
+# This section runs the asynchronous optimization routine. We first setup the [ask/tell optimizer](ask_tell_optimization.ipynb). Notice that we use **AsyncEfficientGlobalOptimization** rule specifically designed for asynchronous scenarios. Next we create thread-safe queues for points and observations, and run the optimization loop.
+#
+# Crucially, even though we are using batch acquisition function Local Penalization, we specify batch size of 1. This is because we don't really want a batch. Since the amout of workers we have is fixed, whenever we see a new observation we only need one point back. However this process can only be done with acquisition funcitons that implement greedy batch collection strategies, because they are able to take into account points that are currently being observed (in Trieste we call them "pending"). Trieste currently provides two such functions: Local Penalization and GIBBON.
```
hstojic: amout -> amount

hstojic: why do we then even allow setting `num_query_points` to anything else than 1?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+# Number of iterations to run the sycnhronous scenario for
+num_iterations = 10
+# Number of observations to collect in asynchronous scenario
+num_observations = num_workers * num_iterations
+# Set this flag to False to disable sleep delays in case you want the notebook to execute quickly
+enable_sleep_delays = True
+
+# %% [markdown]
+# ## Asynchronous optimization
+# This section runs the asynchronous optimization routine. We first setup the [ask/tell optimizer](ask_tell_optimization.ipynb). Notice that we use **AsyncEfficientGlobalOptimization** rule specifically designed for asynchronous scenarios. Next we create thread-safe queues for points and observations, and run the optimization loop.
```
hstojic: additional sentence on why ask-tell is needed here would be useful

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+m = Manager()
+pq = m.Queue()
+oq = m.Queue()
+# keep track of all workers we have launched
+observer_processes = []
+# counter to keep track of collected observations
+points_observed = 0
+
+start = timeit.default_timer()
+try:
```
hstojic, line 198: a bit of education for me - can you clarify why is `try` needed here?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+sync_lp_time = stop - start
+print(f"Got {len(sync_lp_observations)} observations in {sync_lp_time:.2f}s")
+
+
+# %% [markdown]
+# ## Comparison
+# To compare outcomes of sync and async runs, let's plot their respective regrets side by side, and print out the running time. We expect async scenario to run a little bit faster for this toy problem.
+
+# %%
+from trieste.objectives import SCALED_BRANIN_MINIMUM
```
hstojic: imported already, right?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+#     display_name: 'Python 3.7.5 64-bit (''.venv'': venv)'
+#     name: python3
+# ---
+
+# %% [markdown]
+# # Asynchronous Bayesian optimization with Trieste
+#
+# In this notebook we show how Bayesian optimization can be done asynchronuosly. That is pertinent in scenarios when the objective function we are aiming to optimize can be run for several points in parallel, and observations might return back at different times. To avoid wasting resources we immediatelly request the next point asynchronuosly, taking into account points that still being evaluated.
+#
+# To contrast this approach with regular [batch optimization](batch_optimization.ipynb), this notebook also shows how to run parallel synchronous batch approach.
```
hstojic, line 19: it would be good to add a sentence on potential performance differences, async one will have more information as some observations would have had arrived and model could be improved, right?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+sync_lp_observations = (
+    sync_bo.to_result().try_get_final_dataset().observations - SCALED_BRANIN_MINIMUM
+)
+sync_lp_time = stop - start
+print(f"Got {len(sync_lp_observations)} observations in {sync_lp_time:.2f}s")
+
+
+# %% [markdown]
+# ## Comparison
+# To compare outcomes of sync and async runs, let's plot their respective regrets side by side, and print out the running time. We expect async scenario to run a little bit faster for this toy problem.
```
hstojic: a comment on expected differences in performance?

hstojic (commented):
what I could assess looks very good - mostly notebook, multiprocessing in python an AsyncEfficientGlobalOptimization itself I canno really evaluate

I have left lots of small suggestions for improvement in the notebook, nothing major

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+# 
+# The overall setup is illustrated in this diagram:
+# ![multiprocessing setup](figures/async_bo.png)
+
+# %%
+# Necessary multiprocessing primitives
+from multiprocessing import Manager, Process
+
+# %% [markdown]
+# We now define several common functions to implement the described setup. First we define a worker function that will be running a single observation in a separate process. Worker takes both queues as an input, reads next point from the points queue, makes an observation, and inserts observed data into the observations queue.
```
hstojic, line 115: not sure if its needed, but perhaps it would be useful to add something about assumptions - that the user actually has multiple threads/cores

perhaps add a link to a useful python documentation on multiprocessing

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+# these imports will be used later for optimization
+from trieste.acquisition import LocalPenalizationAcquisitionFunction
+from trieste.acquisition.rule import AsyncEfficientGlobalOptimization, EfficientGlobalOptimization
+from trieste.ask_tell_optimization import AskTellOptimizer
+
+
+# %% [markdown]
+# ## Multiprocessing setup
+#
+# To keep this notebook as reproducible as possible, we will only be using Python's multiprocessing package here. In this section we will explain our setup and define some common code to be used later.
```
hstojic, line 103: perhaps mention in passing what would be alternatives to Python multiprocessing package?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
henrymoss: In this notebook we demonstrate Trieste's ability to perform asynchronous Bayesian optimisation, as is suitable for scenarios where the objective function  can be run for several points in parallel but where observations might return back at different times. To avoid wasting resources waiting for the evaluation of the whole batch, we immediately request the next point asynchronously, taking into account points that still being evaluated.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
henrymoss: First, let's define a simple objective that will emulate evaluations taking variable time. We will be using a classic BO benchmark function [Branin](https://www.sfu.ca/~ssurjano/branin.html) with a sleep call inserted in the middle of the calculation to emulate delay. Our sleep delay is a scaled sum of all input values to make sure delays are uneven.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+#
+# In both synchronous and asynchronous scenarios we will have a fixed set of worker processes performing observations. We will also have a main process responsible for optimization process with Trieste. When Trieste suggests a new point, it is inserted into a points queue. One of the workers picks this point from the queue, performs the observation, and inserts the output into observations queue. The main process then picks up the observation from the queue, at which moment it either waits for the rest of the points in the batch to come back (synchronous scenario) or immediately suggests a new point (asynchrunous scenario). This process continues either for a certain number of iterations or until we accumulate necessary number of observations.
```
henrymoss: This section is much clearer!

henrymoss: In both synchronous and asynchronous scenarios we will have a fixed number of worker processes collecting observations. We will also have a main process responsible for thee Bayesian optimization process through Trieste. When Trieste suggests a new point, it is inserted into a points queue. One of the workers picks this point from the queue, performs the observation, and inserts the output into observations queue. The main process then picks up the observation from the queue, at which moment it either waits for the rest of the points in the batch to come back (synchronous scenario) or immediately suggests a new point (asynchronous scenario). This process continues either for a certain number of iterations or until we accumulate necessary number of observations.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+
+# %% [markdown]
+# Finally we set some common parameters. See comments below for explanation of what each one means.
+# %%
+# Number of worker processes to run simultaneously
+# Setting this to 1 will turn both setups into non-batch sequential optimization
+num_workers = 3
+# Number of iterations to run the sycnhronous scenario for
+num_iterations = 10
+# Number of observations to collect in asynchronous scenario
```
henrymoss: in the asynchronous scenario

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+num_observations = num_workers * num_iterations
+# Set this flag to False to disable sleep delays in case you want the notebook to execute quickly
+enable_sleep_delays = True
```
henrymoss, line 168: Is there a way to set this to be False, when we run our notebook tests? i.e checking an environment variable? @hstojic @uri-granta ?

henrymoss: We could do something similar in most of our notebooks, as they are currently super slow when testing the package.

henrymoss: In the BoTorch notebooks they use 
`SMOKE_TEST = os.environ.get("SMOKE_TEST")`
and then write things like
`candidate_set_size = 1000 if not SMOKE_TEST else 5`

henrymoss: they actually tell us how to do it https://github.com/pytorch/botorch/commit/5629ac4201cfdb1e3f51d42e25d190150f8c8dd9

docs/notebooks/asynchronous_bayesian_optimization.pct.py
henrymoss: I agree. There is no need to allow batches for asynchronous

henrymoss (commented):
Getting there! Notebook is much better

trieste/acquisition/rule.py
apaleyes: of course, thanks

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: that was silly of me to miss

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: good call

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+import time
+import timeit
+
+
+# %% [markdown]
+# First, let's define a simple objective that will emulate workload of variable time. We will be using class BO benchmark function [Branin](https://www.sfu.ca/~ssurjano/branin.html), and insert sleep call in the middle of the calculation to emulate delay. Our sleep delay is a scaled sum of all input values to make sure delays are uneven.
+# %%
+def objective(points, sleep=True):
+    if points.shape[1] != 2:
+        raise ValueError(f"Incorrect input shape, expected (*, 2), got {points.shape}")
```
vdutor, line 43: Minor: following TensorFlow syntax for tensor shapes.
```suggestion
        raise ValueError(f"Incorrect input shape, expected [None, 2], got {points.shape}")
```

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+        observation = scaled_branin(point)
+        observations.append(observation)
+
+    return np.array(observations)
+
+# test the defined objective function
+objective(np.array([[0.1, 0.5]]), sleep=False)
+
+# %% [markdown]
+# As always, we need to prepare model and some initial data to kick-start the optimization process.
```
vdutor:
```suggestion
# As always, we need to prepare the model and some initial data to kick-start the optimization process.
```

docs/notebooks/asynchronous_bayesian_optimization.pct.py
vdutor:
```suggestion
# In both synchronous and asynchronous scenarios we will have a fixed set of worker processes performing observations. We will also have a main process responsible for optimization process with Trieste. When Trieste suggests a new point, it is inserted into a points queue. One of the workers picks this point from the queue, performs the observation, and inserts the output into the observations queue. The main process then picks up the observation from the queue, at which moment it either waits for the rest of the points in the batch to come back (synchronous scenario) or immediately suggests a new point (asynchronous scenario). This process continues either for a certain number of iterations or until we accumulate necessary number of observations.
```

docs/notebooks/asynchronous_bayesian_optimization.pct.py
```diff
@@ -0,0 +1,373 @@
...
+
+# %% [markdown]
+# ## Multiprocessing setup
+#
+# To keep this notebook as reproducible as possible, we will only be using Python's multiprocessing package here. In this section we will explain our setup and define some common code to be used later.
+#
+# In both synchronous and asynchronous scenarios we will have a fixed set of worker processes performing observations. We will also have a main process responsible for optimization process with Trieste. When Trieste suggests a new point, it is inserted into a points queue. One of the workers picks this point from the queue, performs the observation, and inserts the output into observations queue. The main process then picks up the observation from the queue, at which moment it either waits for the rest of the points in the batch to come back (synchronous scenario) or immediately suggests a new point (asynchrunous scenario). This process continues either for a certain number of iterations or until we accumulate necessary number of observations.
+# 
+# The overall setup is illustrated in this diagram:
+# ![multiprocessing setup](figures/async_bo.png)
```
vdutor, line 108: Figure is not displayed in ipynb.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
vdutor: The "ask/tell optimizer" link doesn't seem to point anything?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: It isn't. This worker sits in a separate process and listens to the queue. When a new point arrives, it processes it, puts the output in the queue, and sits waiting again. In the end we terminate it to clean up resources.

What you have in mind is a different way to implement it: firing workers for each new point and taking them down immediately after the observation is done. It's also a reasonable approach, I've just implemented a different one here.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: Actually setting this to 2 will increase the computation time. More workers - faster compute.

trieste/acquisition/rule.py
apaleyes: You are not, it isn't used. It's a remnant of my attempts of writing general StatefulEGO. Removed

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: yes, removed

docs/notebooks/asynchronous_bayesian_optimization.pct.py
vdutor: Suggestion: a sentence on what the AsyncEfficientGlobalOptimization does differently could be interesting.

apaleyes: added

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: to clean up running processes in case anything goes wrong and the code falls with exception. Note that we are not actually using `except`, only `try/finally`

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: yup! removed

vdutor (commented):
*Only went through the notebook.* Looks really good to me - only left a few minor suggestions.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: I don't know if we can clearly make this call. Although [this paper](https://arxiv.org/abs/1901.10452) claims async is actually more efficient in both runtime and sample efficiency, i wouldn't rush to set user expectations in general. Any ideas how to phrase it without claiming much?

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: see reply above

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: No assumptions really, that code can run anywhere  Python can. But in case of 1 CPU there will be no performance gain, in fact it may degrade. I'll add a short note in the last section.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: My plan is to provide a second notebook that will use one of the alternatives. Then we'll link the two.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: Was pondering the same question waiting for the builds to run. Will research

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: "asynchronous" is tricky to spell out, thanks!

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: good point! this is now re-written slightly because we renamed it to make it obvious it only works with greedy functions. hopefully the re-write made it clearer

apaleyes: Unfortunate side effect of using pct.py instead of ipynb files in the codebase. The link will work in the final render of the docs! There is a tutorial on using ask/tell in Trieste which we refer to here.

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: Weird, it does for me. I'll build the docs locally to verify

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: good point, fixed

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: So, we can use env variable, or even [papermill](https://github.com/yaananth/papermill) which allows to parametrize notebooks. But we will have to add something ugly in the tutorial, along these lines:
```
if "async-bo-enable-sleep" in os.environ:
    # long comment explaining the reader why this is here, about integ tests and what not
    enable_sleep_delays = os.getenv("async-bo-enable-sleep")
else:
    enable_sleep_delays = True
```

i find this ugly in a tutorial. Shall we just downscale the delay?

henrymoss: we could just have one single environmental variable that can be used to scale down the runtimes for all notebooks (not a specific "async-bo-enable-sleep")

henrymoss: if we name the variable "running_CI" or something similar, then we dont need a comment to explain whats going on.

henrymoss: I would vote for having this line at the start of all our notebooks, as they are taking 20 mins to run at the moment. Any thoughts @uri-granta @hstojic @vpicheny @sebastianober ???

docs/notebooks/asynchronous_bayesian_optimization.pct.py
hstojic: I think its enough to put a stress on "potential" and give usual "depending on the use case" etc - I do think its important to say smth about why would you use it - link to that paper would also be good to add

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: i'll create an issue to discuss, it is definitely a good question, one that deserves a separate discussion

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: agreed and removed after some further thinking

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: After local build image and all links render properly.

apaleyes:
Hopefully all the comments are addressed. Noticeable updates:
* Renamed rule to AsynchronousGreedy. No abbreviations, no shortcuts, only hardcore!
* Removed batch size parameter.
* I've run into [this bug](https://stackoverflow.com/questions/64462924/how-do-i-fix-raise-valueerrorduplicate-signature-r-signature-when-i-try) when building docs locally, and had to update version of jupyter_client (and resolve a conflict)
* Many-many fixes and updates in notebook's write-up

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: #371

docs/notebooks/asynchronous_bayesian_optimization.pct.py
apaleyes: Added something that is hopefully vague enough

hstojic (approved):
notebook looks great!

rule.py and function.py hopefully others were able to properly cover

henrymoss (approved):

PR #367, opened 2021-10-05
Support for product spaces

henrymoss:
I have added a product space, which takes in any number of other SearchSpace objects and builds a single space. Each individual space is given a name and we can still access the individual subspaces when required, as well as using the product space.

` space_A = Box([-1, -2], [2, 3])`
`space_B = DiscreteSearchSpace(tf.constant([[-0.5, 0.5]]))`
`ProductSearchSpace(spaces=[("A", space_A), ("A", space_B)])`

This functionality has at least two immediate applications:
1. it allows mixtures of continuous and discrete search spaces and so allows optimization over mixed search spaces.
2. it allows rules that can handle different search space inputs differently, e.g. we might have a product of two spaces --- one representing a space of decision variables (which we will want to optimize over) and the other representing another space of context variables which can be handled differently, e.g. when performing multi-fidelity or profile optimization. 

I have added unit tests for all the new methods.

### DIFF
--- a/tests/unit/test_space.py
+++ b/tests/unit/test_space.py
@@ -24,7 +24,7 @@ import tensorflow as tf
 from typing_extensions import Final
 
 from tests.util.misc import TF_DEBUGGING_ERROR_TYPES, ShapeLike, various_shapes
-from trieste.space import Box, DecomposableSearchSpace, DiscreteSearchSpace, SearchSpace
+from trieste.space import Box, DiscreteSearchSpace, ProductSearchSpace, SearchSpace
 
 
 class Integers(SearchSpace):
@@ -43,7 +43,7 @@ class Integers(SearchSpace):
         return Integers(self.limit * other.limit)
 
     @property
-    def num_input_features(self) -> tf.Tensor:
+    def dimension(self) -> tf.Tensor:
         pass
 
 
@@ -92,16 +92,16 @@ def test_discrete_search_space_does_not_contain_other_points(point: tf.Tensor) -
 
 
 @pytest.mark.parametrize(
-    "space, num_input_features",
+    "space, dimension",
     [
         (DiscreteSearchSpace(tf.constant([[-0.5], [0.2], [1.2], [1.7]])), 1),  # 1d
         (DiscreteSearchSpace(tf.constant([[-0.5, -0.3], [1.2, 0.4]])), 2),  # 2d
     ],
 )
-def test_discrete_search_space_returns_correct_num_input_features(
-    space: DiscreteSearchSpace, num_input_features: int
+def test_discrete_search_space_returns_correct_dimension(
+    space: DiscreteSearchSpace, dimension: int
 ) -> None:
-    assert space.num_input_features == num_input_features
+    assert space.dimension == dimension
 
 
 @pytest.mark.parametrize(
@@ -287,15 +287,15 @@ def test_box_raises_if_any_lower_bound_is_not_less_than_upper_bound(
 
 
 @pytest.mark.parametrize(
-    "space, num_input_features",
+    "space, dimension",
     [
         (Box([-1], [2]), 1),  # 1d
         (Box([-1, -2], [1.5, 2.5]), 2),  # 2d
         (Box([-1, -2, -3], [1.5, 2.5, 3.5]), 3),  # 3d
     ],
 )
-def test_box_returns_correct_num_input_features(space: Box, num_input_features: int) -> None:
-    assert space.num_input_features == num_input_features
+def test_box_returns_correct_dimension(space: Box, dimension: int) -> None:
+    assert space.dimension == dimension
 
 
 def test_box_bounds_attributes() -> None:
@@ -469,90 +469,91 @@ def test_box_deepcopy() -> None:
     npt.assert_allclose(box.upper, box_copy.upper)
 
 
-def test_decomposable_space_bounds_attributes() -> None:
+def test_product_space_raises_for_non_unqique_subspace_names() -> None:
+    space_A = Box([-1, -2], [2, 3])
+    space_B = DiscreteSearchSpace(tf.constant([[-0.5, 0.5]]))
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        ProductSearchSpace(spaces=[("A", space_A), ("A", space_B)])
+
+
+def test_product_space_subspace_tags_attribute() -> None:
     decision_space = Box([-1, -2], [2, 3])
     context_space = DiscreteSearchSpace(tf.constant([[-0.5, 0.5]]))
-    decomposable_space = DecomposableSearchSpace(
-        context_space=context_space, decision_space=decision_space
+    product_space = ProductSearchSpace(
+        spaces=[("context", context_space), ("decision", decision_space)]
     )
 
-    assert isinstance(decomposable_space.decision_space, type(decision_space))
-    assert isinstance(decomposable_space.context_space, type(context_space))
-
-    npt.assert_allclose(decomposable_space.decision_space.lower, [-1, -2])
-    npt.assert_allclose(decomposable_space.decision_space.upper, [2, 3])
-
-    npt.assert_allclose(decomposable_space.context_space.points, tf.constant([[-0.5, 0.5]]))
+    npt.assert_array_equal(product_space.subspace_tags, ["context", "decision"])
 
 
 @pytest.mark.parametrize(
-    "context_space, decision_space, num_input_features",
+    "spaces, dimension",
     [
-        (
-            DiscreteSearchSpace(tf.constant([[-0.5]])),
-            DiscreteSearchSpace(tf.constant([[-0.5, -0.3], [1.2, 0.4]])),
-            3,
-        ),
-        (DiscreteSearchSpace(tf.constant([[-0.5]])), Box([-1], [2]), 2),
-        (Box([-1, -2], [2, 3]), DiscreteSearchSpace(tf.constant([[-0.5]])), 3),
-        (Box([-1, -2], [2, 3]), Box([-1, -2], [2, 3]), 4),
+        ([("A", DiscreteSearchSpace(tf.constant([[-0.5, -0.3], [1.2, 0.4]])))], 2),
+        ([("A", DiscreteSearchSpace(tf.constant([[-0.5]]))), ("B", Box([-1], [2]))], 2),
+        ([("A", Box([-1, -2], [2, 3])), ("B", DiscreteSearchSpace(tf.constant([[-0.5]])))], 3),
+        ([("A", Box([-1, -2], [2, 3])), ("B", Box([-1, -2], [2, 3])), ("C", Box([-1], [2]))], 5),
     ],
 )
-def test_decomposable_search_space_returns_correct_num_input_features(
-    context_space: SearchSpace, decision_space: SearchSpace, num_input_features: int
+def test_product_search_space_returns_correct_dimension(
+    spaces: dict[str, SearchSpace], dimension: int
 ) -> None:
-    space = DecomposableSearchSpace(context_space=context_space, decision_space=decision_space)
-    assert space.num_input_features == num_input_features
+    space = ProductSearchSpace(spaces=spaces)
+    assert space.dimension == dimension
+
+
+def test_product_space_get_subspace() -> None:
+    space_A = Box([-1, -2], [2, 3])
+    space_B = DiscreteSearchSpace(tf.constant([[-0.5, 0.5]]))
+    space_C = Box([-1], [2])
+    product_space = ProductSearchSpace(spaces=[("A", space_A), ("B", space_B), ("C", space_C)])
+
+    subspace_A = product_space.get_subspace("A")
+    assert isinstance(subspace_A, Box)
+    npt.assert_array_equal(subspace_A.lower, [-1, -2])
+    npt.assert_array_equal(subspace_A.upper, [2, 3])
+
+    subspace_B = product_space.get_subspace("B")
+    assert isinstance(subspace_B, DiscreteSearchSpace)
+    npt.assert_array_equal(subspace_B.points, tf.constant([[-0.5, 0.5]]))
+
+    subspace_C = product_space.get_subspace("C")
+    assert isinstance(subspace_C, Box)
+    npt.assert_array_equal(subspace_C.lower, [-1])
+    npt.assert_array_equal(subspace_C.upper, [2])
 
 
 @pytest.mark.parametrize(
-    "context_space, decision_space, num_context_features, num_decision_features",
+    "spaces, subspace_dim_range",
     [
+        ([("A", DiscreteSearchSpace(tf.constant([[-0.5]])))], {"A": [0, 1]}),
         (
-            DiscreteSearchSpace(tf.constant([[-0.5, 0.5]])),
-            DiscreteSearchSpace(tf.constant([[-0.5, -0.3], [1.2, 0.4]])),
-            2,
-            2,
+            [
+                ("A", DiscreteSearchSpace(tf.constant([[-0.5]]))),
+                ("B", DiscreteSearchSpace(tf.constant([[-0.5, -0.3], [1.2, 0.4]]))),
+            ],
+            {"A": [0, 1], "B": [1, 3]},
         ),
-        (DiscreteSearchSpace(tf.constant([[-0.5]])), Box([-1], [2]), 1, 1),
-        (Box([-1, -2], [2, 3]), DiscreteSearchSpace(tf.constant([[-0.5]])), 2, 1),
-        (Box([-1], [2]), Box([-1, -2], [2, 3]), 1, 2),
+        (
+            [
+                ("A", Box([-1, -2], [2, 3])),
+                ("B", DiscreteSearchSpace(tf.constant([[-0.5]]))),
+                ("C", Box([-1], [2])),
+            ],
+            {"A": [0, 2], "B": [2, 3], "C": [3, 4]},
+        ),
     ],
 )
-def test_decomposable_space_can_get_context_component(
-    context_space: SearchSpace,
-    decision_space: SearchSpace,
-    num_context_features: int,
-    num_decision_features: int,
+def test_product_space_can_get_subspace_components(
+    spaces: dict[str, SearchSpace],
+    subspace_dim_range: dict[str, int],
 ) -> None:
-    space = DecomposableSearchSpace(context_space=context_space, decision_space=decision_space)
-    points = tf.random.uniform([10, num_context_features + num_decision_features])
-    npt.assert_array_equal(space.get_context_component(points), points[:, :num_context_features])
+    space = ProductSearchSpace(spaces)
+    points = tf.random.uniform([10, space.dimension])
 
-
-@pytest.mark.parametrize(
-    "context_space, decision_space, num_context_features, num_decision_features",
-    [
-        (
-            DiscreteSearchSpace(tf.constant([[-0.5, 0.5]])),
-            DiscreteSearchSpace(tf.constant([[-0.5, -0.3], [1.2, 0.4]])),
-            2,
-            2,
-        ),
-        (DiscreteSearchSpace(tf.constant([[-0.5]])), Box([-1], [2]), 1, 1),
-        (Box([-1, -2], [2, 3]), DiscreteSearchSpace(tf.constant([[-0.5]])), 2, 1),
-        (Box([-1], [2]), Box([-1, -2], [2, 3]), 1, 2),
-    ],
-)
-def test_decomposable_space_can_get_decision_component(
-    context_space: SearchSpace,
-    decision_space: SearchSpace,
-    num_context_features: int,
-    num_decision_features: int,
-) -> None:
-    space = DecomposableSearchSpace(context_space=context_space, decision_space=decision_space)
-    points = tf.random.uniform([10, num_context_features + num_decision_features])
-    npt.assert_array_equal(space.get_decision_component(points), points[:, -num_decision_features:])
+    for tag in space._spaces.keys():
+        subspace_points = points[:, subspace_dim_range[tag][0] : subspace_dim_range[tag][1]]
+        npt.assert_array_equal(space.get_subspace_component(tag, points), subspace_points)
 
 
 @pytest.mark.parametrize(
@@ -562,13 +563,11 @@ def test_decomposable_space_can_get_decision_component(
         tf.constant([2.0, 3.0, -0.5, 0.5], dtype=tf.float64),
     ],
 )
-def test_decomposable_space_contains_point(point: tf.Tensor) -> None:
-    context_space = Box([-1.0, -2.0], [2.0, 3.0])
-    decision_space = DiscreteSearchSpace(tf.constant([[-0.5, 0.5]], dtype=tf.float64))
-    decomposable_space = DecomposableSearchSpace(
-        context_space=context_space, decision_space=decision_space
-    )
-    assert point in decomposable_space
+def test_product_space_contains_point(point: tf.Tensor) -> None:
+    space_A = Box([-1.0, -2.0], [2.0, 3.0])
+    space_B = DiscreteSearchSpace(tf.constant([[-0.5, 0.5]], dtype=tf.float64))
+    product_space = ProductSearchSpace(spaces=[("A", space_A), ("B", space_B)])
+    assert point in product_space
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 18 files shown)


trieste/space.py
```diff
@@ -104,6 +108,11 @@ def points(self) -> TensorType:
         """All the points in this space."""
         return self._points
 
+    @property
+    def num_input_features(self) -> TensorType:
```
vpicheny: why not `dimension`?

trieste/space.py
```diff
@@ -329,3 +343,121 @@ def __mul__(self, other: Box) -> Box:
...
         return self
+
+
+class DecomposableSearchSpace(SearchSpace):
+    r"""
+    Decomposable :class:`SearchSpace` consisting of separate decision and
+    context :class:`SearchSpace`s. This class provides functionality for using
+    either the resulting combined search space or each individual space.
+
+    Note that this class assumes individual points are represented with their context
```
hstojic: This seems like a potential problem down the line, it will make things more difficult for any downstream function that needs to differentiate between them (models, acquisition functions etc). Why not having a `NamedTuple` of tensors? Then those functions that are able or need to use context will use the additional attribute in the tuple.

trieste/space.py
```diff
@@ -51,7 +51,11 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
             dimensionality from this :class:`SearchSpace`.
         """
 
+    @property
     @abstractmethod
+    def num_input_features(self) -> TensorType:
+        """The input dimensionality of this search space."""
+
     def __mul__(self: SP, other: SP) -> SP:
```
apaleyes, line 60: Oops! Did `@abstractmethod` disappear from `__mul__` deliberately?

trieste/space.py
apaleyes: I guess the problem here is that this object has to allow to be treated as a regular search space. So at some point it will be given or asked for just a point `x`, and without assumptions it's impossible to tell which part of this `x` stands for what

trieste/space.py
```diff
@@ -329,3 +343,121 @@ def __mul__(self, other: Box) -> Box:
 
     def __deepcopy__(self, memo: dict[int, object]) -> Box:
         return self
+
+
+class DecomposableSearchSpace(SearchSpace):
```
apaleyes: Are you going to add `set_context` or something along these lines? That's probably the main point of having context variables

trieste/space.py
```diff
@@ -329,3 +343,121 @@ def __mul__(self, other: Box) -> Box:
...
+    @property
+    def num_input_features(self) -> TensorType:
+        """The input dimensionality of this search space."""
+        return self._context_space.num_input_features + self._decision_space.num_input_features
+
+    def get_context_component(self, values: TensorType) -> TensorType:
+        """
+        Returns the context components of ``values``, assuming that the context variables
+        are listed first.
+        :param value: Points from the :class:`DecomposableSearchSpace` of shape [N,D].
```
apaleyes: You need an empty line here for docs build to pass

trieste/space.py
```diff
@@ -329,3 +343,121 @@ def __mul__(self, other: Box) -> Box:
...
+        :return: The context components of ``values` of shape [N_1, D], where
+            N_1 is the dimensionality of the context space.
+        """
+        return values[:, : self.context_space.num_input_features]
+
+    def get_decision_component(self, values: TensorType) -> TensorType:
+        """
+        Returns the decision components of ``values``, assuming that the context variables
+        are listed first.
+        :param value: Points from the :class:`DecomposableSearchSpace` of shape [N,D].
```
apaleyes: ditto

trieste/space.py
```diff
@@ -329,3 +343,121 @@ def __mul__(self, other: Box) -> Box:
...
+    def decision_space(self) -> SearchSpace:
+        """The domain of the decision variables."""
+        return self._decision_space
+
+    @property
+    def num_input_features(self) -> TensorType:
+        """The input dimensionality of this search space."""
+        return self._context_space.num_input_features + self._decision_space.num_input_features
+
+    def get_context_component(self, values: TensorType) -> TensorType:
```
apaleyes: These components methods look good, but it's a bit hard to judge how useful they are without an example. Can you explain why they are needed?

They are used below in `__contains__`, but since they are declared as public, I was wondering why would anyone use them outside this class.

trieste/space.py
```diff
@@ -329,3 +343,121 @@ def __mul__(self, other: Box) -> Box:
...
+        :return: The decision components of ``values` of shape [N_2, D], where
+            N_2 is the dimensionality of the decision space.
+        """
+        return values[:, -self.decision_space.num_input_features :]
+
+    def __contains__(self, value: TensorType) -> bool | TensorType:
+        """
+        Return `True` if ``value`` is a member of this search space, else `False`. A point is a
+        member if its decision components lie in the decision space and its context components lie
+        in its context space.
```
apaleyes: Worth mentioning the order assumption again?

trieste/space.py
```diff
@@ -329,3 +343,121 @@ def __mul__(self, other: Box) -> Box:
...
+        :param decision_space: the domain of the decision variables.
+        """
+        self._context_space = context_space
+        self._decision_space = decision_space
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""DecomposableSearchSpace(
+        {self._context_space.__repr__()},
+        {self._decision_space.__repr__()})
```
apaleyes: How about this?

```suggestion
        return f"""DecomposableSearchSpace(
        ContextSpace = {self._context_space.__repr__()},
        DecisionSpace = {self._decision_space.__repr__()})
```

tests/unit/test_space.py
```diff
@@ -438,3 +467,224 @@ def test_box_deepcopy() -> None:
...
+@pytest.mark.parametrize(
+    "point",
+    [
+        tf.constant([-1.1, 0.0, -0.5, 0.5], dtype=tf.float64),  # just outside context space
+        tf.constant([-10, 10.0, -0.5, 0.5], dtype=tf.float64),  # well outside context space
+        tf.constant([2.0, 3.0, 2.0, 7.0], dtype=tf.float64),  # outside decision space
+        tf.constant([-10.0, -10.0, -10.0, -10.0], dtype=tf.float64),  # outside both
+    ],
+)
+def test_decomposable_space_does_not_contains_point(point: tf.Tensor) -> None:
```
apaleyes:
```suggestion
def test_decomposable_space_does_not_contain_point(point: tf.Tensor) -> None:
```

tests/unit/test_space.py
```diff
@@ -438,3 +467,224 @@ def test_box_deepcopy() -> None:
...
+    assert point in decomposable_space
+
+
+@pytest.mark.parametrize(
+    "point",
+    [
+        tf.constant([-1.1, 0.0, -0.5, 0.5], dtype=tf.float64),  # just outside context space
+        tf.constant([-10, 10.0, -0.5, 0.5], dtype=tf.float64),  # well outside context space
+        tf.constant([2.0, 3.0, 2.0, 7.0], dtype=tf.float64),  # outside decision space
+        tf.constant([-10.0, -10.0, -10.0, -10.0], dtype=tf.float64),  # outside both
```
apaleyes, line 599: Fun test case - wrong order of context and decision components

tests/unit/test_space.py
```diff
@@ -438,3 +467,224 @@ def test_box_deepcopy() -> None:
...
+    ],
+)
+def test_decomposable_space_contains_raises_on_point_of_different_shape(
+    context_space: SearchSpace, decision_space: SearchSpace, num_input_features: int
+) -> None:
+    space = DecomposableSearchSpace(context_space=context_space, decision_space=decision_space)
+    for wrong_input_shape in [num_input_features - 1, num_input_features + 1]:
+        point = tf.zeros([wrong_input_shape])
+        with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+            space.__contains__(point)
```
apaleyes: why not `_ = point in space`?

apaleyes (commented):
That looks quite nice! A bit hard to say how well this will work in practice though without examples. In particular, just like you mention above, we may expect a lot of that to change once mixed spaces come about.

Are you sure about the name? "Decomposable" can mean many different things, and doesn't immediately hint at contexts.

trieste/space.py
hstojic: I get that, I'm just thinking whether we might want to move away from that - I didn't spell it out above, but if we start using `NamedTuple` here we would probably need to use them for other search spaces, but I think that might not be too much of a change

perhaps its too hard to say what would be more useful without seeing models/acquisitions examples that would handle both context and decision spaces

trieste/space.py
```diff
@@ -94,6 +98,7 @@ def __init__(self, points: TensorType):
         """
         tf.debugging.assert_shapes([(points, ("N", "D"))])
         self._points = points
+        self._dimension = tf.shape(self.points)[-1]
```
apaleyes:
```suggestion
        self._dimension = tf.shape(self._points)[-1]
```

trieste/space.py
```diff
@@ -51,7 +51,11 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
             dimensionality from this :class:`SearchSpace`.
         """
 
+    @property
     @abstractmethod
+    def dimension(self) -> TensorType:
+        """The input dimensionality of this search space."""
+
     def __mul__(self: SP, other: SP) -> SP:
```
apaleyes: Still missing `@abstractmethod`!

trieste/space.py
```diff
@@ -202,6 +212,8 @@ def __init__(
 
         tf.debugging.assert_less(self._lower, self._upper)
 
+        self._dimension = tf.shape(self.upper)[-1]
```
apaleyes:
```suggestion
        self._dimension = tf.shape(self._upper)[-1]
```

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+        r"""
+        :param spaces: A list of tag-space tuples summarizing the names and domains of
+            the space's subspaces.
+        """
+
+        tags = [tag for (tag, space) in spaces]
+        tf.debugging.assert_equal(
+            len(tags),
+            len(set(tags)),
+            message="subspace names must be unique",
```
apaleyes: ideally you'd what to tell the user which tag is repeated

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+        self._subspace_sizes = dict(zip(self._tags, subspace_sizes))
+        self._subspace_starting_indicies = dict(
+            zip(self._tags, tf.cumsum(subspace_sizes, exclusive=True))
+        )
+        self._dimension = tf.reduce_sum(subspace_sizes)
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""DecomposableSearchSpace(
+        {[space.__repr__()+"," for space in self._spaces]})
```
apaleyes: That iterates over keys, iirc, so the output will be just tags. Also, I'll repeat my suggestion of outputting tags, in a format of `space_tag: space_repr`

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+        :return: `True` if ``value`` is a member of this search space, else `False`. May return a
+            scalar boolean `TensorType` instead of the `bool` itself.
+        :raise ValueError (or tf.errors.InvalidArgumentError): If ``value`` has a different
+            dimensionality from the search space.
+        """
+
+        tf.debugging.assert_equal(
+            tf.shape(value),
+            self.dimension,
+            message="value must have same dimensionality as search space",
```
apaleyes: Adding some runtime information to logs and errors is always nice for the end users

```suggestion
            message=f"Dimensionality mismatch: space is {self.dimension}, value is {tf.shape(value}",
```

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+        )
+        value = tf.expand_dims(value, 0)
+        in_each_subspace = [
+            self._spaces[tag].__contains__(self.get_subspace_component(tag, value)[0, :])
```
apaleyes, line 465: Why call `__contains__` explicitly instead of 
```
self.get_subspace_component(tag, value)[0, :] in self._spaces[tag]
```

Also, if you have to do this `[0, :]` thingy, perhaps that shall happen in `get_subspace_component`

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+        """
+
+        starting_index_of_subspace = self._subspace_starting_indicies[tag]
+        ending_index_of_subspace = starting_index_of_subspace + self._subspace_sizes[tag]
+        return values[:, starting_index_of_subspace:ending_index_of_subspace]
+
+    def __contains__(self, value: TensorType) -> bool | TensorType:
+        """
+        Return `True` if ``value`` is a member of this search space, else `False`. A point is a
+        member if each of its subspace components lie in each subspace.
```
apaleyes, line 444:
```suggestion
        member if each of its subspace components lie in corresponding subspace.
```

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+
+        Recall that individual points in product spaces are represented with their inputs in the
+        same order as specified when initializing the space.
+
+        :param value: A point to check for membership of this :class:`SearchSpace`.
+        :return: `True` if ``value`` is a member of this search space, else `False`. May return a
+            scalar boolean `TensorType` instead of the `bool` itself.
+        :raise ValueError (or tf.errors.InvalidArgumentError): If ``value`` has a different
+            dimensionality from the search space.
```
apaleyes, line 453: So which one? It can't be both, can it?

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+        """
+        Sample randomly from the space by sampling from each subspace
+        and concatenating the resulting samples.
+
+        :param num_samples: The number of points to sample from this search space.
+        :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
+            from this search space with shape '[num_samples, D]' , where D is the search space
+            dimension.
+        """
+        tf.debugging.assert_non_negative(num_samples)
```
apaleyes, line 480: 0 samples is ok? Just checking, I don't know if it is

tests/unit/test_space.py
```diff
@@ -438,3 +467,212 @@ def test_box_deepcopy() -> None:
...
+    ],
+)
+def test_product_space_can_get_subspace_components(
+    spaces: list[tuple[str, SearchSpace]],
+    subspace_dim_range: dict[str, list[int]],
+) -> None:
+    space = ProductSearchSpace(spaces)
+    points = tf.random.uniform([10, space.dimension])
+
+    for tag in space._spaces.keys():
```
apaleyes: Accessing private `_space` shall not be necessary. Maybe just iterate over `spaces` instead?

tests/unit/test_space.py
apaleyes: So that's a "no"?

tests/unit/test_space.py
apaleyes: Question still stands

tests/unit/test_space.py
```diff
@@ -438,3 +467,212 @@ def test_box_deepcopy() -> None:
...
+    space_B = DiscreteSearchSpace(tf.ones([100, 2], dtype=tf.float64))
+    product_space = ProductSearchSpace(spaces=[("A", space_A), ("B", space_B)])
+
+    dss = product_space.discretize(num_samples)
+    samples = dss.sample(num_samples)
+
+    assert len(samples) == num_samples
+
+    with pytest.raises(tf.errors.InvalidArgumentError):
+        dss.sample(num_samples + 1)
```
apaleyes, line 692: That shall be a separate test case

apaleyes (commented):
This looks to become quite general, nice!

trieste/space.py
hstojic: is it necessary at all, downstream `sample` methods are doing these checks, right?

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+
+        subspace_samples = [self._spaces[tag].sample(num_samples) for tag in self._tags]
+        return tf.concat(subspace_samples, -1)
```
hstojic, line 483: it might be very cumbersome to keep track to which tag samples are associated with, should we return tag-samples here?

henrymoss: If you want to get samples from a particular subspace, you can call `space.get_subspace("tag").sample(10)`

henrymoss: its important that the sample method of this class just returns a tensor, as this will be used when maximizing acquisition functions over the product space

tests/unit/test_space.py
henrymoss: done

trieste/space.py
henrymoss: I dont have this method for the product space, so its not needed for all space classes

trieste/space.py
apaleyes: There is no leeway here, clients of this class will just call `space.sample()` and expect the same result format no matter which space implementation it is.

trieste/space.py
henrymoss: all other samplers support 0 and this is tested for this class

trieste/space.py
henrymoss: so `return f"""DecomposableSearchSpace({[str((tag, self.get_subspace(tag)))+"," for tag in self.subspace_tags]})` ?

apaleyes: Not quite. Ideally I'd like to see soemthing like this:
```
ProductSpace:
    Box: [[1, 2], [1, 2]]
    DiscreteSpace: [[1, 1], [2,2], [3,3]]
```
makes sense?

trieste/space.py
apaleyes: Is that true? then let's discuss what spaces can implement it, and move it appropriately.

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
 
     def __deepcopy__(self, memo: dict[int, object]) -> Box:
         return self
+
+
+class ProductSearchSpace(SearchSpace):
```
uri-granta: Maybe make explicit the fact that the subspaces are tagged? (which differentiates this e.g. from the more limited multiplication method on spaces)

```suggestion
class TaggedProductSearchSpace(SearchSpace):
```

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+    Product :class:`SearchSpace` consisting of a product of
+    multiple :class:`SearchSpace`s. This class provides functionality for
+    accessing either the resulting combined search space or each individual space.
+
+    Note that this class assumes that individual points in product spaces are
+    represented with their inputs in the same order as specified when initializing
+    the space.
+    """
+
+    def __init__(self, spaces: list[tuple[str, SearchSpace]]):
```
uri-granta: This syntax isn't very Pythonic. How about the following?

```suggestion
    def __init__(self, spaces: Sequence[SearchSpace], tags: Optional[Sequence[str]] = None):
```

You would then write TaggedProductSearchSpace([space_A, space_B], ["A", "B"]), which is I think is cleaner. The initialiser can verify that the lists have the same lengths (just as it already verifies tag uniqueness).

You could then also provide default tags (e.g. "0", "1", etc) in the cases where this doesn't matter. Which raises the possibility of making `SearchSpace.__mul__` handle multiplying arbitrary search spaces, not just those of the same kind, falling back to returning a TaggedProductSearchSpace([self, other]) if there's no better impementation.

?

---

If you don't like spitting the tags and spaces, then another alternative is:

```suggestion
    def __init__(self, spaces: OrderedDict[str, SearchSpace]):
```

which I think is a bit cleaner than using lists of tuples (plus it checks tag uniqueness for you).

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+
+        tags = [tag for (tag, space) in spaces]
+        tf.debugging.assert_equal(
+            len(tags),
+            len(set(tags)),
+            message=f"subspace names must be unique, received {tags}.",
+        )
+
+        self._tags = tags
+        self._spaces = dict(zip(self._tags, [space for (tag, space) in spaces]))
```
uri-granta: You can initialise a dict directly from a list of tupes.

```suggestion
        self._spaces = dict(spaces)
```

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+        subspace_sizes = [self._spaces[tag].dimension for tag in self._tags]
+        self._subspace_sizes = dict(zip(self._tags, subspace_sizes))
+        self._subspace_starting_indicies = dict(
+            zip(self._tags, tf.cumsum(subspace_sizes, exclusive=True))
+        )
+        self._dimension = tf.reduce_sum(subspace_sizes)
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""DecomposableSearchSpace(
```
uri-granta: Make sure to use the actual class name (whatever that ends up as)

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+    the space.
+    """
+
+    def __init__(self, spaces: list[tuple[str, SearchSpace]]):
+        r"""
+        :param spaces: A list of tag-space tuples summarizing the names and domains of
+            the space's subspaces.
+        """
+
+        tags = [tag for (tag, space) in spaces]
```
uri-granta: Better store these in a tuple, so nobody can modify it by mistake (since you're exposing it in a property):

```suggestion
        tags = tuple(tag for (tag, space) in spaces)
```

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+        self._dimension = tf.reduce_sum(subspace_sizes)
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""DecomposableSearchSpace(
+        {[str((tag, self.get_subspace(tag)))+"," for tag in self.subspace_tags]})
+        """
+
+    @property
+    def subspace_tags(self) -> list[str]:
```
uri-granta: see comment above:

```suggestion
    def subspace_tags(self) -> tuple[int, ...]:
```

trieste/space.py
```diff
@@ -329,3 +346,124 @@ def __mul__(self, other: Box) -> Box:
...
+    def discretize(self, num_samples: int) -> DiscreteSearchSpace:
+        """
+        :param num_samples: The number of points in the :class:`DiscreteSearchSpace`.
+        :return: A discrete search space consisting of ``num_samples`` points sampled
+            uniformly across the space.
+        """
+        return DiscreteSearchSpace(points=self.sample(num_samples))
+
+    def __deepcopy__(self, memo: dict[int, object]) -> ProductSearchSpace:
+        return self
```
uri-granta: Not 100% sure about this. The other classes implement this method, though I'm not certain why: perhaps it's to avoid e.g. copying all the points of a discrete search space when we can just reuse the same object? If however we ever actually need to copy something in a search space type to make it work, then having this here would prevent that being called for a subspace of that type.

Maybe just remove this and use the default copy behaviour instead?

trieste/space.py
uri-granta: (My comment below about the constructor and default tags would make it trivial to get this working for product spaces.)

uri-granta (commented):
A few comments about the code (not looked at the tests yet). Tell me what you think.

trieste/space.py
henrymoss: Yep I like this idea

trieste/space.py
henrymoss: these are all good suggestions!

I will split up spaces and tags, so that we can get the default behaviour. If we go down the orderedDict route, then we cannot do the searchspace.__mul__

trieste/space.py
henrymoss: actually, now that we can have default tagging behaviour, I will keep it as ProductSearchSpace

trieste/space.py
henrymoss: nice trick!

trieste/space.py
henrymoss: Gone a slightly different way based on Uri's suggestions

trieste/space.py
```diff
@@ -359,23 +360,34 @@ class ProductSearchSpace(SearchSpace):
...
+                    received {len(tags)} tags but {len(spaces)} subspaces.
+                """,
+            )
+            tf.debugging.assert_equal(
+                len(tags),
+                len(set(tags)),
+                message=f"Subspace names must be unique but received {tags}.",
+            )
+
+        self._spaces = dict(zip(tags, [space for space in spaces]))
```
```suggestion
        self._spaces = dict(zip(tags, spaces))
```

trieste/space.py
```diff
@@ -359,23 +360,34 @@ class ProductSearchSpace(SearchSpace):
...
+        self._tags = tuple(tag for tag in tags)  # fix to avoid accidental modification
```
```suggestion
        self._tags = tuple(tags)  # fix to avoid accidental modification
```

trieste/space.py
```diff
@@ -465,5 +478,22 @@ def discretize(self, num_samples: int) -> DiscreteSearchSpace:
...
+            TaggedProductSearchSpace(spaces=[Box([0.0,1.0]),Box[0.0,2.0]], tags=["0", "1"])
+            >>> new_box.get_subspace("1")
+            TaggedProductSearchSpace(spaces=[Box([0.0,3.0]),Box[0.0,4.0]], tags=["0", "1"])
+            >>> new_box.get_subspace("0").get_subspace("0")
+            Box([0.0,1.0]
+
+        :param other: A search space of the same type as this search space.
+        :return: The Cartesian product of this search space with the ``other``.
+        """
+        return TaggedProductSearchSpace(spaces=[self, other])
```
uri-granta, line 501: Nice. If you've got the energy, you can now easily simplify the abstract definition of `__mul__` so that it supports multiplying arbitrary searchspaces together, returning a single space when possible and a product space otherwise:

1. delete the definition of SP and change the abstract definition to:

```python
    def __mul__(self, other: SearchSpace) -> SearchSpace:
```

and

```python
    def __pow__(self, other: int) -> SearchSpace:
```

2. add a default implementation

```python
```

3. update the overriden `__mul__` methods to check if they're being passed the same type, otherwise delegate to super: for example

```python
    if not isinstance(other, DiscreteSearchSpace) or self.points.dtype is not other.points.dtype:
        return super().__mul__(self, other)
```

4 remove the `__mul__` override for TaggedProductSearchSpace as it's no longer necessary.

uri-granta (approved):
Looks good, just one more suggestion to make it even gooder! (plus the docs issue which I'll look at next)

trieste/space.py
```diff
@@ -329,3 +347,153 @@ def __mul__(self, other: Box) -> Box:
...
+            uniformly across the space.
+        """
+        return DiscreteSearchSpace(points=self.sample(num_samples))
+
+    def __mul__(self, other: SearchSpace) -> TaggedProductSearchSpace:
+        r"""
+        Return the Cartesian product of the two :class:`TaggedProductSearchSpace`\ s,
+        building a tree of :class:`TaggedProductSearchSpace`\ s. For example:
+
+            >>> space_0 = TaggedProductSearchSpace(spaces=[Box([0.0,1.0]),Box[0.0,2.0]])
```
```suggestion
            >>> space_0 = TaggedProductSearchSpace(spaces=[Box([0.0],[1.0]),Box([0.0],[2.0]])
```

trieste/space.py
henrymoss: Yep will do this!

trieste/space.py
```diff
@@ -329,3 +347,153 @@ def __mul__(self, other: Box) -> Box:
 
     def __deepcopy__(self, memo: dict[int, object]) -> Box:
         return self
+
+
+class TaggedProductSearchSpace(SearchSpace):
+    r"""
+    Product :class:`SearchSpace` consisting of a product of
+    multiple :class:`SearchSpace`s. This class provides functionality for
```
uri-granta: Apparently you can't attach an s to a reference like this.

```suggestion
    multiple :class:`SearchSpace`. This class provides functionality for
```

trieste/space.py
```diff
@@ -329,3 +347,153 @@ def __mul__(self, other: Box) -> Box:
...
+    accessing either the resulting combined search space or each individual space.
+
+    Note that this class assumes that individual points in product spaces are
+    represented with their inputs in the same order as specified when initializing
+    the space.
+    """
+
+    def __init__(self, spaces: Sequence[SearchSpace], tags: Optional[Sequence[str]] = None):
+        r"""
+        TODO discuss
```
uri-granta: Needs blank line (also actual text!)

```suggestion
        TODO discuss
        
```

trieste/space.py
```diff
@@ -329,3 +347,153 @@ def __mul__(self, other: Box) -> Box:
...
+    def get_subspace(self, tag: str) -> SearchSpace:
+        """Return the domain of a particular subspace."""
+        return self._spaces[tag]
+
+    def get_subspace_component(self, tag: str, values: TensorType) -> TensorType:
+        """
+        Returns the components of ``values`` lying in a particular subspace.
+
+        :param value: Points from the :class:`DecomposableSearchSpace` of shape [N,D].
+        :return: The context components of ``values` of shape [M, D], where
```
```suggestion
        :return: The context components of ``values`` of shape [M, D], where
```

uri-granta (commented):
Docs issues

trieste/space.py
henrymoss: Actually decided not to do this. THis can be a future PR. I am worried about this allowing users to build product spaces with subspaces operating on different dtypes, e.g. tf.float32 and tf.float64

henrymoss: I have opened an issue about this

trieste/space.py
```diff
@@ -329,3 +347,160 @@ def __mul__(self, other: Box) -> Box:
...
+        """
+
+        if tags is None:
+            tags = [str(i) for i in range(len(spaces))]
+        else:
+            tf.debugging.assert_equal(
+                len(tags),
+                len(spaces),
+                message=f"""
+                    Number of tags must match number if search spaces but
```
ChrisMorter:
```suggestion
                    Number of tags must match number of search spaces but
```

trieste/space.py
```diff
@@ -329,3 +347,160 @@ def __mul__(self, other: Box) -> Box:
...
+            tf.debugging.assert_equal(
+                len(tags),
+                len(set(tags)),
+                message=f"Subspace names must be unique but received {tags}.",
+            )
+
+        self._spaces = dict(zip(tags, spaces))
+        self._tags = tuple(tags)  # avoid accidental modification by users
+
+        subspace_sizes = [self._spaces[tag].dimension for tag in list(self._tags)]
```
ChrisMorter, line 400: It's slightly confusing that `subspace_sizes` has  a different type to `self._subspace_sizes`, maybe `self._subspace_sizes` would be better name `self._subspace_sizes_by_tag`?

trieste/space.py
```diff
@@ -329,3 +347,160 @@ def __mul__(self, other: Box) -> Box:
...
+            tf.shape(value),
+            self.dimension,
+            message="value must have same dimensionality as search space",
+        )
+        value = tf.expand_dims(value, 0)
+        in_each_subspace = [
+            self._spaces[tag].__contains__(self.get_subspace_component(tag, value)[0, :])
+            for tag in self._tags
+        ]
+        return tf.math.reduce_all(in_each_subspace)
```
ChrisMorter: NIT: There's a reduce_all in the top level tf module, as well as tf.math, so this would be ever so slightly more succinct

```suggestion
        return tf.reduce_all(in_each_subspace)
```

trieste/space.py
```diff
@@ -329,3 +347,160 @@ def __mul__(self, other: Box) -> Box:
...
+    def get_subspace(self, tag: str) -> SearchSpace:
+        """Return the domain of a particular subspace."""
+        return self._spaces[tag]
+
+    def get_subspace_component(self, tag: str, values: TensorType) -> TensorType:
+        """
+        Returns the components of ``values`` lying in a particular subspace.
+
+        :param value: Points from the :class:`DecomposableSearchSpace` of shape [N,D].
+        :return: The context components of ``values`` of shape [M, D], where
```
ChrisMorter: I'm not convinced that this docstring is right, should it not be [N, Dprod] -> [N, Dsub], where Dprod is the dimension of the product space, and Dsub is the the dimension of the subspace?

trieste/space.py
```diff
@@ -329,3 +347,160 @@ def __mul__(self, other: Box) -> Box:
...
+        :raise ValueError (or tf.errors.InvalidArgumentError): If ``value`` has a different
+            dimensionality from the search space.
+        """
+
+        tf.debugging.assert_equal(
+            tf.shape(value),
+            self.dimension,
+            message="value must have same dimensionality as search space",
+        )
+        value = tf.expand_dims(value, 0)
```
ChrisMorter: NIT: Personally I find `value[tf.newaxis, ...]` a bit more readable than `tf.expand_dims(value, 0)`

ChrisMorter (approved):
LGTM

trieste/space.py
henrymoss: Never really worked out why they include it twice

trieste/space.py
henrymoss: Good spot!

trieste/space.py
```diff
@@ -51,6 +51,11 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
             dimensionality from this :class:`SearchSpace`.
         """
 
+    @property
+    @abstractmethod
+    def dimension(self) -> TensorType:
+        """The input dimensionality of this search space."""
```
apaleyes: What does "input dimensionality" mean for a space?

trieste/space.py
```diff
@@ -329,3 +347,150 @@ def __mul__(self, other: Box) -> Box:
...
+    Product :class:`SearchSpace` consisting of a product of
+    multiple :class:`SearchSpace`. This class provides functionality for
+    accessing either the resulting combined search space or each individual space.
+
+    Note that this class assumes that individual points in product spaces are
+    represented with their inputs in the same order as specified when initializing
+    the space.
+    """
+
+    def __init__(self, spaces: Sequence[SearchSpace], tags: Optional[Sequence[str]] = None):
```
apaleyes, line 366: Since this is very dict-like structure, how about class method `from_dictionary(spaces: OrderedDict[str, SearchSpace])`? I suspect it will be easier for users to specify spaces in a dict in practice. And you can still have the default behavior as discussed before

trieste/space.py
```diff
@@ -329,3 +347,150 @@ def __mul__(self, other: Box) -> Box:
...
+        :param spaces: A list of tag-space tuples summarizing the names and domains of
+            the space's subspaces.
+        :param tags: An optional list of tags giving the unique identifiers of
+            the space's subspaces.
+        :raise ValueError (or tf.errors.InvalidArgumentError): If ``spaces`` has a different
+            length to ``tags`` when ``tags`` is provided or if ``tags`` contains duplicates.
+        """
+
+        if tags is None:
+            tags = [str(i) for i in range(len(spaces))]
```
apaleyes: slightly more pythonic

```suggestion
            tags = [str(i) for i, _ in enumerate(spaces))]
```

trieste/space.py
```diff
@@ -329,3 +347,150 @@ def __mul__(self, other: Box) -> Box:
...
+                    received {len(tags)} tags and {len(spaces)} subspaces.
+                """,
+            )
+            tf.debugging.assert_equal(
+                len(tags),
+                len(set(tags)),
+                message=f"Subspace names must be unique but received {tags}.",
+            )
+
+        self._spaces = dict(zip(tags, spaces))
```
apaleyes, line 397: After seeing this line I've learned that starting from 3.7 python dicts guarantee order

apaleyes (commented):
No major comments, although apparently some suggestions from previous review still stand. Added few more, but all minor

trieste/space.py
henrymoss: will make clearer

trieste/space.py
henrymoss: its back

trieste/space.py
henrymoss: Will do in a future PR when the need arises

trieste/space.py
henrymoss: get_context_component will be used for rules that want to do different things with different subspaces, e.g. profile optimisation

trieste/space.py
henrymoss: done

trieste/space.py
henrymoss: no longer relevant with the more general class design

trieste/space.py
henrymoss: this is sorted now

trieste/space.py
henrymoss: I now list the tags in the error

trieste/space.py
henrymoss: done

apaleyes (approved):

PR #368, opened 2021-10-07
qehvi-greedy batch support

TsingQAQ:
This draft PR is a refactoring of `BatchMonteCarloExpectedHypervolumeImprovement`, the main change is now the batch points are acquired not using `batchify` but a sequential greedy approach, which mitigate the difficulty of acq optimization in case of batch size and/or input dimension is big (with the effort of more acq opt time).

The bo iteration and metric score for integration test has been changed demonstrating its effectiveness.


TsingQAQ:
These are the changes would like to made before the merge of #360, please check if these changes are not hampering the effort of avoiding retracing @uri-granta, thanks.

trieste/acquisition/function.py
```diff
@@ -1385,6 +1265,138 @@ def update_acquisition_function(
         )
 
 
+class BatchMonteCarloExpectedHypervolumeImprovement(SingleModelGreedyAcquisitionBuilder):
```
apaleyes: Would it be possible to move this class back where it used to be within this file? This diff makes it difficult to see what changes were actually made to the implementation of this acquisition.

TsingQAQ: Indeed, though it seems not very easy to move back as we have changed the super class (i.e., now it shall be defined after `SingleModelGreedyAcquisitionBuilder`), please let me know if there is some elegant way to show the changes more clearly.

i may highlight the changes made in this PR, which is actually restricted to only several lines of code:
- add `pending_points` at [1303](https://github.com/TsingQAQ/trieste/blob/e532d7212a5adfd894d48d5610f0b54cf367fc66/trieste/acquisition/function.py#L1303)/[1330](https://github.com/TsingQAQ/trieste/blob/e532d7212a5adfd894d48d5610f0b54cf367fc66/trieste/acquisition/function.py#L1330) to store those chosen yet evaluated location
- main change is: *switch [1061](https://github.com/secondmind-labs/trieste/blob/a4b86cf58781ffb2e97760f9f42bc737ee987a68/trieste/acquisition/function.py#L1061) to [1353-1359](https://github.com/TsingQAQ/trieste/blob/e532d7212a5adfd894d48d5610f0b54cf367fc66/trieste/acquisition/function.py#L1353-L1359)*: instead of doing [1359](https://github.com/TsingQAQ/trieste/blob/e532d7212a5adfd894d48d5610f0b54cf367fc66/trieste/acquisition/function.py#L1359) as before, if we have pending points, we now concat them with current candidate `at` and sample them together to get the joint distribution 

If this is still not easy to review, maybe i can temporarily move `batch_ehvi` back, which is able to = show the difference partially, please let me know which suits better.

TsingQAQ: Now it seems more easier to see the diff.

trieste/acquisition/function/multi_objective.py
```diff
@@ -279,20 +287,27 @@ def batch_ehvi(
...
 
         def gen_q_subset_indices(q: int) -> tf.RaggedTensor:
             # generate all subsets of [1, ..., q] as indices
             indices = list(range(q))
             return tf.ragged.constant([list(combinations(indices, i)) for i in range(1, q + 1)])
 
-        samples = sampler.sample(at, jitter=sampler_jitter)  # [..., S, B, num_obj]
+        if pending_points is not None:  # jointly sample pending points and new candidate
+            aug_at = tf.concat(
+                [at, tf.repeat(pending_points[tf.newaxis], tf.shape(at)[:-2], axis=0)], axis=-2
```
TsingQAQ, line 306: This is a bit dirty but the current found workaround for batch concat.

TsingQAQ:
Have just done some recheck on this functionality: found the current `batchify_joint` acquisition function optimizer already has a comparable performance against greedy batch optimization (using the same integration test in `test_multi_objective_bayesian_optimization.py`, which is up to q=4).  

Perhaps we are already satisfied with its performance and don't need a greedy batch version for qEHVI (which means we can close this PR for now)?  @henrymoss @vpicheny

PR #370, opened 2021-10-08
Add logging/tensorboard support

uri-granta:
Add tensorboard summary logging support to trieste:

* **trieste.logging** allows setting/overriding a tf.summary.SummaryWriter and an optimizatino step number.
* If a summary writer is specified, then after each optimization step, **BayesianOptimizer** logs the best new objective values, the best overall objective values, and any model-specific parameters.
* **ProbabilisticModel** has a new log method for logging model-specific parameters (and ModelStack calls it on all submodels). **GPflowPredictor** logs the kernel variance and each of lengthscales.
**EfficientGlobalOptimization**.acquire logs the maximum found acquisition function value at each step (or multiple values for greedy functions).

Things left to do, but not necessarily in this PR:

* Add more scalar logs, as well as perhaps some text logs and histogram logs of non-scalar tensors.
* Update AskTellOptimizer to support logging (involves keeping track of optimization step somehow)
* Consider what sort of _custom_ logging users might want apart from model-specific parameters.
* Support visualising computation graph?


apaleyes:
Just a high-level question - to which extent this can be supported in AskTellOptimizer?

trieste/acquisition/rule.py
```diff
@@ -174,6 +182,8 @@ def acquire(
         search_space: SP_contra,
         datasets: Mapping[str, Dataset],
         models: Mapping[str, ProbabilisticModel],
+        step_number: int,
+        summary_writer: Optional[tf.summary.SummaryWriter] = None,
```
hstojic: it seems like it should be possible to specify a log method like for the models, we would need to save some useful results that we want to log in variables that we initialize at init, and which `acquire` saves into

hstojic (commented):
looks like a good start!

I wonder if its too much to demand that every model specifies what information is to be logged, that will probably be highly dependent on the user, perhaps it would be possible to set up a listener in the bayesian optimisation loop, exposing model - acqusition - optimisation objects to it, and then user would pass a function that would extract wanted information from these objects; we could define several default listeners with common metrics to track

trieste/acquisition/rule.py
uri-granta: Not sure I understand what you're suggesting. Since rules only expose an acquire method (which is called very optimisation step) it fees more flexible to allow them to choose when/whether to log to tensorboard during that step (e.g. we may want to log the acquisition function mutiple times during the acquire method execution).

uri-granta:
> Just a high-level question - to which extent this can be supported in AskTellOptimizer?

Should be easy to support AskTellOptimizer too: I'll make sure to update it.

uri-granta:
> I wonder if its too much to demand that every model specifies what information is to be logged, that will probably be highly dependent on the user, perhaps it would be possible to set up a listener in the bayesian optimisation loop, exposing model - acqusition - optimisation objects to it, and then user would pass a function that would extract wanted information from these objects; we could define several default listeners with common metrics to track

It's a bit more tricky than that: since the different rules handle models and acquisition functions differently, there isn't a single logging interface that we can expose and apply. Extending model logging is easy: a user can just subclass the model class with an overridden logging method. To figure out how a user might extend logging acquisition functions (both greedy and not) or anything else it would be useful to have an example. Can you think of one?

hstojic:
> > I wonder if its too much to demand that every model specifies what information is to be logged, that will probably be highly dependent on the user, perhaps it would be possible to set up a listener in the bayesian optimisation loop, exposing model - acqusition - optimisation objects to it, and then user would pass a function that would extract wanted information from these objects; we could define several default listeners with common metrics to track
> 
> It's a bit more tricky than that: since the different rules handle models and acquisition functions differently, there isn't a single logging interface that we can expose and apply. Extending model logging is easy: a user can just subclass the model class with an overridden logging method. To figure out how a user might extend logging acquisition functions (both greedy and not) or anything else it would be useful to have an example. Can you think of one?

subclassing a model does not seem ideal, but perhaps good enough for now
TF Agents has a fairly user friendly system, although I'm not sure how exactly it works: https://www.tensorflow.org/agents/api_docs/python/tf_agents/metrics

usually the available metrics worked for us, but its easy to define additional metrics by writing a new metrics class object
perhaps its a useful case that we can emulate to some extent?

(Keras callbacks is perhaps  another example: https://keras.io/api/callbacks/, although those callbacks are used more broadly, for controlling training dynamically for example, not just the logging)

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -0,0 +1,113 @@
...
+
+# %% [markdown]
+# The data remains available when the optimization is over.
+#
+# ![TensorBoard graphs](figures/tensorboard_bo.png)
+
+# %% [markdown]
+# ## Logging additional model parameters
+#
+# When logging is enabled, trieste decides what information is interesting enough to log. This includes objective and acquisition function values, and some (but not all) model parameters. To log additional model parameters, you can define your own model subclass and override the `log` method. For example, the following GPR subclass also logs the average lengthscale at each step.
```
hstojic, line 76: its not possible to do smth similar for the acquisition function? If not I would mention it, if yes it would be good to see an example as well.

trieste/bayesian_optimizer.py
```diff
@@ -386,6 +396,21 @@ def optimize(
                     model.update(dataset)
                     model.optimize(dataset)
 
+                if self._summary_writer:
+                    with self._summary_writer.as_default():
```
hstojic: could we instead of passing `summary_writer` into all of the objects use this `with` staement when defining the BayesianOptimizer and calling optimize? I think that might work...

uri-granta: Using the with at definition time won't persist until runtime. However based on a suggestion by @apaleyes, I've changed it so that we can configure the logger globally (either using `trieste.logging.set_tensorboard_writer` or in a context using `with trieste.logging.using_tensorboard_writer`) which means we no longer need to pass the summary writers around.

trieste/logging.py
```diff
@@ -0,0 +1,56 @@
...
+    Set a :class:`~tf.summary.SummaryWriter` instance to use for logging
+    to TensorBoard, or `None` to disable.
+
+    :param summary_writer: optional summary writer instance.
+    """
+    global _TENSORBOARD_WRITER
+    _TENSORBOARD_WRITER = summary_writer
+
+
+def get_tensorboard_writer() -> Optional[tf.summary.SummaryWriter]:
```
apaleyes, line 37: That's really nice, thanks for making it available globally!

Next idea would be to maintain the set of logging meta info in similar fashion. At the moment it will only contain a step number. So in `optimizer` we could call `logging.set_step_number()` or `logging.set_optimization_meta_data`, and then whenever it is needed call it as `logging.get_step_number` or `logging.get_optimization_meta_data` (may return None if not set, or dictionary of Nones, or custom object).

That would make it not necessary to pass step number around, and also make it trivial to support step numbers in ask/tell. But there might be some good reason not to do it. What do you think?

hstojic: that's a great idea!

trieste/logging.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Returns a :class:`~tf.summary.SummaryWriter` instance to use for logging
+    to TensorBoard, or `None`.
+
+    :return: optional summary writer instance.
+    """
+    return _TENSORBOARD_WRITER
+
+
+@contextmanager
```
hstojic, line 47: have no clue what is this about - a bit of education please :)

trieste/models/gpflow/interface.py
```diff
@@ -92,19 +93,20 @@ def optimize(self, dataset: Dataset) -> None:
...
-            lengthscales = self.get_kernel().lengthscales
-            if tf.rank(lengthscales) == 0:
-                tf.summary.scalar(f"{context}.kernel.lengthscale", lengthscales)
-            elif tf.rank(lengthscales) == 1:
-                for i, lengthscale in enumerate(lengthscales):
-                    tf.summary.scalar(f"{context}.kernel.lengthscale.{i}", lengthscale)
+        summary_writer = get_tensorboard_writer()
+        if summary_writer:
+            with summary_writer.as_default(step=step_number):
+                tf.summary.scalar(f"{context}.kernel.variance", self.get_kernel().variance)
```
hstojic, line 105: at the moment the user would need to subclass the model and modify the `log` method to record something else, another approach would be that user passes some kind of ModelMetric objects that would work with self object here and extract some useful metrics, we would define some usual metrics and user could pass a list of them - custom ones would then be added bu subclassing the ModelMetric etc

not sure if this alternative is substantively different, but it seems cleaner to me, metrics related code would be encapsulated in different objects

hstojic: since this would be very generic, this method could be pushed up to `TrainableProbabilisticModel`, different model classes would then have various useful metrics defined that a user could use, say in gpflow/metrics.py

trieste/bayesian_optimizer.py
```diff
@@ -154,11 +156,16 @@ class BayesianOptimizer(Generic[SP]):
...
+    def __init__(
+        self,
+        observer: Observer,
+        search_space: SP,
+    ):
         """
         :param observer: The observer of the objective function.
         :param search_space: The space over which to search. Must be a
             :class:`~trieste.space.SearchSpace`.
+        :param summary_writer: An optional summary writer object for logging to.
```
hstojic: leftover, no?

trieste/bayesian_optimizer.py
```diff
@@ -386,6 +395,22 @@ def optimize(
                     model.update(dataset)
                     model.optimize(dataset)
 
+                summary_writer = get_tensorboard_writer()
+                if summary_writer:
+                    with summary_writer.as_default():
+                        for tag in datasets:
+                            models[tag].log(step, f"{tag}.model")
+                            tf.summary.scalar(
```
hstojic, line 403: following my comment on logging related to models in gpflow/interface.py, for optimization here we could again apply a list of (potentially user defined) metrics, we would then define these as say OptimizationMetrics, as a collection of useful default metrics

trieste/logging.py
uri-granta: This is a decorator for easily defining context managers (see https://docs.python.org/3/reference/datamodel.html#context-managers). It means we can write:

```python
with using_tensorboard_writer(summary_writer):
    ....
```

and the indented code will use the given summary writer.

trieste/models/gpflow/interface.py
uri-granta: I'm wary of making an abstract Metrics class straight away, before we've had a chance to figure out 

1. the full range of metrics we're actually interested in extracting (at the moment we're just logging scalar step-based metrics to tensorboard)
2. a wider range of places where we might want to generate metrics (at the moment we do it just in the bayes opt loop, acquisition rule and model)

Maybe I'll create an issue to go back and refactor once we have a better idea?

trieste/logging.py
hstojic: Thanks!

trieste/models/gpflow/interface.py
hstojic: sounds good

docs/tutorials.rst
```diff
@@ -44,6 +44,7 @@ The following tutorials (or sections thereof) explain how to use and extend spec
 * :doc:`How do I set up a batch Bayesian optimization routine?<notebooks/batch_optimization>`
 * :ref:`How do I make a custom acquisition function?<notebooks/failure_ego:Create a custom acquisition function>`
 * :doc:`How do I recover a failed optimization loop?<notebooks/recovering_from_errors>`
+* :doc:`How do I track and visualize an optimization loop in TensorBoard?<notebooks/visualizing_with_tensorboard>`
```
hstojic: I would probably add smth like tracking "in real time"...

tests/integration/test_bayesian_optimization.py
```diff
@@ -166,28 +168,32 @@ def build_model(data: Dataset) -> GaussianProcessRegression:
...
     model = build_model(initial_data)
 
-    dataset = (
-        BayesianOptimizer(observer, search_space)
-        .optimize(num_steps, initial_data, model, acquisition_rule)
-        .try_get_final_dataset()
-    )
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        summary_writer = tf.summary.create_file_writer(tmpdirname)
+        with tensorboard_writer(summary_writer):
```
apaleyes, line 173: Does this need to happen for each integration test we are running? I don't actually know if there is any noticeable overhead for additional I/O. But if there is, perhaps having a separate integ test that just tests tensorboard logging once would suffice?

tests/unit/test_logging.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2021 The Trieste Contributors
```
apaleyes, line 1: Tests for step number too?

trieste/acquisition/rule.py
```diff
@@ -594,6 +612,13 @@ def state_func(
                     new_optimization_step=False,
                 )
                 new_point = self._optimizer(search_space, self._acquisition_function)
+                if summary_writer:
+                    with summary_writer.as_default(step=step_number):
+                        batched_point = tf.expand_dims(new_point, axis=0)
+                        value = self._acquisition_function(batched_point)[0][0]
+                        tf.summary.scalar(
+                            f"AsyncGreedy.acquisition_function.maximum_found.{i}", value
```
apaleyes, line 620: Following the EGO example, this shall be
```suggestion
                            f"AsyncGreedy.acquisition_function.maximum_found.{i+1}", value
```

trieste/ask_tell_optimization.py
```diff
@@ -260,6 +260,7 @@ def ask(self) -> TensorType:
         #   which, when called, returns state and points
         # so code below is needed to cater for both cases
 
+        # TODO: handle step number for logging?
```
apaleyes: With configurable step number now we can just off-load that to the user. Some explanation in the docstring of the optimizer would suffice.

trieste/acquisition/rule.py
```diff
@@ -218,6 +227,12 @@ def acquire(
                 chosen_point = self._optimizer(search_space, self._acquisition_function)
                 points = tf.concat([points, chosen_point], axis=0)
 
+                if summary_writer:
+                    with summary_writer.as_default(step=step_number):
+                        batched_points = tf.expand_dims(chosen_point, axis=0)
+                        value = self._acquisition_function(batched_points)[0][0]
+                        tf.summary.scalar(f"EGO.acquisition_function.maximum_found.{i+1}", value)
```
apaleyes, line 234: Feels like the prefix here shall somehow specify the fact that we are in the middle of the batch. Perhaps `...maximum_found.batch.{i+1}`?

trieste/models/interfaces.py
```diff
@@ -99,6 +99,14 @@ def get_kernel(self) -> gpflow.kernels.Kernel:
         """
         raise NotImplementedError(f"Model {self!r} does not provide a kernel")
 
+    def log(self, context: str) -> None:
```
apaleyes, line 102: That's really minor, and probably down to taste, but for me "log" normally means text-based logging, and this feel more like reporting metrics.

apaleyes (commented):
That looks really nice. Several comments, but nothing major

tests/integration/test_bayesian_optimization.py
uri-granta: There's no significant time overhead (it just logs a small number of values to file)

trieste/models/interfaces.py
uri-granta: What would you suggest calling it? At the moment it's as genera as users want it to be, though I expect after we play with this for a while we may end up refactoring out some metric specific logic like hrvoje suggested.

apaleyes: I am not sure if it is useful to abstract to any kind of logging here. E.g. with usual logging it isn't uncommon to do thing like
```
logger.log(f"Calling optimizer for {self.model} and dataset {dataset}", DEBUG)
```
so some very low level stuff that one wouldn't want to send to tensorboard. However we are not using Python logging (yet?), so all that is just hypothetical. Whereas this `model.log` has very specific purpose of writing numerical data to graph. Unless I misunderstood it, which is entirely possible of course.

I guess my suggestion here is to be a bit more specific with this method. `(log|write)_to_tensorboard`?

trieste/acquisition/rule.py
uri-granta: EGO (where we don't always have batches) numbers them "maximum_found", "maximum_found.1", "maximum_found.2", etc.

AsyncGreedy (where we do always have batches) numbers them "maximum_found.0", "maximum_found.1", "maximum_found.2", etc.

I don't think we should worry too much about this until people start giving using it and giving feedback, as we may end up changing the hierarchy of metrics much more significantly.

trieste/models/interfaces.py
hstojic: actually, files that are saved don't necessarily need to be used by tensorboard, you could extract the data and make your own plots etc, so I think it can stay more general, but I agree with Andrei that there might be a confusion with usual python logging which is not the intended purpose - I would vote for metrics

trieste/acquisition/rule.py
apaleyes: fair enough

hstojic:
what would be useful is to also record a configuration of the algorithm (perhaps also toolbox version, not sure what to use here, commit hash?), so if we have several experiment files in the same summary folder (tensorboard will automatically read them in and we can compare them), it will be easy to differentiate them

trieste/models/interfaces.py
uri-granta: model.log doesn't have to be used just to write metrics to tensorboard: you could use it for any sort of model logging (in the broad sense) that you want to happen after every BO loop. Though as I said, I expect we'll refactor the code soon to handling metrics more explicit, at which point this will be renamed.

apaleyes: ok, let's leave it for now

apaleyes (approved):
this looks like a great start, keeping in mind that some changes might occur when people start using it

trieste/models/interfaces.py
uri-granta: Happy to rename it though. Though it should be a verb, as it's writing metrics, not returning them. `write_metrics`?

uri-granta: `generate_metrics`? `save_metrics`?

uri-granta:
> what would be useful is to also record a configuration of the algorithm (perhaps also toolbox version, not sure what to use here, commit hash?), so if we have several experiment files in the same summary folder (tensorboard will automatically read them in and we can compare them), it will be easy to differentiate them

Great suggestion, though I'd rather do that in a subsequent PR.

trieste/models/interfaces.py
hstojic: write_metrics is more precise

hstojic (approved):
it looks ready for a test drive, lets just raise an issue with some tentative ideas for future development, so we keep a nicer record and don't forget coming back to this

Update: I gave few more comments for improving the notebook

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -0,0 +1,124 @@
...
+
+# %% [markdown]
+# ## Using Tensorboard with Ask-Tell Optimization
+#
+# To use Tensorboard logging with the [Ask-Tell interface](ask_tell_optimization.ipynb), you must also explicitly set the optimization step number before calling ask or tell:
+
+# %%
+for step in range(num_steps):
+    trieste.logging.set_step_number(step)
+    # new_point = ask_tell.ask()
```
hstojic: this is still commented out? you need to import asktell as well...

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -0,0 +1,124 @@
...
+# %% [markdown]
+# ## Setting up TensorBoard
+#
+# Before running the optimization loop, we decide where TensorBoard summary logs should be stored, and set a summary writer to do this.
+
+# %%
+# Clear any logs from previous runs
+# !rm -rf logs/tensorboard
+
+summary_writer = tf.summary.create_file_writer("logs/tensorboard")
```
hstojic: this would show how you can get automatic comparison of several different runs/experiments
```suggestion
summary_writer = tf.summary.create_file_writer("logs/tensorboard/exp1")
```

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -0,0 +1,124 @@
...
+                tf.summary.scalar(
+                    f"{context}.kernel.lengthscales.mean",
+                    np.mean(self.get_kernel().lengthscales)
+                )
+
+
+model = GPRExtraLogging(gpr)
+
+# %% [markdown]
+# Running with this model now also produces logs for the mean lengthscale.
```
hstojic:
```suggestion
# Running with this model now also produces logs for the mean lengthscale. We mark this optimization run as a separate experiment by creating a new summary writer, Tensorboard will conveniently offer an automatic comparison of experiments.
```

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -0,0 +1,124 @@
...
+
+# %%
```
hstojic, line 102:
```suggestion
# %%
summary_writer = tf.summary.create_file_writer("logs/tensorboard/exp2")
trieste.logging.set_tensorboard_writer(summary_writer)

```

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -0,0 +1,124 @@
...
+
+# %% [markdown]
+# ![TensorBoard custom graphs](figures/tensorboard_custom.png)
+
+# %% [markdown]
+# ## Using Tensorboard with Ask-Tell Optimization
+#
+# To use Tensorboard logging with the [Ask-Tell interface](ask_tell_optimization.ipynb), you must also explicitly set the optimization step number before calling ask or tell:
+
+# %%
```
hstojic, line 117:
```suggestion
# %%
summary_writer = tf.summary.create_file_writer("logs/tensorboard/exp3")
trieste.logging.set_tensorboard_writer(summary_writer)

```

hstojic:
> > what would be useful is to also record a configuration of the algorithm (perhaps also toolbox version, not sure what to use here, commit hash?), so if we have several experiment files in the same summary folder (tensorboard will automatically read them in and we can compare them), it will be easy to differentiate them
> 
> Great suggestion, though I'd rather do that in a subsequent PR.

Fair enough, lets add this in the issue

docs/notebooks/visualizing_with_tensorboard.pct.py
uri-granta: this is just an example rather than full running code (I'll add a comment to make that clear)

uri-granta: actually, i'll turn it into a working example (and use a new summary writer too)

PR #372, opened 2021-10-12
Decomposing models - optimizers leftovers

hstojic:
This PR is tying up some final pieces for refactoring the `models` package to support multiple models more easily. It also replaces PR #343.

I would welcome some opinions. There are few points in particular:
- `Optimizer` and `BatchOptimizer` (ex `TFOptimizer`) are currently used by GPflow only so we could make them GPflow specific. However, these are fairly general classes and most likely other models would use them (e.g. some additional model implemented by a user), so perhaps it makes sense to leave them in their current location, shared by all models.  
- To me it doesn't seem like we need an interface. One reason for an interface would be to simplify configs, as otherwise models would need to override some config methods (e.g. GPflux does that). On the other hand, models consume these optimizers and they tend to do it in idiosyncratic ways, e.g. in case of Keras it would be cumbersome to use optimizers via `Optimizer` class.   

There are some small fixes to gpflow tests and minor change in gpflux default optimizers


tests/unit/models/gpflow/test_models.py
```diff
@@ -423,15 +425,15 @@ def test_gaussian_process_regression_default_optimize(
 
 
 @random_seed
-@pytest.mark.parametrize("optimizer", [gpflow.optimizers.Scipy(), tf.optimizers.Adam(), None])
```
hstojic, line 426: Not sure why we were testing `None` here, optimizer is not an optional argument in `Optimizer`, the only reason why it was passing was that `create_optimizer` function didn't have `NotImplementedError` as a default

trieste/models/config.py
```diff
@@ -61,6 +61,14 @@ def supported_models(
         """
         raise NotImplementedError
 
+    def create_optimizer(self) -> Optimizer:
```
hstojic, line 64: added a method to make it easier for model-specific configs to override the default one (e.g. GPflux case, Keras probably as well)

trieste/models/gpflow/__init__.py
```diff
@@ -20,4 +20,9 @@
 from .config import GPflowModelConfig
 from .interface import GPflowPredictor
 from .models import GaussianProcessRegression, SparseVariational, VariationalGaussianProcess
+# from .optimizer import (
```
hstojic: this is a current solution for the issue with importing `create_optimizer` and `create_loss_function` with the singledispatch decorator from `models/optimizers.py`, we don't want to export these internal functions, but it seems the file is not executed at all otherwise ( because we are explicitly importing things in this file, rather than leaving it empty, right?)

trieste/models/gpflow/optimizer.py
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Any, Dict
+
+import gpflow
+from gpflow.models import ExternalDataTrainingLossMixin, InternalDataTrainingLossMixin
+
+from ..optimizer import LossClosure, Optimizer, TrainingData, create_loss_function, create_optimizer
```
hstojic, line 27: importing `create_optimizer` and `create_loss_function` with the singledispatch decorator doesn't seem to work (had to export these internal functions explicitly in `__init__.py`), I also tried decorating them again here but that didn't help, any thoughts?

trieste/models/gpflux/interface.py
```diff
@@ -30,6 +30,26 @@ class GPfluxPredictor(ProbabilisticModel, tf.Module, ABC):
...
     dtype errors."""
 
+    def __init__(self, optimizer: tf.optimizers.Optimizer | None = None):
+        """
+        :param optimizer: The optimizer with which to train the model. Defaults to
+            :class:`~tf.optimizers.Optimizer` with :class:`~tf.optimizers.Adam`.
+        """
+        super().__init__()
+
+        if optimizer is None:
```
hstojic, line 40: moved default optimizer definition to the interface level rather than the model level

trieste/models/optimizer.py
```diff
@@ -41,27 +46,31 @@
...
 """
 
 
 @dataclass
 class Optimizer:
     """Optimizer for training models with all the training data at once."""
 
-    optimizer: gpflow.optimizers.Scipy | tf.optimizers.Optimizer
-    """ The underlying optimizer to use. """
+    optimizer: Any
```
hstojic, line 59: seems better to rely on TF as a default here, but not sure if we should specify `tf.optimizers.Optimizer` type, that would force models to subclass `Optimizer` and change the default to Scipy optimizer, right? 

also, I think all optimizers in `tf.optimizers.Optimizer` are meant to be used on mini-batches, not on the whole dataset - from that perspective, this class could be in GPflow only

trieste/models/optimizer.py
```diff
@@ -86,27 +95,31 @@ def optimize(self, model: tf.Module, dataset: Dataset) -> OptimizeResult:
 
 
 @dataclass
-class TFOptimizer(Optimizer):
+class BatchOptimizer(Optimizer):
```
hstojic, line 100: seems like a more suitable name...

trieste/models/optimizer.py
hstojic: on aside, why GPflow has its own scipy optimizer, is it any different than the actual scipy one?

trieste/models/optimizer.py
```diff
@@ -139,65 +152,51 @@ def train_fn() -> None:
...
+    decorator together with its register method to make a (model-)specific optimizer available.
 
-    :param optimizer: The optimizer with which to train the model.
-    :param optimizer_args: The keyword arguments to pass to the optimizer wrapper..
+    :param optimizer: The optimizer with which to train the model. Note that a flexible type
+        `Any` is used to allow for vartious optimizers that specific models might need to use.
+    :param optimizer_args: The keyword arguments to pass to the optimizer wrapper.
     :return: The :class:`Optimizer` wrapper.
     """
+    raise NotImplementedError(f"Unknown optimizer {optimizer} passed for creating an optimizer")
```
hstojic, line 171: anyone knows why we were not raising `NotImplementedError` previously?
on the other hand, now with `Any` type does this make any sense? perhaps yes, as it forces you to register at least one instance and gives a nice error for passing `None`, since `Optimizer` does not allow `None`

trieste/models/optimizer.py
```diff
@@ -139,65 +152,51 @@ def train_fn() -> None:
...
     optimizer_args: Dict[str, Any],
 ) -> Optimizer:
-    return Optimizer(optimizer, **optimizer_args)
+    return BatchOptimizer(optimizer, **optimizer_args)
 
 
 @singledispatch
 def create_loss_function(
-    model: gpflow.models.GPModel, dataset: TrainingData, compile: bool = False
+    model: tf.Module, dataset: TrainingData, compile: bool = False
```
hstojic: models here are our wrappers, so this was incorrect previously

trieste/models/optimizer.py
```diff
@@ -139,65 +153,43 @@ def train_fn() -> None:
...
 ) -> LossClosure:
     """
     Generic function for building a loss function for a specified `model` and `dataset`.
-    The implementations depends on the type of the model.
+    The implementations depends on the type of the model, which should use this function as a
```
apaleyes, line 188:
```suggestion
    The implementations depend on the type of the model, which should use this function as a
```

trieste/models/optimizer.py
apaleyes: Then why not use TrainableProbabilisticModel or some other wrapper class?

trieste/models/optimizer.py
```diff
@@ -139,65 +153,43 @@ def train_fn() -> None:
 
 @singledispatch
 def create_optimizer(
-    optimizer: gpflow.optimizers.Scipy | tf.optimizers.Optimizer,
+    optimizer: Any,
     optimizer_args: Dict[str, Any],
 ) -> Optimizer:
     """
     Generic function for creating a :class:`Optimizer` wrapper from a specified
     `optimizer` and `optimizer_args`. The implementations depends on the type of the
```
apaleyes:
```suggestion
    `optimizer` and `optimizer_args`. The implementations depend on the type of the
```

trieste/models/optimizer.py
```diff
@@ -139,65 +153,43 @@ def train_fn() -> None:
...
-    underlying optimizer.
+    underlying optimizer, which can be model-specific. This function needs to be used as a
+    decorator together with its register method to make a (model-)specific optimizer available.
 
-    :param optimizer: The optimizer with which to train the model.
-    :param optimizer_args: The keyword arguments to pass to the optimizer wrapper..
+    :param optimizer: The optimizer with which to train the model. Note that a flexible type
+        `Any` is used to allow for vartious optimizers that specific models might need to use.
```
apaleyes:
```suggestion
        `Any` is used to allow for various optimizers that specific models might need to use.
```

trieste/models/optimizer.py
apaleyes: That's a good catch about not implemented. But not sure about `Any`. Why does the type need to change at all?

trieste/models/optimizer.py
```diff
@@ -86,27 +96,31 @@ def optimize(self, model: tf.Module, dataset: Dataset) -> OptimizeResult:
...
+
         def creator_fn(data: TrainingData) -> LossClosure:
             return create_loss_function(model, data, self.compile)
 
-        if self.dataset_builder is None and self.batch_size is None:
-            x = tf.convert_to_tensor(dataset.query_points)
-            y = tf.convert_to_tensor(dataset.observations)
-            return creator_fn((x, y))
-        elif self.dataset_builder is None:
+        if self.dataset_builder is None:
```
apaleyes, line 124: Why was the not None case removed?

trieste/models/optimizer.py
apaleyes: looks good!

trieste/models/optimizer.py
apaleyes: Again, not sure I agree that these types had to change to Any. What is not covered between `gpflow.optimizers.Scipy | tf.optimizers.Optimizer`, what additional flexibility do we need? I might be asking out of ignorance here actually

apaleyes (commented):
Left some comments below, but some things are not very clear. I guess I just lack context of previous PRs on model decomposition

trieste/models/optimizer.py
hstojic: true enough, thanks

trieste/models/optimizer.py
hstojic: There are a lot more types of optimizers that are not covered here, e.g. MCMC, Particle filters, evolutionary algorithms. I'm not saying you would use these with say GPs, but as a user you might implement an additional model which needs a different optimizer. My thinking was that `Any` then allows passing whatever optimizer, rather than subclassing `Optimizer` and setting the optimizer to your type. I also tried to contain usage of gpflow within `models/gpflow` sub-package, it made sense to me, but perhaps there is not a strong argument for that.

trieste/models/optimizer.py
hstojic: see my comment above

trieste/models/optimizer.py
hstojic: it seemed superflous given that `Optimizer` class was doing that - if you don't specify a batch size or dataset builder you fall back to it, but then why having a mini-batch specific version, if we have that part then we might as well collapse Optimizer and BatchOptimizer into a single class - what do you think?

trieste/models/optimizer.py
apaleyes: that change makes more sense now, thanks!

trieste/models/optimizer.py
uri-granta: `NotImplementedError` wasn't required before since there were overloads for all the supported types (SciPy and Optimizer). Changing it to `Any` means you now need it.

tests/unit/models/gpflow/conftest.py
```diff
@@ -38,15 +38,15 @@
 
 
 @pytest.fixture(
-    name="gpr_interface_factory",
+    name="gpflow_interface_factory",
```
uri-granta, line 41: (thanks for this!)

uri-granta (approved):
LGTM! Agree on leaving Optimizer and BatchOptimizer where they are (for now at least).

PR #374, opened 2021-10-13
Non-greedy async optimization

apaleyes:
This PR contains implementation for a second rule for async optimization, conveniently named AsynchronousOptimization. This isn't expected to be high-performant (and we warn users in docs about it), but it gives access to non-greedy batch functions in async mode.

Although not a part of the PR, new rule was verified to run in the async notebook in addition to all tests.

Plus there is a bit of a clean up for blunders I made in #366 .

### DIFF
--- a/trieste/acquisition/rule.py
+++ b/trieste/acquisition/rule.py
@@ -368,6 +368,7 @@ class AsynchronousOptimization(
 
             if state.has_pending_points:
                 pending_points: TensorType = state.pending_points
+
                 def function_with_pending_points(x: TensorType) -> TensorType:
                     # stuff below is quite tricky, and thus deserves an elaborate comment
                     # we receive unknown number N of points to evaluate
--- a/trieste/acquisition/rule.py
+++ b/trieste/acquisition/rule.py
@@ -218,7 +218,7 @@ class AsynchronousRuleState:
     These are points which were requested but are not observed yet.
     """
 
-    pending_points: TensorType
+    pending_points: Optional[TensorType] = None
 
     def __post_init__(self) -> None:
         if self.pending_points is None:
@@ -420,10 +420,9 @@ class AsynchronousOptimization(
                     all_points = tf.concat([pending_points_repeated, x], axis=1)
                     return cast(AcquisitionFunction, self._acquisition_function)(all_points)
 
-                acquisition_function = function_with_pending_points
+                acquisition_function = cast(AcquisitionFunction, function_with_pending_points)
             else:
-                # mypy gets confused about types of these functions
-                acquisition_function = self._acquisition_function  # type: ignore
+                acquisition_function = cast(AcquisitionFunction, self._acquisition_function)
 
             new_point = self._optimizer(search_space, acquisition_function)
             state = state.add_pending_points(new_point)
--- a/trieste/acquisition/rule.py
+++ b/trieste/acquisition/rule.py
@@ -236,31 +236,48 @@ class AsynchronousRuleState:
         """Returns `True` if there is at least one pending point, and `False` otherwise."""
         return (self.pending_points is not None) and tf.size(self.pending_points) > 0
 
-    def subtract_points(self, other_points: TensorType) -> AsynchronousRuleState:
-        """Removes all rows from current `pending_points` that are present in `other_points`.
+    def remove_points(self, points_to_remove: TensorType) -> AsynchronousRuleState:
+        """Removes all rows from current `pending_points` that are present in `points_to_remove`.
+        If a point to remove occurs multiple times in the list of pending points,
+        only first occurrence of it will be removed.
 
-        :param other_points: Points to remove.
+        :param points_to_remove: Points to remove.
         :return: New instance of `AsynchronousRuleState` with updated pending points.
         """
+
+        def _remove_point(pending_points: TensorType, point_to_remove: TensorType) -> TensorType:
+            # find indexes of all occurrences of the point to remove
+            indexes_of_point = tf.where(
+                tf.reduce_all(tf.equal(pending_points, point_to_remove), axis=1)
+            )
+            if tf.equal(tf.size(indexes_of_point), 0):
+                return pending_points
+
+            # we only want to remove first occurrence
+            # and know that shape of indexes is [N, 1]
+            # that allows us to do this trick below
+            first_index = indexes_of_point[0, 0]
+            return tf.concat(
+                [pending_points[:first_index, :], pending_points[first_index + 1 :, :]], axis=0
+            )
+
         if not self.has_pending_points:
             # nothing to do if there are no pending points
             return self
 
         tf.debugging.assert_shapes(
-            [(self.pending_points, [None, "D"]), (other_points, [None, "D"])],
-            message=f"""Other points shall be 2D and have same last dimension as pending points.
+            [(self.pending_points, [None, "D"]), (points_to_remove, [None, "D"])],
+            message=f"""Point to remove shall be 1xD where D is the last dimension of pending points.
                         Got {tf.shape(self.pending_points)} for pending points
-                        and {tf.shape(other_points)} for other points.""",
+                        and {tf.shape(points_to_remove)} for other points.""",
         )
 
-        @tf.function
-        def is_not_in_points(x: TensorType) -> TensorType:
-            return tf.reduce_all(tf.reduce_any(tf.not_equal(other_points, x), axis=1))
+        new_pending_points = tf.foldl(
+            _remove_point, points_to_remove, initializer=self.pending_points
+        )
+        return AsynchronousRuleState(new_pending_points)
 
-        mask = tf.map_fn(is_not_in_points, self.pending_points, fn_output_signature=tf.bool)
-        return AsynchronousRuleState(tf.boolean_mask(self.pending_points, mask))
-
-    def add_points(self, new_points: TensorType) -> AsynchronousRuleState:
+    def add_pending_points(self, new_points: TensorType) -> AsynchronousRuleState:
         """Adds `new_points` to the already known pending points.
 
         :param new_points: Points to add.
@@ -295,7 +312,10 @@ class AsynchronousOptimization(
     We call such points "pending", and consider them a part of acquisition state.
     We use :class:`AsynchronousRuleState` to store these points.
 
-    AsynchronousOptimization works with non-greedy batch acquisition functions.
+    `AsynchronousOptimization` works with non-greedy batch acquisition functions.
+    For example, it would work with
+    :class:`~trieste.acquisition.BatchMonteCarloExpectedImprovement`,
+    but cannot be used with :class:`~trieste.acquisition.ExpectedImprovement`.
     If there are P pending points, the acquisition function is used with batch size P+1.
     During optimization first P points are fixed to pending, and thus optimization
     is done over the last point only, which is then returned.
@@ -378,7 +398,7 @@ class AsynchronousOptimization(
             if state is None:
                 state = AsynchronousRuleState(None)
 
-            state = state.subtract_points(datasets[OBJECTIVE].query_points)
+            state = state.remove_points(datasets[OBJECTIVE].query_points)
 
             if state.has_pending_points:
                 pending_points: TensorType = state.pending_points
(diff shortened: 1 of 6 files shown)


trieste/acquisition/rule.py
```diff
@@ -211,31 +212,208 @@ def acquire(
...
+        if not self.has_pending_points:
+            return AsynchronousRuleState(new_points)
+
+        tf.debugging.assert_shapes(
+            [(self.pending_points, [..., "D"]), (new_points, [..., "D"])],
+            message=f"""Last dimension mismatch between pending points
+                        {tf.shape(self.pending_points)} and given points {tf.shape(new_points)}""",
+        )
+
+        new_pending_points = tf.concat([self.pending_points, new_points], axis=0)
```
vpicheny, line 296: shouldn't we also assert the rest of the shapes? otherwise this may error without useful message?

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,208 @@ def acquire(
...
 
     To make the best decision about next point to observe, acquisition function
     needs to be aware of currently running observations.
-    We call such points "pending". AsyncEGO provides a ``AsynchronousGreedy.State``
-    object that keeps track of pending points.
+    We call such points "pending", and consider them a part of acquisition state.
+    We use :class:`AsynchronousRuleState` to store these points.
+
+    AsynchronousOptimization works with non-greedy batch acquisition functions.
+    If there are P pending points, the acquisition function is used with batch size P+1.
```
vpicheny, line 319: actually we can set P+B if we want, right?

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,208 @@ def acquire(
...
-    @dataclass(frozen=True)
-    class State:
-        pending_points: TensorType
+    def __init__(
+        self,
+        builder: Optional[AcquisitionFunctionBuilder | SingleModelAcquisitionBuilder] = None,
+        optimizer: AcquisitionOptimizer[SP_contra] | None = None,
+    ):
+        """
+        :param builder: Batch acquisition function builder. Defaults to
```
vpicheny, line 333: This code only works if the acquisition function can deal with batches. For instance, this will not work with ExpectedImprovement. Not sure how clear this fact is from the docstring. Also, could we have a way to check this early?

trieste/acquisition/rule.py
apaleyes: This was also discussed in previous PR, where we actually removed batch size parameter. There is no real use for it in async scenarios, you can just ask for the next point as many time as you'd like and all previous pending points will be taken into account. Also, normally you ask for a point because your worker just finished working on the previous observation, so there is no need for more than one anyway.

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,208 @@ def acquire(
...
+
+        def state_func(
+            state: AsynchronousRuleState | None,
+        ) -> tuple[AsynchronousRuleState | None, TensorType]:
+            tf.debugging.Assert(self._acquisition_function is not None, [])
+
+            if state is None:
+                state = AsynchronousRuleState(None)
+
+            state = state.subtract_points(datasets[OBJECTIVE].query_points)
```
vpicheny: I'm not sure this covers the corner case where we have duplicated observations. In that case we do want to remove the pending points that have just been queried but not the pending points that have duplicates in the past.

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,208 @@ def acquire(
...
+
+            if state.has_pending_points:
+                pending_points: TensorType = state.pending_points
+
+                def function_with_pending_points(x: TensorType) -> TensorType:
+                    # stuff below is quite tricky, and thus deserves an elaborate comment
+                    # we receive unknown number N of points to evaluate
+                    # note that we don't deal with batches here
+                    # so the shape of `x` is [N, 1, D]
```
vpicheny, line 410: do we really not deal with batches? we could have a case where the user wants to wait until he's got at least 2 workers free to choose the next batch of 2 points?

trieste/acquisition/rule.py
apaleyes: Docstring says "AsynchronousOptimization works with non-greedy batch acquisition functions." Shall we add something else?

For validation, we don't have a way in Trieste to validate if a function is a batch or non-batch. So the only way to do early validation is to compare against concrete classes. This is ugly and bound to break when we have more batch acquisitions, so for now we rely on acquisition function failing when called with more than one point.

trieste/acquisition/rule.py
apaleyes: why would we want to keep duplicate points? and why would duplicates arise in the first place?

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,208 @@ def acquire(
...
+        new_pending_points = tf.concat([self.pending_points, new_points], axis=0)
+        return AsynchronousRuleState(new_pending_points)
+
+
+class AsynchronousOptimization(
+    AcquisitionRule[State[Optional["AsynchronousRuleState"], TensorType], SP_contra]
 ):
-    """AsynchronousGreedy rule, as name suggests,
-    is designed for asynchronous BO scenarios.
+    """AsynchronousOptimization rule is designed for asynchronous BO scenarios.
```
hstojic, line 303: can we point the users to some useful references?

apaleyes: Sure will add a reference or two

trieste/acquisition/rule.py
apaleyes: good point, will do. this also needs post-init validation of pending points.

trieste/acquisition/rule.py
apaleyes: After an offline discussion we agreed to re-introduce batch sizes to both async rules in a separate PR

trieste/acquisition/rule.py
vpicheny: maybe something to think about in the future... maybe for now just add that it will fail with standard EI and work with MCEI?

trieste/acquisition/rule.py
vpicheny: if you have noise, it can make sense to choose to sample twice the same point. In some cases, the maximiser of the acquisition function can be an existing point. 
I think that making sure that we don't remove more points than the newly acquired ones should do the trick. E.g. if you have 3 identical pending points and you get 1 back, it doesn't matter which pending point you're substracting but you should not remove more than one.

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,222 @@ def acquire(
...
+        )
+
+        @tf.function
+        def is_not_in_points(x: TensorType) -> TensorType:
+            return tf.reduce_all(tf.reduce_any(tf.not_equal(other_points, x), axis=1))
+
+        mask = tf.map_fn(is_not_in_points, self.pending_points, fn_output_signature=tf.bool)
+        return AsynchronousRuleState(tf.boolean_mask(self.pending_points, mask))
+
+    def add_points(self, new_points: TensorType) -> AsynchronousRuleState:
```
vpicheny: `add_pending_points`?

vpicheny (commented):
we're going to want to document this in a notebook, but I guess this is another PR?

apaleyes:
Yes, the notebook is in process. I'd prefer to make a separate PR for it, to make these PRs reasonably sized. Smaller and focused PRs are easier to review.

trieste/acquisition/rule.py
apaleyes: done

trieste/acquisition/rule.py
apaleyes: this is now implemented, but turned out to be very hard to do efficiently, so i suspect current implementation may trigger tf graph evaluation prematurely. If anyone has better ideas on how to implement it, please suggest

trieste/acquisition/rule.py
apaleyes: sure

trieste/acquisition/rule.py
```diff
@@ -153,7 +154,7 @@ def __init__(
         ):
             builder = builder.using(OBJECTIVE)
 
-        if isinstance(builder, AcquisitionFunctionBuilder):
+        if isinstance(builder, AcquisitionFunctionBuilder) and num_query_points > 1:
```
uri-granta, line 157: :+1:

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,242 @@ def acquire(
...
+            message=f"""Pending points are expected to be a 2D tensor,
+                        instead received tensor of shape {tf.shape(self.pending_points)}""",
+        )
+
+    @property
+    def has_pending_points(self) -> bool:
+        """Returns `True` if there is at least one pending point, and `False` otherwise."""
+        return (self.pending_points is not None) and tf.size(self.pending_points) > 0
+
+    def remove_points(self, points_to_remove: TensorType) -> AsynchronousRuleState:
```
uri-granta, line 239: If you think it's a good idea, you could also define `__add__` and `__sub__` so you can write `state += points` and `state -= points`.

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,242 @@ def acquire(
...
 
-class AsynchronousGreedy(
-    AcquisitionRule[State[Optional["AsynchronousGreedy.State"], TensorType], SP_contra]
+@dataclass(frozen=True)
+class AsynchronousRuleState:
+    """Stores pending points for asynchronous rules.
+    These are points which were requested but are not observed yet.
+    """
+
+    pending_points: TensorType
```
uri-granta: Definitely mention that it can be `None` in the type. Whether to make it optional is up to you.

```suggestion
    pending_points: Optional[TensorType] = None
```

trieste/acquisition/rule.py
```diff
@@ -297,38 +509,26 @@ def acquire(
...
-                pending_points = points
-            else:
-                pending_points = tf.concat([pending_points, points], axis=0)
-
-            state = AsynchronousGreedy.State(pending_points=pending_points)
+            new_point = self._optimizer(search_space, self._acquisition_function)
+            state = state.add_pending_points(new_point)
 
-            return state, points
+            return state, new_point
```
uri-granta, line 530: cleaner

trieste/acquisition/rule.py
```diff
@@ -211,31 +212,242 @@ def acquire(
...
+                    # pending points are 2D, we need 3D and repeat along first axis
+                    expanded = tf.expand_dims(pending_points, axis=0)
+                    pending_points_repeated = tf.repeat(expanded, [tf.shape(x)[0]], axis=0)
+                    all_points = tf.concat([pending_points_repeated, x], axis=1)
+                    return cast(AcquisitionFunction, self._acquisition_function)(all_points)
+
+                acquisition_function = function_with_pending_points
+            else:
+                # mypy gets confused about types of these functions
+                acquisition_function = self._acquisition_function  # type: ignore
```
uri-granta: why not cast it like above? (or alternatively, type ignore the previous cast)

(btw mypy is correct to be confused because  this is happening inside a closure and there's no guarantee that by the time we execute state_func sef._acquisition_function won't be None again)

uri-granta (commented):
Haven't looked at the tests yet, but code looks good!

trieste/acquisition/rule.py
uri-granta: (just realised this is ambiguous: I just meant the code looks cleaner now, which is good!)

trieste/acquisition/rule.py
apaleyes: Good point, but I'd like to keep state a bit more general, in case we want to add another state data to it.

trieste/acquisition/rule.py
apaleyes: Great catch, updated, as it can certainly be None

trieste/acquisition/rule.py
apaleyes: Cast here doesn't work because mypy works out some confusing type for `acquisition_function` based on assignment two lines above. I ended up casting to AcquisitionFunction on both of these lines

vpicheny (approved):

PR #377, opened 2021-10-18
Shrink DGP integ test search space

### LINKED ISSUE #375: Make deep gp integ test more stable
The `test_two_layer_dgp_optimizer_finds_minima_of_michalewicz_function` test in `test_bayesian_optimization` still fails sometimes. But always comes back when you restart it.

What can we do to make it stable? Since this is the integration test, I propose we reduce the search space to ensure it finds the minimum at `x*`. At the moment it uses `Box([0, 0], [pi, pi])` as is standard for Michalewicz functions. Perhaps for an integration test we could use `Box(x* - [0.5, 0.5], x* + [0.5, 0.5])`.

apaleyes:
This PR hopefully makes Deep GP integration test a bit more stable by bringing the search space from `([0, pi], [0, pi])` down to `global_min±0.5`.

Resolves #375

### DIFF
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -136,7 +136,6 @@ from trieste.types import State, TensorType
         ],
     ),
 )
-@pytest.mark.skip()
 def test_optimizer_finds_minima_of_the_scaled_branin_function(
     num_steps: int,
     acquisition_rule: AcquisitionRule[TensorType, SearchSpace]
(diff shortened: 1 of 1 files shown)


tests/integration/test_bayesian_optimization.py
```diff
@@ -137,6 +136,7 @@
         ],
     ),
 )
+@pytest.mark.skip()
```
hstojic: we can mark it slow rather than skip, `@pytest.mark.slow` no?

apaleyes: oh sorry, that was a debugging leftover. shouldn't be there at all, thanks for catching!

hstojic (approved):
looks good

PR #381, opened 2021-10-20
Improve function that removes pending points

apaleyes:
This PR improves code that removes first occurrence of a point from pending points in async state. The function was verified to compile as expected, and not trigger retracing warning.


trieste/acquisition/rule.py
```diff
@@ -245,18 +245,19 @@ def remove_points(self, points_to_remove: TensorType) -> AsynchronousRuleState:
...
 
+        @tf.function
         def _remove_point(pending_points: TensorType, point_to_remove: TensorType) -> TensorType:
-            # find indexes of all occurrences of the point to remove
-            indexes_of_point = tf.where(
-                tf.reduce_all(tf.equal(pending_points, point_to_remove), axis=1)
-            )
-            if tf.equal(tf.size(indexes_of_point), 0):
+            # find all points equal to the one we need to remove
+            are_points_equal = tf.reduce_all(tf.equal(pending_points, point_to_remove), axis=1)
```
hstojic, line 260: I'm not really familiar with the async BO code, but just a thought - would it make sense to create a dictonary for the pending points, assign them unique indices as keys and then removing them would be way easier, no?

apaleyes: This code would be much simpler, yes. But that would also mean that users need to keep track of those indexes, which is really inconvenient. Complexity of tracking pending points is one of few things library can actually do for async BO without being opinionated.

hstojic: ah, ok, I thought users might not need to know about the indices

apaleyes: I think they need, because once the observation comes back the user needs to know which point was it for. So users would need to do some accounting with workers and points, which sounds complicated. But I might be wrong

vpicheny (approved):

PR #382, opened 2021-10-20
notebook with data normalization

### LINKED ISSUE #379: Data preprocessing / transformations / normalisations
Support for data pre-processing / transformations / normalisation. Several common normalisations should be offered by trieste but with flexibility to define a custom transformation - this could be part of the model interface where predictions would be automatically transformed back to the non-transformed space, or `dataset_builder` in `TFOptimizer` could potentially be used as well. 

A common transformation is normalisation which is often employed to deal with numerical issues in training the models, to speed up training or improve convergence. Some models offer this out of the box, but GPflow doesn't. In regression problems its easy enough to deal with this as you have to do it once before the training, but in BayesOpt setting it needs to be done inside the optimization loop.

Current alternative is to use the Ask-Tell interface and deal with any transformation. This is OK, but suboptimal, as its a common operation for training the models. Its also not a trivial thing to implement, because of active sampling that shifts the distribution of the data.

**Notes**
- Distribution is being shifted by active sampling - important to address!
- GPflowOpt used to rescale at each iteration  
- ANy other BayesOpt toolbox dealing with this?

hstojic:
This is related to #379, showing how to data standardisation as an example of data transformation using the ask-tell interface. Until we figure out how to address the issue more completely it will be a useful resource for users.

### DIFF
--- a/docs/notebooks/data_transformation.pct.py
+++ b/docs/notebooks/data_transformation.pct.py
@@ -1,10 +1,9 @@
 # %% [markdown]
 # # Data transformation with the help of Ask-Tell interface.
 
+# %%
 import gpflow
 import matplotlib.pyplot as plt
-
-# %%
 import numpy as np
 import tensorflow as tf
 import tensorflow_probability as tfp
@@ -30,10 +29,9 @@ tf.random.set_seed(1794)
 #
 # In regression problem it is easy to perform data transformations as you do it once before the training. In Bayesian optimization this is more complex, as the data is added with each iteration and would need to be transformed again before the model is updated. At the moment Trieste cannot do such transformations for the user. Luckily, this can be easily done by using the [Ask-Tell interface](ask_tell_optimization.ipynb), as this is exactly the case where we want to have greater control of the optimization loop. The disadvantage is that it is up to the user to take care of all the data transformation.
 #
-# As an example, we will be searching for a minimum of a 10-dimensional [Trid functions](https://www.sfu.ca/~ssurjano/trid.html). The range of variation of the Trid function values is large. It varies from values of $10^5$ to its global minimum $f(x^∗) = −210$. This large variation range makes it difficult for Bayesian optimization with Gaussian processes to find the global minimum. However, with data normalisation it becomes possible (see <cite data-cite="hebbal2019bayesian">[Hebbal et al. 2019](https://arxiv.org/abs/1905.03350)</cite>).
+# As an example, we will be searching for a minimum of a 10-dimensional [Trid function](https://www.sfu.ca/~ssurjano/trid.html). The range of variation of the Trid function values is large. It varies from values of $10^5$ to its global minimum $f(x^∗) = −210$. This large variation range makes it difficult for Bayesian optimization with Gaussian processes to find the global minimum. However, with data normalisation it becomes possible (see <cite data-cite="hebbal2019bayesian">[Hebbal et al. 2019](https://arxiv.org/abs/1905.03350)</cite>).
 
 # %%
-
 function = trid_10
 F_MINIMUM = TRID_10_MINIMUM
 search_space = TRID_10_SEARCH_SPACE
@@ -137,7 +135,7 @@ def plot_ask_tell_regret(dataset):
 
     ax.set_yscale("log")
     ax.set_ylabel("Regret")
-    ax.set_ylim(0.001, 10000)
+    ax.set_ylim(0.001, 100000)
     ax.set_xlabel("# evaluations")
 
 
@@ -161,7 +159,7 @@ def standardise(x, mean=None, std=None):
 #
 # Note that we also need to modify the search space, from the original $[-100, 100]$ for all 10 dimensions to the standardised space. For illustration, $[-1,1]$ will suffice here.
 
-# %
+# %%
 search_space = Box([-1], [1]) ** 10
 
 
@@ -171,7 +169,7 @@ search_space = Box([-1], [1]) ** 10
 #
 # We are using a simple approach whereby we normalize the initial data and use estimated mean and standard deviation from the initial normalization for transforming the new points that Bayesian optimization loop adds to the dataset.
 
-# %
+# %%
 x_sta, x_mean, x_std = standardise(initial_data.query_points)
 y_sta, y_mean, y_std = standardise(initial_data.observations)
 standardised_data = Dataset(query_points=x_sta, observations=y_sta)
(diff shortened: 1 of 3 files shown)


docs/notebooks/data_transformation.pct.py
```diff
@@ -0,0 +1,227 @@
...
+observations = dataset.observations.numpy()
+
+arg_min_idx = tf.squeeze(tf.argmin(observations, axis=0))
+
+print(f"query point: {query_points[arg_min_idx, :]}")
+print(f"observation: {observations[arg_min_idx, :]}")
+
+
+# %% [markdown]
+# Once ask-tell optimization is over, you can extract an optimization result object and perform whatever analysis you need, just like with regular Trieste optimization interface. For instance, here we will plot regret for each optimization step.
```
apaleyes: This section (and code below) needs to be updated. It came straight from ask tell notebook, right?

docs/notebooks/data_transformation.pct.py
```diff
@@ -0,0 +1,227 @@
...
+    ax.set_xlabel("# evaluations")
+
+
+plot_ask_tell_regret(dataset)
+
+
+# %% [markdown]
+# # Data transformation with the help of Ask-Tell interface
+#
+# We first write a simple function for doing the standardisation of the data, that is, we scale the data to have a zero mean and a variance equal to 1. We also return the mean and standard deviation parameters as we will use them to transform new points.
```
apaleyes: Context switch is a bit hard here. Perhaps start with "We will now show how data normalization can improve results achieved by BO."

docs/notebooks/data_transformation.pct.py
```diff
@@ -0,0 +1,227 @@
...
+
+# %%
+def standardise(x, mean=None, std=None):
+    if mean is None and std is None:
+        mean = tf.math.reduce_mean(x, 0, True)
+        std = tf.math.sqrt(tf.math.reduce_variance(x, 0, True))
```
apaleyes:
```suggestion
def standardise(x, mean=None, std=None):
    if mean is None:
        mean = tf.math.reduce_mean(x, 0, True)
    if std is None:
        std = tf.math.sqrt(tf.math.reduce_variance(x, 0, True))
```

docs/notebooks/data_transformation.pct.py
```diff
@@ -0,0 +1,227 @@
...
+plot_ask_tell_regret(dataset)
+
+
+# %% [markdown]
+# # Data transformation with the help of Ask-Tell interface
+#
+# We first write a simple function for doing the standardisation of the data, that is, we scale the data to have a zero mean and a variance equal to 1. We also return the mean and standard deviation parameters as we will use them to transform new points.
+
+# %%
+def standardise(x, mean=None, std=None):
```
apaleyes: So, shall we call whatever is going on with the data in this notebook "normalization" (as in PR title) or "standardisation" (as in here)? Probably good to just stick to one name throughout

docs/notebooks/data_transformation.pct.py
```diff
@@ -0,0 +1,227 @@
...
+
+# %%
+x_sta, x_mean, x_std = standardise(initial_data.query_points)
+y_sta, y_mean, y_std = standardise(initial_data.observations)
+standardised_data = Dataset(query_points=x_sta, observations=y_sta)
+
+dataset = initial_data
+for step in range(num_acquisitions):
+
+    # Retraining the model from scratch every 10 steps
```
apaleyes: Why?

docs/notebooks/data_transformation.pct.py
```diff
@@ -0,0 +1,227 @@
...
+    if step % 10 == 0:
+        model = build_gp_model(standardised_data)
+        model.optimize(standardised_data)
+    else:
+        model.update(standardised_data)
+        model.optimize(standardised_data)
+
+    # Asking for a new point to observe
+    ask_tell = AskTellOptimizer(search_space, standardised_data, model)
```
apaleyes: Another illustration why we need something like `tell_model` @henrymoss !

apaleyes (commented):
Reads very clear and explains the general idea well. Some phrasing could be better, but @henrymoss and @saadhamidml can comment on that better than me. Nevertheless, some notes on content below.

docs/notebooks/data_transformation.pct.py
hstojic: right, needs to be slightly adapted

docs/notebooks/data_transformation.pct.py
hstojic: fair enough, lets stick to the more general term normalisation

docs/notebooks/data_transformation.pct.py
hstojic: i'll check if thats really needed

hstojic: not needed, removed

apaleyes (approved):

docs/notebooks/data_transformation.pct.py
```diff
@@ -0,0 +1,250 @@
...
+os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"
+tf.get_logger().setLevel("ERROR")
+
+
+# %% [markdown]
+# ## Describe the problem
+#
+# In this notebook, we show how to perform data transformation during Bayesian optimization. This is usually required by the models. A very common example is normalising the data before fitting the model, either min-max or standard normalization. This is usually done for numerical stability, or to improve or speed up the convergence.
+#
+# In regression problem it is easy to perform data transformations as you do it once before the training. In Bayesian optimization this is more complex, as the data is added with each iteration and would need to be transformed again before the model is updated. At the moment Trieste cannot do such transformations for the user. Luckily, this can be easily done by using the [Ask-Tell interface](ask_tell_optimization.ipynb), as this is exactly the case where we want to have greater control of the optimization loop. The disadvantage is that it is up to the user to take care of all the data transformation.
```
saadhamid96: Would suggest minor change to:
In regression problems it is easy to perform data transformations as you do it once before training. In Bayesian optimization this is more complex, as the data added with each iteration and needs to be transformed as well before the model is updated. At the moment Trieste cannot do such transformations for the user. Luckily, this can be easily done by using the Ask-Tell interface, as it provides greater control of the optimization loop. The disadvantage is that it is up to the user to take care of all the data transformation.

docs/notebooks/data_transformation.pct.py
```diff
@@ -0,0 +1,250 @@
...
+
+# %%
+search_space = Box([-1], [1]) ** 10
+
+
+# %% [markdown]
+#
+# Next we have to define our own Bayesian optimization loop where Ask-Tell optimizer performs optimisation, and we take care of data transformation and model fitting.
+#
+# We are using a simple approach whereby we normalize the initial data and use estimated mean and standard deviation from the initial normalization for transforming the new points that Bayesian optimization loop adds to the dataset.
```
saadhamid96: Would suggest minor change to:
We are using a simple approach whereby we normalize the initial data and use estimated mean and standard deviation from the initial normalization for transforming the new points that the Bayesian optimization loop adds to the dataset.

PR #383, opened 2021-10-21
Uri/optional datasets

uri-granta:
Update `AcquisitionRule.acquire` and the various  `prepare_acquisition_function` and `update_acquisition_function` acquisition function builder methods so that dataset is an optional keyword argument. Some builders and rules insist on data being present; others don't.

### DIFF
--- a/tests/unit/acquisition/test_function.py
+++ b/tests/unit/acquisition/test_function.py
@@ -505,10 +505,7 @@ def test_min_value_entropy_search_chooses_same_as_probability_of_improvement() -
 def test_negative_lower_confidence_bound_builder_builds_negative_lower_confidence_bound() -> None:
     model = QuadraticMeanAndRBFKernel()
     beta = 1.96
-    acq_fn = NegativeLowerConfidenceBound(beta).prepare_acquisition_function(
-        model,
-        dataset=Dataset(tf.zeros([0, 1]), tf.zeros([0, 1])),
-    )
+    acq_fn = NegativeLowerConfidenceBound(beta).prepare_acquisition_function(model)
     query_at = tf.linspace([[-10]], [[10]], 100)
     expected = -lower_confidence_bound(model, beta)(query_at)
     npt.assert_array_almost_equal(acq_fn(query_at), expected)
@@ -518,21 +515,14 @@ def test_negative_lower_confidence_bound_builder_updates_without_retracing() -> 
     model = QuadraticMeanAndRBFKernel()
     beta = 1.96
     builder = NegativeLowerConfidenceBound(beta)
-    acq_fn = builder.prepare_acquisition_function(
-        model,
-        dataset=Dataset(tf.zeros([0, 1]), tf.zeros([0, 1])),
-    )
+    acq_fn = builder.prepare_acquisition_function(model)
     assert acq_fn._get_tracing_count() == 0  # type: ignore
     query_at = tf.linspace([[-10]], [[10]], 100)
     expected = -lower_confidence_bound(model, beta)(query_at)
     npt.assert_array_almost_equal(acq_fn(query_at), expected)
     assert acq_fn._get_tracing_count() == 1  # type: ignore
 
-    up_acq_fn = builder.update_acquisition_function(
-        acq_fn,
-        model,
-        dataset=Dataset(tf.ones([0, 1]), tf.ones([0, 1])),
-    )
+    up_acq_fn = builder.update_acquisition_function(acq_fn, model)
     assert up_acq_fn == acq_fn
     npt.assert_array_almost_equal(acq_fn(query_at), expected)
     assert acq_fn._get_tracing_count() == 1  # type: ignore
@@ -584,7 +574,7 @@ def test_probability_of_feasibility(threshold: float, at: tf.Tensor, expected: f
 @pytest.mark.parametrize("threshold", [-2.3, 0.2])
 def test_probability_of_feasibility_builder_builds_pof(threshold: float, at: tf.Tensor) -> None:
     builder = ProbabilityOfFeasibility(threshold)
-    acq = builder.prepare_acquisition_function(QuadraticMeanAndRBFKernel(), empty_dataset([1], [1]))
+    acq = builder.prepare_acquisition_function(QuadraticMeanAndRBFKernel())
     expected = probability_of_feasibility(QuadraticMeanAndRBFKernel(), threshold)(at)
 
     npt.assert_allclose(acq(at), expected)
@@ -622,11 +612,11 @@ def test_probability_of_feasibility_builder_updates_without_retracing(
     builder = ProbabilityOfFeasibility(threshold)
     model = QuadraticMeanAndRBFKernel()
     expected = probability_of_feasibility(QuadraticMeanAndRBFKernel(), threshold)(at)
-    acq = builder.prepare_acquisition_function(model, dataset=empty_dataset([1], [1]))
+    acq = builder.prepare_acquisition_function(model)
     assert acq._get_tracing_count() == 0  # type: ignore
     npt.assert_allclose(acq(at), expected)
     assert acq._get_tracing_count() == 1  # type: ignore
-    up_acq = builder.update_acquisition_function(acq, model, dataset=empty_dataset([2], [2]))
+    up_acq = builder.update_acquisition_function(acq, model)
     assert up_acq == acq
     npt.assert_allclose(acq(at), expected)
     assert acq._get_tracing_count() == 1  # type: ignore
@@ -1818,10 +1808,7 @@ def test_echvi_is_constraint_when_no_feasible_points() -> None:
 
 def test_predictive_variance_builder_builds_predictive_variance() -> None:
     model = QuadraticMeanAndRBFKernel()
-    acq_fn = PredictiveVariance().prepare_acquisition_function(
-        model,
-        dataset=Dataset(tf.zeros([0, 1]), tf.zeros([0, 1])),
-    )
+    acq_fn = PredictiveVariance().prepare_acquisition_function(model)
     query_at = tf.linspace([[-10]], [[10]], 100)
     _, covariance = model.predict_joint(query_at)
     expected = tf.linalg.det(covariance)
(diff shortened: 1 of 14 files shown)


tests/unit/acquisition/test_rule.py
```diff
@@ -161,19 +166,21 @@ def update_acquisition_function(
...
 
 class _JointBatchModelMinusMeanMaximumSingleBuilder(AcquisitionFunctionBuilder):
     def prepare_acquisition_function(
-        self, dataset: Mapping[str, Dataset], model: Mapping[str, ProbabilisticModel]
+        self,
+        models: Mapping[str, ProbabilisticModel],
+        datasets: Optional[Mapping[str, Dataset]] = None,
     ) -> AcquisitionFunction:
-        return lambda at: -tf.reduce_max(model[OBJECTIVE].predict(at)[0], axis=-2)
+        return lambda at: -tf.reduce_max(models[OBJECTIVE].predict(at)[0], axis=-2)
```
henrymoss, line 184: good spot!

trieste/acquisition/function.py
```diff
@@ -72,21 +72,23 @@ class AcquisitionFunctionBuilder(ABC):
...
+        models: Mapping[str, ProbabilisticModel],
+        datasets: Optional[Mapping[str, Dataset]] = None,
     ) -> AcquisitionFunction:
         """
         Prepare an acquisition function.
 
-        :param datasets: The data from the observer.
-        :param models: The models over each dataset in ``datasets``.
+        :param models: The models for each tag.
+        :param datasets: The data from the observer. (optional)
```
henrymoss: perhaps have (optional) before the full stop ;)

trieste/acquisition/function.py
```diff
@@ -72,21 +72,23 @@ class AcquisitionFunctionBuilder(ABC):
 
     @abstractmethod
     def prepare_acquisition_function(
-        self, datasets: Mapping[str, Dataset], models: Mapping[str, ProbabilisticModel]
+        self,
+        models: Mapping[str, ProbabilisticModel],
+        datasets: Optional[Mapping[str, Dataset]] = None,
     ) -> AcquisitionFunction:
         """
         Prepare an acquisition function.
```
henrymoss: Maybe say something along the lines of "we assume that acq functions always require at least models, but sometimes also want data "

trieste/acquisition/rule.py
```diff
@@ -79,33 +79,36 @@ def acquire(
             :class:`~trieste.space.SearchSpace` depends on the specific :class:`AcquisitionRule`.
 
         :param search_space: The local acquisition search space for *this step*.
-        :param datasets: The known observer query points and observations for each tag.
-        :param models: The model to use for each :class:`~trieste.data.Dataset` in ``datasets``
-            (matched by tag).
+        :param models: The model for each tag.
+        :param datasets: The known observer query points and observations for each tag. (optional)
```
henrymoss: Is it worth saying that the necessity of passing datasets is specified by the chosen acquisition function?

trieste/acquisition/rule.py
```diff
@@ -386,18 +392,30 @@ def acquire(
...
         """
+        if models.keys() != {OBJECTIVE}:
+            raise ValueError(
+                f"dict of models must contain the single key {OBJECTIVE}, got keys {models.keys()}"
+            )
+        if datasets is None or datasets.keys() != {OBJECTIVE}:
+            raise ValueError(
+                f"""datasets must be provided and contain the single key {OBJECTIVE}"""
+            )
+
```
henrymoss, line 411: do we definitely need datasets for all async acquisition functions? @apaleyes

trieste/acquisition/rule.py
```diff
@@ -521,27 +540,41 @@ def acquire(
...
             points from the previous acquisition state.
         """
+        if models.keys() != {OBJECTIVE}:
+            raise ValueError(
+                f"dict of models must contain the single key {OBJECTIVE}, got keys {models.keys()}"
+            )
+        if datasets is None or datasets.keys() != {OBJECTIVE}:
+            raise ValueError(
+                f"""datasets must be provided and contain the single key {OBJECTIVE}"""
+            )
```
henrymoss, line 558: see above

henrymoss (approved):
LGTM

trieste/acquisition/rule.py
hstojic: I agree, it would be useful to clarify what optional means

trieste/acquisition/function.py
```diff
@@ -170,30 +181,39 @@ def __repr__(self) -> str:
...
         :param dataset: The data from the observer. Must be populated.
-        :param model: The model over the specified ``dataset``.
         :return: The expected improvement function. This function will raise
             :exc:`ValueError` or :exc:`~tf.errors.InvalidArgumentError` if used with a batch size
             greater than one.
         :raise tf.errors.InvalidArgumentError: If ``dataset`` is empty.
         """
-        tf.debugging.assert_positive(len(dataset))
+        tf.debugging.Assert(dataset is not None, [])
+        assert dataset is not None
```
hstojic: isnt this covered in the line above?

trieste/acquisition/function.py
```diff
@@ -170,30 +181,39 @@ def __repr__(self) -> str:
...
         """
         :param function: The acquisition function to update.
-        :param dataset: The data from the observer.
-        :param model: The model over the specified ``dataset``.
+        :param model: The model.
+        :param dataset: The data from the observer.  Must be populated.
         """
-        tf.debugging.assert_positive(len(dataset))
+        tf.debugging.Assert(dataset is not None, [])
+        assert dataset is not None
```
hstojic: again, covered in the line above?

trieste/acquisition/function.py
```diff
@@ -248,30 +268,39 @@ def __repr__(self) -> str:
...
         :param dataset: The data from the observer. Must be populated.
-        :param model: The model over the specified ``dataset``.
         :return: The expected improvement function. This function will raise
             :exc:`ValueError` or :exc:`~tf.errors.InvalidArgumentError` if used with a batch size
             greater than one.
         :raise tf.errors.InvalidArgumentError: If ``dataset`` is empty.
         """
-        tf.debugging.assert_positive(len(dataset))
+        tf.debugging.Assert(dataset is not None, [])
+        assert dataset is not None
```
hstojic: same as above....
it seems like its repeated often below

trieste/acquisition/function.py
```diff
@@ -170,30 +181,39 @@ def __repr__(self) -> str:
         return "ExpectedImprovement()"
 
     def prepare_acquisition_function(
-        self, dataset: Dataset, model: ProbabilisticModel
+        self,
+        model: ProbabilisticModel,
+        dataset: Optional[Dataset] = None,
```
hstojic, line 186: do we have to declare it as optional for the acq fncs where we know Dataset is not optional?

trieste/acquisition/function.py
```diff
@@ -248,30 +268,39 @@ def __repr__(self) -> str:
...
         """
         :param function: The acquisition function to update.
-        :param dataset: The data from the observer.
-        :param model: The model over the specified ``dataset``.
+        :param model: The model.
+        :param dataset: The data from the observer. Must be populated.
         """
-        tf.debugging.assert_positive(len(dataset))
+        tf.debugging.Assert(dataset is not None, [])
+        assert dataset is not None
```
hstojic, line 302: same as above

trieste/acquisition/function.py
uri-granta: The first line produces better debugging at runtime, but the second line is for mypy, as otherwise it doesn't infer that dataset is not None past this point. Not sure if there's a better solution (elsewhere we've sometimes cast the variable to the known type).

trieste/acquisition/function.py
uri-granta: Type-wise it still has to be optional, as we're telling calling code that we're implementing the AcquisitionFunctionBuilder API. It's only at runtime that we can decide that it's required.

trieste/acquisition/rule.py
uri-granta: Strictly speaking it's up to the rule, as the rule may not even use an acquisition function! But I can add that to the EGO docstring, and include a more generic explanation here (similar to the comment about functions above).

trieste/acquisition/rule.py
uri-granta: We currently use the datasets. If they're not needed we can always fix it in a separate PR?

(Similarly, while doing this PR I noticed that the combination acquisition function builders defined in combination.py currently still retrace their constituent parts. Again, I'll fix this is in a separate PR.)

apaleyes: Yes, at the moment we expect dataset for input to track pending points in async scenarios. We can revisit that if necessary, but agreed with @uri-granta that's out of scope for that PR

trieste/acquisition/function.py
apaleyes: Can overloads be used here to avoid that? That, and also occasions where dataset isn't even used, but is declared anyway. Just a wild thought, I don't really know if that's possible

trieste/acquisition/function.py
hstojic: ok, I see - I think simple `assert`  might cause issues if any of them are used inside `tf.function` call https://github.com/tensorflow/tensorflow/issues/34715 

didn't check that carefully, hopefully we aren't :)

tests/unit/acquisition/test_function.py
```diff
@@ -17,7 +17,7 @@
 import math
```
hstojic, line 17: I saw on few occasions you tested in passing that not passing the dataset to an acq fnc would work fine
would it be useful to have one explicit test where we test all function that require Dataset, that they raise an error if not passed?

hstojic (approved):
looks good - I have one more concrete suggestion in test_function that you might want to take a look at

trieste/acquisition/function.py
uri-granta: The tests are all passing, but I've replaced the asserts with casts, just in case.

apaleyes (approved):
That was impressively quick! This PR achieves functionally equivalent code while making dataset optional, which is exactly what we wanted. As can be seen in the comments, there are a few possible follow ups, but they can be a subject to a separate effort.

PR #385, opened 2021-10-25
Allow specify fixed reference points | Callable functions to extract reference point

TsingQAQ:
This draft PR for achieving #353: we add a new optional input arg: `ref_point_specification` for:
- `ExpectedHypervolumeImprovement`
- `BatchMonteCarloExpectedHypervolumeImprovement`
- `ExpectedConstrainedHypervolumeImprovement`

The `ref_point_specification` can either be: 
1.  a static reference point (e.g., `TensorType`, `Sequence`) throughout the whole BO iteration (with test supported)
2.  a user-defined callable function that takes the input of `observations` and returns a reference point in each iteration. 

Other changes are:
1. We now provide the `reference_point_spec`  *all the existing observations* (in case of `ExpectedConstrainedHypervolumeImprovement`, we provide all the `feasible_mean`), so that user can extract reference point either only from the front, or from the whole (feasible) observations.
2. store ref point in acq. 

By default, we still use the `get_reference_point` as a dynamic reference point extraction method in each BO iter.


vpicheny (commented):
This is a very nice functionality. I am a bit unsure about what the tests do, ignore me if I didn't understand them properly, do we test properly the case where `ref_point_specification` is a (new) callable?

Also, I think that restricting the function input to be a Pareto front doesn't allow for enough flexibility. In some cases, we'd want to use the Nadir of the observations for instance?

However, I'm not sure actually how much we should generalise. We could simply allow the function to accept a set of observations, or we could pass the models and do all the `mean, _ = model.predict(dataset.query_points)` there, because it's the only place where it's used.

trieste/acquisition/function.py
```diff
@@ -888,9 +889,35 @@ class ExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder):
...
+        """
+        :param ref_point_specification: this method is used to determine how the reference point is
+            calculated. If a Callable function specified, it is expected to take existing pareto
+            front and return a reference point with shape [D] (D represents number of objectives).
+            If the Pareto front location is known, this arg can be used to specify a fixed
+            reference point in each bo iteration. A dynamic reference point updating strategy is
+            used by default to set a reference point according to the datasets.
+        """
+        if callable(ref_point_specification):
+            self._ref_point_specification = ref_point_specification
```
uri-granta: May be worth giving an explicit type here (but ok not to if you don't get any mypy errors after the suggested change below)

```suggestion
            self._ref_point_specification : TensorType | Callable[[TensorType], TensorType] = ref_point_specification
```

trieste/acquisition/function.py
```diff
@@ -888,9 +889,35 @@ class ExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder):
     follows :cite:`yang2019efficient`
     """
 
+    def __init__(
+        self,
+        ref_point_specification: Sequence[float]
```
uri-granta: The parameter name reads a bit weird to me. How about this?

```suggestion
        reference_point_spec: Sequence[float]
```

trieste/acquisition/function.py
```diff
@@ -908,10 +935,19 @@ def prepare_acquisition_function(
         mean, _ = model.predict(dataset.query_points)
 
         _pf = Pareto(mean)
-        _reference_pt = get_reference_point(_pf.front)
+        if isinstance(self._ref_point_specification, FunctionType):
```
uri-granta: FunctionType is only for user defined functions so won't work for e.g. class instances with a call method (or built in functions).

```suggestion
        if callable(self._ref_point_specification):
```

trieste/acquisition/function.py
```diff
@@ -888,9 +889,35 @@ class ExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder):
...
+            self._ref_point_specification = tf.convert_to_tensor(ref_point_specification)
+        self._ref_point = None
+
     def __repr__(self) -> str:
         """"""
-        return "ExpectedHypervolumeImprovement()"
+        if isinstance(self._ref_point_specification, FunctionType):
+            return (
+                f"ExpectedHypervolumeImprovement("
+                f"{self._ref_point_specification.__name__!r})".replace("'", "")
```
```suggestion
                f"{self._ref_point_specification.__name__})"
```

trieste/acquisition/function.py
```diff
@@ -908,10 +935,19 @@ def prepare_acquisition_function(
         mean, _ = model.predict(dataset.query_points)
 
         _pf = Pareto(mean)
-        _reference_pt = get_reference_point(_pf.front)
+        if isinstance(self._ref_point_specification, FunctionType):
+            self._ref_point = self._ref_point_specification(_pf.front)
+        else:
+            assert isinstance(
```
uri-granta: This assert shouldn't be necessary as the type should be inferrable (assuming you make the changes above)

trieste/acquisition/function.py
```diff
@@ -908,10 +935,19 @@ def prepare_acquisition_function(
         mean, _ = model.predict(dataset.query_points)
 
         _pf = Pareto(mean)
-        _reference_pt = get_reference_point(_pf.front)
+        if isinstance(self._ref_point_specification, FunctionType):
+            self._ref_point = self._ref_point_specification(_pf.front)
```
uri-granta: Do you want to cast the result to the right dtype like for the fixed case?

trieste/acquisition/function.py
```diff
@@ -932,8 +968,17 @@ def update_acquisition_function(
         mean, _ = model.predict(dataset.query_points)
 
         _pf = Pareto(mean)
-        _reference_pt = get_reference_point(_pf.front)
-        _partition_bounds = prepare_default_non_dominated_partition_bounds(_reference_pt, _pf.front)
+        if isinstance(self._ref_point_specification, FunctionType):
```
uri-granta: Same comments as above

trieste/acquisition/function.py
```diff
@@ -1053,10 +1098,24 @@ class BatchMonteCarloExpectedHypervolumeImprovement(SingleModelAcquisitionBuilde
     follows :cite:`daulton2020differentiable`
     """
 
-    def __init__(self, sample_size: int, *, jitter: float = DEFAULTS.JITTER):
+    def __init__(
```
uri-granta: Same comments as for ExpectedHypervolumeImprovement

trieste/acquisition/function.py
```diff
@@ -1176,13 +1257,48 @@ class ExpectedConstrainedHypervolumeImprovement(ExpectedConstrainedImprovement):
     ExpectedHypervolumeImprovement.
     """
 
+    def __init__(
```
uri-granta: Same comments again

tests/unit/acquisition/test_function.py
```diff
@@ -916,6 +918,57 @@ def test_ehvi_builder_updates_expected_hv_improvement_using_pareto_from_model()
     assert acq_fn.__call__._get_tracing_count() == 1  # type: ignore
 
 
+@pytest.mark.parametrize(
+    "specify_ref_points",
+    [
+        pytest.param(
+            get_reference_point,
```
uri-granta: Maybe add a test for a custom non-default callable? (maybe even a callable class instance to check that it works)

```python
class GetReferencePoint:
   def __call__(self, front: TensorType) -> TensorType:
       ...
```

uri-granta (approved):
Looks good (and useful)! Just a few comments.

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -74,17 +74,21 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
 
 
-def get_reference_point(front: TensorType) -> TensorType:
+def get_reference_point(
+    observations: TensorType,
+) -> TensorType:
     """
-    reference point calculation method. Note if the front only contains one point, this reference
-    point calculation method will result a reference point the same as the input.
+    default reference point calculation method: a Pareto front will firstly be
```
henrymoss: Default reference point calculation method that calculates the reference point according to a Pareto front extracted from set of observations.

henrymoss: This will do weird things if there are noisy observations?

trieste/acquisition/function/multi_objective.py
```diff
@@ -53,9 +53,32 @@ class ExpectedHypervolumeImprovement(SingleModelAcquisitionBuilder[Probabilistic
...
+        | TensorType
+        | Callable[..., TensorType] = get_reference_point,
+    ):
+        """
+        :param reference_point_spec: this method is used to determine how the reference point is
+            calculated. If a Callable function specified, it is expected to take existing
+            observations and return a reference point with shape [D] (D represents number of
+            objectives). If the Pareto front location is known, this arg can be used to specify
+            a fixed reference point in each bo iteration. A dynamic reference point updating
+            strategy is used by default to set a reference point according to the datasets.
```
henrymoss: Very nice!

henrymoss (commented):
LGTM, I just wonder what we should be doing in the case of noisy observations?

trieste/acquisition/multi_objective/pareto.py
TsingQAQ: >This will do weird things if there are noisy observations?  

Probably, we already pass the posterior mean-based `observations` hopefully to screen out some observation noise, but whether is the most ideal strategy remains unknown.

Am not immediately aware of an existing dynamic reference updating strategy that is able to take noisy obs into account.    In case of noisy observations, probably an alternative (and published) strategy is just to specify ref point beforehand assuming the knowledge of the rough location of the Pareto frontier (ref page 4 of [qNEHVI](https://proceedings.neurips.cc/paper/2021/hash/11704817e347269b7254e744b5e22dac-Abstract.html)), that means a temp strategy that can be used is we do not use this `get_reference_point` in that setting,

TsingQAQ:
> LGTM, I just wonder what we should be doing in the case of noisy observations?

There is no specific ref point specification strategy tailored for noisy observation in this PR. In practice as far as i can see, the user might either:

- Specify a conservative fixed ref point beforehand (as qNEHVI does)
- Dynamically specify a ref point by self defines some reference specification method. However, this functionality is limited supported in this PR since `reference_point_spec` does not take the model as its input (i.e., it cannot use `model.predict` internally).

trieste/acquisition/multi_objective/pareto.py
henrymoss: please can you add some comments saying that the observations input is actuially the posterior mean-based observations. say this gives better perfomance for noisy problems

TsingQAQ: The documentation of `reference_point_spec` has been modified accordingly in the latest [commit](https://github.com/secondmind-labs/trieste/pull/385/commits/3200d3b006bd592b7ec2483a1f8cb7743e125a57). Since the posterior mean-based observation is actually extracted at the acquisition function level, it maybe makes more sense to describe it there for easier usage.  Please see if this makes sense.

PR #387, opened 2021-10-26
optimizer for TaggedProductSearchSpaces

henrymoss:
This PR allows the use of our continuous_optimizer for optimizing over TaggedProductSearchSpaces.  For a mixed space containing both continuous and discrete subspaces, we do a large random search over all the subspaces to find a promising candidate and then perform gradient optimization across the space's Box subspaces whilst keeping its discrete subspaces fixed to match the promising candidate.

This PR contains
- removing redundant functionality for sigmoid-based optimizer. This is a bad optimizer that we have no real use for and its inclusion makes the imp[implementation of this new optimizer much more tricky.
- Moving the upper/lower method to be an abstract method for all search spaces and defining upper/lower for DiscreteSearchSpace, as required for computing the optimization domain.
- A new method for TaggedProductSearchSpace, that allows the `fixing` of one of its subspaces to take a fixed set of values.
- Allows the sampling of a larger number of samples than the number of points in a DiscreteSearchSpace.
- unit tests for all above
- integration test for mixed-space optimization problem


trieste/space.py
```diff
@@ -411,25 +432,49 @@ def __repr__(self) -> str:
...
 
     def get_subspace(self, tag: str) -> SearchSpace:
         """Return the domain of a particular subspace."""
         return self._spaces[tag]
 
+    def fix_subspace(self, tag: str, values: TensorType) -> TaggedProductSearchSpace:
+        """
+        Return a new :class:`TaggedProductSearchSpace` with the specified subspace replaced with
+        a :class:`DiscreteSearchSpace` containing ``values`` as its points.
+        """
```
apaleyes, line 481: docstring needs to describe the parameters

trieste/space.py
```diff
@@ -411,25 +432,49 @@ def __repr__(self) -> str:
...
-    def dimension(self) -> TensorType:
+    def dimension(self) -> int:
         """The number of inputs in this product search space."""
         return self._dimension
 
     def get_subspace(self, tag: str) -> SearchSpace:
         """Return the domain of a particular subspace."""
         return self._spaces[tag]
 
+    def fix_subspace(self, tag: str, values: TensorType) -> TaggedProductSearchSpace:
```
apaleyes, line 473: This method probably needs to validate that the subspace to fix contains provided values. As well as normal stuff: tag is known, values are not empty

trieste/acquisition/optimizer.py
```diff
@@ -214,11 +216,39 @@ def _perform_optimization(starting_point: TensorType) -> OptimizeResult:
...
+    if isinstance(space, TaggedProductSearchSpace):  # convert discrete subspaces to box spaces.
+        space_with_fixed_discrete = space
+        for tag in space.subspace_tags:
+            if isinstance(space.get_subspace(tag), DiscreteSearchSpace):
+                subspace_value = space.get_subspace_component(tag, current_point)
+                space_with_fixed_discrete = space_with_fixed_discrete.fix_subspace(
+                    tag, subspace_value
+                )
+        return spo.Bounds(space_with_fixed_discrete.lower, space_with_fixed_discrete.upper)
+    else:  # if already a box space then just return bounds
```
apaleyes: check if it is a Box explicitly, and then throw an error if it isn't

trieste/acquisition/optimizer.py
```diff
@@ -214,11 +216,39 @@ def _perform_optimization(starting_point: TensorType) -> OptimizeResult:
...
+        return tf.convert_to_tensor(chosen_point)
 
     return optimize_continuous
 
 
+def get_bounds_of_box_relaxation_around_point(
+    space: TaggedProductSearchSpace | Box, current_point: TensorType
+) -> spo.Bounds:
+    """
+    A function to return the bounds of a continuous relaxation of
```
apaleyes, line 234: I don't know what "continuous relaxation" means. Can the docstring explain that?

That method could also use some validation

trieste/acquisition/optimizer.py
```diff
@@ -214,11 +216,39 @@ def _perform_optimization(starting_point: TensorType) -> OptimizeResult:
...
+    a :class:`TaggedProductSearchSpace' space. All :class:`DiscreteSearchSpace'
+    subspaces are replaced with a new :class:`DiscreteSearchSpace' containing only their
+    respective component of the specified `current_point'. Note that
```
```suggestion
    subspaces are replaced with a new :class:`DiscreteSearchSpace' fixed at their
    respective component of the specified `current_point'. Note that
```

trieste/acquisition/optimizer.py
```diff
@@ -165,34 +170,29 @@ def optimize_continuous(space: Box, target_func: AcquisitionFunction) -> TensorT
...
         def _objective() -> TensorType:
-            return -target_func(bijector.forward(variable[:, None, :]))  # [1]
+            return -target_func(variable[:, None, :])  # [1]
 
-        def _perform_optimization(starting_point: TensorType) -> OptimizeResult:
-            variable.assign(bijector.inverse(starting_point))  # [1, D]
+        def _perform_optimization(starting_point: TensorType, bounds: spo.Bounds) -> OptimizeResult:
+            opt_kwargs = {"bounds": bounds}
+            variable.assign(starting_point)  # [1, D]
             return gpflow.optimizers.Scipy().minimize(_objective, (variable,), **opt_kwargs)
```
apaleyes: Since the kwargs aren't defined at runtime, you can just specify them directly

```suggestion
            variable.assign(starting_point)  # [1, D]
            return gpflow.optimizers.Scipy().minimize(_objective, (variable,), bounds=bounds)
```

apaleyes (commented):
Looks reasonable, although didn't take super close look at tests. One general comment is that new methods need validation of inputs.

tests/integration/test_mixed_space_bayesian_optimization.py
```diff
@@ -0,0 +1,126 @@
...
+    "num_steps, acquisition_rule",
+    cast(
+        List[
+            Tuple[
+                int,
+                AcquisitionRule[TensorType, TaggedProductSearchSpace],
+            ]
+        ],
+        [
+            (15, EfficientGlobalOptimization()),
```
vpicheny, line 57: Would we want to test that with `TrustRegions` as well? TR redefines the search space, which makes it more likely to break...

vpicheny: I'm not sure TR supports discrete input spaces though

henrymoss: TR requires Box (at least at the moment!)

trieste/space.py
```diff
@@ -125,12 +145,13 @@ def sample(self, num_samples: int) -> TensorType:
         :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
             from this search space.
         """
-        tf.debugging.assert_less_equal(
-            num_samples,
-            len(self._points),
-            message="Number of samples cannot be greater than the number of points in search space",
-        )
-        return tf.random.shuffle(self._points)[:num_samples, :]
+        if num_samples == 0:
```
uri-granta, line 148: Is this if branch necessary? (i.e. won't calling tf.random.categorical and tf.gather on empty samples and indices work?)

trieste/space.py
```diff
@@ -411,25 +432,49 @@ def __repr__(self) -> str:
                 tags = {self.subspace_tags})
                 """
 
+    @property
+    def lower(self) -> TensorType:
+        """The lowest values taken by each space dimension, concatenated across subspaces."""
+        lower_for_each_subspace = [self.get_subspace(tag).lower for tag in self.subspace_tags]
+        return tf.concat(lower_for_each_subspace, -1)
```
uri-granta: (we usually write it like this, no?)

```suggestion
        return tf.concat(lower_for_each_subspace, axis=-1)
```

trieste/space.py
```diff
@@ -411,25 +432,49 @@ def __repr__(self) -> str:
...
+
+    @property
+    def upper(self) -> TensorType:
+        """The largest values taken by each space dimension, concatenated across subspaces."""
+        upper_for_each_subspace = [self.get_subspace(tag).upper for tag in self.subspace_tags]
+        return tf.concat(upper_for_each_subspace, -1)
```
```suggestion
        return tf.concat(upper_for_each_subspace, axis=-1)
```

trieste/space.py
uri-granta: :+1: (and remember to add unit tests for the error cases)

trieste/acquisition/optimizer.py
uri-granta: A high level explanation of what the purpose of this might be useful too. E.g. why do we fix only discrete subspaces and not boxes?

trieste/space.py
```diff
@@ -125,12 +145,13 @@ def sample(self, num_samples: int) -> TensorType:
         :return: ``num_samples`` i.i.d. random points, sampled uniformly, and without replacement,
```
uri-granta: Following this change is the description ("without replacement") still correct?

tests/unit/test_space.py
```diff
@@ -121,32 +151,12 @@ def test_discrete_search_space_contains_raises_for_invalid_shapes(
...
-) -> None:
-    assert all(sample in search_space for sample in samples)
-    assert len(samples) == num_samples
-
-    unique_samples = set(tuple(sample.numpy().tolist()) for sample in samples)
-
-    assert len(unique_samples) == len(samples)
-
-
 @pytest.mark.parametrize("num_samples", [0, 1, 3, 5, 6])
```
uri-granta: Also test for num_sampes > len(search_space)?

tests/unit/test_space.py
```diff
@@ -517,6 +535,39 @@ def test_product_search_space_returns_correct_dimension(
...
+            tf.constant([2.0, 3.0, -0.5]),
+        ),
+        (
+            [Box([-1, -2], [2, 3]), Box([-1, -2], [2, 3]), Box([-1], [2])],
+            tf.constant([-1.0, -2.0, -1.0, -2.0, -1.0]),
+            tf.constant([2.0, 3.0, 2.0, 3.0, 2.0]),
+        ),
+    ],
+)
+def test_product_space_returns_correct_bounds(
```
uri-granta, line 563: You should probably also have a similar test for plain boxes.

tests/unit/test_space.py
```diff
@@ -540,6 +591,56 @@ def test_product_space_get_subspace() -> None:
...
+        Box([-1], [2]),
+    ]
+    tags = ["A", "B", "C"]
+    product_space = TaggedProductSearchSpace(spaces=spaces, tags=tags)
+
+    for tag in tags:
+        product_space_with_fixed_subspace = product_space.fix_subspace(tag, points)
+        remaining_tags = tags
+        remaining_tags.remove(tag)
+        for non_target_tag in remaining_tags:
```
uri-granta: Simpler to just use an if test:

```suggestion
        for other_tag in tags:
            if other_tag != tag:
```

trieste/space.py
```diff
@@ -105,13 +115,23 @@ def __repr__(self) -> str:
         """"""
         return f"DiscreteSearchSpace({self._points!r})"
 
+    @property
+    def lower(self) -> TensorType:
+        """The lowest value taken across all points by each search space dimension."""
+        return tf.reduce_min(self.points, -2)
```
vpicheny, line 121: does this make sense

tests/unit/test_space.py
```diff
@@ -540,6 +591,56 @@ def test_product_space_get_subspace() -> None:
...
+            assert isinstance(
```
uri-granta: Is there any reason why we don't define equality on search spaces? (i.e. same type, same bounds for Boxes, same points for discrete, same subspaces for tagged)

If we don't want to do that, you could perhaps also check that the bounds haven't changed?

tests/integration/test_mixed_space_bayesian_optimization.py
```diff
@@ -0,0 +1,126 @@
...
+        .optimize(num_steps, initial_data, model, acquisition_rule)
+        .try_get_final_dataset()
+    )
+
+    arg_min_idx = tf.squeeze(tf.argmin(dataset.observations, axis=0))
+
+    best_y = dataset.observations[arg_min_idx]
+    best_x = dataset.query_points[arg_min_idx]
+
+    relative_minimizer_err = tf.abs((best_x - BRANIN_MINIMIZERS) / BRANIN_MINIMIZERS)
```
uri-granta, line 116: I'm getting worried about how long our checks take to run now. We should have another chat about this at some point soon.

uri-granta (approved):
LGTM. Just a few small comments/questions.

trieste/space.py
henrymoss: unfortunately not

trieste/space.py
henrymoss: good spot!

tests/unit/test_space.py
henrymoss: we already do

tests/unit/test_space.py
henrymoss: Yeah we should do this! but in another PR for sure

trieste/space.py
```diff
@@ -105,13 +115,23 @@ def __repr__(self) -> str:
...
+
+    @property
+    def upper(self) -> TensorType:
+        """The lowest value taken across all points by each search space dimension."""
```
hstojic: lowest -> highest?

trieste/space.py
```diff
@@ -411,25 +432,58 @@ def __repr__(self) -> str:
...
+        Return the domain of a particular subspace.
+
+        :param tag: The tag specifying the target subspace.
+        """
+
         return self._spaces[tag]
 
+    def fix_subspace(self, tag: str, values: TensorType) -> TaggedProductSearchSpace:
+        """
+        Return a new :class:`TaggedProductSearchSpace` with the specified subspace replaced with
```
hstojic, line 475: its a bit hard to see from this description what is this used for - some additional clarifications perhaps?

trieste/space.py
```diff
@@ -122,15 +142,16 @@ def __contains__(self, value: TensorType) -> bool | TensorType:
...
-        tf.debugging.assert_less_equal(
-            num_samples,
-            len(self._points),
-            message="Number of samples cannot be greater than the number of points in search space",
-        )
-        return tf.random.shuffle(self._points)[:num_samples, :]
+        if num_samples == 0:
+            return self.points[:0, :]
+        else:
+            sampled_indicies = tf.random.categorical(
```
hstojic: nitpicking: indicies -> indices

trieste/acquisition/optimizer.py
```diff
@@ -99,19 +98,17 @@ def optimize_discrete(space: DiscreteSearchSpace, target_func: AcquisitionFuncti
...
     num_recovery_runs: int = 5,
-) -> AcquisitionOptimizer[Box]:
+) -> AcquisitionOptimizer[Box | TaggedProductSearchSpace]:
     """
-    Generate a gradient-based acquisition optimizer for :class:'Box' spaces and batches
-    of size of 1. We perfom gradient-based optimization starting from the best location
-    across a sample of `num_initial_samples` random points.
+    Generate a gradient-based optimizer for :class:'Box' and :class:'TaggedProductSearchSpace'
+    spaces and batches of size 1. We perform gradient-based optimization across
+    all :class:'Box' subspaces, starting from the best location found across a sample
```
hstojic: it might be good to clarify that these subspaces apply to TaggedProductSearchSpace:
 We perform gradient-based optimization, across all :class:'Box' subspaces in case of :class:`TaggedProductSearchSpace`, ...

trieste/acquisition/optimizer.py
```diff
@@ -99,19 +98,17 @@ def optimize_discrete(space: DiscreteSearchSpace, target_func: AcquisitionFuncti
 
 def generate_continuous_optimizer(
     num_initial_samples: int = 1000,
-    sigmoid: bool = False,
```
hstojic, line 102: why dropping the option for sigmoid? are we certain its always better to use L-BFGS-B?

trieste/acquisition/optimizer.py
```diff
@@ -273,13 +308,7 @@ def optimize_random(space: SP, target_func: AcquisitionFunction) -> TensorType:
                 [..., 1].
         :return: The **one** point in ``space`` that maximises ``target_func``, with shape [1, D].
         """
-
-        if isinstance(space, DiscreteSearchSpace) and (num_samples > len(space.points)):
```
hstojic, line 277: this check is moved to the space definition, right?

trieste/acquisition/optimizer.py
```diff
@@ -214,11 +222,38 @@ def _perform_optimization(starting_point: TensorType) -> OptimizeResult:
                     """
                 )
 
-        return chosen_point  # [1, D]
+        return tf.convert_to_tensor(chosen_point)  # convert chosen point back from a variable
 
     return optimize_continuous
 
 
+def get_bounds_of_box_relaxation_around_point(
```
hstojic, line 230: perhaps we should have a `util.py` with this and batchify function, potentially some other aux functions from rule.py and function.py?

trieste/acquisition/optimizer.py
```diff
@@ -165,34 +170,32 @@ def optimize_continuous(space: Box, target_func: AcquisitionFunction) -> TensorT
...
         def _objective() -> TensorType:
-            return -target_func(bijector.forward(variable[:, None, :]))  # [1]
+            return -target_func(variable[:, None, :])  # [1]
 
-        def _perform_optimization(starting_point: TensorType) -> OptimizeResult:
-            variable.assign(bijector.inverse(starting_point))  # [1, D]
-            return gpflow.optimizers.Scipy().minimize(_objective, (variable,), **opt_kwargs)
+        def _perform_optimization(starting_point: TensorType, bounds: spo.Bounds) -> OptimizeResult:
+            variable.assign(starting_point)  # [1, D]
+            return gpflow.optimizers.Scipy().minimize(_objective, (variable,), bounds=bounds)
```
hstojic, line 180: we should leave `**opt_kwargs` in and allow user to pass arguments to it by adding this argument in `generate_continuous_optimizer`, @sebastianober for example needed to switch compilation off

trieste/acquisition/optimizer.py
```diff
@@ -214,11 +222,38 @@ def _perform_optimization(starting_point: TensorType) -> OptimizeResult:
...
+    """
+    tf.debugging.Assert(isinstance(space, TaggedProductSearchSpace), [])
+
+    space_with_fixed_discrete = space
+    for tag in space.subspace_tags:
+        if isinstance(
+            space.get_subspace(tag), DiscreteSearchSpace
+        ):  # convert discrete subspaces to box spaces.
+            subspace_value = space.get_subspace_component(tag, current_point)
+            space_with_fixed_discrete = space_with_fixed_discrete.fix_subspace(tag, subspace_value)
```
hstojic, line 253: we need to really make sure this fixing is done correctly...

trieste/acquisition/optimizer.py
henrymoss: we never had access to **opt_kwargs anyway. I agree that this is important, but should be another PR. Ive added an issue

trieste/acquisition/optimizer.py
henrymoss: Yeh. Extensive experiments in Trieste and internally show its not really fit for purpose in contained space (for which we always have for acquisition function maximisation)

trieste/acquisition/optimizer.py
henrymoss: Yep. something to bear in mind for another PR

trieste/acquisition/optimizer.py
henrymoss: Agreed. I have lots of unit tests to check this. Do you have any additional requests?

trieste/space.py
```diff
@@ -411,25 +432,66 @@ def __repr__(self) -> str:
...
+        Return a new :class:`TaggedProductSearchSpace` with the specified subspace replaced with
+        a :class:`DiscreteSearchSpace` containing ``values`` as its points. This is useful if you
+        wish to restrict subspaces to sets of representative points.
+
+        :param tag: The tag specifying the target subspace.
+        :param values: The  values used to populate the new discrete subspace.z
+        """
+
+        new_spaces = [
+            self.get_subspace(t) if t != tag else DiscreteSearchSpace(points=values)
```
hstojic, line 484: my understanding from the optimizer part is that we would use fixing only for the discrete variables, not continuous ones - here to me it seems you could convert Box into DiscreteSearchSpace, is that desired behaviour? just added flexibility that might come handy later on?

tests/unit/test_space.py
```diff
@@ -540,6 +600,55 @@ def test_product_space_get_subspace() -> None:
...
+
+
+@pytest.mark.parametrize(
+    "points",
+    [
+        tf.ones((1, 5), dtype=tf.float64),
+        tf.ones((2, 3), dtype=tf.float64),
+    ],
+)
+def test_product_space_fix_subspace_doesnt_fix_undesired_subspace(points: tf.Tensor) -> None:
```
hstojic, line 633: I would probably consider adding a test of bounds - after fixing you should know all the bounds and could easily test it and I think its a bit more comprehensive test. This being said, these two tests are probably fine by themselves, the method shouldn't affect bounds of non-fixed dimensions - at the moment at least, perhaps as a defensive measure against future changes of the method :)

trieste/space.py
henrymoss: added flexibility :)

trieste/acquisition/optimizer.py
hstojic: its a quick change, no? unless you think it would require writing another test?
if so, then sure, best left for another PR if you are in a hurry

tests/unit/test_space.py
hstojic: ah, right, you do that in `test_optimizer`...

trieste/acquisition/optimizer.py
henrymoss: okay will leave for now

trieste/acquisition/optimizer.py
```diff
@@ -165,34 +170,32 @@ def optimize_continuous(space: Box, target_func: AcquisitionFunction) -> TensorT
...
-        chosen_point = bijector.forward(variable)  # [1, D]
+        chosen_point = variable  # [1, D]
         chosen_point_score = target_func(chosen_point[:, None, :])  # [1, 1]
 
         for i in tf.range(
             num_optimization_runs
         ):  # perform optimization for each chosen starting point
-            opt_result = _perform_optimization(initial_points[i : i + 1])
+            if isinstance(space, TaggedProductSearchSpace):
+                bounds = get_bounds_of_box_relaxation_around_point(space, initial_points[i : i + 1])
```
hstojic, line 190: at the moment search in discrete dimensions is restricted to initial points - I wonder whether a sort of iterative procedure would make sense, after finding good values for cont dimensions, fix those and return original bounds for discrete ones, then run random search again on discrete ones, and so on, certain number of iterations of this procedure

hstojic (commented):
Looks quite good, few smaller comments left to resolve

PR #389, opened 2021-10-27
refactoring configs

### LINKED ISSUE #373: Removing `models/config.py`?
I would like to discuss whether our config system is actually useful in its current form, if not perhaps we should dispense with it, simplify the codebase and reduce our maintenance.

After talking with almost everyone, it is not entirely clear why we have `ModelConfig` and were allowing specifying models as a dictionary via `create_model`. Was this aimed at simplifying life for the novice users or at facilitating things for power users that know the modelling toolboxes and trieste wrappers very well?

My best guess is that a user can be ignorant about model and optimizer wrappers, one only needs to specify a GPflow model and our model factory will do the rest - no need to know the names of the wrappers, and no need to handle two objects (first create an Optimizer which is then passed to a model wrapper). For power users, if you know all the arguments to the wrappers I guess this could simplify things a bit. For novice users the issue with this is that you still have to look up the wrappers if you need to pass some additional arguments so you might as well use them (in my view it is also too black-boxy and is not incentivising users to get to know the toolbox). Also, you still need to be quite an expert to set up a model. If we look at our usual examples in the notebooks, building a model requires 5-10 lines of non-trivial GPflow code. A useful config system would take care of building a model as well. 

There is potentially another usage. Configs might facilitate coding up large scale experiments where we compare lots of algorithms. @apaleyes gave an example on how setting the parameters of the experiment could be easier than working with objects directly
```python
basic_config = blah

for params in candidate_param_list:
    basic_config["model_params"] = params
    do_banchmark(basic_config)
```
However, it doesn't seem like anyone was really using configs for this purpose so far (anyone?), so not sure if this is a realistic us case. Perhaps not in a current form and if this would be useful, we could think of extending configs. We could try exploring this with a notebook?

In the discussions there were also mentions of having sensible defaults for (almost) everything so that its easy to set things up and run an BO algorithm for an unexperienced BO user. We should definitely do this but I think this is not relevant for the current config system.

As I see it we have 4 options:
1a. Do nothing for the time being, keep maintaining it until we collect more info on how users use configs (how do we do that though?). Perhaps try making dictionary option possible again.
1b. If we agree that this is a tool for power users, then to reduce chances of confusing novice users we should remove dictionary/model config way of setting up a model from notebooks and write a new notebook aimed at explaining configs, how to use them and when.
2. Remove it and rely on working with model and optimizer wrappers directly. Perhaps focus on making sure we provide sensible defaults wherever we can.
3. Develop configs further, perhaps something where we can configure a whole BO algorithm, both underlying GPflow model, acquisition function, optimization process etc. @avullo might have some useful pointers here.

Any feedback would be much appreciated!
Pointers to config systems in other toolboxes are welcome as well.

hstojic:
This PR will close #373 where we chose to go with option 1a and 1b

Implemented
1. added a registry of models to enable config via dictionary again
2. improved tests
3. removed `create_optimizer`, now dealt with via model registry
4. notebook explaining configs
5. sorted out optimizers in gpflux (incompatible with ModelConfig)


docs/notebooks/failure_ego.pct.py
```diff
@@ -117,26 +115,19 @@ def create_classification_model(data):
...
-        },
-        "optimizer": tf.optimizers.Adam(1e-3),
-        "optimizer_args": {
-            "max_iter": 50,
-        },       
-    }),
+    OBJECTIVE: GaussianProcessRegression(regression_model),
+    FAILURE: VariationalGaussianProcess(
+        classification_model,
+        BatchOptimizer(tf.optimizers.Adam(1e-3), max_iter=50),
```
apaleyes: That looks like slight step back in terms of user experience. Now they need to know which optimizer class to use, which isn't obvious. E.g. I don't have a slight clue why we use `BatchOptimizer` here

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,146 @@
...
+tf.get_logger().setLevel("ERROR")
+
+np.random.seed(1793)
+tf.random.set_seed(1793)
+
+
+# %% [markdown]
+# ## Finding a minimum of the Branin function
+#
+# In this example, as in most other tutorial, we look to find the minimum value of the familiar two-dimensional Branin function over the hypercube $[0, 1]^2$.
```
apaleyes:
```suggestion
# In this example, as in many other tutorials, we look to find the minimum value of the familiar two-dimensional Branin function over the hypercube $[0, 1]^2$.
```

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,146 @@
+# %% [markdown]
+# # Building models with configuration dictionaries
+#
+# If you are an expert user of Trieste and some modelling library, GPflow for example, then building models via a configuration dictionary might be a useful alternative to working with model and optimizer interfaces. Here we provide an overview of how to use configuration dictionaries.
```
apaleyes:
```suggestion
# If you are an expert user of Trieste and some modelling library, GPflow for example, then building models via a configuration dictionary might be a useful alternative to working with model and optimizer wrappers. Here we provide an overview of how to use configuration dictionaries.
```

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,146 @@
...
+initial_query_points = search_space.sample_sobol(num_initial_points)
+initial_data = observer(initial_query_points)
+
+
+# %% [markdown]
+# ## Standard way of setting up a model of the objective function
+#
+# The Bayesian optimization procedure estimates the next best points to query by using a probabilistic model of the objective. We'll use Gaussian Process (GP) regression in this tutorial, as provided by GPflow.
+#
+# GPflow models cannot be used directly in our Bayesian optimization routines, only through a valid model interface. Trieste typically has a separate interface for each model. For instance, `GPR` model from GPflow has to be used with `GaussianProcessRegression` interface. These interfaces standardise outputs from all models, deal with preparation of the data and implement additional methods needed for Bayesian optimization.
```
apaleyes: That sounds slightly weird, as though each model class in GPflow has a corresponding class in Trieste. Maybe "model type"?

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,146 @@
...
+    gpr = gpflow.models.GPR(data.astuple(), kernel, noise_variance=1e-5)
+    return gpr
+
+
+gpflow_model = build_model(initial_data)
+
+# %% [markdown]
+# Before using the model interface we might want to set the optimizer for it as well
+# and set some arguments that we want to pass to it. We need to use Trieste's optimizer
+# interfaces, here `Optimizer` would be the suitable interface. We'll optimize our model with GPflow's Scipy optimizer.
```
apaleyes: Same comment as above - even for simple use cases we seem to require knowledge of a suitable optimizer. Can this be simplified?

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,146 @@
...
+
+# %%
+bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+result = bo.optimize(2, initial_data, model)
+
+
+# %% [markdown]
+# ## Using configuration dictionaries
+#
+# Instead of working direclty with model and optimization interfaces, after you know them sufficiently well, you can skip them by using configuration dictionary. It consists of a dictionary with same four arguments that can be passed to any model interface: `model`, `model_args`, `optimizer` and `optimizer_args`.
```
apaleyes:
```suggestion
# Instead of working directly with model and optimization interfaces, after you know them sufficiently well, you can skip them by using configuration dictionary. It consists of a dictionary with same four arguments that can be passed to any model interface: `model`, `model_args`, `optimizer` and `optimizer_args`.
```

trieste/models/optimizer.py
```diff
@@ -14,16 +14,15 @@
 
 r"""
 This module contains common optimizers based on :class:`~tf.optimizers.Optimizer` that can be used
-with models. Specific models can sub-class these optimizers or implement their own, and should
-register them using a :func:`create_optimizer`, as well as register their loss functions using
-a :func:`create_loss_function`.
+with models. Specific models can also sub-class these optimizers or implement their own, and should
+register register their loss functions using a :func:`create_loss_function`.
```
apaleyes:
```suggestion
register their loss functions using a :func:`create_loss_function`.
```

trieste/models/gpflow/config.py
```diff
@@ -14,35 +14,22 @@
...
-            VGP: VariationalGaussianProcess,
-            SVGP: SparseVariational,
-        }
-        return models_mapping
+# Here we list all the GPflow models currently supported by model interfaces
+# and optimizers, and register them for usage with ModelConfig.
+_SUPPORTED_MODELS: Dict[Type[Any], Tuple[Type[TrainableProbabilisticModel], Type[Optimizer]]] = {
+    GPR: (GaussianProcessRegression, Optimizer),
+    SGPR: (GaussianProcessRegression, Optimizer),
+    VGP: (VariationalGaussianProcess, BatchOptimizer),
```
apaleyes: As far as I can tell you can only know this correspondence either from here or from expert knowledge. Neither are reasonable to expect from non-power users. I think that makes our default class-based method quite hard to use. Can we consider maybe default optimizers for models, or using registry for models? Or am I overthinking this?

apaleyes (commented):
If you don't count the notebook, this PR removes more lines than it adds. Great refactor!

However I do have two main comments:
1. Notebook not very convincing, as it doesn't explain why would anyone want to use configs or a registry. Can we at least simulate a bit more exciting use case?
2. Most of notebooks used to not require expert knowledge of which optimizer wrapper shall be used. Now they do. I feel that's a step back in user experience. Can we avoid that?

docs/notebooks/model_config.pct.py
hstojic: well, only SGPR and GPR share the wrapper GaussianProcessRegression, SVGP and VGP have their own wrapper...

docs/notebooks/failure_ego.pct.py
hstojic: here we need to explicitly pass the `BatchOptimizer` object because we are using non-default settings (`max_iter=50`), otherwise the wrapper tells you which `Optimizer` you need to use

I think there is no way around it, as soon as they need a non-default optimizer they need to know about it
what we could do is the following:
1. Explicitly educate users in one of the intro notebooks on difference between `Optimizer` and `BatchOptimizer` and when to use them
2. Eliminate it from the rest of the notebooks and simply rely on defaults set in the interface, notebooks should work fine with the defaults

docs/notebooks/model_config.pct.py
hstojic: in simple cases we don't, only when we want to use non-default settings - perhaps I can spell that out here?

trieste/models/gpflow/config.py
hstojic: does looking up a wrapper to figure out what arguments are needed, say `GaussianProcessRegression`, constitute an expert knowledge?

each wrapper does use default optimizers, perhaps we are just not doing ourselves a favour by not using them in notebooks?

hstojic:
> If you don't count the notebook, this PR removes more lines than it adds. Great refactor!
> 
> However I do have two main comments:
> 
> 1. Notebook not very convincing, as it doesn't explain why would anyone want to use configs or a registry. Can we at least simulate a bit more exciting use case?
> 2. Most of notebooks used to not require expert knowledge of which optimizer wrapper shall be used. Now they do. I feel that's a step back in user experience. Can we avoid that?

I completely agree about 1), but not sure we can come up with a better use case for configs. I can try setting up a use case for running experiments that we discussed, perhaps that would end up being a better one? 

Regarding 2), we could either use default optimizers in notebooks and not specify them, or go back to using dictionaries which would imply that setting up models with a dictionary is for beginners, not experts? I would be up for the former option, I think even beginners need some exposure to model wrappers so that they know where to look up the info about more advanced usage

docs/notebooks/failure_ego.pct.py
apaleyes: If there is a nice default, i'd go with option 2.

docs/notebooks/model_config.pct.py
apaleyes: But at least it isn't designed "class per model", right? In theory many models can share one wrapper

docs/notebooks/model_config.pct.py
apaleyes: That would be suitable place to explain this indeed

trieste/models/gpflow/config.py
apaleyes: Somewhat... let's use defaults in notebooks as much as we can, so that readers don't start asking "why" or "is that important" a lot.

apaleyes:
> I completely agree about 1), but not sure we can come up with a better use case for configs. I can try setting up a use case for running experiments that we discussed, perhaps that would end up being a better one?
> 
> Regarding 2), we could either use default optimizers in notebooks and not specify them, or go back to using dictionaries which would imply that setting up models with a dictionary is for beginners, not experts? I would be up for the former option, I think even beginners need some exposure to model wrappers so that they know where to look up the info about more advanced usage

Just mentioning it in the notebook would be fine for now. And yeah, let's go with defaults where we can

docs/notebooks/failure_ego.pct.py
hstojic: ah, these weren't meant to be options, I was proposing doing both things :)

docs/notebooks/model_config.pct.py
hstojic: that's true, I'll word it more carefully

hstojic:
Addressed all the comments, here are the highlights: 
- removed custom optimizer from most notebooks (not really needed) 
- added a more detailed explanation on role of wrappers in an intro notebook
- added another use case to the model_config notebook

docs/notebooks/active_learning.pct.py
```diff
@@ -45,7 +45,7 @@
 # %% [markdown]
 # ## Surrogate model
 #
-# Just like in sequential optimization, we fit a surrogate Gaussian process model to the initial data.
+# Just like in sequential optimization, we fit a surrogate Gaussian process model as implemented in GPflow to the initial data. As usual, the GPflow models cannot be used directly in our Bayesian optimization routines, only through a valid model wrapper. Below we construct a `GPR` model from GPflow and pass it to the appropriate `GaussianProcessRegression` wrapper.
```
apaleyes:
```suggestion
# Just like in sequential optimization, we fit a surrogate Gaussian process model as implemented in GPflow to the initial data. The GPflow models cannot be used directly in our Bayesian optimization routines, only through a valid model wrapper. Below we construct a `GPR` model from GPflow and pass it to the appropriate `GaussianProcessRegression` wrapper.
```

it's not usual, they just can't

docs/notebooks/batch_optimization.pct.py
```diff
@@ -35,13 +35,13 @@
 
 # %% [markdown]
 # ## Surrogate model
-# Just like in purely sequential optimization, we fit a surrogate Gaussian process model to the initial data.
+#
+# Just like in purely sequential optimization, we fit a surrogate Gaussian process model to the initial data. As usual, the GPflow models cannot be used directly in our Bayesian optimization routines, only through a valid model wrapper. Below we construct a `GPR` model from GPflow and pass it to the appropriate `GaussianProcessRegression` wrapper.
```
apaleyes: Is it actually worth adding this bit in many notebooks? We can be fairly detailed in one notebook, and refer to it thereafter.

docs/notebooks/model_config.pct.py
```diff
@@ -110,19 +113,79 @@ def build_model(data):
...
 
 # %%
 bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 result = bo.optimize(2, initial_data, model_config)
 
 
+# %% [markdown]
+# ## Using configuration dictionaries for setting up experiments
+#
+# Another use case is in setting up experiments, where it becomes easier to benchmark Bayesian optimization algorithms. The advantage is that we can easily change the models and set up any argument for them from one condition to another - only object with the experiment specification needs to change (`experiment_conditions` below), while the rest of the code for executing experiments can stay the same.
```
apaleyes: Second sentence is a bit hard to read, can you simplify it? Also what are these "conditions"?

docs/notebooks/model_config.pct.py
```diff
@@ -110,19 +113,79 @@ def build_model(data):
...
+        "minimize_args": {"options": dict(maxiter=100)},
+    },
+}
+
+# here we specify our experiments
+experiment_conditions = [
+    {"model_args": {"num_kernel_samples": 500}},
+    {"model_args": {"num_kernel_samples": 1000}},
+    {"model": build_dgp_model(initial_data)},
+]
```
apaleyes: To make sure this doesn't affect CI you may want to leave only one, or none at all, for quick test purposes.

apaleyes (commented):
Looks good! And experimentation example in the notebook illustrates utility of configs well.

docs/notebooks/active_learning.pct.py
uri-granta: I expect the purpose of the "as usual" is to highlight that this bit of information is not new, but one the reader should have previously encountered when reading the tutorials. Though I agree with your later comment that we don't need to be so verbose each time we mention this.

docs/notebooks/deep_gaussian_processes.pct.py
```diff
@@ -69,14 +69,16 @@
...
 #
-# Since DGPs can be hard to build, Trieste provides some basic architectures: here we use the `build_vanilla_deep_gp` method.
+# Few other useful notes regarding building a DGP model. The DGP model requires us to specify the number of inducing points, as we don't have the true posterior. To train the model we have to use a stochastic optimizer; Adam is used by default, but we can use other stochastic optimizers from TensorFlow. GPflux allows us to use the Keras `fit` method, which makes optimizing a lot easier - this method is used in the background for training the model. For this problem we need to modify the default optimizer settings slightly, so we initialize a new optimizer wrapper instance (`Optimizer`) with custom minimization arguments `minimize_args` which are passed to Keras' `fit` method (check [Keras API documentation](https://keras.io/api/models/model_training_apis/#fit-method) for a list of possible arguments).
```
uri-granta, line 74: (we should probably be more consistent regarding whether we say trieste or Trieste)

docs/notebooks/failure_ego.pct.py
```diff
@@ -105,38 +104,29 @@ def create_classification_model(data):
...
 classification_model = create_classification_model(initial_data[FAILURE])
 
+
 # %% [markdown]
 # ## Build Trieste models
 #
 # We now specify how Trieste will use our GPflow models within the BO loop.
 #
-# For our `GPR` model, we will use a standard L-BFGS optimizer from Scipy, whereas we will optimze our `VGP` model using alternate Adam steps (to optimize kernel parameter) and NatGrad steps (to optimize variational parameters).
+# For our `VGP` model we will use a non-default optimization: alternate Adam steps (to optimize kernel parameter) and NatGrad steps (to optimize variational parameters). For this we need to use the `BatchOptimizer` wrapper and set the `use_natgrads` model argument to `True` in our `VariationalGaussianProcess` model wrapper.
```
uri-granta, line 116: so is this notebook currently only working because we use a custom acquisition function that's not compiled (which means it doesn't hit the bug I'm trying to resolve)?

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,209 @@
...
+initial_query_points = search_space.sample_sobol(num_initial_points)
+initial_data = observer(initial_query_points)
+
+
+# %% [markdown]
+# ## Standard way of setting up a model of the objective function
+#
+# The Bayesian optimization procedure estimates the next best points to query by using a probabilistic model of the objective. We'll use Gaussian Process (GP) regression in this tutorial, as provided by GPflow.
+#
+# The GPflow models cannot be used directly in our Bayesian optimization routines, only through a valid model wrapper. Trieste has wrappers that support several popular model. For instance, `GPR` and `SGPR` models from GPflow have to be used with `GaussianProcessRegression` wrapper. These wrappers standardise outputs from all models, deal with preparation of the data and implement additional methods needed for Bayesian optimization.
```
uri-granta:
```suggestion
# The GPflow models cannot be used directly in our Bayesian optimization routines, only through a valid model wrapper. Trieste has wrappers that support several popular models. For instance, `GPR` and `SGPR` models from GPflow have to be used with the `GaussianProcessRegression` wrapper. These wrappers standardise outputs from all models, deal with preparation of the data and implement additional methods needed for Bayesian optimization.
```

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,209 @@
...
+    variance = tf.math.reduce_variance(data.observations)
+    kernel = gpflow.kernels.Matern52(variance=variance, lengthscales=[0.2, 0.2])
+    gpr = gpflow.models.GPR(data.astuple(), kernel, noise_variance=1e-5)
+    return gpr
+
+
+gpflow_model = build_model(initial_data)
+
+# %% [markdown]
+# Usually constructing a GPflow model would be enough, as it is the only required argument for the model wrappers. Wrappers have other arguments - an `optimizer` argument as a rule and potentially some additional model arguments (for example, `num_kernel_samples` in `GaussianProcessRegression`). These arguments are set to sensible defaults and hence typically we can simplify the model building.
```
uri-granta: (or if you're really wedded to using dashes then use an em-dash — )

```suggestion
# Usually constructing a GPflow model would be enough, as it is the only required argument for the model wrappers. Wrappers have other arguments: an `optimizer` argument as a rule and potentially some additional model arguments (for example, `num_kernel_samples` in `GaussianProcessRegression`). These arguments are set to sensible defaults and hence typically we can simplify the model building. 
```

docs/notebooks/model_config.pct.py
uri-granta: Agree. Also, replace - with : or —.

docs/notebooks/model_config.pct.py
uri-granta: Agree. Feel free to ask me if you're unsure how to set up the quicktest overrides.

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,209 @@
...
+# here we specify our experiments
+experiment_conditions = [
+    {"model_args": {"num_kernel_samples": 500}},
+    {"model_args": {"num_kernel_samples": 1000}},
+    {"model": build_dgp_model(initial_data)},
+]
+
+results = []
+for exp in experiment_conditions:
+    model_config = deepcopy(basic_config)
```
uri-granta, line 176: Should we provide a `combine_dict_configs` function that merges two dict configs? This would simplify this code and also handle properly merging nested configs (e.g. if one of the experiments tried to override maxiter).

```
def merge_dict_configs(config1: ModelDictConfig, config2: ModelDictConfig) -> ModelDictConfig:
   ...
```

trieste/models/config.py
```diff
@@ -12,100 +12,181 @@
...
-        for model_type, model_interface in self.supported_models().items():
-            if isinstance(self.model, model_type):
-                return model_interface(self.model, optimizer, **self.model_args)  # type: ignore
+        return model_interface(self.model, optimizer, **self.model_args)  # type: ignore
 
-        raise NotImplementedError(f"Not supported type {type(self.model)}")
 
-
-ModelSpec = Union[ModelConfig, TrainableProbabilisticModel]
+ModelSpec = Union[Dict[str, Any], ModelConfig, TrainableProbabilisticModel]
```
uri-granta: Might be useful to also define an alias

```
ModelDictConfig = Dict[str, Any]
```

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,209 @@
...
+    for key in exp:
+        model_config[key] = exp[key]
+    results.append(run_experiment(model_config))
+
+
+# %% [markdown]
+# ## Registry of supported models
+#
+# Configuration dictionaries are made possible with the `ModelRegistry` that contains mapping between each model (e.g. GPflow or GPflux) and corresponding model wrapper and optimizer wrapper. All models that Trieste currently supports are registered there.
```
uri-granta:
```suggestion
# Configuration dictionaries are made possible with the `ModelRegistry` that contains mapping between each model (e.g. GPflow or GPflux) and the corresponding model wrapper and optimizer wrapper. All models that Trieste currently supports are registered there.
```

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,209 @@
...
+#
+# You can add new models to the registry, in case you have custom models with which you wish to use the configuration dictionaries. Let see an example of this. We will register the `GPMC` model from GPflow that is currently not supported. You would likely need to create a new model wrapper and perhaps a new optimizer wrapper as well, but just for the sake of an example we will borrow here existing wrappers, `GaussianProcessRegression` and `Optimizer`.
```
uri-granta:
```suggestion
# You can add new models to the registry, in case you have custom models with which you wish to use the configuration dictionaries. Let's see an example of this. We will register the `GPMC` model from GPflow that is currently not supported. You would likely need to create a new model wrapper and perhaps a new optimizer wrapper as well, but just for the sake of an example we will borrow here existing wrappers, `GaussianProcessRegression` and `Optimizer`.
```

docs/notebooks/model_config.pct.py
```diff
@@ -0,0 +1,209 @@
...
+
+# %%
+from trieste.models import ModelRegistry
+
+# adding the GPMC model to the registry
+ModelRegistry.register_model(gpflow.models.GPMC, GaussianProcessRegression, Optimizer)
+
+# check if it has been registered
+print(gpflow.models.GPMC in ModelRegistry.get_registered_models())
```
uri-granta, line 196: Is the print necessary?

```suggestion
gpflow.models.GPMC in ModelRegistry.get_registered_models()
```

trieste/models/gpflow/config.py
```diff
@@ -14,35 +14,22 @@
...
+    SVGP: (SparseVariational, BatchOptimizer),
+}
+for model_type, interface_optimizer in _SUPPORTED_MODELS.items():
```
uri-granta: You can do this:

```suggestion
for model_type, (interface, optimizer) in _SUPPORTED_MODELS.items():
```

trieste/models/gpflux/config.py
```diff
@@ -14,35 +14,19 @@
...
-        return models_mapping
-
-    def create_optimizer(self) -> tf.optimizers.Optimizer:
-        return self.optimizer
+# Here we list all the GPflux models currently supported by model interfaces
+# and optimizers, and register them for usage with ModelConfig.
+_SUPPORTED_MODELS: Dict[Type[Any], Tuple[Type[TrainableProbabilisticModel], Type[Optimizer]]] = {
+    DeepGP: (DeepGaussianProcess, Optimizer),
+}
+for model_type, interface_optimizer in _SUPPORTED_MODELS.items():
```
uri-granta:
```suggestion
for model_type, (interface, optimizer) in _SUPPORTED_MODELS.items():
```

trieste/models/config.py
```diff
@@ -12,100 +12,181 @@
...
+        Register a new model type. Note that this will overwrite a registry
+        entry if the model has already been registered.
+
+        :param model_type: The model type.
+        :param interface_type: The interface to be used with the model type.
+        :param optimizer_type: The optimizer to be used with the model type.
+        """
+        if model_type in cls._REGISTRY.keys():
+            warn(
+                f"Model {model_type} has already been registered, " f"you have now overwritten it. "
```
uri-granta:
```suggestion
                f"Model {model_type} has already been registered, you have now overwritten it."
```

tests/util/models/gpflow/models.py
```diff
@@ -116,6 +116,10 @@ def get_kernel(self) -> tfp.math.psd_kernels.PositiveSemidefiniteKernel:
         return self.kernel
 
 
+class GPRcopy(GPR):
```
uri-granta: Is this only used for testing the registry? If so it probably belongs there rather than here (as it's not a generally useful class).

tests/unit/models/test_config.py
```diff
@@ -14,18 +14,61 @@
...
+    ModelRegistry.register_model(GPRcopy, GaussianProcessRegression, Optimizer)
+    model_type = type(gpr_copy_model())
+
+    assert ModelRegistry.get_interface(model_type) == GaussianProcessRegression
+    assert ModelRegistry.get_optimizer(model_type) == Optimizer
+
+
+def test_model_registry_register_model_warning() -> None:
+
+    with pytest.warns(UserWarning) as record:
```
uri-granta, line 70: It's generally bad to have tests that depend on other tests being run first. You can fix it here (and below) by explicitly (re)registering first. (It would be even better to ensure the global state is put back at the end of each of the tests, but that's harder without adding an unregister_model which probably isn't worth it.)

tests/unit/models/gpflow/test_config.py
```diff
@@ -14,72 +14,39 @@
...
+    ]
+) -> None:
 
-    expected_optimizer = tf.optimizers.Adam
+    model_type, interface, optimizer = supported_models
 
-    assert isinstance(model_config.optimizer, expected_optimizer)
+    assert model_type in ModelRegistry.get_registered_models()
+    assert ModelRegistry.get_interface(model_type) == interface
+    assert ModelRegistry.get_optimizer(model_type) == optimizer
```
uri-granta, line 52: Any reason not to (also) build a ModelConfig like before and check the types of the actual constructed optimizer?

tests/unit/models/gpflux/test_config.py
```diff
@@ -14,66 +14,29 @@
...
+    ],
+)
+def test_supported_gpflow_models_are_correctly_registered(
+    supported_models: Tuple[Type[DeepGP], Type[TrainableProbabilisticModel], Type[Optimizer]]
+) -> None:
+
+    model_type, interface, optimizer = supported_models
+
+    assert model_type in ModelRegistry.get_registered_models()
+    assert ModelRegistry.get_interface(model_type) == interface
```
uri-granta, line 41: (same comment as above)

uri-granta (approved):
Overall looks good. Various comments (feel free to ask for clarifications if required).

docs/notebooks/deep_gaussian_processes.pct.py
hstojic: do we use trieste or Trieste everywhere?

docs/notebooks/failure_ego.pct.py
hstojic: probably

docs/notebooks/model_config.pct.py
hstojic: it is

docs/notebooks/model_config.pct.py
hstojic: this is really an illustration only, I don't want to say this is the best way to do it - I'll add a comment reflecting this

tests/unit/models/gpflow/test_config.py
hstojic: huh, not sure whether this is the right place anymore for that - wrappers providing defaults for the optimizers, so this should probably be done in `test_models.py`, not really related to configs
(before it was related to model config's default for the optimizer, but we don't have that anymore, that's something that wrappers should deal with)

docs/notebooks/model_config.pct.py
hstojic: gave it a try, @uri-granta how can I check whether quickrun is working properly?

docs/notebooks/deep_gaussian_processes.pct.py
hstojic: went with Trieste and changed few more notebooks that used trieste

docs/notebooks/model_config.pct.py
apaleyes: Maybe you could try running quicktest.py locally and see what notebook it creates?

hstojic: its not working, probably regular expression in the quickrun.py is not that flexible, I suggest we leave it as is, its two BO runs now with 2 iterations only, so should be quick enough

PR #390, opened 2021-10-28
Fast notebooks for CI

### LINKED ISSUE #371: Improve time it takes to build notebooks
It might be possible to decrease the time some notebooks take to run during the CI build. We may investigate parameters with [papermill](https://github.com/yaananth/papermill) or just use environment variables, and make special arrangements (fewer iterations, less delays, etc) for build environment.

This needs careful thinking, because these notebooks are shown to users as tutorials. So it would be best to avoid exposing some very specific CI shenanigans in the middle of the tutorial.

henrymoss:
Add a DUMMY_RUN  environment variable to tox that sets notebook parameters to dummy values when running CI, e.g. we set number of BO steps to be 2


tox.ini
```diff
@@ -27,6 +27,7 @@ basepython = python3.7
 deps = pip~=20.3
 whitelist_externals = bash
 setenv =
+    DUMMY_RUN = True
```
uri-granta: It's quite ugly to have this strewn in all the tutorials, I think. Users shouldn't have to know or care about our continuous integration system to read the tutorials!

At the very least we should call it something less informal: maybe

```suggestion
    PARTIAL_RUN = True
```

And then use it less conspicuously (and with no mention of testing):

```python
FULL_RUN = not os.environ.get("PARTIAL_RUN")  # full execution or quick partial run?

num_query_points = 10 if FULL_RUN else 2
```

tox.ini
```diff
@@ -27,6 +27,7 @@ basepython = python3.7
...
     CUDA_VISIBLE_DEVICES =
 commands =
```
uri-granta: The other thing I don't like is that we're now not running the full notebooks anymore, which will prevent us from catching a crash that happens e.g. on the 5th loop. One way to fix this is to declare two tox jobs: docs (which does a fulll run) and quickdocs (which does a partial run). We can then run quickdocs in the pre-merge checks and (full) docs after merging, so we still catch these errors before we try releasing.

If you're ok with this, I'm happy to add the docs/quickdocs mechanism into my PR (though we'll need your PR to actually make quickdocs any quicker than docs). Also happy to merge your PR into mine and do them both.

docs/notebooks/active_learning.pct.py
```diff
@@ -10,6 +10,9 @@
 import tensorflow as tf
 import pandas as pd
 
+import os
```
uri-granta: `import os` should really go at the top, before the other imports.

uri-granta (commented):
A few comments. Happy to chat if you like.

hstojic (commented):
Good job!

I have few comments regarding this line in every notebook:
`FULL_RUN = os.environ.get("PARTIAL_RUN") # full execution or quick partial run?`

1. If somebody is running this interactively and this env variable is not defined default will be PARTIAL_RUN, to me it makes more sense that default is FULL_RUN, as a user you want to see the intended results, right? I would make it True unless overriden by the environment variable
2. PARTIAL_RUN is a bit confusing, in the quickdocs tox task you set it to 1, which means FULL_RUN = 1 and then longer versions of notebooks are executed, right? Perhaps use FULL_RUN as the env variable and set it to 0? 
3. Seeing this bit of the code will be confusing to the users, we need a more detailed explanation in a comment above the statement, rather than the current very brief one (`# full execution or quick partial run?`), what is the purpose and in what way they should ignore this piece of code

uri-granta:
The reason it's so confusing at the moment is because there's a rather significant typo! It should say:

> FULL_RUN = not os.environ.get("PARTIAL_RUN") # full execution or quick partial run?

That way it does a full run uness PARTIAL_RUN is explicitly set.

hstojic:
The solution is still a bit ugly, a lot of code in the notebooks that users do not care about and would be confusing.

Why not using smth like papermill? (https://pypi.org/project/papermill/)
Probably more work on our side, specifying separate call for building each notebook with the changed parameters, but would be much cleaner.

I would favour this option, when we go on to create a notebook it would be part of the job. To make it easier we could keep the names of variables the same across notebooks, or we could specify parameters in a YAML file - there are probably ways to do this elegantly.

uri-granta:
Agree that we don't want to expose any of this in the tutorials if we can avoid it, and parametrised notebooks look like they're exactly what we want, thanks! Will investigate papermill.

hstojic:
while we are at it, can we add to the top of each file:
```
import os

os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"
import tensorflow as tf

tf.get_logger().setLevel("ERROR")
```

It would eliminate a lot of irrelevant warnings you tend to see.
Thanks to @apaleyes who started using that :+1:

uri-granta:
> while we are at it, can we add to the top of each file:

Note that this is already done automatically when building the docs. Adding it explicitly only affect opening the notebooks in jupyter-notebook.

.github/workflows/quality-checks.yaml
```diff
@@ -61,5 +61,4 @@ jobs:
         wget -O "$TEMP_DEB" 'https://github.com/jgm/pandoc/releases/download/2.10.1/pandoc-2.10.1-1-amd64.deb' &&
         sudo dpkg -i "$TEMP_DEB"
         rm -f "$TEMP_DEB"
-    - run: tox -e docs
-
+    - run: tox -e quickdocs
```
hstojic, line 64: perhaps rename the task from docs to quickdocs?

docs/notebooks/deep_gaussian_processes.pct.py
```diff
@@ -288,7 +290,7 @@ def build_gp_model(data):
 observer = mk_observer(function)
 
 num_initial_points = 50
-num_acquisitions = 50
+num_steps = 50
```
hstojic, line 293: num_steps appears above as well with a different number, we might want to make sure variable names we are changing have unique names, or somehow ignore the numbers that follow if its ok to substitute with the same nuimber at both places

docs/notebooks/expected_improvement.pct.py
```diff
@@ -226,7 +227,8 @@ def build_model(data):
 # If we need more iterations for better convergence, we can run the optimizer again using the data produced from the last run, as well as the model. We'll visualise the final data.
 
 # %%
-result = bo.optimize(5, result.try_get_final_dataset(), result.try_get_final_model())
+num_steps = 10
```
hstojic, line 230: again repeated with a different number

tox.ini
```diff
@@ -42,3 +42,6 @@ commands =
     docs: pip install . -r notebooks/requirements.txt -c notebooks/constraints.txt
     docs: pip install -r docs/requirements.txt -c docs/constraints.txt
     docs: bash -c "cd docs; make html"
+    quickdocs: pip install . -r notebooks/requirements.txt -c notebooks/constraints.txt
+    quickdocs: pip install -r docs/requirements.txt -c docs/constraints.txt
+    quickdocs: bash -c "cd docs; python notebooks/quickrun/quickrun.py; make html; python notebooks/quickrun/quickrun.py --revert"
```
hstojic, line 47: why is revert needed? I would have thought that GH actions does this manipulation on a server side, not really changing the documents in the repo

hstojic (approved):
looks awesome, great work!
tested and works fine? whats the running time?

docs/notebooks/deep_gaussian_processes.pct.py
hstojic: right, scratch this, you do the former :)

hstojic:
> > while we are at it, can we add to the top of each file:
> 
> Note that this is already done automatically when building the docs. Adding it explicitly only affect opening the notebooks in jupyter-notebook.

exactly, users will be doing that when running them interactively, so its helpful for users if all the warnings are hidden

tox.ini
uri-granta: It's not needed for github, but if you run it locally it's nice to leave the documents unchanged.

docs/notebooks/deep_gaussian_processes.pct.py
uri-granta: The config already ignores the numbers that follow by specifying "num_steps = \d+" as the regex to replace. You're right that if we wanted the "after" values to differ then we'd need to rename the variables.

.github/workflows/quality-checks.yaml
uri-granta: `tox -e docs` is still the recommended way for end users to build a local copy of the documentation, so it should generate the correct tutorials. I've updated the docs README with some details about quickdocs.

uri-granta:
> looks awesome, great work! tested and works fine? whats the running time?

Down to 10 minutes (from over an hour at the start of the week).

apaleyes (approved):
That is quite amazing change, looks great. The only thing that this now lacks is documentation, as we shall have a documented guideline of how to make use of this new technique while implementing a new notebook.

PR #391, opened 2021-10-28
Remove superfluous quality checks and update docs README

uri-granta:
The GitHub merge checks are now taking far too long. Speedups implemented so far:

1. Remove the notebook check, as that's subsumed by the docs check (the former just executes the python in the notebooks, while the latter also generates the docs).
2. Make the tox docs build rune make html just once. We were running it twice before due to bibtex which isn't important for finding errors; we'll still run it twice when we deploy new documentation (not 100% sure it's necessary but no harm).
3. Add some notes in the docs README about how to generate the API docs without running the notebooks at all, which could save people lots of time when debugging docs issues.

### DIFF
--- a/.github/workflows/quality-checks.yaml
+++ b/.github/workflows/quality-checks.yaml
@@ -48,16 +48,6 @@ jobs:
     - run: pip install tox
     - run: tox -e tests
 
-  notebooks:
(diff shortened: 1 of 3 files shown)


henrymoss (approved):
LGTM

PR #394, opened 2021-10-28
Wrapper for models that handles data normalisation

### LINKED ISSUE #379: Data preprocessing / transformations / normalisations

saadhamid96:
This PR introduces the `NormalizationMixin` mixin for handling data transforms for `TrainableProbabilisticModels`. `NormalizationMixin` needs to be subclassed by the user to implement a data transform in practice. The `StandardizationMixin` is provided as an example that normalises query points and observations to zero mean and unit variance. The `data_transformation` notebook is also modified to provide an example that demonstrates usage of the mixin.

Outstanding tasks:
- Allow for updating hyperparameter priors when the normalisation parameters are updated.
- Write docstrings.
- Write tests.


docs/notebooks/data_transformation.pct.py
```diff
@@ -244,6 +244,89 @@ def normalise(x, mean=None, std=None):
 print(f"observation: {observations[arg_min_idx, :]}")
 
 
+# %% [markdown]
+#
+# Now lets try managing data transformation with a `NormalizationMixin`.
+#
+# First, we subclass the mixin to handle our normalization. See `trieste.models.normalization.StandardizationMixin` for an example that normalizes both query points and observations to have zero mean and unit variance. We modify this further to add support for updating the priors over hyperparameters.
```
hstojic: An average python user probably doesn't know about mixins - that might need a bit of explanation or avoid using the term altogether.

trieste/models/normalization.py
```diff
@@ -0,0 +1,193 @@
...
+import tensorflow as tf
+from gpflow import default_float
+from gpflow.utilities.traversal import read_values, multiple_assign
+
+from .interfaces import TrainableProbabilisticModel
+from ..data import Dataset
+from ..types import S, TensorType
+
+
+class NormalizationMixin(TrainableProbabilisticModel):
```
hstojic: I think class name is not descriptive enough and mixin is a technical term, we should try to think of something better that would tell immediately to the user what is used for - what do you think about `ModelWithDataTransform`?

trieste/models/normalization.py
```diff
@@ -0,0 +1,193 @@
...
+        **kwargs
+    ) -> None:
+        normalization_parameters = (
+            {} if normalization_parameters is None else normalization_parameters
+        )
+        self.__dict__.update(normalization_parameters)
+        super().__init__(*args, **kwargs)
+        self._initialize_model_and_normalization_parameters(dataset)
+
+    def _transform_query_points(self, query_points: TensorType) -> TensorType:
```
hstojic: it will be a bit difficult for users to create their custom transforms without docstrings, I suggest we have a set of public ones with docstrings and detailed explanation what they are meant to do to guide the user as well as possible, and a set of private ones with _ that are used in the public ones - then declare private ones as abstract methods. Perhaps we should leave covariance and update parameters as optional, as thats a more complex scenario?

trieste/models/normalization.py
```diff
@@ -0,0 +1,193 @@
...
+        self._update_model_and_normalization_parameters(dataset)
+        transformed_dataset = self._transform_dataset(dataset)
+        super().update(transformed_dataset)  # Will update data only.
+
+    def optimize(self, dataset: Dataset) -> None:
+        transformed_dataset = self._transform_dataset(dataset)
+        return super().optimize(transformed_dataset)
+
+
+class StandardizationMixin(NormalizationMixin):
```
hstojic: we should offer min-max and standardisation, I think that's a bare minimum

hstojic (commented):
I took a quick look and looks quite good so far, thanks for taking a lead here!

trieste/models/normalization.py
saadhamid96: Sure, I think `DataTransformWrapper` might be good since it makes it clear that a subclass that inherits from both the wrapper and the desired model is what the user wants.

docs/notebooks/data_transformation.pct.py
saadhamid96: Sure, I'll use the term wrapper instead

trieste/models/normalization/model_wrappers.py
```diff
@@ -0,0 +1,232 @@
...
+        _process_hyperparameter_dictionary
+        _get_unnormalized_hyperparameter_priors
+        _update_hyperparameter_priors
+    """
+
+    def __init__(
+        self,
+        dataset: Dataset,
+        *model_args: Any,
+        query_point_transformer: DataTransformer,
```
hstojic: I guess its a realistic case that you might want different type of transformations for inputs and outputs, if we go one step further, even different dimensions?

trieste/models/normalization/model_wrappers.py
```diff
@@ -0,0 +1,232 @@
...
+
+from trieste.models.gpflow.models import GaussianProcessRegression
+
+from .transformers import DataTransformer
+from ..interfaces import TrainableProbabilisticModel
+from ...data import Dataset
+from ...types import S, TensorType
+
+
+class DataTransformWrapper(TrainableProbabilisticModel):
```
hstojic: better but not there yet :)
I think its better to have a word model in there as well, so that user can immediately see that this is something that is used on a model

trieste/models/normalization/model_wrappers.py
```diff
@@ -0,0 +1,232 @@
...
+    def optimize(self, dataset: Dataset, *args: Any, **kwargs: Any) -> None:
+        """Wrap the model's `optimize` method to pass in a normalized dataset.
+        
+        :param dataset: Unnormalized dataset.
+        """
+        transformed_dataset = self._transform_dataset(dataset)
+        return super().optimize(transformed_dataset, *args, **kwargs)
+
+
+class GPRDataTransformWrapper(DataTransformWrapper, GaussianProcessRegression):
```
hstojic: huh, this is getting messier, would we need a wrapper per model? supporting this would be a nightmare...

hstojic (commented):
quick interim feedback:
1. I'm not sure that we need a whole package for this, a module with a generic wrapper and transformers we offer could do the trick - I would call it smth like `data_transformation.py`. Model specific wrappers should go in model specific sub-packages, e.g. models/gpflow/data_transformation.py`. If you think we need a package I would give it a more general name, like `data_transformation` or `transformers`, as users might use it for more than just data normalization.
2. I think supporting the more complex option (updating the parameters in each iteration, correcting the model etc) could be too complicated to support, perhaps we should also think of separating a basic case and the more iterative one

docs/notebooks/data_transformation.pct.py
henrymoss: Yeah, make sure you get rid of "mixin"

henrymoss: I don't think we every want to normalize the query points? We just want to be able to scale the search space to [0,1]^D ?

docs/notebooks/data_transformation.pct.py
```diff
@@ -244,6 +244,89 @@ def normalise(x, mean=None, std=None):
...
+#
+# We then create our new model class that handles normalization automatically.
+
+# %%
+from trieste.models.normalization import StandardizationMixin
+
+class StandardizedGaussianProcessRegression(StandardizationMixin, GaussianProcessRegression):
+    def __init__(self, dataset, model, optimizer, num_kernel_samples):
+        super().__init__(dataset, model, optimizer, num_kernel_samples)
+
```
henrymoss: This seems quite complicated for a standard user.  Maybe we could just have a normalizer input to all our model classes?

trieste/objectives/single_objectives.py
```diff
@@ -587,3 +587,37 @@ def trid_10(x: TensorType) -> TensorType:
 The search space for :func:`trid` function is defined over :math:`[-d^2, d^2]` for all i=1,...,d.
 Here, we define it specifically for the 10-dimensional variant.
 """
+
+
+def beale(x: TensorType) -> TensorType:
+    """
+    The Beale function over :math:`[-4.5, 4.5]^2` is multimodal with sharp peaks at the corners of
+    its domain.
```
hstojic: reference for this function?

docs/notebooks/data_transformation.pct.py
```diff
@@ -97,17 +97,18 @@ def build_gp_model(data, x_std = 1.0, y_std = 0.1):
...
-            minimize_args={"options": dict(maxiter=100)}
-        ),
-        num_kernel_samples=100,
-    )
+    return gpr
 
 # build the model
-model = build_gp_model(initial_data, 20, 10000)
+gp_model = build_gp_model(initial_data, 20, 10000)
+model = GaussianProcessRegression(
```
hstojic, line 104: was this really needed? leads to duplication of code below...

docs/notebooks/data_transformation.pct.py
```diff
@@ -244,6 +255,71 @@ def normalise(x, mean=None, std=None):
 print(f"observation: {observations[arg_min_idx, :]}")
 
 
+# %% [markdown]
+# # Data transformation with the DataTransformModelWrapper class
+#
+# Actually, we can achieve the same effect with a standard `BayesianOptimizer` by using a `DataTransformModelWrapper`.
```
hstojic: perhaps stress what is the difference/advantage? e.g. no need to manage the loop, several standard normalisers defined, predictions/samples are rescaled automatically...

docs/notebooks/data_transformation.pct.py
```diff
@@ -244,6 +255,71 @@ def normalise(x, mean=None, std=None):
...
+#
+# First, we create a new model class that inherits from the model wrapper class and the kind of `TrainableProbabilisticModel` that we want to use.
+
+# %%
+from trieste.models.normalization import DataTransformModelWrapper
+
+class GPRwithDataNormalization(DataTransformModelWrapper, GaussianProcessRegression):
```
hstojic: I'm thinking we should offer these so users don't need to define them, say in `models/gpflow/transformers.py`, this is easy to maintain...

docs/notebooks/data_transformation.pct.py
```diff
@@ -244,6 +255,71 @@ def normalise(x, mean=None, std=None):
...
+    pass
+
+
+# %% [markdown]
+#
+# Now we create `DataTransformer`s for query_points and observations, and pass these into our new model class when we construct it.
```
hstojic: for the sake of illustration, we could use max-min transformation on query points and standardisation on observations

docs/notebooks/data_transformation.pct.py
```diff
@@ -244,6 +255,71 @@ def normalise(x, mean=None, std=None):
...
+    observation_transformer=observation_transformer,
+)
+
+# %% [markdown]
+#
+# Then we run Bayesian Optimization as usual and display the results.
+#
+# Again, normalization has improved BO performance.
+
+
```
hstojic, line 313: perhaps adding an illustration of how this is more useful if you want to inspect the model, as now predictions and samples are automatically in the original space, not the normalised space?

docs/notebooks/data_transformation.pct.py
```diff
@@ -244,6 +255,71 @@ def normalise(x, mean=None, std=None):
...
+observations = dataset.observations.numpy()
+
+arg_min_idx = tf.squeeze(tf.argmin(observations, axis=0))
+
+plot_regret_with_min(dataset)
+
+print(f"query point: {query_points[arg_min_idx, :]}")
+print(f"observation: {observations[arg_min_idx, :]}")
+
+
```
hstojic, line 329: lets add a small paragraph explaining you can be more sophisticated - just in few words how, and then point to the more advanced notebook

docs/notebooks/data_transformation.pct.py
hstojic: a small separate paragraph below that explains in some details what happens below the hood?

docs/notebooks/data_transformation.pct.py
```diff
@@ -199,18 +192,36 @@ def normalise(x, mean=None, std=None):
...
 normalised_data = Dataset(query_points=x_sta, observations=y_sta)
 
+# %% [markdown]
+#
+# Note that we also need to modify the search space, from the original $[-100, 100]$ for all 10 dimensions to the normalised space.
+
+# %%
+normalised_lower_boundary, _, _ = normalise(search_space.lower, x_mean, x_std)
+normalised_upper_boundary, _, _ = normalise(search_space.upper, x_mean, x_std)
+normalised_search_space = Box(tf.squeeze(normalised_lower_boundary), tf.squeeze(normalised_upper_boundary))
```
hstojic: thanks for this, more precise :)

trieste/models/normalization.py
```diff
@@ -0,0 +1,350 @@
...
+        """Transform the variance of normalized data to the variance of unnormalized data.
+
+        :param variance: The variance of the unnormalized data. Shape [..., D] if transform
+        operates over multiple dimensions else any shape.
+        :return: The variance of the normalized data. Same shape as `data`.
+        """
+        return variance * self._delta ** 2
+
+
+class DataTransformModelWrapper(TrainableProbabilisticModel):
```
hstojic: lets move this up to the top of the file, those data transformers are feeding into this one, its more sensible to start with this object when inspecting the file

trieste/models/normalization.py
```diff
@@ -0,0 +1,350 @@
...
+    """
+    A wrapper that handles data transformations for :class: `TrainableProbabilisticModel`s.
+
+    Usage requires creating a new class that inherits from this class and the desired model, e.g.
+    ```
+    class GPRwithDataNormalization(DataTransformModelWrapper, GaussianProcessRegression):
```
hstojic: I commented in the notebook, lets provide these simple new objects in model specific directory

docs/notebooks/data_transformation.pct.py
saadhamid96: I've added an explanation that this makes model inspection easier, but it's difficult to illustrate this for a 10D function... Is there a way you would suggest doing the inspection?

docs/notebooks/data_transformation.pct.py
saadhamid96: I agree, there were actually a few methods that aren't part of the common interface which I've now wrapped for the appropriate classes. (The main one is `covariance_between_points` for `GaussianProcessRegression`)

trieste/models/normalization.py
saadhamid96: The issue with this is that the types of the transformers, i.e. `DataTransformer` need to be included in the type hints, so this needs to be defined earlier in the file.

A potential alternative is to move `DataTransfomModelWrapper` into `trieste.models.interfaces`. Or do you think it is too awkward a fit for that module?

trieste/models/normalization.py
saadhamid96: Thanks for the comments! I have addressed most of them but there are still a few minor points that need further discussion

trieste/objectives/single_objectives.py
```diff
@@ -587,3 +587,37 @@ def trid_10(x: TensorType) -> TensorType:
 The search space for :func:`trid` function is defined over :math:`[-d^2, d^2]` for all i=1,...,d.
 Here, we define it specifically for the 10-dimensional variant.
 """
+
+
+def beale(x: TensorType) -> TensorType:
```
hstojic, line 592: is there a specific reason why we need this one?

docs/notebooks/advanced_data_transformation.pct.py
```diff
@@ -0,0 +1,231 @@
...
+    def _update_model_and_normalization_parameters(self, dataset):
+        """Update the model and normalization parameters based on the new dataset.
+        i.e. Denormalize using the old parameters, and renormalize using parameters set from
+        the new dataset.
+
+        :param dataset: New, unnormalized, dataset.
+        """
+        unnormalized_mean = self._observation_transformer.inverse_transform(self._model.mean_function.c)
+        self._observation_transformer.set_parameters(dataset.observations)
+        self._model.mean_function.c.assign(self._observation_transformer.transform(unnormalized_mean))
```
avullo, line 182: Doesn't this assume a constant mean function?

trieste/objectives/single_objectives.py
saadhamid96: Yes, for the advanced data transforms notebook. The Beale function is a bit more interesting than the Trid function since it has multiple modes as well as having high dynamic range. Updating the normalisation and model parameters improves BO performance for the Beale function, but doesn't for the Trid function.

docs/notebooks/advanced_data_transformation.pct.py
saadhamid96: Yes, it does assume constant mean function. In general, updating the normalisation and model parameters like this will depend on the model, which is why this functionality needs to be specified by the user rather than being something we provide a standard implementation for.

hstojic:
this is complicated topic so we will not implement this for the time being, hence we are closing it for now but keeping it around if it comes up again

PR #396, opened 2021-10-29
Fixed to get new_point_score for acquisition from optimization result

sebastianober:
Compiling the `target_func` in the acquisition optimizer can cause things to break when calling the `target_func` to obtain the `chosen_point_score`. This PR changes it so that we get the `chosen_point_score` directly from the Scipy optimizer.


hstojic (commented):
I think this is fine, these two should be equivalent - @henrymoss and @vpicheny is there any acquisition function where this might not be the case?

henrymoss (approved):
LGTM

PR #398, opened 2021-11-01
Refactor of model - sampler interactions

henrymoss:
This PR is a prerequisite for @sebastianober's MCEI acquisition function PR and adding in the functionality from out S-GP-TS paper. 

Previously we had a set of models and a separate set of samplers (of varying types) defined over with our acquisition functions. Certain samplers only worked for certain models and we dealt with this by having lots of error traps to check the suitability of models when attempting to sample from them. This is not a scalable setup, especially as we have @sebastianober's new deepGP models and their associated custom samplers. 

We have three types of samplers
1. reparameterization samplers: as used when optimizing Monte-Carlo acquisition functions, i.e `BatchReparameterizationSampler` and @sebastianober 's new reparam sample for deepGPs
2. trajectory samplers: e.g. using random Fourier features to generate sample paths from models. Note that, decoupled Thompson sampling (from GPflux) would also count as a trajectory samplers.
3. Thompson Samplers: samplers that return samples from the min value of the model, e.g. Gumbel/ExactThompsonSampling. 

This PR has two key parts:
1. I have split our sampler into the three separate types explained above. Note that this change required enabling a jitter param for our `IndependentReparameterizationSampler`, which is probably a nice feature to have (to make it consistent with the batch version) and perhaps not that contentious as `IndependentReparameterizationSampler` is not actually used in our code base!
2. Secondly, I have created two new methods for our models: `reparam_sampler` and `trajectory_sampler` that are defined when defining a new model and call the reparameterization and trajectory samplers relevant for the model. This makes an explicit link between the models and their supported samplers and has allowed me to move the relevant samplers into the models part of the code base. For example, the RFF sampler goes near GaussianProcessRegression and `BatchReparameterizationSampler` now lives near `GPFlowPredictor`. Note that the Thompson samplers still live near acquisition functions.

Crucially, this PR paves the way for @sebastianober to define his custom GPflux trajectory/reparam samplers in a way that means they are easily accessible by the GPFlux model and his new MCEI acquisition functions can work with our existing models and our MC acquisition functions can work with his (i.e. they all just require a model with a model.reparam_sampler method rather than requiring a list of explicit names for supported models).

I have update the sampling used within the MES and GIBBON acquisition functions and the discrete Thompson sampling ruled **to work** with the new changes, however, this part of the code will be made much nicer as soon as this PR is closed. Using the work from this PR, we can now just pass in our desired type of Thompson Sampler when defining these functions/rules which is more Pythonic and requires a lot less code.


trieste/acquisition/rule.py
```diff
@@ -611,22 +611,22 @@ class DiscreteThompsonSampling(AcquisitionRule[TensorType, SearchSpace]):
 
     The model is sampled either exactly (with an :math:`O(N^3)` complexity), or sampled
     approximately through a random Fourier `M` feature decompisition
-    (with an :math:`O(\min(n^3,M^3))` complexity for a model trained on `n` points).
+    (with an :math:`O(\min(n^3,M^3))` complexity for a model trained on `n` points). The number
+    `M` of Fourier features is specified when building the model.
```
uri-granta, line 646: Why does the number of features now live on the model rather than in the rule/acquisition function?

trieste/acquisition/sampler.py
```diff
@@ -19,49 +19,16 @@
...
-    @abstractmethod
-    def sample(self, at: TensorType) -> TensorType:
-        """
-        :param at: Input points that define the sampler.
-        :return: Samples.
-        """
-
-
-class ThompsonSampler(Sampler, ABC):
+class ThompsonSampler(ABC):
```
uri-granta, line 31: It's a bit confusing that the Thompson samplers are here and the others are all now under model. Is there a reason why these can't be moved too (even though they're not generated by the model)?

uri-granta (commented):
(still looking, but a couple of early questions)

trieste/acquisition/rule.py
henrymoss: This is open question about how we want to be able to define what sort of trajectory sampler we package with a model. I thought that defining this when we build the model would be a nice way to do this, e.g. if we included other types of fourier features at some point then we could tell the model which ones it will use. Any thoughts @vpicheny @sebastianober ?

trieste/acquisition/sampler.py
henrymoss: The Thompson samplers are defined for any model and are used just in our acquisition functions. I think it makes sense to keep them here. My next PR, where I tidy up the Thompson sampling rule will make this neater

trieste/models/interfaces.py
```diff
@@ -226,3 +249,91 @@ def optimize(self, dataset: Dataset) -> None:
...
+
+    def __repr__(self) -> str:
+        """"""
+        return f"{self.__class__.__name__}({self._model!r})"
+
+    @abstractmethod
+    def get_trajectory(self) -> TrajectoryFunction:
+        """
+        :return: A trajectory function representing an approximate trajectory from the Gaussian
+            process, taking an input of shape `[N, D]` and returning shape `[N, 1]`
```
sebastianober: Should be more broad than "Gaussian process" right?

trieste/models/interfaces.py
```diff
@@ -226,3 +249,91 @@ def optimize(self, dataset: Dataset) -> None:
...
+        """"""
+        return f"{self.__class__.__name__}({self._sample_size!r}, {self._model!r})"
+
+    @abstractmethod
+    def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorType:
+        """
+        :param at: Input points that define the sampler.
+        :param jitter: The size of the jitter to use when stabilising the Cholesky
+            decomposition of the covariance matrix.
+        :return: Samples.
```
sebastianober: It's not clear to me why we want jitter here. Most of the reparameterization samplers will end up calling the model's `predict` method to get `mean` and `var` with `full_cov=False`, so any need for jitter should be the model's responsibility

trieste/models/interfaces.py
```diff
@@ -226,3 +249,91 @@ def optimize(self, dataset: Dataset) -> None:
...
+        Return a reparametrization sampler providing `num_samples` samples.
+
+        Note that this is not supported by all models.
+
+        :param num_samples: The desired number of samples.
+        :return: The reparametrization sampler.
+        """
+        raise NotImplementedError(f"Model {self!r} does not have a reparametrization sampler")
+
+
```
sebastianober, line 294: I feel like it would be better to have the base classes for the samplers in their own file `samplers.py`, but this would probably cause circular imports?

trieste/models/gpflow/sampler.py
```diff
@@ -0,0 +1,350 @@
...
+            the number of latent model dimensions.
+        :raise ValueError (or InvalidArgumentError): If ``at`` has an invalid shape or ``jitter``
+            is negative.
+        """
+        tf.debugging.assert_shapes([(at, [..., 1, None])])
+        tf.debugging.assert_greater_equal(jitter, 0.0)
+
+        mean, var = self._model.predict(at[..., None, :, :])  # [..., 1, 1, L], [..., 1, 1, L]
+        var = var + jitter
+
```
sebastianober, line 82: Why would this need jitter?

trieste/acquisition/sampler.py
sebastianober: I'm fine with the location of these, but it's perhaps a bit strange that these two quite different samplers will subclass the same `ThompsonSampler` class. Is this the sort of thing you're thinking of tidying?

trieste/models/interfaces.py
henrymoss: jitter has always been something that we specify when calling the sampler. But I agree with you @sebastianober. WHat do you think @vpicheny ?

trieste/models/gpflow/sampler.py
henrymoss: I added this in, so we could have a single ReparameterizationClass. I agree that it doesnt make sense. Any way round this @uri-granta ?

trieste/acquisition/rule.py
```diff
@@ -680,16 +675,14 @@ def acquire(
...
                 self._num_query_points, models[OBJECTIVE]
             )
         else:  # Perform approximate Thompson sampling
-            thompson_sampler = RandomFourierFeatureThompsonSampler(
+            thompson_sampler = ThompsonSamplerFromTrajectory(
                 self._num_query_points,
                 models[OBJECTIVE],
-                datasets[OBJECTIVE],
-                num_features=self._num_fourier_features,
             )
```
sebastianober: I'm a bit confused by this. Why is there a different case for the using Fourier features versus not using Fourier features? Shouldn't we just let the model decide how to get the Thompson samples, since this is discrete Thompson sampling?

sebastianober: For instance the current trajectory sampler for the DGP models doesn't use Fourier features at all

henrymoss: This isn't new for this PR. But we previously allowed the use of RFF for getting the discrete Thompson samples if you had a big model or wanted a coarse grid of candidate points. This will be refactored properly in the next PR

tests/unit/models/gpflow/test_sampler.py
```diff
@@ -0,0 +1,313 @@
...
+        gpflow.kernels.RBF()
+    )  # need a gpflow kernel object for random feature decompositions
+    dataset = Dataset(
+        tf.constant([[-2.0]], dtype=tf.float64), tf.constant([[4.1]], dtype=tf.float64)
+    )
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        RandomFourierFeatureTrajectorySampler(model, dataset, num_features=num_features)
+
+
+def test_rff_trajecotry_sampler_raises_for_a_non_gpflow_kernel() -> None:
```
sebastianober: "trajecotry"
```suggestion
def test_rff_trajectory_sampler_raises_for_a_non_gpflow_kernel() -> None:
```

tests/unit/models/gpflow/test_sampler.py
```diff
@@ -0,0 +1,313 @@
...
+def test_batch_reparametrization_sampler_sample_raises_for_inconsistent_batch_size() -> None:
+    sampler = BatchReparametrizationSampler(100, QuadraticMeanAndRBFKernel())
+    sampler.sample(tf.constant([[0.0], [1.0], [2.0]]))
+
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        sampler.sample(tf.constant([[0.0], [1.0]]))
+
+
+@pytest.mark.parametrize("num_features", [0, -2])
+def test_rff_trajecotry_sampler_raises_for_invalid_number_of_features(
```
sebastianober:
```suggestion
def test_rff_trajectory_sampler_raises_for_invalid_number_of_features(
```

tests/unit/models/gpflow/test_sampler.py
sebastianober: basically the same for all attempts at "trajectory" in this file...

sebastianober:
Looks good to me in general. My main issue is that `TrajectorySampler` implicitly seems to be assuming we'll use RFF (or similar) for all models, but I don't think this should necessarily be the case, and should allow for a broader range of trajectory samplers.

henrymoss:
> Looks good to me in general. My main issue is that `TrajectorySampler` implicitly seems to be assuming we'll use RFF (or similar) for all models, but I don't think this should necessarily be the case, and should allow for a broader range of trajectory samplers.

Yeah. So the idea will be to define what sort of trajectory sampler you want when you define the model. For now that is defaulting to RFF, but as more methods become available (e.g. quadrature or even exact sampling) you will be able to choose between them. This is similar to Gpflux, where you need to define you chosen kernel decomposition when building the model.

trieste/acquisition/rule.py
sebastianober: @henrymoss we discussed this right? i think we can resolve

trieste/acquisition/rule.py
```diff
@@ -611,22 +611,22 @@ class DiscreteThompsonSampling(AcquisitionRule[TensorType, SearchSpace]):
...
-        num_fourier_features: Optional[int] = None,
+        use_fourier_features: bool = False,
     ):
         """
         :param num_search_space_samples: The number of points at which to sample the posterior.
         :param num_query_points: The number of points to acquire.
-        :num_fourier_features: The number of features used to approximate the kernel. We
-            recommend first trying 1000 features, as this typically perfoms well for a wide
-            range of kernels. If None, then we perfom exact Thompson sampling.
+        :nrandom_fourier_features: If True then approximate the kernel using random Fourier
```
hstojic: nrandom_fourier_features -> use_fourier_features?

trieste/acquisition/sampler.py
```diff
@@ -194,330 +172,48 @@ def binary_search(val: float) -> float:  # Find empirical interquartile range
...
-points (each of dimension `D`) i.e. takes input of shape `[N, D]` and returns
-shape `[N, 1]`.
-
-A key property of these trajectory functions is that the same sample draw is evaluated
-for all queries. This property is known as consistency.
-"""
-
-
-class RandomFourierFeatureThompsonSampler(ThompsonSampler):
+class ThompsonSamplerFromTrajectory(ThompsonSampler):
```
hstojic, line 185: I'm a bit confused by the name - why do we call these samples trajectories?

trieste/models/gpflow/__init__.py
```diff
@@ -21,4 +20,9 @@
 from .config import GPflowModelConfig
 from .interface import GPflowPredictor
 from .models import GaussianProcessRegression, SparseVariational, VariationalGaussianProcess
+from .sampler import (
```
hstojic, line 23: you mean .. not . ?

trieste/models/gpflow/interface.py
```diff
@@ -91,3 +92,11 @@ def optimize(self, dataset: Dataset) -> None:
         :param dataset: The data with which to optimize the `model`.
         """
         self.optimizer.optimize(self.model, dataset)
+
+    def reparam_sampler(self, num_samples: int) -> ReparametrizationSampler:
```
hstojic, line 96: what about having a generic `sampler` method and you have to pass a `ModelSampler` object? 

depending on the need a user could then pass either an RFF based sampler or reparametrization based method or something else, if you don't use acq functions that need these then you don't pass anything

defining a method for each type  of a sampler does not seem like a great solution, is there a problem with decoupling model and sampler implementation like this?

hstojic: also, if you have this sampler, why not `trajectory_sampler` as well?

trieste/models/interfaces.py
```diff
@@ -92,6 +94,27 @@ def get_observation_noise(self) -> TensorType:
         """
         raise NotImplementedError(f"Model {self!r} does not provide scalar observation noise")
 
+    def reparam_sampler(self, num_samples: int) -> ReparametrizationSampler:
```
hstojic, line 97: what about having a generic `sampler` method and you have to pass say a `ModelSampler` object (a parent abstract class for all samplers)?

depending on the need a user could then pass either an RFF based sampler or reparametrization based method or something else, if you don't use acq functions that need these then you don't pass anything

defining a method for each type of a sampler does not seem like a great solution, is there a problem with decoupling model and sampler implementation like this?

trieste/models/interfaces.py
hstojic: I agree, samplers are distinct enough to warrant a separate file - what about only a generic `ModelSampler` class defined here and used above as output type, and move all of these to a `sampler.py` file (as subclasses) - that would avoid circular import issue

trieste/models/gpflow/models.py
```diff
@@ -38,7 +39,11 @@ class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel):
     """
 
     def __init__(
-        self, model: GPR | SGPR, optimizer: Optimizer | None = None, num_kernel_samples: int = 10
+        self,
+        model: GPR | SGPR,
+        optimizer: Optimizer | None = None,
+        num_kernel_samples: int = 10,
+        num_rff_features: int = 1000,
```
hstojic, line 48: choice of a sampler type seems to some extent independent from the model, or at least you could have more than 1 option, so these kind of arguments would be more appropriate at the sampler level...

hstojic:
had a quick look (not yet in details), here are two main comments:  
1. I would also move TS part to the `models` package, yes they are used by acq functions, but so are these samplers that you have moved. All of them are sampling from the model, one way or another.  I would be confused as a user by having samplers in two different places.
2. instead of having a number of `*_sampler` methods in the model lets have a single generic `sampler` and pass a sampler object at initialising a model, if one wants to use acq functions that require them. Acq functions don't necessarily care about the type of sampler so this seems like a better solution. We could even dispense with this and pass `Sampler` object directly to the acquisition function, it would amount to the same thing.

henrymoss:
> had a quick look (not yet in details), here are two main comments:
> 
>     1. I would also move TS part to the `models` package, yes they are used by acq functions, but so are these samplers that you have moved. All of them are sampling from the model, one way or another.  I would be confused as a user by having samplers in two different places.
> 
>     2. instead of having a number of `*_sampler` methods in the model lets have a single generic `sampler` and pass a sampler object at initialising a model, if one wants to use acq functions that require them. Acq functions don't necessarily care about the type of sampler so this seems like a better solution. We could even dispense with this and pass `Sampler` object directly to the acquisition function, it would amount to the same thing.

Cheers. 

1. the TS samplers work for **any** model and so it would be unclear where to put these in the models bit, I guess in the top level of models?
2. Its completely likely that we might want to use both a reparam_sampler and a trajectory_sampler for the same model. Your proposed solution would not allow this. 

Moreover, acquisiton functions **do care** about the type of sampler. For example, the Monte-Carlo based acquisition funtions **need** a reparam sampler for them to work, a trajectory sampler would not really be appropriate in this case. Whereas,a  new Thompson sampling function (for cts Thompson sampling) would need trajectories (i.e. sample paths), but would not want a reparam sampler!

trieste/acquisition/sampler.py
henrymoss: Trajectories is a well used term in the literature referring to sample paths from the GP. These are smooth cts function draws that can be optimized (using gradients) to find their optima, e.g. for Thompson sampling.

henrymoss: Note that these are distinct to what is provided by a reparam sampler, which just provides (repeatable) samples at a fixed set of points.

trieste/acquisition/sampler.py
henrymoss: indeed

trieste/acquisition/sampler.py
hstojic: I don't see why people would call them a trajectory, besides that they look like that visually, we can be more precise than that e.g. calling them function draws 

but fair enough, if its a widespread name, sounds good

hstojic:
> > had a quick look (not yet in details), here are two main comments:
> > ```
> > 1. I would also move TS part to the `models` package, yes they are used by acq functions, but so are these samplers that you have moved. All of them are sampling from the model, one way or another.  I would be confused as a user by having samplers in two different places.
> > 
> > 2. instead of having a number of `*_sampler` methods in the model lets have a single generic `sampler` and pass a sampler object at initialising a model, if one wants to use acq functions that require them. Acq functions don't necessarily care about the type of sampler so this seems like a better solution. We could even dispense with this and pass `Sampler` object directly to the acquisition function, it would amount to the same thing.
> > ```
> 
> Cheers.
> 
> 1. the TS samplers work for **any** model and so it would be unclear where to put these in the models bit, I guess in the top level of models?
> 2. Its completely likely that we might want to use both a reparam_sampler and a trajectory_sampler for the same model. Your proposed solution would not allow this.
> 
> Moreover, acquisiton functions **do care** about the type of sampler. For example, the Monte-Carlo based acquisition funtions **need** a reparam sampler for them to work, a trajectory sampler would not really be appropriate in this case. Whereas,a new Thompson sampling function (for cts Thompson sampling) would need trajectories (i.e. sample paths), but would not want a reparam sampler!

Re 1) yes, that was my thinking, putting them at the parent level, say `models/sampler.py`, following up on some comments above, we could have reparametrization sampler and some others there - if we opt for having a generic `ModelSampler` in `models/interfaces.py`

Re 2) that's true, it wouldn't work if we would need both - I couldn't think of such a case, where would you need both? perhaps if you want to reuse same model for trying more than one acq fnc? Regarding the other arguments, for TS I completely agree, but for MCEI its not clear to me that it needs to be reparametrization based sampler. In any case, I was just thinking of making things more general here. Perhaps we can just take practical route and do it this way and if we see a better way down the line, we can modify it again

trieste/models/interfaces.py
```diff
@@ -226,3 +249,102 @@ def optimize(self, dataset: Dataset) -> None:
...
+"""
+
+
+class TrajectorySampler(ABC):
+    r"""
+    This class builds functions that approximate a trajectory sampled from an
+    underlying :class:`ProbabilisticModel`.
+
+    Unlike the :class:`ReparametrizationSampler`, a :class:`TrajectorySampler` provides
+    consistent samples (i.e ensuring that the same sample draw is used for all evalutions
```
apaleyes:
```suggestion
    consistent samples (i.e ensuring that the same sample draw is used for all evaluations
```

trieste/models/gpflow/sampler.py
apaleyes: Can you explain a bit more what the problem is? As in, what classes would we have if you didn't add jitter here?

trieste/models/gpflow/sampler.py
```diff
@@ -0,0 +1,350 @@
...
+        try:
+            self._noise_variance = model.get_observation_noise()
+            self._kernel = model.get_kernel()
+        except (NotImplementedError, AttributeError):
+            raise ValueError(
+                """
+            Thompson sampling with random Fourier features only currently supports models
+            with a Gaussian likelihood and an accessible kernel attribute.
+            """
```
apaleyes: something funny with indentation here

trieste/models/gpflow/sampler.py
```diff
@@ -0,0 +1,350 @@
...
+        )  # prep feature functions at data
+
+        if (
+            self._num_features < self._num_data
+        ):  # if m < n  then calculate posterior in design space (an m*m matrix inversion)
+            self._theta_posterior = self._prepare_theta_posterior_in_design_space()
+        else:  # if n <= m  then calculate posterior in gram space (an n*n matrix inversion)
+            self._theta_posterior = self._prepare_theta_posterior_in_gram_space()
+
+        self._pre_calc = False  # Flag so we only calculate the posterior for the weights once.
```
apaleyes: that doesn't seem to be used anywhere

trieste/acquisition/function.py
```diff
@@ -1098,13 +1095,21 @@ def prepare_acquisition_function(
...
         _partition_bounds = prepare_default_non_dominated_partition_bounds(_reference_pt, _pf.front)
 
-        sampler = BatchReparametrizationSampler(self._sample_size, model)
+        try:
+            sampler = model.reparam_sampler(self._sample_size)
+        except (NotImplementedError):
+            raise ValueError(
+                """
+                The batch Monte-Carlo expected hyper-volume improvment acquisition function
+                only supports models with a reparam_sampler method.
```
apaleyes:
```suggestion
                The batch Monte-Carlo expected hyper-volume improvement acquisition function
                only supports models that implement a reparam_sampler method.
```

trieste/acquisition/function.py
```diff
@@ -1303,7 +1307,18 @@ def __init__(self, sample_size: int, model: ProbabilisticModel, eta: TensorType,
...
         self._sample_size = sample_size
-        self._sampler = BatchReparametrizationSampler(sample_size, model)
+
+        try:
+            sampler = model.reparam_sampler(self._sample_size)
+        except (NotImplementedError):
+            raise ValueError(
+                """
+                The batch Monte-Carlo expected improvment acquisition function
+                only supports models with a reparam_sampler method.
```
apaleyes: ditto

trieste/acquisition/function.py
```diff
@@ -1844,30 +1859,28 @@ def __init__(
...
-            - ``num_fourier_features`` is negative or zero, or
-            - ``num_fourier_features`` is specified and ``use_thompson`` is `False`
+            - ``use_fourier_features`` is `True` and ``use_thompson`` is `False`
         """
         tf.debugging.assert_positive(num_samples)
         tf.debugging.assert_positive(grid_size)
 
-        if num_fourier_features is not None:
+        if use_fourier_features:
             tf.debugging.Assert(use_thompson, [])
```
apaleyes: AN error message would be nice, explaining that one param can only be used if the other one is True

trieste/models/gpflow/models.py
```diff
@@ -47,6 +52,10 @@ def __init__(
         :param num_kernel_samples: Number of randomly sampled kernels (for each kernel parameter) to
             evaluate before beginning model optimization. Therefore, for a kernel with `p`
             (vector-valued) parameters, we evaluate `p * num_kernel_samples` kernels.
+        :param num_rff_features: The number of random Foruier features used to approximate the
```
apaleyes, line 57:
```suggestion
        :param num_rff_features: The number of random Fourier features used to approximate the
```

trieste/models/interfaces.py
apaleyes: Second that, method per sampler type doesn't sound great at the first glance

apaleyes (commented):
Found it very hard to follow at the first attempt, so will try again a bit later

trieste/acquisition/rule.py
vpicheny: this is also how I coded it in my quantile experiment

trieste/models/interfaces.py
vpicheny: no strong opinion, but we need to make sure that we can tweak those jitters when needed, they sometimes have a too strong influence on the results.

trieste/acquisition/function.py
```diff
@@ -1293,8 +1298,7 @@ def update_acquisition_function(
 class batch_monte_carlo_expected_improvement(AcquisitionFunctionClass):
     def __init__(self, sample_size: int, model: ProbabilisticModel, eta: TensorType, jitter: float):
         """
-
-        :param sampler:  BatchReparametrizationSampler.
+        :param sampler:  ReparametrizationSampler.
```
vpicheny: you do need a batch sampler here, right?

trieste/acquisition/sampler.py
vpicheny: My first impression is that this is the right call. Thompson samplers are different in nature because we assume a particular usage of the model (e.g. it is an objective and we want to minimise it), while the rest is a modelling thing (how do you get a sample from the joint distribution of your model).

trieste/acquisition/sampler.py
vpicheny: Trajectory or path are equally good I think.

trieste/models/gpflow/sampler.py
henrymoss: Removed jitter for indep

trieste/models/gpflow/models.py
henrymoss: In the PR when we introduce more than one trajectory sampler type to Trieste, I will make sense to add support for this. Will leave as is for now

trieste/acquisition/function.py
henrymoss: Actually not. This is a type guide. BatchReparameterizationSampler is the one specific to Gaussian models

trieste/acquisition/function.py
henrymoss: outdated

trieste/models/interfaces.py
uri-granta: Except reparametrization and trajectory samplers don't behave the same (they expose completely different methods) so it would be weird to have a method that can return either one of them, even if models didn't support both.

trieste/models/interfaces.py
```diff
@@ -241,6 +264,105 @@ def log(self) -> None:
...
+        Return a reparametrization sampler providing `num_samples` samples across
+        all the models in the model stack.
+
+        :param num_samples: The desired number of samples.
+        :return: The reparametrization sampler.
+        :raise NotImplementedError: If the models in the stack do not share the
+            same :meth:`reparam_sampler`.
+        """
+
+        sampler_types = [type(model.reparam_sampler(1)) for model in self._models]
```
uri-granta: As discussed, this optimization is only guaranteed to work for BatchReparameterizationSamplers, so you should restrict it to that. Also if possible you should definitely fall back to generating all the reparameterization samplers individually and  creating a stack sampler that combines the samples. 

Furthermore, since sampler initialisation is cheap for BatchReparameterizationSamplers but won't necessarily be for or other types, you might as well generate the actual samples when checking:

```suggestion
        samplers = [model.reparam_sampler(num_samples) for model in self._models]
        if all(isinstance(sampler, BatchReparameterizationSampler) for sampler in samplers):
            return BatchReparameterizationSampler(num_samples, self)
        else:
            ...
```

trieste/models/interfaces.py
```diff
@@ -241,6 +264,105 @@ def log(self) -> None:
...
+                """
+            )
+        else:  # return the shared sampler rebuilt for the whole model stack
+            shared_sampler = sampler_types[0]
+            return shared_sampler(num_samples, self)
+
+
+class ReparametrizationSampler(ABC):
+    r"""
+    This sampler employs the *reparameterization trick* to draw samples from a
```
uri-granta:
```suggestion
    These samplers employs the *reparameterization trick* to draw samples from a
```

trieste/models/interfaces.py
```diff
@@ -241,6 +264,105 @@ def log(self) -> None:
...
+            raise NotImplementedError(
+                f"""
+                Reparameterization sampling is only currently supported for model
+                stacks built from models that use the same reparameterization sampler,
+                however, received samplers of types {unique_sampler_types}.
+                """
+            )
+        else:  # return the shared sampler rebuilt for the whole model stack
+            shared_sampler = sampler_types[0]
+            return shared_sampler(num_samples, self)
```
uri-granta: Can ModelStacks also generate a stack trajectory sampler from individual trajectory samplers in the stack?

trieste/models/gpflow/sampler.py
```diff
@@ -0,0 +1,348 @@
...
+from ...utils import DEFAULTS
+from ..interfaces import (
+    ProbabilisticModel,
+    ReparametrizationSampler,
+    TrajectoryFunction,
+    TrajectorySampler,
+)
+
+
+class IndependentReparametrizationSampler(ReparametrizationSampler):
```
uri-granta, line 36: What's the purpose of this, given that it's not the reparam sampler of any mode in trieste? How and when are people supposed to us it?

trieste/models/gpflow/sampler.py
```diff
@@ -0,0 +1,348 @@
...
+
+        # _eps is essentially a lazy constant. It is declared and assigned an empty tensor here, and
+        # populated on the first call to sample
+        self._eps = tf.Variable(
+            tf.ones([0, 0, sample_size], dtype=tf.float64), shape=[None, None, sample_size]
+        )  # [0, 0, S]
+
+        # for some reason graph compilation is resulting in self._eps reporting the wrong shape
+        # we therefore use an extra boolean variable to keep track of whether it's initialised
+        self._initialized = tf.Variable(False)
```
uri-granta: IMPORTANT: because batch_monte_carlo_expected_improvement previously only used BatchReparametrizationSampler it could reset it by setting _initialized to False. Now that it can use arbitrary samplers, we probably need to add a reset_sampler() method to the ABC and have it reset self._initialized to False when called, to ensure new samplers don't forget to do this.

trieste/acquisition/sampler.py
```diff
@@ -186,358 +174,55 @@ def binary_search(val: float) -> float:  # Find empirical interquartile range
...
-
         if self._sample_min_value:
             thompson_samples = tf.zeros([0, 1], dtype=at.dtype)  # [0,1]
         else:
             thompson_samples = tf.zeros([0, tf.shape(at)[1]], dtype=at.dtype)  # [0,D]
 
-        for _ in tf.range(self._sample_size):
-            sampled_trajectory = self.get_trajectory()
+        for _ in tf.range(sample_size):
+            sampled_trajectory = trajectory_sampler.get_trajectory()
```
uri-granta, line 225: (this file all looks good!)

trieste/acquisition/function/function.py
```diff
@@ -674,7 +671,18 @@ def __init__(self, sample_size: int, model: ProbabilisticModel, eta: TensorType,
...
+            sampler = model.reparam_sampler(self._sample_size)
+        except (NotImplementedError):
+            raise ValueError(
+                """
+                The batch Monte-Carlo expected improvement acquisition function
+                only supports models that implement a reparam_sampler method.
+                """
+            )
+
+        self._sampler = sampler
```
uri-granta, line 687: (see comment above about the need to reset sampler in the update method)

trieste/acquisition/function/entropy.py
```diff
@@ -362,13 +342,8 @@ def _update_quality_term(
...
         query_points = tf.concat([dataset.query_points, query_points], 0)
-        self._min_value_samples = _get_min_value_samples(
-            model,
-            dataset,
-            sampled_points=query_points,
-            num_samples=self._num_samples,
-            use_thompson=self._use_thompson,
-            num_fourier_features=self._num_fourier_features,
+        self._min_value_samples = self._min_value_sampler.sample(
+            model, self._num_samples, query_points
```
uri-granta, line 346: (this file definitely looks better!)

tests/integration/test_bayesian_optimization.py
```diff
@@ -94,9 +95,10 @@ def OPTIMIZER_PARAMS() -> Tuple[
             (
                 22,
                 EfficientGlobalOptimization(
-                    MinValueEntropySearch(BRANIN_SEARCH_SPACE, num_fourier_features=1000).using(
-                        OBJECTIVE
-                    )
+                    MinValueEntropySearch(
+                        BRANIN_SEARCH_SPACE,
+                        min_value_sampler=ThompsonSamplerFromTrajectory(sample_min_value=True),
+                    ).using(OBJECTIVE)
```
uri-granta, line 101: as an aside, we don't need to explicitly use `using` when constructing an EGO

tests/unit/models/test_interfaces.py
```diff
@@ -177,3 +178,36 @@ def _assert_data(self, dataset: Dataset) -> None:
...
+def test_model_stack_reparam_sampler_raises_for_submodels_without_reparam_sampler() -> None:
+    model01 = _QuadraticModel([0.0, 0.5], [1.0, 0.3])
+    model2 = QuadraticMeanAndRBFKernel()
+    model_stack, _ = ModelStack((model01, 2), (model2, 1)), (model01, model2)  # type: ignore
+
+    with pytest.raises(NotImplementedError):
+        model_stack.reparam_sampler(1)
+
+
+def test_model_stack_reparam_sampler() -> None:
```
uri-granta, line 192: (model stack would need more tests if you make the changes I suggested)

tests/util/acquisition/sampler.py
```diff
@@ -1,7 +1,7 @@
 import tensorflow as tf
 
 from tests.util.models.gpflow.models import QuadraticMeanAndRBFKernel
-from trieste.acquisition.sampler import BatchReparametrizationSampler
+from trieste.models.gpflow import BatchReparametrizationSampler
```
uri-granta, line 4: Can you write somewhere (either in the PR notes or in the commit log) what breaking import changes there are so we can include them in the next release notes?

tests/util/models/gpflow/models.py
```diff
@@ -123,6 +139,32 @@ def mock_data() -> tuple[tf.Tensor, tf.Tensor]:
...
+        super().__init__(
+            x_shift=x_shift, kernel_amplitude=kernel_amplitude, noise_variance=noise_variance
+        )
+        self._dataset = dataset
+
+    def trajectory_sampler(self) -> TrajectorySampler:
+        return RandomFourierFeatureTrajectorySampler(self, self._dataset, 100)
+
+    def reparam_sampler(self, num_samples: int) -> ReparametrizationSampler:
+        return BatchReparametrizationSampler(num_samples, self)
```
uri-granta, line 165: (might it be useful to have this use IndependentReparametrizationSampler instead or does that not make sense?)

uri-granta (commented):
Looks good overall! A few comments.

trieste/models/interfaces.py
henrymoss: Made your change. Will leave the general solution for stacks of models with different samplers for when we include another type of reparam sampler and can see what it actually looks like!

trieste/models/interfaces.py
henrymoss: Yeah they probably could. Will leave this for when we need to implement it though!

trieste/models/gpflow/sampler.py
henrymoss: Its currently relied on in other projects, but we will be able to remove it soonish!

trieste/models/gpflow/sampler.py
henrymoss: which ABC?

henrymoss: Done!

tests/unit/models/test_interfaces.py
henrymoss: Left changes for other PR (as described above)

henrymoss:
Thanks @uri-granta , I've made your changes. Note, that I had to make a new sampler.py file in the models folder to avoid cicular imports from `BatchReparametrizationSampler`

tests/util/models/gpflow/models.py
henrymoss: This needs to be the batch one, e.g. for batch Monte-Carlo EI

trieste/models/interfaces.py
```diff
@@ -272,25 +272,104 @@ def reparam_sampler(self, num_samples: int) -> ReparametrizationSampler:
...
-        samplers = [model.reparam_sampler(num_samples) for model in self._models]
-
-        if all(isinstance(sampler, BatchReparametrizationSampler) for sampler in samplers):
-            return BatchReparametrizationSampler(
-                num_samples, self
-            )  # return the shared sampler rebuilt for the whole model stack
-        else:
+        sampler_types = [type(model.reparam_sampler(1)) for model in self._models]
+        unique_sampler_types = set(sampler_types)
+        if len(unique_sampler_types) > 1:
```
uri-granta, line 282: I'm confused. Did you make the change and then revert it?

(also as an aside the code should actually check for type equality, not instance, as the optimization doesn't support using subtypes of BatchReparameterizationSampler)

```python
if all(type(sampler) is BatchReparameterizationSampler for sampler in samplers):
```

trieste/models/gpflow/sampler.py
```diff
@@ -0,0 +1,348 @@
...
+        """
+        super().__init__(sample_size, model)
+
+        # _eps is essentially a lazy constant. It is declared and assigned an empty tensor here, and
+        # populated on the first call to sample
+        self._eps = tf.Variable(
+            tf.ones([sample_size, 0], dtype=tf.float64), shape=[sample_size, None]
+        )  # [S, 0]
+
+        self._initialized = tf.Variable(False)
```
uri-granta: Surely this isn't required any more given it's set automatically in the ABC?

trieste/models/gpflow/sampler.py
uri-granta: (ditto, and the comment can go too)

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+    @abstractmethod
+    def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorType:
+        """
+        :param at: Input points that define the sampler.
+        :param jitter: The size of the jitter to use when stabilising the Cholesky
+            decomposition of the covariance matrix.
+        :return: Samples.
+        """
+
+    def reset_sampler(self) -> None:
```
uri-granta, line 334: Looks good.

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+
+        :param num_samples: The desired number of samples.
+        :return: The reparametrization sampler.
+        :raise NotImplementedError: If the models in the stack do not share the
+            same :meth:`reparam_sampler`.
+        """
+
+        sampler_types = [type(model.reparam_sampler(1)) for model in self._models]
+        unique_sampler_types = set(sampler_types)
+        if len(unique_sampler_types) > 1:
```
uri-granta: I'm confused. Did you make this change then revert it? Happy to discuss if doesn't seem to work.

Also, my original suggestion isn't quite right. You should check for type equality, not instance, as the optimization won't work correctly if any of the samplers are subtypes of BatchReparameterizationSampler:

```python
if all(type(sampler) is BatchReparameterizationSampler for sampler in samplers):
```

uri-granta (commented):
Reset function looks good, remaining question about stack model.

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+class ReparametrizationSampler(ABC):
+    r"""
+    Thes samplers employ the *reparameterization trick* to draw samples from a
+    :class:`ProbabilisticModel`\ 's predictive distribution  across a discrete set of
+    points.
+    """
+
+    def __init__(self, sample_size: int, model: ProbabilisticModel):
+        """
+        :param sample_size: The desired number of samples.
```
uri-granta, line 309: Maybe add a comment that ModelStack currently assumes that all ReparametrizationSampler constructors have just these parameters and will not work with more complicated constructors.

trieste/acquisition/__init__.py
```diff
@@ -74,10 +74,3 @@
     SingleModelGreedyAcquisitionBuilder,
     UpdatablePenalizationFunction,
 )
-from .sampler import (
-    BatchReparametrizationSampler,
-    ExactThompsonSampler,
```
hstojic, line 79: how come you don't need to import Thompson samplers from this folder?

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+        as their reparameterization sampler.
+
+        :param num_samples: The desired number of samples.
+        :return: The reparametrization sampler.
+        :raise NotImplementedError: If the models in the stack do not share the
+            same :meth:`reparam_sampler`.
+        """
+
+        sampler_types = [type(model.reparam_sampler(1)) for model in self._models]
+        unique_sampler_types = set(sampler_types)
```
uri-granta: I would still simplify this and prepare it for when/if we support heterogeneous sampler types:

```python
samplers = [model.reparam_sampler(num_samples) for model in self._models]
unique_sampler_types = set(type(sampler) for sampler in sampler_types)
if len(unique_sampler_types) == 1:
    # currently assume that all sampler constructors look the same
    shared_sampler_type = type(samplers[0])
    return shared_sampler_type(num_samples, self)
else:
    raise NotImplementedError(...)
```

trieste/acquisition/function/entropy.py
```diff
@@ -61,39 +56,38 @@ def __init__(
         search_space: SearchSpace,
         num_samples: int = 5,
         grid_size: int = 1000,
-        use_thompson: bool = True,
-        num_fourier_features: Optional[int] = None,
+        min_value_sampler: Optional[ThompsonSampler] = None,
```
hstojic, line 59: entropy function look cleaner now, great job!

uri-granta (approved):
Just a couple more of comments

trieste/acquisition/function/function.py
```diff
@@ -664,8 +662,7 @@ def update_acquisition_function(
 class batch_monte_carlo_expected_improvement(AcquisitionFunctionClass):
     def __init__(self, sample_size: int, model: ProbabilisticModel, eta: TensorType, jitter: float):
         """
-
-        :param sampler:  BatchReparametrizationSampler.
+        :param sampler:  ReparametrizationSampler.
```
hstojic, line 667: arguments are not right, we need `sample_size` and `model` no?

trieste/models/gpflow/models.py
```diff
@@ -27,9 +27,10 @@
 from ...data import Dataset
 from ...types import TensorType
 from ...utils import DEFAULTS, jit
-from ..interfaces import FastUpdateModel, TrainableProbabilisticModel
+from ..interfaces import FastUpdateModel, TrainableProbabilisticModel, TrajectorySampler
```
hstojic, line 30: same as with `FastUpdateModel` we should probably have `ReparametrizationSamplerModel` and `TrajectorySamplerModel` interface - then we can also be more precise in each acquisition function what type of probabilistic model user needs 

(although we still need to check existence of methods at runtime as users might not use mypy)

trieste/models/gpflow/models.py
```diff
@@ -236,6 +251,16 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
 
+    def trajectory_sampler(self) -> TrajectorySampler:
+        """
+        Return a trajectory sampler. For :class:`GaussianProcessRegression`, we build
+        trajectories using a random Fourier feature approximation.
+
+        :return: The trajectory sampler.
+        """
+        models_data = Dataset(self.model.data[0].value(), self.model.data[1].value())
+        return RandomFourierFeatureTrajectorySampler(self, models_data, self._num_rff_features)
```
hstojic, line 262: this works only with GPR at the moment and that's why its here and not in the `GPflowPredictor` interface?

trieste/models/gpflow/models.py
hstojic: shouldn't `num_rff_features` be optional? 
if its costly to initialise the sampler, we probably shouldn't create the sampler by default

trieste/models/interfaces.py
```diff
@@ -92,6 +94,27 @@ def get_observation_noise(self) -> TensorType:
...
+        Return a reparametrization sampler providing `num_samples` samples.
+
+        Note that this is not supported by all models.
+
+        :param num_samples: The desired number of samples.
+        :return: The reparametrization sampler.
+        """
+        raise NotImplementedError(f"Model {self!r} does not have a reparametrization sampler")
+
+    def trajectory_sampler(self) -> TrajectorySampler:
```
hstojic, line 108: now that we have the `FastUpdateModel` interface, we should probably move these methods to new interfaces that we would put here, see my comment above...

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+                """
+            )
+        else:  # return the shared sampler rebuilt for the whole model stack
+            shared_sampler = sampler_types[0]
+            return shared_sampler(num_samples, self)
+
+
+class ReparametrizationSampler(ABC):
+    r"""
+    Thes samplers employ the *reparameterization trick* to draw samples from a
```
hstojic:
```suggestion
    These samplers employ the *reparameterization trick* to draw samples from a
```

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+    :class:`ProbabilisticModel`\ 's predictive distribution  across a discrete set of
+    points.
```
hstojic: we should put a reference here that explains the reparamtrization trick in greater detail

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+        self._initialized = tf.Variable(False)
+
+    def __repr__(self) -> str:
+        """"""
+        return f"{self.__class__.__name__}({self._sample_size!r}, {self._model!r})"
+
+    @abstractmethod
+    def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorType:
+        """
+        :param at: Input points that define the sampler.
```
hstojic: perhaps indicate the dimensions for `at` and return tensors?

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+        :param jitter: The size of the jitter to use when stabilising the Cholesky
+            decomposition of the covariance matrix.
+        :return: Samples.
+        """
```
hstojic, line 330: why not raising `NotImplementedError`? 
seems safer to me

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+        """"""
+        return f"{self.__class__.__name__}({self._model!r})"
+
+    @abstractmethod
+    def get_trajectory(self) -> TrajectoryFunction:
+        """
+        :return: A trajectory function representing an approximate trajectory from the
+            model, taking an input of shape `[N, D]` and returning shape `[N, 1]`
+        """
+
```
hstojic, line 381: why not raising `NotImplementedError`?

trieste/models/interfaces.py
```diff
@@ -241,6 +264,113 @@ def log(self) -> None:
...
+        """
+        :param sample_size: The desired number of samples.
+        :param model: The model to sample from.
+        :raise ValueError (or InvalidArgumentError): If ``sample_size`` is not positive.
+        """
+        tf.debugging.assert_positive(sample_size)
+
+        self._sample_size = sample_size
+        self._model = model
+        self._initialized = tf.Variable(False)
```
hstojic: can you add a comment, what is this object for?

trieste/models/__init__.py
```diff
@@ -19,4 +19,10 @@
 
 from . import gpflow, gpflux, optimizer
 from .config import ModelConfig, ModelRegistry, ModelSpec, create_model
-from .interfaces import ModelStack, ProbabilisticModel, TrainableProbabilisticModel
+from .interfaces import (
+    ModelStack,
+    ProbabilisticModel,
+    ReparametrizationSampler,
+    TrainableProbabilisticModel,
+    TrajectorySampler,
```
hstojic, line 29: this seems to be missing FastUpdateModel and TrajectoryFunction, no?

trieste/models/gpflow/sampler.py
```diff
@@ -0,0 +1,348 @@
...
+    This sampler employs the *reparameterization trick* to approximate samples from a
+    :class:`ProbabilisticModel`\ 's predictive distribution as
+
+    .. math:: x \mapsto \mu(x) + \epsilon \sigma(x)
+
+    where :math:`\epsilon \sim \mathcal N (0, 1)` is constant for a given sampler, thus ensuring
+    samples form a continuous curve.
+    """
+
+    def __init__(self, sample_size: int, model: ProbabilisticModel):
```
hstojic, line 47: if the model can be a generic one and does not need to be a `GPflowPredictor` instance for these three samplers then we can move the whole file to the parent folder, no?
(perhaps `RandomFourierFeatureTrajectorySampler` is the only one that is GP specific...)

trieste/models/gpflow/sampler.py
```diff
@@ -0,0 +1,348 @@
...
+    calculation strategies depending on the number of considered features m and the number
+    of data points n.
+
+    If :math:`m<n` then we follow Appendix A of :cite:`hernandez2014predictive` and calculate the
+    posterior distribution for :math:`\theta` following their Bayesian linear regression motivation,
+    i.e. the computation revolves around an O(m^3)  inversion of a design matrix.
+
+    If :math:`n<m` then we use the kernel trick to recast computation to revolve around an O(n^3)
+    inversion of a gram matrix. As well as being more efficient in early BO
+    steps (where :math:`n<m`), this second computation method allows must larger choices
```
hstojic:
```suggestion
    steps (where :math:`n<m`), this second computation method allows much larger choices
```

trieste/acquisition/function/multi_objective.py
```diff
@@ -255,13 +254,21 @@ def prepare_acquisition_function(
         # hypervolume improvement in this area
         _partition_bounds = prepare_default_non_dominated_partition_bounds(_reference_pt, _pf.front)
 
-        sampler = BatchReparametrizationSampler(self._sample_size, model)
+        try:
+            sampler = model.reparam_sampler(self._sample_size)
```
hstojic, line 258: if we implement the `ReparametrizationSamplerModel` interface, then we can also be more precise above in specifying the type of the probabilistic model

trieste/acquisition/sampler.py
```diff
@@ -19,100 +19,77 @@
...
+    minimiser :math:`x^*` over a discrete set of input locations.
 
     Although exact Thompson sampling is costly (incuring with an :math:`O(N^3)` complexity to
     sample over a set of `N` locations), this method can be used for any probabilistic model
     with a sampling method.
 
     """
 
-    def sample(self, at: TensorType) -> TensorType:
+    def sample(self, model: ProbabilisticModel, sample_size: int, at: TensorType) -> TensorType:
```
hstojic, line 76: why this change in arguments?
initialising with the model and the sample size is not working any more?

trieste/acquisition/sampler.py
```diff
@@ -186,358 +174,55 @@ def binary_search(val: float) -> float:  # Find empirical interquartile range
...
-        super().__init__(sample_size, model)
-
-        # _eps is essentially a lazy constant. It is declared and assigned an empty tensor here, and
-        # populated on the first call to sample
-        self._eps = tf.Variable(
-            tf.ones([sample_size, 0], dtype=tf.float64), shape=[sample_size, None]
-        )  # [S, 0]
-
-    def sample(self, at: TensorType) -> TensorType:
+    def sample(self, model: ProbabilisticModel, sample_size: int, at: TensorType) -> TensorType:
```
hstojic, line 193: if we implement the new interface, then we can be more precise here with `TrajectorySamplerModel`

trieste/acquisition/sampler.py
```diff
@@ -186,358 +174,55 @@ def binary_search(val: float) -> float:  # Find empirical interquartile range
...
         if self._sample_min_value:
             thompson_samples = tf.zeros([0, 1], dtype=at.dtype)  # [0,1]
         else:
             thompson_samples = tf.zeros([0, tf.shape(at)[1]], dtype=at.dtype)  # [0,D]
 
-        for _ in tf.range(self._sample_size):
-            sampled_trajectory = self.get_trajectory()
+        for _ in tf.range(sample_size):
+            sampled_trajectory = trajectory_sampler.get_trajectory()
             evaluated_trajectory = sampled_trajectory(at)  # [N, 1]
```
hstojic, line 226: why do we first create an instance and then pass `at`? it seems to me we could simplify things and pass it to `get_trajectory` method and get evaluation from there?

tests/unit/models/gpflow/test_interface.py
```diff
@@ -84,3 +88,25 @@ def test_gpflow_predictor_sample() -> None:
...
     assert samples.shape == (0, 1, 1)
+
+
+def test_gpflow_reparam_sampler_returns_a_param_sampler() -> None:
+    sampler = _QuadraticPredictor().reparam_sampler(10)
+    assert isinstance(sampler, BatchReparametrizationSampler)
+    assert sampler._sample_size == 10
+
+
+def test_gpflow_reparam_sampler_returns_param_sampler_with_correct_samples() -> None:
```
hstojic:
```suggestion
def test_gpflow_reparam_sampler_returns_reparam_sampler_with_correct_samples() -> None:
```

trieste/acquisition/sampler.py
henrymoss: So that you can specify a ThompsonSampler class without having access to the model it is going to be used on, i..e. when specifying a sampler as an argument when initializing rule or acquisition function.

trieste/acquisition/sampler.py
henrymoss: In the future we will want to optimize sampled_trajectory using a gradient optimizer instead of just picking the best over a discrete grid (as happens here). Therefore, we need to be able to return the whole trajecotry

trieste/models/gpflow/sampler.py
henrymoss: These reparam sampler all assume a Gaussian posterior and so only work for GPs. Moreover, they dont even work for DeepGP (Seb wrote his own for gpflux)

tests/unit/models/gpflow/test_models.py
```diff
@@ -359,6 +360,54 @@ def test_find_best_model_initialization_improves_likelihood(
     npt.assert_array_less(post_init_loss, pre_init_loss)
 
 
+@random_seed
+def test_gaussian_process_regression_trajectory_sampler_returns_a_trajectory_sampler(
+    gpflow_interface_factory: ModelFactoryType, dim: int
```
hstojic, line 365: since this is for GPR only lets not use the factory at all, below use the GPR wrapper directly

tests/unit/models/gpflow/test_models.py
```diff
@@ -359,6 +360,54 @@ def test_find_best_model_initialization_improves_likelihood(
...
+        pytest.skip("trajectory_sampler is only currently implemented for the GPR models.")
+
+    trajectory_sampler = model.trajectory_sampler()
+
+    assert isinstance(trajectory_sampler, TrajectorySampler)
+
+
+@random_seed
+def test_gaussian_process_regression_trajectory_sampler_has_correct_samples(
+    gpflow_interface_factory: ModelFactoryType,
```
hstojic, line 384: same here...

tests/unit/acquisition/test_sampler.py
```diff
@@ -448,24 +193,23 @@ def test_rff_sampler_returns_correctly_shaped_samples(
 @random_seed
 def test_rff_thompson_samples_are_minima() -> None:
```
hstojic:
```suggestion
def test_thompson_trajectory_samples_are_minima() -> None:
```

trieste/models/gpflow/sampler.py
hstojic: probabilistic neural nets can give a Gaussian posterior, it should work for them..

henrymoss: When needed (in another PR) we can move this up and rename it GaussianBatchRepar......

PR #400, opened 2021-11-02
Fix description of the failure dataset

apaleyes:
In the notebook that shows how to model failure regions, when describing the failure dataset, the code assigns labels in the opposite way compared to what the text says. But it seems that the code is actually correct, so we only need to fix the text.


vpicheny (approved):
good catch

PR #404, opened 2021-11-03
Mark branin test as slow running and add simpler quadratic test

### LINKED ISSUE #256: Run additional tests on develop after merging

uri-granta:
when reviewing this make sure to add ?w=1 to the url to ignore whitespace


tests/integration/test_bayesian_optimization.py
```diff
@@ -142,58 +140,101 @@
...
+                    )
+                    # these accuracies are the current best for the given number of optimization
+                    # steps, which makes this is a regression test
+                    assert tf.reduce_any(
+                        tf.reduce_all(relative_minimizer_err < 0.05, axis=-1), axis=0
+                    )
+                    npt.assert_allclose(best_y, SCALED_BRANIN_MINIMUM, rtol=0.005)
+                else:
+                    # currently assume that every rule should be able to solve this in 5 steps
+                    npt.assert_allclose(best_x, tf.constant([1.0, 1.0], tf.float64), rtol=0.05)
```
hstojic: we might need to pass tolerances as arguments, if we want to have different objectives, no?

uri-granta: There are only two hard-coded objectives (controlled by the branin parameter), each with its own hard-coded tolerance. If we need more flexibility we can easily refactor.

hstojic (approved):
looks good so far...
is there a specific reason for defining these tests in a class?

uri-granta:
> looks good so far... is there a specific reason for defining these tests in a class?

I originally tried to parametrize the class directly (which should be possible), but that didn't work. Personally I quite like grouping related tests into classes, but not too fussed either way.

hstojic:
> > looks good so far... is there a specific reason for defining these tests in a class?
> 
> I originally tried to parametrize the class directly (which should be possible), but that didn't work. Personally I quite like grouping related tests into classes, but not too fussed either way.

Since we aren't doing that with any other test, for consistency I wouldn't use a class here if its not necessary

tests/integration/test_ask_tell_optimization.py
```diff
@@ -46,9 +46,7 @@
 from trieste.space import Box, SearchSpace
 from trieste.types import State, TensorType
 
-
-@random_seed
-@pytest.mark.parametrize(
+BRANIN_OPTIMIZER_PARAMS = (
```
apaleyes: How about more neutral name, e.g. `TEST_SCENARIOS` or simply `OPTIMIZER_PARAMS`? They are not super specific to branin, and we use them as inputs to both branin and quadratic tests

tests/integration/test_ask_tell_optimization.py
```diff
@@ -97,7 +95,39 @@
...
+    acquisition_rule_fn: Callable[[], AcquisitionRule[TensorType, SearchSpace]]
+    | Callable[
+        [],
+        AcquisitionRule[State[TensorType, AsynchronousRuleState | TrustRegion.State], Box],
+    ],
+) -> None:
+    _test_ask_tell_optimization_finds(False, min(num_steps, 5), reload_state, acquisition_rule_fn)
+
+
+def _test_ask_tell_optimization_finds(
```
apaleyes:
```suggestion
def _test_ask_tell_optimization_finds_minima(
```

tests/integration/test_ask_tell_optimization.py
```diff
@@ -97,7 +95,39 @@
...
+def test_ask_tell_optimizer_finds_minima_of_simple_quadratic(
+    num_steps: int,
+    reload_state: bool,
+    acquisition_rule_fn: Callable[[], AcquisitionRule[TensorType, SearchSpace]]
+    | Callable[
+        [],
+        AcquisitionRule[State[TensorType, AsynchronousRuleState | TrustRegion.State], Box],
+    ],
+) -> None:
+    _test_ask_tell_optimization_finds(False, min(num_steps, 5), reload_state, acquisition_rule_fn)
```
apaleyes: Move this 5 to a constant, e.g. `simple_quadratic_bo_steps`?

tests/integration/test_ask_tell_optimization.py
```diff
@@ -113,6 +143,11 @@ def test_ask_tell_optimization_finds_minima_of_the_scaled_branin_function(
 
     search_space = BRANIN_SEARCH_SPACE
 
+    # for speed reasons we sometimes test with a simple quadratic defined on the
+    # same search space as branin
+    def simple_quadratic(x: TensorType) -> TensorType:
+        return -tf.math.reduce_sum(x, axis=-1, keepdims=True) ** 2
```
apaleyes: Maybe move it to test/utils package?

tests/integration/test_ask_tell_optimization.py
```diff
@@ -97,7 +95,39 @@
...
+    | Callable[
+        [],
+        AcquisitionRule[State[TensorType, AsynchronousRuleState | TrustRegion.State], Box],
+    ],
+) -> None:
+    _test_ask_tell_optimization_finds(False, min(num_steps, 5), reload_state, acquisition_rule_fn)
+
+
+def _test_ask_tell_optimization_finds(
+    branin: bool,
```
apaleyes:
```suggestion
    optimize_branin: bool,
```

tests/integration/test_ask_tell_optimization.py
```diff
@@ -164,8 +199,13 @@ def build_model(data: Dataset) -> GaussianProcessRegression:
...
+    if branin:
+        relative_minimizer_err = tf.abs((best_x - BRANIN_MINIMIZERS) / BRANIN_MINIMIZERS)
+        # these accuracies are the current best for the given number of optimization steps,
+        # which makes this is a regression test
+        assert tf.reduce_any(tf.reduce_all(relative_minimizer_err < 0.05, axis=-1), axis=0)
+        npt.assert_allclose(best_y, SCALED_BRANIN_MINIMUM, rtol=0.005)
+    else:
+        # currently assume that every rule should be able to solve this in 5 steps
+        npt.assert_allclose(best_x, tf.constant([1.0, 1.0], tf.float64), rtol=0.05)
+        npt.assert_allclose(best_y, tf.constant([-4.0], tf.float64), rtol=0.05)
```
apaleyes: Again, would be nice to put these to constants and move along the function definition. In fact, maybe it's worth having this funciton just as any other objective in `single_objectives`?

apaleyes (commented):
Looks great! Below are just some ideas about location and naming of things

tests/integration/test_ask_tell_optimization.py
apaleyes: another option: `is_branin`

apaleyes (approved):

PR #405, opened 2021-11-03
Fix optimizer bug in VGP models

henrymoss:
After model optimizer refactor, a bug was introduced that raises a spurious error message.  We require the **base** optimizer of an optimzer to be a tensorflow one (i.e. not scipy), however we were checking that the optimizer itself BatchOptimizer instead.

I have also added a valid default for the VGP optimizer. Before, it was defaulting to a scipy optimizer.

Also included an extra test to make sure this doesn't happen again.


vpicheny (approved):
LGTM

PR #407, opened 2021-11-03
Use name scopes, and add another tutorial example

uri-granta:


hstojic (approved):
looks good - how come context is not needed any more? tensorboard automatically uses name scopes?

uri-granta:
> looks good - how come context is not needed any more? tensorboard automatically uses name scopes?

Yes. Happened to spot that the tf.summary name parameter docs says "the summary tag used for TensorBoard will be this name prefixed by any active name scopes."

PR #408, opened 2021-11-04
Split up function.py

### LINKED ISSUE #221: Split out GPFlow-specific models from the general model_interfaces file
As a Trieste contributor, I would like Trieste to be more maintainable and extensible by keeping general models separate from GPFlow-specific ones.

uri-granta:
Split up function.py into five parts:

1. `trieste.acquisition.interfaces`: interfaces
2. `trieste.acquisition.function.multi-objective`: multi-objective functions  
3. `trieste.acquisition.function.entropy`: entropy-based functions
4.  `trieste.acquisition.function.local_penalization`: local penalization functions and penalizers
5. `trieste.acquisition.function.function`: all the other functions (4 improvement based functions plus lcb, pof and predictive variance; can be split more later if we wish)

Any existing code that tried to import the interfaces directly from `trieste.acquisition.function` now has to either import them from `trieste.acquisition` or from `trieste.acquisition.interfaces`. Otherwise, there are no breaking changes.


uri-granta:
(For anyone looking at this: I still need to migrate the unit tests.)

hstojic:
Looks good, except for:

> 5. `trieste.acquisition.single-objective.function`: all the other functions (4 improvement based functions plus lcb, pof and predictive variance)

I would split this one up further into `improvement` and `active`, improvement based ones are default and there is quite a few of them already, active learning will be expanded so we might as well put it in a separate file already

Did you make any changes or everything was just about moving things around?

uri-granta:
> I would split this one up further into `improvement` and `active`, improvement based ones are default and there is quite a few of them already, active learning will be expanded so we might as well put it in a separate file already

To be honest I'm not really sure what active learning means. Which of these builders fall under that category?

> Did you make any changes or everything was just about moving things around?

Just moving things.

apaleyes:
That was a ton of work, thanks for doing this!

Just one thought. Consider the very first notebook in the diff:

<pre><code>
from trieste.acquisition.<b>optimizer</b> import generate_continuous_optimizer
from trieste.acquisition.<b>rule</b> import EfficientGlobalOptimization
from trieste.acquisition<b> nothing here!</b> import PredictiveVariance
</code></pre>

Whereas before the last line would have `function`. I found that very intuitive, since Trieste has this rather unique distinction between a function and a rule. I'd suggest we keep it, by having all acquisition functions inside a function package.

Admittedly this could be up to taste, feel free to ignore.

uri-granta:
I did consider including a dummy `trieste.acquisition.function` package but that's ugly (and is easy to forget to maintain). I also considered making multi-objective and single-objective nested _inside_ a function package, but then the hierarchy becomes unwieldily deep (`trieste.acqusition.function.single_objective.entropy`). If you think it's clearer, I could make the import more explicit as follows, but I'm also happy with acquisition functions, which are defined in `trieste.acquisition.interfaces`, naturally residing in `trieste.acquisition`, while rules, etc reside deeper down (and may end up being moved one day).

```
from trieste.acquisition.optimizer import generate_continuous_optimizer
from trieste.acquisition.rule import EfficientGlobalOptimization
from trieste.acquisition.single_objective import PredictiveVariance
```

apaleyes:
Your second option is what I had in mind indeed. Even though users will never see this level of nesting, it's a good point that it can become a pain for development.

Let's leave it as it is for now then. Personally I don't see much difference between what the PR does right now and the third option.

apaleyes:
> > I would split this one up further into `improvement` and `active`, improvement based ones are default and there is quite a few of them already, active learning will be expanded so we might as well put it in a separate file already
> 
> To be honest I'm not really sure what active learning means. Which of these builders fall under that category?

Afaik, active learning here refers to the regime where our goal is not to find the optimum, but to learn the function. So these acquisition functions tend to be very explorative and aimed at reducing variance of the model. Separating it out provisionally is a good idea, although there isn't much certainty (in my head at least) about how widely will Trieste support it. #294 where this was introduced feels rather experimental.

If we choose to develop it in Trieste further, I'd imagine we'll have many interface changes to separate it from optimization anyway. So I'd say let's leave it where this PR puts it for now. @hstojic thoughts?

hstojic:
> > > I would split this one up further into `improvement` and `active`, improvement based ones are default and there is quite a few of them already, active learning will be expanded so we might as well put it in a separate file already
> > 
> > 
> > To be honest I'm not really sure what active learning means. Which of these builders fall under that category?
> 
> Afaik, active learning here refers to the regime where our goal is not to find the optimum, but to learn the function. So these acquisition functions tend to be very explorative and aimed at reducing variance of the model. Separating it out provisionally is a good idea, although there isn't much certainty (in my head at least) about how widely will Trieste support it. #294 where this was introduced feels rather experimental.
> 
> If we choose to develop it in Trieste further, I'd imagine we'll have many interface changes to separate it from optimization anyway. So I'd say let's leave it where this PR puts it for now. @hstojic thoughts?

We do have plans to develop it more seriously, but its ok to leave it as it is for the time being - we can restructure things as we make some progress there

hstojic:
> Your second option is what I had in mind indeed. Even though users will never see this level of nesting, it's a good point that it can become a pain for development.

I would agree with the 2nd option, its more intuitive to have imports from `functions`, it's a bit more painful to develop with such nested structure but not a huge hassle in my opinion

apaleyes (approved):

uri-granta:
> I would agree with the 2nd option, its more intuitive to have imports from `functions`, it's a bit more painful to develop with such nested structure but not a huge hassle in my opinion

How about this then?

* put all the functions in `trieste.acquisition.function` but without extra subdirectories for single and multi objective functions
* multi objective functions can go in `trieste.acquisition.function.multi_objective`, the others can keep their existing module names (e.g. `trieste.acquisition.function.entropy`)
* the multi objective _helper utilities_ remain in `trieste.acquisition.multi_objective`

hstojic:
> > I would agree with the 2nd option, its more intuitive to have imports from `functions`, it's a bit more painful to develop with such nested structure but not a huge hassle in my opinion
> 
> How about this then?
> 
> * put all the functions in `trieste.acquisition.function` but without extra subdirectories for single and multi objective functions
> * multi objective functions can go in `trieste.acquisition.function.multi_objective`, the others can keep their existing module names (e.g. `trieste.acquisition.function.entropy`)
> * the multi objective _helper utilities_ remain in `trieste.acquisition.multi_objective`

To me that sounds good.

hstojic (approved):
looks good

henrymoss (approved):
LGTM, Thanks for doing this @uri-granta !

PR #410, opened 2021-11-05
added kwargs argument that user can pass to scipy optimizer (for optimizing acq functions)

### LINKED ISSUE #392: Allow specification of scipy optimizer kwargs when building continuous optimizers
Allow specification of scipy optimizer kwargs when building continuous optimizers

hstojic:
I have added kwargs argument that user can pass to scipy optimizer (for optimizing acq functions), this would close #392

I have also moved hard coded consts into `utils/misc`

### DIFF
--- a/trieste/acquisition/optimizer.py
+++ b/trieste/acquisition/optimizer.py
@@ -1,4 +1,4 @@
-# Copyright 2020 The Trieste Contributors
+# Copyright 2021 The Trieste Contributors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -11,13 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 r"""
 This module contains functionality for optimizing
 :data:`~trieste.acquisition.AcquisitionFunction`\ s over :class:`~trieste.space.SearchSpace`\ s.
 """
+
 from __future__ import annotations
 
-from typing import Any, Callable, Optional, TypeVar
+import copy
+from typing import Any, Callable, TypeVar
 
 import gpflow
 import scipy.optimize as spo
@@ -26,11 +29,26 @@ from scipy.optimize import OptimizeResult
 
 from ..space import Box, DiscreteSearchSpace, SearchSpace, TaggedProductSearchSpace
 from ..types import TensorType
-from ..utils import DEFAULTS
 from .function import AcquisitionFunction
 
 SP = TypeVar("SP", bound=SearchSpace)
 """ Type variable bound to :class:`~trieste.space.SearchSpace`. """
+
+
+NUM_SAMPLES_MIN: int = 5000
+"""
+The default minimum number of initial samples for :func:`generate_continuous_optimizer` function in
+:func:`automatic_optimizer_selector`, used for determining the number of initial samples in the
+multi-start acquisition function optimization.
+"""
+
+
+NUM_SAMPLES_DIM: int = 1000
+"""
+The default minimum number of initial samples per dimension of the search space for
+:func:`generate_continuous_optimizer` function in :func:`automatic_optimizer_selector`, used for
+determining the number of initial samples in the multi-start acquisition function optimization.
+"""
 
 
 class FailedOptimizationError(Exception):
@@ -64,9 +82,7 @@ def automatic_optimizer_selector(
         return optimize_discrete(space, target_func)
 
     elif isinstance(space, (Box, TaggedProductSearchSpace)):
-        num_samples = tf.minimum(
-            DEFAULTS.NUM_SAMPLES_MIN, DEFAULTS.NUM_SAMPLES_DIM * tf.shape(space.lower)[-1]
-        )
+        num_samples = tf.minimum(NUM_SAMPLES_MIN, NUM_SAMPLES_DIM * tf.shape(space.lower)[-1])
         return generate_continuous_optimizer(num_samples)(space, target_func)
 
     else:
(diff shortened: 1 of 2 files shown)


sebastianober:
Is there a specific motivation for this? I'm interested because of the possible relation to PR #396

hstojic:
> Is there a specific motivation for this? I'm interested because of the possible relation to PR #396

this was motivated by your issues with acq function optimizer function not being flexible enough - users should have greater control and pass arguments to the optimizer (e.g. compile, scipy_kwargs...)

feel free to review it :)

trieste/utils/misc.py
```diff
@@ -180,3 +180,15 @@ class DEFAULTS:
...
+    """
+    The default jitter, typically used to stabilise computations near singular points, such as in
+    Cholesky decomposition.
+    """
+
+    NUM_SAMPLES_DIM: int = 1000
+    """
+    The default jitter, typically used to stabilise computations near singular points, such as in
+    Cholesky decomposition.
+    """
```
apaleyes: jitter has many disguises

trieste/acquisition/optimizer.py
```diff
@@ -100,18 +103,19 @@ def generate_continuous_optimizer(
...
 ) -> AcquisitionOptimizer[Box | TaggedProductSearchSpace]:
     """
     Generate a gradient-based optimizer for :class:'Box' and :class:'TaggedProductSearchSpace'
     spaces and batches of size 1. In the case of a :class:'TaggedProductSearchSpace', We perform
     gradient-based optimization across all :class:'Box' subspaces, starting from the best location
     found across a sample of `num_initial_samples` random points.
 
-    This optimizer supports Scipy's L-BFGS-B optimizer, which optimizes directly within and up to
-    the bounds of the search space.
+    This optimizer supports Scipy's L-BFGS-B optimizer (used via GPflow's wrapper), which optimizes
```
apaleyes: It is probably a good idea to specify how scipy optimizer is used. But "gpflow wrapper" needs to be a bit more specific, because in Trieste context wrappers are around models. at the very least "gpflow Scipy optimizer wrapper"

trieste/acquisition/optimizer.py
```diff
@@ -63,7 +64,9 @@ def automatic_optimizer_selector(
         return optimize_discrete(space, target_func)
 
     elif isinstance(space, (Box, TaggedProductSearchSpace)):
-        num_samples = tf.minimum(5000, 1000 * tf.shape(space.lower)[-1])
+        num_samples = tf.minimum(
+            DEFAULTS.NUM_SAMPLES_MIN, DEFAULTS.NUM_SAMPLES_DIM * tf.shape(space.lower)[-1]
+        )
```
henrymoss: It is slightly problematic that a user cant just look at the optimizer and see what the default behaviour is (i.e. they would have to hop into the utils script)?

henrymoss (commented):
LGTM, just the one comment above about DEFAULTS

trieste/acquisition/optimizer.py
```diff
@@ -177,7 +184,8 @@ def _objective() -> TensorType:
 
         def _perform_optimization(starting_point: TensorType, bounds: spo.Bounds) -> OptimizeResult:
             variable.assign(starting_point)  # [1, D]
-            return gpflow.optimizers.Scipy().minimize(_objective, (variable,), bounds=bounds)
+            optimizer_args["bounds"] = bounds
+            return gpflow.optimizers.Scipy().minimize(_objective, (variable,), **optimizer_args)
```
apaleyes: Slightly concerned about modifying global (relatively to this function) dictionary. This call leaves the args dict modified after the call, which is a side effect that can be dangerous. Would be safer to create a copy of the dict.

trieste/acquisition/optimizer.py
hstojic: I kinda agree but then why do we have that `DEFAULTS` object, we either use it for our defaults or not and put the constants in a file where they are used, from developer perspective same file option would be better, from user perspective and when you want to modify multiple constants it might be better to have a single object where you change default behaviour - but then in the `DEFAULTS` object we have very little at the moment

trieste/utils/misc.py
hstojic: :)

trieste/acquisition/optimizer.py
hstojic: perhaps we use it in files at the moment, I'll check all the files and if we see that we have more constants spread around I'll collect them in DEFAULTS in a separate PR

henrymoss: I think the JITTER default makes sense as it isnt something that a user would really ever change and is used in loads of location across the code. This is a very specific parameter (only used here) that is likely to be changed often by users?

apaleyes: If that's really a very specific constant only used here, and it doesn't make sense in any other context in Trieste, I would agree with Henry. Maybe move it to a constant local to this file, but don't move it outside.

trieste/acquisition/optimizer.py
hstojic: thanks for spotting that, resolved!

trieste/acquisition/optimizer.py
```diff
@@ -32,6 +35,22 @@
...
+"""
+
+
+NUM_SAMPLES_DIM: int = 1000
+"""
+The default minimum number of initial samples per dimension of the search space for
+:func:`generate_continuous_optimizer` function in :func:`automatic_optimizer_selector`, used for
+determining the number of initial samples in the multi-start acquisition function optimization.
+"""
+
```
henrymoss, line 52: Perhaps also use this for the random search optimizer?

henrymoss (approved):
LGTM

apaleyes (approved):

trieste/acquisition/optimizer.py
hstojic: not sure about that, its not per dimension and we don't do that in `generate_continuous_optimizer` either, defaults for functions are set already there

henrymoss: It should be per dim? Why not make it consistent?

PR #411, opened 2021-11-08
Fix showing pareto points only bug, docstring

apaleyes:
The actual fix is on the last line of the diff - the code was trying to index a TF Tensor with an array, and failed with this error:

>TypeError: Only integers, slices (`:`), ellipsis (`...`), tf.newaxis (`None`) and scalar tf.int32/tf.int64 tensors are valid indices, got array([20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 33, 34, 35, 36, 37,
       38, 39, 41, 42, 43, 44, 45, 46, 47, 48, 49])

Now we convert to numpy array if the tensor was passed in. Also updated the docstring while at it.


hstojic (approved):
looks good

PR #412, opened 2021-11-08
temporary fix for dependency versions

tensorlicious:
pin specific version of `gpflux` and restrict version of `tensorflow-probability` until `trieste` is updated to `gpflux` v0.2.7+


hstojic:
> pin specific version of `gpflux` and restrict version of `tensorflow-probability` until `trieste` is updated to `gpflux` v0.2.7+

why not updating to the latest gpflux then?

tensorlicious:
> > pin specific version of `gpflux` and restrict version of `tensorflow-probability` until `trieste` is updated to `gpflux` v0.2.7+
> 
> why not updating to the latest gpflux then?

Yes, we should, but that requires some work and in the meantime the repo shouldn't be broken.

hstojic (approved):
looks good

PR #414, opened 2021-11-09
consistent defaults for acquisition function optimizers

hstojic:


trieste/acquisition/optimizer.py
```diff
@@ -137,7 +137,7 @@ def generate_continuous_optimizer(
     `num_recovery_runs` starting from random locations.
 
     The default behavior of this method is to return a L-BFGS-B optimizer that performs
-    a single optimization from the best of 1000 initial locations. If this optimization fails then
+    a single optimization from the best of 5000 initial locations. If this optimization fails then
```
```suggestion
    a single optimization from the best of `NUM_SAMPLES_MIN` initial locations. If this optimization fails then
```

apaleyes (approved):

PR #416, opened 2021-11-11
Fix VGP wrapper to support tf function

uri-granta:


tests/integration/test_bayesian_optimization.py
```diff
@@ -169,19 +173,26 @@ def test_optimizer_finds_minima_of_simple_quadratic(
 ) -> None:
     # for speed reasons we sometimes test with a simple quadratic defined on the same search space
     # branin; currently assume that every rule should be able to solve this in 5 steps
-    _test_optimizer_finds_minimum(False, min(num_steps, 5), acquisition_rule)
+    _test_optimizer_finds_minimum(min(num_steps, 5), acquisition_rule)
+
+
+@random_seed
+def test_optimizer_finds_minima_with_vgp_model() -> None:
```
hstojic, line 180: perhaps we should do the same with other GPflow models? SGPR and SVGP?

hstojic, line 180: perhaps we should do the same with other GPflow models? SGPR and SVGP?

uri-granta: Yes, definitely. And I'll have a think if this could have also been caught in unit tests.

uri-granta:
Closing PR as issue is actually a bit subtler than implied here. Will open a new one once I have a fix.

Issue #406, opened 2021-11-03
TF compilation causes problems for VGP models


apaleyes:
**Describe the bug**
TF compilation causes problems for models other than GPR when used with EI

**To reproduce**
To reproduce, we start with the EI notebook: https://secondmind-labs.github.io/trieste/notebooks/expected_improvement.html
In this notebook, replace GRP with VGP, by replacing this code:
```
def build_model(data):
    variance = tf.math.reduce_variance(data.observations)
    kernel = gpflow.kernels.Matern52(variance=variance, lengthscales=[0.2, 0.2])
    prior_scale = tf.cast(1.0, dtype=tf.float64)
    kernel.variance.prior = tfp.distributions.LogNormal(tf.cast(-2.0, dtype=tf.float64), prior_scale)
    kernel.lengthscales.prior = tfp.distributions.LogNormal(tf.math.log(kernel.lengthscales), prior_scale)
    gpr = gpflow.models.GPR(data.astuple(), kernel, noise_variance=1e-5)
    gpflow.set_trainable(gpr.likelihood, False)

    return GPflowModelConfig(**{
        "model": gpr,
        "model_args": {
            "num_kernel_samples": 100,
        },
        "optimizer": gpflow.optimizers.Scipy(),
        "optimizer_args": {
            "minimize_args": {"options": dict(maxiter=100)},
        },
    })
```

with this (the smallest possible change to replace GPR with VGP):
```
def build_model(data):
    variance = tf.math.reduce_variance(data.observations)
    kernel = gpflow.kernels.Matern52(variance=variance, lengthscales=[0.2, 0.2])
    prior_scale = tf.cast(1.0, dtype=tf.float64)
    kernel.variance.prior = tfp.distributions.LogNormal(tf.cast(-2.0, dtype=tf.float64), prior_scale)
    kernel.lengthscales.prior = tfp.distributions.LogNormal(tf.math.log(kernel.lengthscales), prior_scale)
    likelihood = gpflow.likelihoods.Gaussian()
    vgp = gpflow.models.VGP(data.astuple(), kernel, likelihood)

    return GPflowModelConfig(**{
        "model": vgp,
        "optimizer": gpflow.optimizers.Scipy(),
        "optimizer_args": {
            "minimize_args": {"options": dict(maxiter=100)},
        },
    })
```

Now run it and observe the following error:
[...]

PR #417, opened 2021-11-12
Bayesian Active Learning By Disagreement Acq

satrialoka:
BALD Active learning acquisition function for Gaussian Process Classification model.
In this PR [BALD](https://arxiv.org/pdf/1112.5745.pdf) is implemented. 

works to do:

- [x] Notebook 
- [x] Test 
- [x] Assert model type in bald
- [x] fix @tf.function error (waiting for #406 fix)


trieste/acquisition/function/function.py
```diff
@@ -762,3 +763,74 @@ def acquisition(x: TensorType) -> TensorType:
         return tf.exp(tf.linalg.logdet(covariance + jitter))
 
     return acquisition
+
+
+class BALD(SingleModelAcquisitionBuilder):
```
hstojic: we should avoid abbreviations if possible, perhaps `LearningByDisagreement`?

trieste/acquisition/function/function.py
```diff
@@ -762,3 +763,74 @@ def acquisition(x: TensorType) -> TensorType:
...
+        """
+
+        :param model: The model of the objective function.
+        :param jitter: The size of the jitter to avoid numerical problem caused by the variance value.
+        :return: The Bayesian Active Learning By Disagreement acquisition function.
+        """
+        self._model = model
+        self._jitter = jitter
+
+    # @tf.function
```
hstojic: there is an issue with VGP that is the root cause of retracing (see #406), it might not be due to the code here, @uri-granta is working on fixing this

docs/notebooks/bald.pct.py
```diff
@@ -0,0 +1,144 @@
+# %%
```
hstojic: I assume docstrings will come :)
can you please use a more descriptive name for the file, not abbreviation

docs/notebooks/bald.pct.py
```diff
@@ -0,0 +1,144 @@
...
+plt.show()
+
+
+# %%
+initial_models = trieste.utils.map_values(create_bo_model, datasets)
+
+# %%
+acq = BALD()
+# %% BO
+numIter = 50
```
hstojic: we now have quick build of notebooks for each commit to the PR, for this to work you have to add a .yaml file in the `docs/notebooks/quickrun` folder, check for example `thompson_sampling.yaml` file, you would change `numIter` to smth like `numIter = 2`

hstojic: by the way `numIter` is the camel case, perhaps switch to the snake one used elsewhere for this kind of variables?

trieste/acquisition/function/function.py
satrialoka: wouldn't `ActiveLearningByDisagreement` be more suitable? since Trieste also has a long-named acquisition function like `BatchMonteCarloExpectedImprovement`

docs/notebooks/bald.pct.py
satrialoka: Yes, I am working on it
Just thinking if it is neater to put this on the active learning notebook along with the predictive variance acquisition function, what do you think?

trieste/acquisition/function/function.py
hstojic: that's fine as well

docs/notebooks/bald.pct.py
hstojic: this one is specific to classificaiton problems, so perhaps we can keep it separate, but you should definitely point one to each other

trieste/acquisition/function/function.py
```diff
@@ -762,3 +763,74 @@ def acquisition(x: TensorType) -> TensorType:
...
+        """
+        return function  # no need to update anything
+
+
+class bald(AcquisitionFunctionClass):
+    def __init__(self, model: ProbabilisticModel, jitter: float):
+        """
+
+        :param model: The model of the objective function.
+        :param jitter: The size of the jitter to avoid numerical problem caused by the variance value.
```
hstojic: jitter is used below to avoid issue in taking logs of values close to zero, its not really related to variance value

trieste/acquisition/function/function.py
```diff
@@ -762,3 +763,74 @@ def acquisition(x: TensorType) -> TensorType:
...
 
     return acquisition
+
+
+class BALD(SingleModelAcquisitionBuilder):
+    """
+    Builder for the *Bayesian Active Learning By Disagreement* acquisition function defined in
+    :cite:`houlsby2011bayesian`. The acquisition function computes the information gains of the predictive entropy.
+
+    This acquisition function is intended to use for Binary Gaussian Process Classification model with Bernoulli likelihood.
```
hstojic: perhaps mention that integrating over nuisance parameters is not currently supported (eq 6 from the paper)?

docs/notebooks/bald.pct.py
```diff
@@ -0,0 +1,144 @@
...
+from trieste.utils import map_values
+
+np.random.seed(1965)
+tf.random.set_seed(1965)
+
+sp = trieste.space.Box([-1, -1], [1, 1])
+inDim = 2
+
+
+def circle(x):
```
hstojic: is this a common toy problem in this literature?
what about replicating the problems from the paper (results from Figure 3 - only the BALD and perhaps random search)?

trieste/acquisition/function/function.py
satrialoka: indeed that's what I want to say, thanks for the correction. fixed.

docs/notebooks/bald.pct.py
satrialoka: not sure if this function is common or not,
The reason we use this toy problem is that, it is more illustrative than for example calculating the accuracy like in the paper. I.e we can see how BALD can sample near the boundary region of the two classes in the circle dataset. In addition, this makes us easily use the continuous search space and EGO acquisition rule. 
What do you think?

hstojic: that's a fair point, but it would still be nice to use a toy problem that has been used in the literature, as widely as possible - I cannot provide much guidance here unfortunately, perhaps @vpicheny ?

docs/notebooks/bald.pct.py
```diff
@@ -0,0 +1,144 @@
...
+
+# %%
+from trieste.models.gpflow import GPflowModelConfig
+
+
+def create_bo_model(data):
+    # kernel = gpflow.kernels.SquaredExponential()
+    # kernel = gpflow.kernels.Matern52()
+    kernel = gpflow.kernels.SquaredExponential()
+    m = gpflow.models.VGP(astuple(data), likelihood=gpflow.likelihoods.Bernoulli(), kernel=kernel)
```
hstojic: with `Dataset` you can do `data.astuple()`

docs/notebooks/bald.pct.py
```diff
@@ -0,0 +1,144 @@
...
+    return trieste.models.create_model(
```
hstojic: lets use the VGP wrapper here directly

docs/notebooks/bald.pct.py
```diff
@@ -0,0 +1,144 @@
...
+from trieste.models import create_model
+from trieste.models.gpflow.models import GaussianProcessRegression, VariationalGaussianProcess
+from trieste.objectives.utils import mk_observer
+from trieste.utils import map_values
+
+np.random.seed(1965)
+tf.random.set_seed(1965)
+
+sp = trieste.space.Box([-1, -1], [1, 1])
+inDim = 2
```
hstojic: camel case

docs/notebooks/bald.pct.py
satrialoka: using the VGP wrapper directly here seems to make the model quite hard to estimate the function. not sure if the problem is in the adam optimiser (I am still investigating this). Do you have a minimal working example for this wrapper for VPG with Bernoulli likelihood?

docs/refs.bib
```diff
@@ -234,3 +234,23 @@ @InProceedings{kandasamy18a
   publisher = {PMLR},
   url = {https://proceedings.mlr.press/v84/kandasamy18a.html}
 }
+
+@misc{houlsby2011bayesian,
+      title={Bayesian Active Learning for Classification and Preference Learning},
```
hstojic: use an indent consistent with the refs above...

trieste/acquisition/function/function.py
```diff
@@ -762,3 +763,79 @@ def acquisition(x: TensorType) -> TensorType:
...
+
+
+class BayesianActiveLearningByDisagreement(SingleModelAcquisitionBuilder):
+    """
+    Builder for the *Bayesian Active Learning By Disagreement* acquisition function defined in
+    :cite:`houlsby2011bayesian`. The acquisition function computes the information gains
+    of the predictive entropy.  
+
+    This acquisition function is intended to use for Binary Gaussian Process Classification
+    model with Bernoulli likelihood. integrating over nuisance parameters is not currently
```
hstojic, line 775:
```suggestion
    Integrating over nuisance parameters is currently not supported (see equation 6 of the paper).
```

hstojic: "Integrating over nuisance parameters is currently not supported (see equation 6 of the paper)."

trieste/acquisition/function/function.py
```diff
@@ -762,3 +763,79 @@ def acquisition(x: TensorType) -> TensorType:
         return tf.exp(tf.linalg.logdet(covariance + jitter))
 
     return acquisition
+
+
+class BayesianActiveLearningByDisagreement(SingleModelAcquisitionBuilder):
+    """
+    Builder for the *Bayesian Active Learning By Disagreement* acquisition function defined in
+    :cite:`houlsby2011bayesian`. The acquisition function computes the information gains
```
hstojic: gains -> gain

trieste/acquisition/function/function.py
```diff
@@ -762,3 +763,79 @@ def acquisition(x: TensorType) -> TensorType:
...
+    def __init__(self, jitter: float = DEFAULTS.JITTER) -> None:
+        """
+        :param jitter: The size of the jitter to avoid numerical problem caused by the
+                log operation if variance is close to zero.
+        """
+        self._jitter = jitter
+
+    def __repr__(self) -> str:
+        """"""
+        return f"BALD(jitter={self._jitter!r})"
```
hstojic:
```suggestion
        return f"BayesianActiveLearningByDisagreement(jitter={self._jitter!r})"
```

trieste/acquisition/function/function.py
```diff
@@ -762,3 +763,79 @@ def acquisition(x: TensorType) -> TensorType:
...
+        """
+        :param function: The acquisition function to update.
+        :param model: The model.
+        :param dataset: Unused.
+        """
+        return function  # no need to update anything
+
+
+class bayesian_active_learning_by_disagreement(AcquisitionFunctionClass):
+    def __init__(self, model: ProbabilisticModel, jitter: float):
```
hstojic: docstring? you can reuse most from the above, this one actually computes it, its not a builder

trieste/acquisition/function/function.py
```diff
@@ -17,6 +17,7 @@
 """
```
hstojic: if you merge in develop you'll see that we now have a new file for active learning in this same directory, `active_learning.py`

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,191 @@
...
+                "optimizer_args": {
+                    "minimize_args": {"options": dict(maxiter=100.0)},
+                },
+            }
+        )
+    )
+
+
+"""
+# model seems harder to train using default VariationalGaussianProcess() optimiser which is adam?
```
hstojic: no, the default is Scipy, if you want to use natgrads then you need to set up Adam with `BatchOptimizer` wrapper, try setting up that and check the performance

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,191 @@
...
+input_dim = 2
+
+
+def circle(x):
+    return tf.cast((tf.reduce_sum(tf.square(x), axis=1, keepdims=True) - 0.5) > 0, tf.float64)
+
+
+# %%
+density = 100
+xx = np.linspace(search_space.lower[0], search_space.upper[0], density)
```
hstojic: can you please use `query_points` instead of xx, `observations` instead of yy

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,191 @@
...
+plt.show()
+
+# %% [markdown]
+# Let's generate some data for our initial model. Here we randomly sample 10 data points.
+
+# %%
+numSamples = 10
+X = search_space.sample(numSamples)
+observer = mk_observer(circle, OBJECTIVE)
+datasets = observer(X)
```
hstojic:
```suggestion
initial_data = observer(X)
```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,191 @@
...
+plt.scatter(xx[np.logical_not(idx)][:, 0], xx[np.logical_not(idx)][:, 1], label="0")
+plt.legend()
+plt.show()
+
+# %% [markdown]
+# Let's generate some data for our initial model. Here we randomly sample 10 data points.
+
+# %%
+numSamples = 10
+X = search_space.sample(numSamples)
```
hstojic:
```suggestion
X = search_space.sample(num_initial_points)
```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,191 @@
...
+plt.scatter(xx[idx][:, 0], xx[idx][:, 1], label="1")
+plt.scatter(xx[np.logical_not(idx)][:, 0], xx[np.logical_not(idx)][:, 1], label="0")
+plt.legend()
+plt.show()
+
+# %% [markdown]
+# Let's generate some data for our initial model. Here we randomly sample 10 data points.
+
+# %%
+numSamples = 10
```
hstojic:
```suggestion
num_initial_points = 10
```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,191 @@
...
+# ## The acquisition process
+#
+# We can construct the BALD acqusition function which maximise information gain about the model parameters, by maximising the mutual information between predictions and model posterior:
+#
+# $$\mathbb{I}\left[y, \boldsymbol{\theta} \mid \mathbf{x}, \mathcal{D}\right]=\mathbb{H}\left[y \mid \mathbf{x}, \mathcal{D}\right]-\mathbb{E}_{p\left(\boldsymbol{\theta} \mid \mathcal{D}\right)}[\mathbb{H}[y \mid \mathbf{x}, \boldsymbol{\theta}]]$$
+#
+# See <cite data-cite="houlsby2011bayesian">[Houlsby et al.](https://arxiv.org/pdf/1112.5745.pdf)</cite> for more details. Then, Trieste's `EfficientGlobalOptimization` is used for the query rule:
+
+# %%
+initial_models = trieste.utils.map_values(create_bo_model, datasets)
```
hstojic: you have a single dataset and model, so no need for this code

```suggestion
initial_models = create_bo_model(datasets)
```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,191 @@
...
+acq = BayesianActiveLearningByDisagreement()
+rule = trieste.acquisition.rule.EfficientGlobalOptimization(acq.using(OBJECTIVE))
```
hstojic:
```suggestion
rule = trieste.acquisition.rule.EfficientGlobalOptimization(acq)
```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,191 @@
...
+)
+plt.plot(
+    final_dataset.query_points[-n_steps:, 0], final_dataset.query_points[-n_steps:, 1], "rx", mew=10
+)
+plt.contour(*grid_xx, np.reshape(yy, [density] * input_dim), levels=[0.5])
+plt.title("Updated Mean")
+plt.show()
+
+# %% [markdown]
+# As expected, BALD will query in important regions like points near the domain boundary and class boundary.
```
hstojic, line 162: please add the usual license part at the end

hstojic:
```
# %% [markdown]
# ## LICENSE
#
# [Apache License 2.0](https://github.com/secondmind-labs/trieste/blob/develop/LICENSE)
```

docs/notebooks/active_learning_for_binary_classification.pct.py
satrialoka: Fixed the performance by setting max_iter to 100, by still using Scipy optimiser

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +246,78 @@ def acquisition(x: TensorType) -> TensorType:
...
+    model with Bernoulli likelihood. integrating over nuisance parameters is currently not 
+    supported (see equation 6 of the paper)
+    """
+
+    def __init__(self, jitter: float = DEFAULTS.JITTER) -> None:
+        """
+        :param jitter: The size of the jitter to avoid numerical problem caused by the
+                log operation if variance is close to zero.
+        """
+        self._jitter = jitter
```
hstojic, line 438: this should be > 0, right? then lets assert its scalar and positive and have a test for raising the exceptions

trieste/acquisition/function/function.py
```diff
@@ -17,6 +17,7 @@
...
 from __future__ import annotations
 
+import math
```
hstojic: no need for this here, right?

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +246,78 @@ def acquisition(x: TensorType) -> TensorType:
...
+        :param model: The model of the objective function.
+        :param jitter: The size of the jitter to avoid numerical problem caused by the
+                log operation if variance is close to zero.
+        :return: The Bayesian Active Learning By Disagreement acquisition function.
+        """
+        self._model = model
+        self._jitter = jitter
+
+    # @tf.function
+    def __call__(self, x: TensorType) -> TensorType:
```
hstojic, line 509: this is not supporting batches, right? lets assert here that x doesn't contain batches (there are other acq fncs that make a similar check) - we'll need a test that raises this exception as well

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +246,78 @@ def acquisition(x: TensorType) -> TensorType:
         return criterion
 
     return acquisition
+
+class BayesianActiveLearningByDisagreement(SingleModelAcquisitionBuilder):
+    """
+    Builder for the *Bayesian Active Learning By Disagreement* acquisition function defined in
```
hstojic: acq fnc is the class below, this is a builder so leave the first sentence here, but move the rest to the docstring in the class below

trieste/acquisition/function/active_learning.py
hstojic: there is also BatchBALD in https://proceedings.neurips.cc/paper/2019/file/95323660ed2124450caaac2c46b5ed90-Paper.pdf, but not sure if the code can be easily extended here...

trieste/acquisition/function/active_learning.py
satrialoka: added, test will be added later along with another tests

trieste/acquisition/function/function.py
satrialoka: we need math.pi for calculating the BALD, is there another alternative?

trieste/acquisition/function/active_learning.py
satrialoka: no it doesnt. will check it

trieste/acquisition/function/active_learning.py
satrialoka: moved

trieste/acquisition/function/function.py
hstojic: I'm saying its now in a wrong file, right?

hstojic:
let me know when you are ready for another round of reviews

trieste/acquisition/function/function.py
satrialoka: sorry my bad

trieste/acquisition/function/active_learning.py
```diff
@@ -418,3 +419,105 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+        :param model: The model of the objective function.
+        :param jitter: The size of the jitter to avoid numerical problem caused by the
+                log operation if variance is close to zero.
+        :return: The Bayesian Active Learning By Disagreement acquisition function.
+        """
+        tf.debugging.assert_positive(jitter, message="Jitter must be positive.")
+        assert (
+            type(model).__name__ == "VariationalGaussianProcess"
+            and type(model.model.likelihood).__name__ == "Bernoulli"  # type: ignore
```
satrialoka: investigating type error

trieste/acquisition/function/active_learning.py
```diff
@@ -323,12 +496,21 @@ def __init__(self, model: ProbabilisticModel, jitter: float):
         :return: The Bayesian Active Learning By Disagreement acquisition function.
         """
         tf.debugging.assert_positive(jitter, message="Jitter must be positive.")
+        assert (
```
hstojic: please use `tf.debugging.Assert` or if condition and then `raise ValueError`

hstojic: also, make sure we have a test for this

tests/unit/acquisition/function/test_active_learning.py
```diff
@@ -392,3 +400,86 @@ def test_integrated_variance_reduction_builder_updates_without_retracing() -> No
...
+    x = to_default_float(tf.zeros([1, 1]))
+    y = to_default_float(tf.zeros([1, 1]))
+    model = VariationalGaussianProcess(vgp_model_bernoulli(x, y))
+    acq_fn = BayesianActiveLearningByDisagreement().prepare_acquisition_function(model)
+    query_at = tf.linspace([[-10]], [[10]], 100)
+    expected = bayesian_active_learning_by_disagreement(model, DEFAULTS.JITTER)(query_at)
+    npt.assert_array_almost_equal(acq_fn(query_at), expected)
+
+
+def test_bayesian_active_learning_by_disagreement_raise_on_non_bernoulli_vgp() -> None:
```
hstojic: we should have a test for non vgp model as well, no? you are checking for that type as well...

hstojic (commented):
looks like we are almost there, some super minor stuff

2 bigger items
1. we still don't have any unit tests that check for correctness of the results
2. just realised there is no integration test, we should have one - you can reuse the code from the notebook for this - you can put it in test_active_learning.py

tests/unit/acquisition/function/test_active_learning.py
satrialoka: yes, I think that also needed.

trieste/acquisition/function/active_learning.py
satrialoka: Sure!

satrialoka:
> looks like we are almost there, some super minor stuff
> 
> 2 bigger items
> 
>     1. we still don't have any unit tests that check for correctness of the results
> 
>     2. just realised there is no integration test, we should have one - you can reuse the code from the notebook for this - you can put it in test_active_learning.py

Sorry just got back from another work.

1. I am not sure how to do this, should I make some dummy problem and check if BALD sample in the right direction?
2. Sure!

tests/unit/acquisition/function/test_active_learning.py
hstojic: by the way, we could also have an SVGP based classifier

hstojic:
> > looks like we are almost there, some super minor stuff
> > 2 bigger items
> > ```
> > 1. we still don't have any unit tests that check for correctness of the results
> > 
> > 2. just realised there is no integration test, we should have one - you can reuse the code from the notebook for this - you can put it in test_active_learning.py
> > ```
> 
> Sorry just got back from another work.
> 
> 1. I am not sure how to do this, should I make some dummy problem and check if BALD sample in the right direction?
> 2. Sure!

It depends on the method, exact equations etc - check out tests for some other acquisition functions to see how we do it elsewhere and give it a try, let me know if you are still stuck

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,187 @@
...
+from trieste.models.gpflow import VariationalGaussianProcess
+from trieste.models.optimizer import Optimizer
+
+optimizer = Optimizer(
+    optimizer=gpflow.optimizers.Scipy(), minimize_args={"options": dict(maxiter=100)}
+)
+
+
+def create_bo_model(data):
+    kernel = gpflow.kernels.SquaredExponential(lengthscales=[0.2, 0.2])
```
hstojic: we now have convenient gpflow model builders in `models/gpflow/builders.py`, you can use those to set up a VGP model with sensible defaults (there is also SVGP classification model)

tests/integration/test_active_learning.py
```diff
@@ -280,3 +288,72 @@ def _get_feasible_set_test_data(
...
+        build_vgp_classifier(initial_data, search_space, noise_free=True)
+    )
+
+
+def svgp_classification_model(initial_data: Dataset, search_space: Box) -> SparseVariational:
+    return SparseVariational(build_svgp(initial_data, search_space, classification=True))
+
+
+@random_seed
+@pytest.mark.parametrize(
```
hstojic, line 299: if it takes more than a minute or soe mark it as a slow test
```
@pytest.mark.slow
```

tests/integration/test_active_learning.py
```diff
@@ -280,3 +288,72 @@ def _get_feasible_set_test_data(
...
+    "num_steps, model_builder",
+    [
+        (20, vgp_classification_model),
+        (70, svgp_classification_model),
+    ],
+)
+def test_optimizer_learns_circle_function(
```
hstojic:
```suggestion
def test_bald_learner_learns_circle_function(
```

trieste/acquisition/function/active_learning.py
```diff
@@ -421,3 +424,107 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        supported (see equation 6 of the paper)
+
+        :param model: The model of the objective function.
+        :param jitter: The size of the jitter to avoid numerical problem caused by the
+                log operation if variance is close to zero.
+        :return: The Bayesian Active Learning By Disagreement acquisition function.
+        """
+        tf.debugging.assert_positive(jitter, message="Jitter must be positive.")
+
+        if type(model) != VariationalGaussianProcess and type(model) != SparseVariational:
```
hstojic: I'm wondering whether it makes sense to require VGP and SVGP, looking at the code below, any classification model that can give you a mean and a variance can be used, no? perhaps we are too restrictive here?

satrialoka: I think it is needed, in this version of BALD we need to use the Gaussian model because it uses specific analytical and Taylor approximate of the entropy terms in equation 2 of the paper: see details on [pages 4-5](https://arxiv.org/pdf/1112.5745.pdf).  Not sure if we use a non-gaussian model (for example deep GP) this approximation and analytical form of the entropy terms are still valid or not.  Correct me if I am wrong.

hstojic: ok, I'll check the paper again and get back to you here

forgot to add, here you should use `isinstance` check then this type construction

hstojic: the paper says "Throughout this section we will assume that we are provided with such a Gaussian approximation from one of these methods, though the active learning algorithm does not care which one." where approximation is of the posterior over f - hence it doesn't strictly need to be GPC, but admittedly it's difficult to check that a model is using a Gaussian approx of the posterior...

we are avoiding these kind of direct checks so I think we should remove the checks (and associated tests) and just add a paragraph in the docstring (potentially also notebook) that its been designed for a VGP classifier with Bernoulli likelihood, but other Gaussian approx of the posterior can be used (SVGP for instance, or some other custom model that is not currently supported by Trieste)

satrialoka: ah yeah,
I'll remove this check (and the tests) and write in docstring about it

hstojic:
let me know when you have removed the checks and the tests!

satrialoka:
> let me know when you have removed the checks and the tests!

sorry forgot to commit and push, done!
I'll try to work on the test today. if I'm stuck, I'll contact you through slack

tests/unit/acquisition/function/test_active_learning.py
```diff
@@ -393,6 +398,41 @@ def test_integrated_variance_reduction_builder_updates_without_retracing() -> No
...
+    "at",
+    [
+        (tf.constant([[[-1.0]]])),
+        (tf.constant([[-0.5]])),
+        (tf.constant([[0.0]])),
+        (tf.constant([[0.5]])),
+        (tf.constant([[1.0]])),
+    ],
+)
+def test_bayesian_active_learning_by_disagreement(at: tf.Tensor) -> None:
```
satrialoka: Not sure if this is the right way to test this, or there are better strategies ? @hstojic

CITATION.cff
```diff
@@ -17,8 +17,8 @@ authors:
   given-names: "Ivo"
 - family-names: "Qing"
   given-names: "Jixiang"
-- family-names: "Satrio"
-  given-names: "Loka"
+- family-names: "Loka"
+  given-names: "Nasrulloh"
```
hstojic, line 21: apologies about this mistake!

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,179 @@
...
+np.random.seed(1793)
+tf.random.set_seed(1793)
+
+# %% [markdown]
+# ## The problem
+
+# %% [markdown]
+# In Trieste, it is also possible to query most interesting points for learning the problem, i.e we want to have as little data as possible to construct the best possible model (active learning). In this tutorial we will try to do active learning for binary classification problem using Bayesain Active Learning by Disagreement (BALD) for a Gaussian Process Classification Model.
+#
+# Our problem is classification problem of circle dataset. In this toturial we assume that the input space is continous so we can use continuous optimiser for our BALD acquisition function.
```
hstojic:
```suggestion
# We will illustrate the BALD algorithm on a synthetic binary classification problem where one class takes shape of a circle in the search space. The input space is continuous so we can use continuous optimiser for our BALD acquisition function.
```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,179 @@
...
+
+# %%
+search_space = trieste.space.Box([-1, -1], [1, 1])
+input_dim = 2
+
+
+def circle(x):
+    return tf.cast((tf.reduce_sum(tf.square(x), axis=1, keepdims=True) - 0.5) > 0, tf.float64)
+
+
```
hstojic, line 38:
```suggestion
# %% [markdown]
# Let's first illustrate how this two dimensional problem looks like.

```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,179 @@
...
+idx = np.squeeze(observations).astype(bool)
+plt.scatter(query_points[idx][:, 0], query_points[idx][:, 1], label="1")
+plt.scatter(
+    query_points[np.logical_not(idx)][:, 0], query_points[np.logical_not(idx)][:, 1], label="0"
+)
+plt.legend()
+plt.show()
+
+# %% [markdown]
+# Let's generate some data for our initial model. Here we randomly sample 10 data points.
```
hstojic:
```suggestion
# Let's generate some data for our initial model. Here we randomly sample a small number of data points.
```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,179 @@
...
+num_initial_points = 5
+X = search_space.sample(num_initial_points)
+observer = mk_observer(circle)
+initial_data = observer(X)
+
+# %% [markdown]
+# ## Modelling the binary classification task
+
+# %% [markdown]
+# For the binary classification model, we use the Variational Gaussian Process with Bernoulli likelihood. For more detail of this model, see <cite data-cite="Nickisch08a">[Nickisch et al.](https://www.jmlr.org/papers/volume9/nickisch08a/nickisch08a.pdf)</cite>. Here we use trieste's gpflow model builder `build_vgp_classifier`.
```
hstojic, line 73: can you try building the docs and check that the references turn out fine?

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,179 @@
...
+# %% [markdown]
+# Lets see our model landscape using only those initial data
+
+# %%
+model.update(initial_data)
+model.optimize(initial_data)
+
+mean, variance = model.predict(query_points)
+
+plt.figure()
```
hstojic: you might be able to use a plot function for this from utils package here in the notebooks folder

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,179 @@
...
+    markersize=10,
+)
+plt.title("Variance")
+plt.show()
+
+
+# %% [markdown]
+# ## The acquisition process
+#
+# We can construct the BALD acqusition function which maximise information gain about the model parameters, by maximising the mutual information between predictions and model posterior:
```
hstojic:
```suggestion
# We can construct the BALD acquisition function which maximises information gain about the model parameters, by maximising the mutual information between predictions and model posterior:
```

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,179 @@
...
+#
+# $$\mathbb{I}\left[y, \boldsymbol{\theta} \mid \mathbf{x}, \mathcal{D}\right]=\mathbb{H}\left[y \mid \mathbf{x}, \mathcal{D}\right]-\mathbb{E}_{p\left(\boldsymbol{\theta} \mid \mathcal{D}\right)}[\mathbb{H}[y \mid \mathbf{x}, \boldsymbol{\theta}]]$$
+#
+# See <cite data-cite="houlsby2011bayesian">[Houlsby et al.](https://arxiv.org/pdf/1112.5745.pdf)</cite> for more details. Then, Trieste's `EfficientGlobalOptimization` is used for the query rule:
+
+# %%
+initial_models = model
```
hstojic: why do you need this?

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -0,0 +1,179 @@
...
+
+# %% [markdown]
+# ## Visualising the result
+# Now, we can visualize our model after the active learning run
+
+# %% Plot BO results
+mean, variance = final_model.predict(query_points)
+
+
+def invlink(f):
```
hstojic: since you use it only once immediately below perhaps there is no need to define a function, just compute it directly below

trieste/acquisition/function/active_learning.py
```diff
@@ -421,3 +423,106 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+        Here :math:`\mathrm{h}(p)` is defined as:
+
+        .. math::
+            \mathrm{h}(p)=-p \log p-(1-p) \log (1-p)
+
+        This acquisition function is intended to use for Binary Gaussian Process Classification
+        model with Bernoulli likelihood. It is designed for VGP but other Gaussian approximation
+        of the posterior can be used. SVGP for instance, or some other model that is not currently
+        supported by Trieste. integrating over nuisance parameters is currently not
```
hstojic:
```suggestion
        supported by Trieste. Integrating over nuisance parameters is currently not
```

trieste/acquisition/function/active_learning.py
```diff
@@ -421,3 +423,106 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        supported (see equation 6 of the paper)
```
hstojic:
```suggestion
        supported (see equation 6 of the paper).
```

trieste/acquisition/function/active_learning.py
```diff
@@ -421,3 +423,106 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+        :param model: The model of the objective function.
+        :param jitter: The size of the jitter to avoid numerical problem caused by the
+                log operation if variance is close to zero.
+        :return: The Bayesian Active Learning By Disagreement acquisition function.
+        """
+        tf.debugging.assert_positive(jitter, message="Jitter must be positive.")
+
+        if not isinstance(model.model.likelihood, Bernoulli):  # type: ignore
+            raise TypeError
```
hstojic: this is GPflow specific so we need to remove this check as well
```suggestion
```

tests/unit/acquisition/function/test_active_learning.py
```diff
@@ -383,3 +396,121 @@ def test_integrated_variance_reduction_builder_updates_without_retracing() -> No
...
+    npt.assert_array_almost_equal(acq_fn(query_at), expected)
+
+
+def test_bayesian_active_learning_by_disagreement_raise_on_non_bernoulli_vgp() -> None:
+    x = to_default_float(tf.zeros([1, 1]))
+    y = to_default_float(tf.zeros([1, 1]))
+    model = VariationalGaussianProcess(vgp_model(x, y))
+
+    with pytest.raises(Exception):
+        BayesianActiveLearningByDisagreement().prepare_acquisition_function(model)
```
hstojic:
```suggestion
```

tests/unit/acquisition/function/test_active_learning.py
```diff
@@ -383,3 +396,121 @@ def test_integrated_variance_reduction_builder_updates_without_retracing() -> No
...
+def test_bayesian_active_learning_by_disagreement(at: tf.Tensor) -> None:
+
+    search_space = Box([-1], [1])
+    x = to_default_float(tf.constant(np.linspace(-1, 1, 8).reshape(-1, 1)))
+    y = to_default_float(tf.reshape(binary_line(x), [-1, 1]))
+    model = VariationalGaussianProcess(
+        build_vgp_classifier(Dataset(x, y), search_space, noise_free=True)
+    )
+
+    p, _ = model.model.predict_y(to_default_float(at))
```
hstojic: `VariationalGaussianProcess` has `predict_y` method, this should work as well
```suggestion
    p, _ = model.predict_y(to_default_float(at))
```

tests/unit/acquisition/function/test_active_learning.py
hstojic: reimplementing equations again is not a way to go, test should try to verify them independently - partly if not fully... if its not really feasible then we rely on integration tests

here I think we can test it independently actually:
1. partial test, for example, it seems like 2nd term can be checked in a different way, see on page 5 this paragraph "Fig. 1 depicts the striking accuracy ..." 
2. partial test again, for the second term only, but probably better MC based test, see "Quantifying Approximation Losses" in section 5

hstojic (commented):
great work! I have left some minor suggestions for text etc

unit test of correctness is not really up to the task, I suggest verifying the approximation of the second term with monte carlo sampling, as in section 5 of the paper

CITATION.cff
satrialoka: no problem, it's quite complicated

tests/unit/acquisition/function/test_active_learning.py
satrialoka: Just replace it with MC based test for both term1 and term2, could you please verify if the code for term2 is indeed right? 
thanks for your patience!

docs/notebooks/active_learning_for_binary_classification.pct.py
```diff
@@ -79,47 +82,56 @@ def circle(x):
...
-plt.plot(
-    initial_data.query_points[:, 0],
-    initial_data.query_points[:, 1],
-    "ko",
-    markersize=10,
-)
-plt.title("Mean")
-plt.colorbar()
-plt.show()
+def plot_active_learning_query(function, points, num_initial_points, title):
```
hstojic: I probobaly should have pointed you to a right function, but was a bit lazy to look it up - `plot_model_predictions_plotly` does exactly what you want here, and to add acq points you can use `add_bo_points_plotly` - see for an example [here](https://github.com/secondmind-labs/trieste/blob/ce84c5e01ccbebfa82ef0490a7277445a8eb1cd2/docs/notebooks/expected_improvement.pct.py#L203)

docs/notebooks/active_learning_for_binary_classification.pct.py
satrialoka: removed

trieste/acquisition/function/active_learning.py
satrialoka: not used anymore

docs/notebooks/active_learning_for_binary_classification.pct.py
satrialoka: Just did it, yes the references are fine

hstojic:
I hope you don't mind that I made some changes to your correctness test, I also updated notebook a bit by reusing existing plotting code and did some other smaller fixes

if you are happy with this, let me know and we can merge it in

satrialoka:
No problems! Like it that way cleaner and easy to read.
Yes I am ready to merge

hstojic (approved):
official stamp of approval :)

great job!

PR #418, opened 2021-11-15
Uri/fix vgp wrapper

### LINKED ISSUE #406: TF compilation causes problems for VGP models

uri-granta:
Fix for https://github.com/secondmind-labs/trieste/issues/406. The issue was that the VGP variational parameters `VGP.m_qu` and `VGP.m_sqrt`, which are stored in `gpflow.Parameters`, need to be `.assigned` to like Variables rather than overriden.

Unfortunately to be able to assign we need to give them variable shapes, just like we do for the data. However `gpflow.Parameter` doesn't support this, so I've hacked a modified version that does (which in turn requires a modified version of `tfp.python.util.TransformedVariable`).

In the case where `use_natgrads` is False there is still an issue if people try to use the Adam optimizer, as the optimizer maintains state (e.g. moment vectors) which can't handle the changing parameter shapes. Suggest we handle this in a separate PR. For now I've removed tests of this combination (which isn't the recommended default).


tests/integration/test_bayesian_optimization.py
```diff
@@ -176,19 +181,44 @@ def test_optimizer_finds_minima_of_simple_quadratic(
 ) -> None:
     # for speed reasons we sometimes test with a simple quadratic defined on the same search space
     # branin; currently assume that every rule should be able to solve this in 5 steps
-    _test_optimizer_finds_minimum(False, min(num_steps, 5), acquisition_rule)
+    _test_optimizer_finds_minimum(min(num_steps, 5), acquisition_rule)
+
+
+@random_seed
+@pytest.mark.parametrize("use_natgrads", [False, True])
+def test_optimizer_with_vgp_model(use_natgrads: bool) -> None:
```
hstojic, line 197: should we rename the main test above then to mention gpr?

tests/integration/test_bayesian_optimization.py
```diff
@@ -198,9 +228,30 @@ def build_model(data: Dataset) -> GaussianProcessRegression:
...
+        elif model_type == "SVGP":
+            Z = search_space.sample_sobol(20)  # Initialize diverse inducing locations
+            svgp = gpflow.models.SVGP(
+                kernel,
+                gpflow.likelihoods.Gaussian(variance=1e-5),
+                Z,
+                num_data=len(data.observations),
+            )
+            gpflow.utilities.set_trainable(svgp.likelihood, False)
+            return SparseVariational(svgp, BatchOptimizer(tf.optimizers.Adam(0.1)))
```
hstojic, line 260: seems like a high learning rate for adam, but if the tests pass I guess its fine

tests/unit/models/gpflow/test_models.py
```diff
@@ -55,6 +55,7 @@
     SparseVariational,
     VariationalGaussianProcess,
 )
+from trieste.models.gpflow.models import NumDataPropertyMixin
```
hstojic, line 59: there's been quite a few updates in this file in latest PR

trieste/models/gpflow/models.py
```diff
@@ -469,9 +479,70 @@ def num_data(self, value: TensorType) -> None:
         self._num_data.assign(value)
 
 
-class SVGPWrapper(SVGP, NumDataPropertyMixin):
-    """A wrapper around GPFlow's SVGP class that stores num_data in a tf.Variable and exposes
-    it as a property."""
+class Parameter(gpflow.Parameter):
```
hstojic, line 489: can we move this class to util.py in this folder? this file is getting rather big...

trieste/models/gpflow/models.py
```diff
@@ -496,7 +567,12 @@ def __init__(self, model: SVGP, optimizer: Optimizer | None = None):
         # GPflow stores num_data as a number. However, since we want to be able to update it
         # without having to retrace the acquisition functions, put it in a Variable instead.
         # So that the elbo method doesn't fail we also need to turn it into a property.
-        if not isinstance(self._model, SVGPWrapper):
+        if not isinstance(self._model, NumDataPropertyMixin):
+
+            class SVGPWrapper(type(self._model), NumDataPropertyMixin):  # type: ignore
```
hstojic, line 581: better solution I think

trieste/models/gpflow/models.py
```diff
@@ -601,27 +672,67 @@ def __init__(
 
         self._use_natgrads = use_natgrads
         self._natgrad_gamma = natgrad_gamma
+        self._ensure_variable_model_data()
+
+    def _ensure_variable_model_data(self) -> None:
```
hstojic, line 692: this whole check could potentially live in util.py as well, takes a single argument (model) and optionally returns three objects with which we update some parts of it

hstojic (commented):
with the caveat that I couldn't assess properly the content of Parameter object and related changes in that _ensure method, it looks ok to me, 2 main comments:
1. I think most of this could go to utils.py, these classes and the file are getting big, and these things are not essential for understanding the VGP class so I don't think we lose much by moving them out of the view
2. not only related to the current PR, but as a more general solution for our sequential decision making use case would be to raise an issue/PR in GPflow, it might be possible to allow for dynamic shapes on their side, without hurting the regular supervised learning use case

tests/unit/models/gpflow/test_models.py
```diff
@@ -547,6 +549,11 @@ def test_gaussian_process_regression_optimize(
     else:
         args = {}
     loss = internal_model.training_loss(**args)
+
+    # VariationalGaussianProcess doesn't currently support Adam() unless use_natgrads is True
+    if isinstance(internal_model, VGP) and isinstance(optimizer, tf.optimizers.Optimizer):
+        return
```
jesnie: Maybe use `pytest.skip`, so we leave a log of this not being executed?

jesnie (approved):
LGTM

trieste/models/gpflow/models.py
apaleyes: Late to the party, but wanted second this. Perhaps just separate parameter.py in the same folder would be appropriate for this class?

PR #419, opened 2021-11-15
Fast update formulas with fantasized data

vpicheny:
This PR is an attempt to write the fast update formulas with fantasized data. Fantasized data is closely related to pending points.
This PR only deals with GPR models.

This PR does essentially 4 things:

- implement the "update equations" for GPR (for now these are new methods of GaussianProcessRegression) - note that I had to modify the covariance_between_points method to allow broadcasting. The new methods come with unit tests.
- implement a new IntegratedVarianceReduction acquisition function that uses the update equations. Also comes with unit tests + a notebook (that should be merged with the active learning one eventually).
- implement a new FantasizeAcquisitionFunction , which is an alternative to LocalPenalizationAcquisitionFunction. Note that this is limited for now to SingleModel but the final version should allow multiple models.
- this new function requires a new model wrapper, FantasizedGPRModel


trieste/models/gpflow/models.py
```diff
@@ -212,6 +214,70 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
         multiple_assign(self.model, current_best_parameters)
 
 
+    def fantasized_predict_f(self, query_points: TensorType, fantasized_data: Dataset) -> tuple[TensorType, TensorType]:
```
hstojic: this is creating a lot of baggage in the `GaussianProcessRegression` wrapper that most acq fnc will not use, I would probably subclass it and add these methods in a new class

vpicheny: as in creating a `FantasizeGPR`class? I was wondering if it made sense to also create a `FantasizeGPflowPredictor` in that case, any opinion?

hstojic: yes, `FantasizeGaussianProcessRegression` if we want to be consistent (names are getting long though :( )

if this would go beyond GPR, then yes, it would be good to standardise methods with a new interface like `FantasizeGPflowPredictor`, then it would be clear what methods a model would need to additionally implement

vpicheny: If we subclass, the issue is that most users won't understand why they need to use this very specific wrapper if they want to use, say, the `tIMSE` acquisition?

hstojic: input to the timse acq function is a model, so perhaps then the right interface level would be `FantasizeTrainableProbabilisticModel` instead of `FantasizeGPflowPredictor` and then use it as a type of the input, that would be good enough indication what wrapper they need to use?

other than that so far we kinda concluded acq fncs will check that the models have the necessary methods, they won't keep a list of valid wrappers, but I agree its not an ideal solution - the user then has to dig through the models to find a wrapper with such a method 

keeping an explicit mapping between acq fncs and models has its own problem, perhaps we could probably come up with a utility function that scans wrappers in `models` package and lists all the wrappers with a given method?

trieste/models/gpflow/models.py
```diff
@@ -464,3 +587,32 @@ def perform_optimization_step() -> None:  # alternate with natgrad optimizations
 
         else:
             self.optimizer.optimize(model, dataset)
+
+
+class FantasizedGPRModel(ProbabilisticModel):
```
apaleyes: Explicitly throw "This model doesn't support update and optimize" for this class

trieste/acquisition/function/function.py
```diff
@@ -762,3 +762,96 @@ def acquisition(x: TensorType) -> TensorType:
...
+        """
+        :param function: The acquisition function to update.
+        :param model: The model.
+        :param dataset: Unused.
+        """
+        return function  # no need to update anything
+
+
+class integrated_variance_reduction(AcquisitionFunctionClass):
+    def __init__(
```
hstojic, line 822: missing a docstring... perhaps also add a reference and one sentence summary of how its working?

trieste/acquisition/function/function.py
```diff
@@ -762,3 +762,96 @@ def acquisition(x: TensorType) -> TensorType:
...
+        self,
+        model: ProbabilisticModel,
+        integration_points: TensorType,
+        threshold: Sequence[float] = None,
+    ):
+        """
+        The reduction of the average of the predicted variance over the integration points
```
hstojic: these two lines should go to a docstring above `__init__`

trieste/acquisition/function/function.py
```diff
@@ -762,3 +762,96 @@ def acquisition(x: TensorType) -> TensorType:
...
+        (a.k.a. IMSE criterion)
+
+        :param model: The model of the objective function.
+        :param integration_points: points over which to integrate the objective prediction variance
+        :param threshold:
+        """
+        self._model = model
+        self._integration_points = integration_points
+        # self._threshold = threshold
+        if threshold is None:
```
hstojic, line 835: how does this make sense without a threshold?

hstojic:
@vpicheny we can close this draft now, right?

PR #424, opened 2021-11-23
Bump version to 0.8.0 and fix docs typo

uri-granta:


apaleyes (approved):
Does citation file need updating? It has version in it too

uri-granta:
> Does citation file need updating? It has version in it too

Good spot. Looks like this was added in 0.6.1 and didn't make into the release process docs (now added).

PR #426, opened 2021-11-23
Conditional / fantasize methods for GPR

vpicheny:
This PR adds 4 methods to the GPR class: `conditional_predict_f, conditional_predict_joint, conditional_predict_f_sample, conditional_predict_y`.

It also updates the method `covariance_between_points` to allow broadcasting.

Note that this refactors [this draft PR.](https://github.com/secondmind-labs/trieste/pull/419)

This PR unlocks writing the IMSE acquisition function and the "kriging believer" strategy, which will come in other PRs.


tests/unit/models/gpflow/test_models.py
```diff
@@ -643,3 +643,62 @@ def test_gpflow_predictor_get_observation_noise_raises_for_likelihood_with_varia
...
+        predj_meani, predj_covi = modeli.predict_joint(query_points)
+        pred_meani, pred_vari = modeli.predict(query_points)
+
+        np.testing.assert_allclose(predj_mean5[i, ...], predj_meani, atol=1e-5)
+        np.testing.assert_allclose(pred_meani, pred_mean5[i, ...], atol=1e-5)
+        np.testing.assert_allclose(pred_vari, pred_var5[i, ...], atol=1e-5)
+        np.testing.assert_allclose(predj_covi, predj_cov5[i, ...], atol=1e-5)
+
+
+test_gaussian_process_regression_conditional_predict_equations()
```
apaleyes: Is that call needed?

trieste/acquisition/function/entropy.py
```diff
@@ -540,11 +540,14 @@ def __call__(self, x: TensorType) -> TensorType:
...
         )  # need predictive variance of observations
-        A = tf.expand_dims(
-            self._model.covariance_between_points(  # type: ignore
-                tf.squeeze(x, -2), self._pending_points
+        A = tf.squeeze(
+            tf.expand_dims(
+                self._model.covariance_between_points(  # type: ignore
+                    tf.squeeze(x, -2), self._pending_points
+                ),
+                -1,
```
apaleyes:
```suggestion
                axis=-1,
```

trieste/acquisition/function/entropy.py
```diff
@@ -540,11 +540,14 @@ def __call__(self, x: TensorType) -> TensorType:
...
             ),
-            -1,
+            0,
```
apaleyes:
```suggestion
            axis=0,
```

trieste/models/gpflow/models.py
```diff
@@ -112,34 +114,52 @@ def covariance_between_points(
...
+
+        if len(tf.shape(K)) == 2:
+            K = tf.expand_dims(K, -3)
+            Kx1 = tf.expand_dims(Kx1, -3)
+            Kx2 = tf.expand_dims(Kx2, -3)
+            K12 = tf.expand_dims(K12, -3)
+        elif len(tf.shape(K)) > 3:
+            raise NotImplementedError(
+                "Covariance between points is not supported "
+                f"for kernels of type "
```
apaleyes:
```suggestion
                "for kernels of type "
```

trieste/models/gpflow/models.py
```diff
@@ -112,34 +114,52 @@ def covariance_between_points(
...
 
-        Kx2 = self.model.kernel(x, query_points_2)
-        Linv_Kx2 = tf.linalg.triangular_solve(L, Kx2)
+        Kx1 = leading_transpose(Kx1, [..., -1, -2])  # [..., L, num_data, N]
+        Linv_Kx1 = tf.linalg.triangular_solve(L, Kx1)  # [..., L, num_data, N]
+        Linv_Kx2 = tf.linalg.triangular_solve(L, Kx2)  # [L, num_data, M]
 
-        K12 = self.model.kernel(query_points_1, query_points_2)
-        cov = K12 - tf.tensordot(tf.transpose(Linv_Kx1), Linv_Kx2, [[-1], [-2]])
+        cov = K12 - tf.einsum("...lji,ljk->...lik", Linv_Kx1, Linv_Kx2)  # [..., L, N, M]
```
apaleyes, line 160: Would you mind leaving a comment here to explain what's going on in this line? `einsum` is not something you see often in code

trieste/models/gpflow/models.py
```diff
@@ -112,34 +114,52 @@ def covariance_between_points(
...
 
         tf.debugging.assert_shapes(
-            [(query_points_1, ["N", "D"]), (query_points_2, ["M", "D"]), (cov, ["N", "M"])]
+            [
+                (query_points_1, [..., "N", "D"]),
+                (query_points_2, ["M", "D"]),
+                (cov, [..., "L", "N", "M"]),
+            ]
```
apaleyes, line 167: Since this is the outcome of the calculation that isn't directly in user's control, if we were to throw an exception here, let's at least give a clue in the message as to why.

trieste/models/gpflow/models.py
```diff
@@ -211,6 +231,169 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+
+        mean_old_diff = additional_data.observations - mean_old  # [..., N, L]
+        mean_old_diff = leading_transpose(mean_old_diff, [..., -1, -2])[..., None]  # [..., L, N, 1]
+        AM = tf.linalg.triangular_solve(L_old, mean_old_diff)  # [..., L, N, 1]
+
+        mean_new = mean_new + leading_transpose(
+            (tf.matmul(A, AM, transpose_a=True)[..., 0]), [..., -1, -2]
+        )  # [..., M, L]
+
+        tf.debugging.assert_shapes(
```
apaleyes, line 295: ditto about a message. i'll stop now, but it applies to method below too

trieste/models/gpflow/models.py
```diff
@@ -211,6 +231,169 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+                (additional_data.observations, [..., "N", "L"]),
+                (query_points, ["M", "D"]),
+                (mean_new, [..., "M", "L"]),
+                (cov_new, [..., "L", "M", "M"]),
+            ]
+        )
+
+        return mean_new, cov_new
+
+    def conditional_predict_f_sample(
```
apaleyes, line 382: this needs a test

trieste/models/gpflow/models.py
```diff
@@ -211,6 +231,169 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+        :param query_points: Set of query points with shape [M, D]
+        :param additional_data: Dataset with query_points with shape [..., N, D] and observations
+                 with shape [..., N, L]
+        :param num_samples: number of samples
+        :return: samples of f at query points, with shape [..., num_samples, M, L]
+        """
+        mean_new, var_new = self.model.conditional_predict_joint(query_points, additional_data)
+        return sample_mvn(mean_new, var_new, full_cov=True, num_samples=num_samples)
+
+    def conditional_predict_y(
```
apaleyes, line 406: needs a test

trieste/models/gpflow/models.py
```diff
@@ -211,6 +231,169 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+        self, query_points: TensorType, additional_data: Dataset
+    ) -> tuple[TensorType, TensorType]:
+        """
+        Generates samples of y from the GP at query_points conditioned on both the model
+        and some additional data.
+
+        :param query_points: Set of query points with shape [M, D]
+        :param additional_data: Dataset with query_points with shape [..., N, D] and observations
+         with shape [..., N, L]
```
apaleyes: That might be failing the docs build, may need an indent

apaleyes (commented):
Since we seem to believe these four methods are very specific to kB,perhaps we could have an interface for now, in general `models/interfaces`, that can be called `Fantasizer` or something like that, that only `GaussianProcessRegression` will implement?

trieste/models/gpflow/models.py
vpicheny: any suggestion about how to do this?

apaleyes: I can only answer in very general terms. Why would you expect this assert to ever fail? If this can be formulated in a phrase, that's your error message. At the very least "received unexpected shapes after calculating blah and blah" shall suffice, if we can't say anything more useful.

vpicheny:
> Since we seem to believe these four methods are very specific to kB

Actually they are not: those are used also for the IMSE criterion.

tests/unit/acquisition/function/test_entropy.py
```diff
@@ -451,3 +451,6 @@ def test_batch_gibbon_is_sum_of_individual_gibbons_and_repulsion_term(
         npt.assert_array_almost_equal(
             calculated_batch_gibbon[i : i + 1], reconstructed_batch_gibbon
         )
+
+
+test_batch_gibbon_is_sum_of_individual_gibbons_and_repulsion_term(True, 0.1)
```
hstojic: I assume this is a leftover?

trieste/models/gpflow/models.py
```diff
@@ -112,34 +114,53 @@ def covariance_between_points(
...
+        Kx1 = self.model.kernel(query_points_1, x)  # [..., N, num_data] or [..., L, N, num_data]
+        Kx2 = self.model.kernel(x, query_points_2)  # [num_data, M] or [L, num_data, M]
+        K12 = self.model.kernel(query_points_1, query_points_2)  # [..., N, M] or [..., L, N, M]
+
+        if len(tf.shape(K)) == 2:
+            K = tf.expand_dims(K, -3)
+            Kx1 = tf.expand_dims(Kx1, -3)
+            Kx2 = tf.expand_dims(Kx2, -3)
+            K12 = tf.expand_dims(K12, -3)
+        elif len(tf.shape(K)) > 3:
```
hstojic, line 146: we should add a test that triggers this error

trieste/models/gpflow/models.py
```diff
@@ -112,34 +114,53 @@ def covariance_between_points(
 
         .. math:: \Sigma_{12} = K_{12} - K_{x1}(K_{xx} + \sigma^2 I)^{-1}K_{x2}
 
-        :param query_points_1: Set of query points with shape [N, D]
+        :param query_points_1: Set of query points with shape [..., N, D]
```
hstojic, line 120: perhaps add a sentence why shape for tensor 1 is different?

trieste/models/gpflow/models.py
```diff
@@ -112,34 +114,53 @@ def covariance_between_points(
...
 
-        K12 = self.model.kernel(query_points_1, query_points_2)
-        cov = K12 - tf.tensordot(tf.transpose(Linv_Kx1), Linv_Kx2, [[-1], [-2]])
+        # The line below is just A^T*B over the last 2 dimensions.
+        cov = K12 - tf.einsum("...lji,ljk->...lik", Linv_Kx1, Linv_Kx2)  # [..., L, N, M]
 
         tf.debugging.assert_shapes(
-            [(query_points_1, ["N", "D"]), (query_points_2, ["M", "D"]), (cov, ["N", "M"])]
+            [
+                (query_points_1, [..., "N", "D"]),
```
hstojic, line 164: shapes for query_points_1 and query_points_2 were tested above and weren't changed in the meantime, we dont need to test them again, right?

trieste/models/gpflow/models.py
hstojic: also, to a general user N, D, M and L might be unclear what it stands for - perhaps putting abbreviation definitions in the class docstring? 
(L is a batch right? usually B is used for that in TF I think)

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
 
         multiple_assign(self.model, current_best_parameters)
 
+    def conditional_predict_f(
+        self, query_points: TensorType, additional_data: Dataset
+    ) -> tuple[TensorType, TensorType]:
+        """
+        Returns the marginal GP distribution at query_points conditioned on both the model
+        and some additional data, using exact formula.
```
hstojic: perhaps a pointer where these formulas can be found?

hstojic: we use shape abbreviations here again, we should be clear somewhere what they stand for

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+                (additional_data.query_points, [..., "N", "D"]),
+                (additional_data.observations, [..., "N", "L"]),
+                (query_points, ["M", "D"]),
+            ],
+            message="additional_data must have query_points with shape [..., N, D]"
+            " and observations with shape [..., N, L], and query_points "
+            "should have shape [M, D]",
+        )
+
+        if isinstance(self.model, SGPR):
```
hstojic, line 265: we need a simple test that triggers this

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+            raise NotImplementedError("Conditional predict f is not supported for SGPR.")
+
+        mean_old, cov_old = self.model.predict_f(
+            additional_data.query_points, full_cov=True
+        )  # [..., N, L], [..., L, N, N]
+        mean_new, var_new = self.model.predict_f(query_points, full_cov=False)  # [M, L], [M, L]
```
hstojic: `mean_new` here is probably better named `mean_old`, since it corresponds to \hat{Z}(x)_old, right?

similarly, `var_new` would be `var_old`, corresponding to \sigma^2_old(x)?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+            ],
+            message="additional_data must have query_points with shape [..., N, D]"
+            " and observations with shape [..., N, L], and query_points "
+            "should have shape [M, D]",
+        )
+
+        if isinstance(self.model, SGPR):
+            raise NotImplementedError("Conditional predict f is not supported for SGPR.")
+
+        mean_old, cov_old = self.model.predict_f(
```
hstojic: for the sake of following equations more easily `mean_old` here is probably better named `mean_new_old`, since it corresponds to \hat{Z}(x_new)_old, right?

similarly, `cov_old` would be `cov_new_old`, corresponding to \sigma_old(x_new,x)?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
         multiple_assign(self.model, current_best_parameters)
 
+    def conditional_predict_f(
+        self, query_points: TensorType, additional_data: Dataset
+    ) -> tuple[TensorType, TensorType]:
+        """
+        Returns the marginal GP distribution at query_points conditioned on both the model
+        and some additional data, using exact formula.
+
+        :param query_points: Set of query points with shape [M, D]
```
hstojic, line 247: are query points meant to be training data here?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+        noise = self.model.likelihood.variance * tf.eye(
+            cov_shape[-2], batch_shape=cov_shape[:-2], dtype=cov_old.dtype
+        )
+        L_old = tf.linalg.cholesky(cov_old + noise)  # [..., L, N, N]
+        A = tf.linalg.triangular_solve(L_old, cov_cross, lower=True)  # [..., L, N, M]
+        var_new = var_new - leading_transpose(
+            tf.reduce_sum(A ** 2, axis=-2), [..., -1, -2]
+        )  # [M, L]
+
+        mean_old_diff = additional_data.observations - mean_old  # [..., N, L]
```
hstojic: how `additional_data.observations` makes sense here for the batch case where we don't really have these observations?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+
+        if isinstance(self.model, SGPR):
+            raise NotImplementedError("Conditional predict f is not supported for SGPR.")
+
+        mean_old, cov_old = self.model.predict_f(
+            additional_data.query_points, full_cov=True
+        )  # [..., N, L], [..., L, N, N]
+        mean_new, var_new = self.model.predict_f(query_points, full_cov=False)  # [M, L], [M, L]
+
+        cov_cross = self.covariance_between_points(
```
hstojic, line 268: this is \Sigma_new, right?

trieste/models/gpflow/models.py
hstojic: actually the covariance is \Sigma_new?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+
+        cov_shape = tf.shape(cov_old)
+        noise = self.model.likelihood.variance * tf.eye(
+            cov_shape[-2], batch_shape=cov_shape[:-2], dtype=cov_old.dtype
+        )
+        L_old = tf.linalg.cholesky(cov_old + noise)  # [..., L, N, N]
+        A = tf.linalg.triangular_solve(L_old, cov_cross, lower=True)  # [..., L, N, M]
+        var_new = var_new - leading_transpose(
+            tf.reduce_sum(A ** 2, axis=-2), [..., -1, -2]
+        )  # [M, L]
```
hstojic: shouldn't this be [..., "M", "L"]?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+        mean_old_diff = leading_transpose(mean_old_diff, [..., -1, -2])[..., None]  # [..., L, N, 1]
+        AM = tf.linalg.triangular_solve(L_old, mean_old_diff)  # [..., L, N, 1]
+
+        mean_new = mean_new + leading_transpose(
+            (tf.matmul(A, AM, transpose_a=True)[..., 0]), [..., -1, -2]
+        )  # [..., M, L]
+
+        tf.debugging.assert_shapes(
+            [
+                (additional_data.query_points, [..., "N", "D"]),
```
hstojic: no need to assert again `additional_data` and `query_points`, it was done above, right?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+        mean_old_diff = additional_data.observations - mean_old  # [..., N, L]
+        mean_old_diff = leading_transpose(mean_old_diff, [..., -1, -2])[..., None]  # [..., L, N, 1]
+        AM = tf.linalg.triangular_solve(L_old, mean_old_diff)  # [..., L, N, 1]
+        mean_new = m_new + leading_transpose(
+            (tf.matmul(A, AM, transpose_a=True)[..., 0]), [..., -1, -2]
+        )  # [..., M, L]
+
+        tf.debugging.assert_shapes(
+            [
+                (additional_data.query_points, [..., "N", "D"]),
```
hstojic: no need to assert again additional_data and query_points, it was done above, right?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+                (additional_data.observations, [..., "N", "L"]),
+                (query_points, ["M", "D"]),
+                (mean_new, [..., "M", "L"]),
+                (var_new, [..., "M", "L"]),
+            ],
+            message="received unexpected shapes computing conditional_predict_f,"
+            "check model kernel structure?",
+        )
+
+        return mean_new, var_new
```
hstojic: looks fine I think when looking at the equations, but not completely confident about some of the transformations when computing the weights etc, if tests are passing lets take them as ok - still, it would be great to have someone from the PM team to check the implementations, perhaps @alansaul?

trieste/models/gpflow/models.py
hstojic: so this is from https://arxiv.org/pdf/1203.6452.pdf
eq 8 to 10, right?

trieste/models/gpflow/models.py
```diff
@@ -211,6 +232,183 @@ def evaluate_loss_of_model_parameters() -> tf.Tensor:
...
+        and some additional data.
+
+        :param query_points: Set of query points with shape [M, D]
+        :param additional_data: Dataset with query_points with shape [..., N, D] and observations
+                 with shape [..., N, L]
+        :param num_samples: number of samples
+        :return: samples of f at query points, with shape [..., num_samples, M, L]
+        """
+        mean_new, cov_new = self.conditional_predict_joint(query_points, additional_data)
+        mean_for_sample = tf.linalg.adjoint(mean_new)  # [..., P, N]
```
hstojic: why do we need  conjugate transpose here?

hstojic: what is P here?

tests/unit/models/gpflow/test_models.py
```diff
@@ -643,3 +643,92 @@ def test_gpflow_predictor_get_observation_noise_raises_for_likelihood_with_varia
     model.model.likelihood = gpflow.likelihoods.Bernoulli()  # does not have variance attribute
     with pytest.raises(NotImplementedError):
         model.get_observation_noise()
+
+
+def test_gaussian_process_regression_conditional_predict_equations() -> None:
```
hstojic, line 675: nice test!

tests/unit/models/gpflow/test_models.py
```diff
@@ -643,3 +643,92 @@ def test_gpflow_predictor_get_observation_noise_raises_for_likelihood_with_varia
...
+
+
+def test_gaussian_process_regression_conditional_predict_f_sample() -> None:
+    x = gpflow.utilities.to_default_float(
+        tf.constant(np.arange(1, 24).reshape(-1, 1) / 8.0)
+    )  # shape: [23, 1]
+    y = fnc_2sin_x_over_3(x)
+
+    model5 = GaussianProcessRegression(gpr_model(x[:5, :], y[:5, :]))
+    additional_data = Dataset(tf.reshape(x[5:, :], [3, 6, -1]), tf.reshape(y[5:, :], [3, 6, -1]))
```
hstojic, line 747: why this reshaping? emulating a batch?

hstojic (commented):
ok, this took some time, implementations looks correct to me although I'm not 100% about some of the TF manipulations - I would have to go through them line by line to check in details, which I'm not sure is worth the time, tests look like they would catch it if something wasn't correct

I left various comments, but mostly minor stuff, cleaning the code a bit, better naming so its easier to follow equations, improvements in the docs etc

still, it would be great if someone from the PM team would check the implementation

tests/unit/models/gpflow/test_models.py
vpicheny: emulating multiple batch - when you optimise over a batch, our optimiser will start by evaluating a set of random batches before running BFGS. This test makes sure we can do this properly.

tests/unit/acquisition/function/test_entropy.py
vpicheny: yep sorry, will remove

trieste/models/gpflow/models.py
vpicheny: right - but I think this test allows to test that all the dimensions are consistent. I took this from some code John wrote so I would assume it's good practice :-)

trieste/models/gpflow/models.py
vpicheny: correct, thx - will update the docstring

trieste/models/gpflow/models.py
vpicheny: Observations are only useful for the mean. If you look at the tIMSE PR, you can see that I used dummy values (zeros) because tIMSE only uses the variance. In the kriging believer PR, we need those values in general, you need to choose the way you emulate them.

trieste/models/gpflow/models.py
vpicheny: right - we still need some of those to get M and L though. Will update the test

trieste/models/gpflow/models.py
hstojic: not sure I follow, you mean the test for the last cov line is now different because we include test for `query_points_1` and `query_points_2` - they are not independent?

trieste/models/gpflow/models.py
hstojic: would it be a significant enough saving if we would have a method for updating the variances alone?

trieste/models/gpflow/models.py
hstojic: perhaps I misunderstand how `assert_shapes` works...

hstojic: yes, I misunderstood how it works :)

trieste/models/gpflow/models.py
vpicheny: I assumed that this checks that the same "N" and "M" for qp1 and qp2 are valid for cov. i.e. this is shorter to write than assert_equal(query_point_1.shape[-2], cov.shape[-2]), etc.
Is it not the case?

tests/unit/models/gpflow/test_models.py
vpicheny: thx!

trieste/models/gpflow/models.py
vpicheny: L is the number of latent GPs. So for GPR it's the number of output dims.

trieste/models/gpflow/models.py
vpicheny: no, they are points at which we predict.

trieste/models/gpflow/models.py
vpicheny: changed the names with add and qp, it's probably much clearer

trieste/models/gpflow/models.py
vpicheny: yep!

trieste/models/gpflow/models.py
vpicheny: it would indeed save one triangular solve and one matmul. I could add a Boolean with skip_mean_update=False, but is it worth the extra code?

trieste/models/gpflow/models.py
vpicheny: sorry, P is L. I used gpflow's code that expects the mean in this shape.

tests/unit/models/gpflow/test_models.py
```diff
@@ -181,6 +181,33 @@ def test_gpr_raises_for_invalid_num_kernel_samples() -> None:
         GaussianProcessRegression(gpr_model(x, y), num_kernel_samples=-1)
 
 
+def test_gpr_raises_for_covariance_between_invalid_query_points_2() -> None:
```
hstojic: since we actually test the wrapper not the udnerlying GPR model

```suggestion
def test_gaussian_process_regression_raises_for_covariance_between_invalid_query_points_2() -> None:
```

tests/unit/models/gpflow/test_models.py
```diff
@@ -181,6 +181,33 @@ def test_gpr_raises_for_invalid_num_kernel_samples() -> None:
...
+    data = mock_data()
+    model = GaussianProcessRegression(gpr_model(*data))
+
+    with pytest.raises(ValueError):
+        model.covariance_between_points(data[0], tf.expand_dims(data[0], axis=0))
+
+
+def test_sgpr_raises_for_conditional_predict() -> None:
```
hstojic:
```suggestion
def test_gaussian_process_regression_raises_for_conditionals_with_sgpr() -> None:
```

trieste/models/gpflow/models.py
```diff
@@ -260,58 +265,58 @@ def conditional_predict_f(
...
         Predicts the joint GP distribution at query_points conditioned on both the model
-        and some additional data, using exact formula.
+        and some additional data, using exact formula. See :cite:`chevalier2014corrected`
+        (eqs. 8-10) for details.
 
         :param query_points: Set of query points with shape [M, D]
         :param additional_data: Dataset with query_points with shape [..., N, D] and observations
                  with shape [..., N, L]
-        :return: mean_new: predictive variance at query_points, with shape [..., M, L],
+        :return: mean_qp: predictive variance at query_points, with shape [..., M, L],
```
hstojic: inconsistency with `cov_new` below is here for a reason? also, here it says "variance", it should be mean, right?

trieste/models/gpflow/models.py
```diff
@@ -237,13 +241,14 @@ def conditional_predict_f(
...
         """
         Returns the marginal GP distribution at query_points conditioned on both the model
-        and some additional data, using exact formula.
+        and some additional data, using exact formula. See :cite:`chevalier2014corrected`
+        (eqs. 8-10) for details.
 
         :param query_points: Set of query points with shape [M, D]
         :param additional_data: Dataset with query_points with shape [..., N, D] and observations
                  with shape [..., N, L]
         :return: mean_new: predictive variance at query_points, with shape [..., M, L],
```
hstojic: variance -> mean?

trieste/models/gpflow/models.py
```diff
@@ -19,19 +19,21 @@
...
 from ...utils import DEFAULTS, jit
-from ..interfaces import TrainableProbabilisticModel
+from ..interfaces import FastUpdateModel, TrainableProbabilisticModel
 from ..optimizer import BatchOptimizer, Optimizer
 from .interface import GPflowPredictor
 from .utils import assert_data_is_compatible, randomize_hyperparameters, squeeze_hyperparameters
 
 
-class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel):
+class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel, FastUpdateModel):
```
hstojic, line 36: since now we have this new interface, we could instead have a subclass, say `FastGaussianProcessRegression` that inherits both `GaussianProcessRegression` and `FastUpdateModel`, wouldn't this be a bit cleaner? @apaleyes any thoughts?

vpicheny: should this be another PR?

trieste/models/gpflow/models.py
vpicheny: good catch

trieste/models/gpflow/models.py
apaleyes: many thoughts, but i guess we agreed to implement it like that to unblock @vpicheny

apaleyes (approved):
thanks for implementing the interface. it looks good for now to unblock further PRs, just make sure these PRs are programmed against the interface and no the GPR class.

hstojic (approved):
good to go!

PR #427, opened 2021-11-23
Don't swallow return code in quickdocs task

### LINKED ISSUE #437: Inequality constraints notebook fails for a random seed 1793
**Describe the bug**
When running inequality constraints notebook with random seed 1793, several people reported failures. It can also be seen in the github build: https://github.com/secondmind-labs/trieste/runs/4309808577

**To reproduce**
1. Clone and install Trieste from sources
2. Go to doc/notebooks/
3. Change random seed in `inequality_constraints.pct.py` to 1793, if not already set to it.
4. Run `python inequality_constraints.pct.py`. Don't forget to close matplotlib images when they appear!
5. Observe following error:
```
FailedOptimizationError: 
                    Acquisition function optimization failed,
                    even after 6 restarts.
```

**Expected behaviour**
The notebook completes without errors.

**System information**
 - OS: ...
 - Python version: 3.7.5
 - Trieste version: 0.8.0
 - TensorFlow version: 2.5

**Additional context**
The build error is now fixed by switching to a different random seed: #435. However the question on why the notebook fails for a certain random seed remains open.

uri-granta:
Turns out the quickdocs tox task was swallowing the return code, meaning it would never mark a build as failed!

This fixes that using an if-then-else. There might be a shorter way to do this by capturing $@, but I'd rather write as little bash as possible and this way is simpler.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -44,4 +44,4 @@ commands =
     docs: bash -c "cd docs; make html"
     quickdocs: pip install . -r notebooks/requirements.txt -c notebooks/constraints.txt
     quickdocs: pip install -r docs/requirements.txt -c docs/constraints.txt
-    quickdocs: bash -c "cd docs; python notebooks/quickrun/quickrun.py; if make html; then python notebooks/quickrun/quickrun.py --revert; else python notebooks/quickrun/quickrun.py --revert; exit1; fi"
+    quickdocs: bash -c "cd docs; python notebooks/quickrun/quickrun.py; if make html; then python notebooks/quickrun/quickrun.py --revert; else python notebooks/quickrun/quickrun.py --revert; exit 1; fi"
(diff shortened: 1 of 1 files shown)


hstojic (approved):
looks good

tox.ini
```diff
@@ -44,4 +44,4 @@ commands =
     docs: bash -c "cd docs; make html"
     quickdocs: pip install . -r notebooks/requirements.txt -c notebooks/constraints.txt
     quickdocs: pip install -r docs/requirements.txt -c docs/constraints.txt
-    quickdocs: bash -c "cd docs; python notebooks/quickrun/quickrun.py; make html; python notebooks/quickrun/quickrun.py --revert"
+    quickdocs: bash -c "cd docs; python notebooks/quickrun/quickrun.py; if make html; then python notebooks/quickrun/quickrun.py --revert; else python notebooks/quickrun/quickrun.py --revert; exit 1; fi"
```
apaleyes, line 47: how about this?
```
if make html; CODE=0; else CODE=1; fi; python notebooks/quickrun/quickrun.py --revert; exit CODE
```

PR #428, opened 2021-11-23
Victor/krigingbeliever

vpicheny:
This PR implements the "kriging believer" / "fantasize greedy" strategy. In short, this strategy 1) greedily optimises an acquisition function, 2) "fantasize" a value at the chosen query points (either by using the GP mean at this qp, or by generating a GP sample), 3) update the posterior of the GP, conditioning on this fantasized data, 4) iterate over 1-3.

We do this by introducing a `FantasizedGPRModel` wrapper. This model can only do predict, not optimize/update.


docs/notebooks/batch_optimization.pct.py
```diff
@@ -1,4 +1,4 @@
-# %% [markdown]
+z# %% [markdown]
```
hstojic: ?

trieste/acquisition/function/greedy.py
```diff
@@ -349,3 +353,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    else:
+        raise NotImplementedError(
+            f"Fantasize method not supported. "
+            f"Expected KB or sample, received "
+            f"{fantasize_method}"
+        )
+    return Dataset(pending_points, fantasized_obs)
+
+
+class FantasizedGPRModel(ProbabilisticModel):
```
hstojic: shouldn't this go to the models/gpflow package?

trieste/models/gpflow/models.py
```diff
@@ -112,34 +114,56 @@ def covariance_between_points(
 
         .. math:: \Sigma_{12} = K_{12} - K_{x1}(K_{xx} + \sigma^2 I)^{-1}K_{x2}
 
-        :param query_points_1: Set of query points with shape [N, D]
+        :param query_points_1: Set of query points with shape [..., N, D]
         :param query_points_2: Sets of query points with shape [M, D]
+<<<<<<< HEAD
```
hstojic: uncleared conflict

trieste/acquisition/function/greedy.py
vpicheny: Good question. This is the only place where it's used, and it' not really a model, just a wrapper to allow KB, so to me it is less confusing to have it here. But happy to reconsider.

trieste/acquisition/function/greedy.py
```diff
@@ -16,17 +16,21 @@
...
-from typing import Callable, Optional, Union, cast
+from typing import Callable, Mapping, Optional, Union, cast
 
+import gpflow
 import tensorflow as tf
 import tensorflow_probability as tfp
 
 from ...data import Dataset
 from ...models import ProbabilisticModel
+from ...models.gpflow import GaussianProcessRegression
```
hstojic, line 27: this is where the new interface `FastUpdateModel` comes in, we can use it instead of `GaussianProcessRegression`

trieste/acquisition/function/greedy.py
```diff
@@ -16,17 +16,21 @@
 """
```
hstojic: name of the file is not descriptive enough - perhaps greedy_batch.py?

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
         p = -5  # following experiments of :cite:`Alvi:2019`.
         penalization = ((pairwise_distances / (self._radius + self._scale)) ** p + 1) ** (1 / p)
         return tf.reduce_prod(penalization, axis=-1)
+
+
+class FantasizeAcquisitionFunction(GreedyAcquisitionFunctionBuilder):
+    r"""
+    Builder of the acquisition function maker for greedily collecting batches, following the
+    kriging believer heuristic.
```
hstojic: it would be easy to add constant liar as well, no?

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    By iteratively building a batch of points though sequentially maximizing
+    this acquisition function,
+    TODO: add stuff
+    it provides diverse batches of candidate points.
+    """
+
+    def __init__(
+        self,
+        acquisition_function_builder: AcquisitionFunctionBuilder = None,
+        fantasize_method: str = "KB",
```
hstojic: we could have a class attribute with a dictionary of these methods and a class method that the user can call to list all available methods, that would also make it easier to do checks of this input and less space for error

another option is an Enum object

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    ):
+        """
+        :param acquisition_function_builder:
+        :raise tf.errors.InvalidArgumentError: If ``fantasize_method`` is not "KB" or "sample".
```
hstojic: `fantasize_method` is missing

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    acquisition function.
+    By iteratively building a batch of points though sequentially maximizing
+    this acquisition function,
+    TODO: add stuff
+    it provides diverse batches of candidate points.
+    """
+
+    def __init__(
+        self,
+        acquisition_function_builder: AcquisitionFunctionBuilder = None,
```
hstojic: for consistency with local penalization: `base_acquisition_function_builder`?

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            self._builder = ExpectedImprovement()
+        else:
+            self._builder = acquisition_function_builder
+
+        # TODO assert fantasize mode
+        self._fantasize_method = fantasize_method
+
+    def prepare_acquisition_function(
+        self,
+        models: Mapping[str, ProbabilisticModel],
```
hstojic: we should use the new interface here `FastUpdateModel`, no?

trieste/acquisition/function/greedy.py
vpicheny: sounds good, will do

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+                tag: _generate_fantasized_data(model, pending_points, self._fantasize_method)
+                for tag, model in models.items()
+            }
+            models = {
+                tag: FantasizedGPRModel(model, fantasized_data[tag])
+                for tag, model in models.items()
+            }
+
+        return self._builder.prepare_acquisition_function(models, datasets)
+
```
hstojic: we still need an `update_acquisition_function` right? to distinguish between starting a new optimisation set where we update the dataset with true observations and pass the real model, and batch collection where we update the batch and use the fantasized model

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+def _generate_fantasized_data(model, pending_points, fantasize_method):
```
hstojic: why is this a separate function? its easier if its a class method I think

trieste/acquisition/function/greedy.py
vpicheny: yes, although this heuristic doesn't come with a good default for the "liar" value. So it's probably not worth it for now?

The very flexible alternative would be to allow passing a callable. But probably an overkill?

trieste/acquisition/function/greedy.py
vpicheny: could you provide an example? not sure what you mean

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    if fantasize_method == "KB":
+        fantasized_obs, _ = model.predict(pending_points)
+    elif fantasize_method == "sample":
+        fantasized_obs = model.sample(pending_points, num_samples=1)  # TODO check dim
+    else:
```
hstojic: we could easily add constant liar here with say min, mean and max of observations

trieste/acquisition/function/greedy.py
hstojic: ok, I understand a bit better now, this model is passed to the base acquisition function which needs the usual methods which we mask here with the conditional ones - I agree then, it will only be used here and users will not need to know about it at all so we could leave it here (we should have a name beginning with _ in that case)

hstojic: if we condition on the new interface, no need to require GPR
if users don't need to know about it lets also make it private
```suggestion
class _fantasized_model(FastUpdateModel):
```

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    return Dataset(pending_points, fantasized_obs)
+
+
+class FantasizedGPRModel(ProbabilisticModel):
+    """
+    Creates new model from an existing GPR model and additional data.
+    This new model posterior is conditioned and the GPR data and the additional one.
+    """
+
+    def __init__(self, model: ProbabilisticModel, fantasized_data: Dataset):
```
hstojic:
```suggestion
    def __init__(self, model: FastUpdateModel, fantasized_data: Dataset):
```

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        """
+
+        :param model: a GPR model
+        :param fantasized_data: additional dataset to condition on
+        :raise NotImplementedError: If model is not of class GaussianProcessRegression.
+        """
+
+        if not isinstance(model, GaussianProcessRegression):
```
hstojic: we don't need to check its GaussianPprocessRegression, since we use FastUpdateModel

we should still check though that conditional methods actually exist so that a meaningful error appears at runtime

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            raise NotImplementedError(
+                f"FantasizedGPRModel only works with GPR models, received " f"{model.__repr__()}"
+            )
+
```
hstojic: super call is missing?

trieste/acquisition/function/greedy.py
```diff
@@ -351,3 +355,127 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        return self._model.get_kernel()
+
+    def log(self) -> None:
+        return self._model.log()
+
+    def update(self, dataset: Dataset) -> None:
+        raise NotImplementedError
+
+    def optimize(self, dataset: Dataset) -> None:
+        raise NotImplementedError
```
hstojic: potentially we can do a super call on the model passed as an argument and that would instantiate all the original methods that we don't want to overwrite

tests/unit/acquisition/function/test_local_penalization.py
```diff
@@ -27,7 +27,7 @@
     PenalizationFunction,
     UpdatablePenalizationFunction,
 )
-from trieste.acquisition.function.local_penalization import (
+from trieste.acquisition.function.greedy import (
```
hstojic: we need to rename the test file as well

trieste/acquisition/function/greedy.py
hstojic: it does show good performance in comparison to KB, I think its worth including it
providing min, mean, max is straightforward, callable would be nice but more work

trieste/acquisition/function/greedy.py
hstojic: check _REGISTRY in ModelRegistry in config.py as an example, we populate it with key-value pairs that act as available methods - names (keys) and any internal representation (values), then its easier to check whether correct method name is used, and we can have a static method that lists available methods (keys) for the user (e.g. get_registered_models in ModelRegistry)

for enum see https://www.tutorialspoint.com/enum-in-python, it would be another object, a bit unnecessary

trieste/acquisition/function/greedy.py
vpicheny: I don't think so? `prepare` checks if there are pending points and updates the model if there are. So `update` and `prepare` are the same... `GreedyAcquisitionFunctionBuilder` `prepare` method has pending points arguments anyway.

trieste/acquisition/function/greedy_batch.py
```diff
@@ -16,17 +16,21 @@
 """
 from __future__ import annotations
 
-from typing import Callable, Optional, Union, cast
+from typing import Callable, Mapping, Optional, Union, cast
 
+import gpflow
```
uri-granta, line 21: (it would be useful if you mention that function.local_penalization has been renamed to function.greedy_batch in the commit message when you merge, as it' make it easier to remember to highlight this in the next release notes)

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,247 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    "sample" uses samples from the model.
+    """
+
+    def __init__(
+        self,
+        base_acquisition_function_builder: Optional[
+            AcquisitionFunctionBuilder[ProbabilisticModel]
+            | SingleModelAcquisitionBuilder[ProbabilisticModel]
+        ] = None,
+        fantasize_method: str = "KB",
```
uri-granta, line 385: Conversation for a later date: do we want to increase our minimum supported Python version from 3.7, which is quite old now (the latest stable version is 3.10). There are some useful features we could use from later version: for example in 3.8 you could specify fantasize_method's type as `Literal["KB", "sample"]`, which is obviously clearer than `str`.

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,247 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+            fantasized_data = {
+                tag: self._generate_fantasized_data(model, pending_points)
+                for tag, model in models.items()
+            }
+            models = {
+                tag: _fantasized_model(model, fantasized_data[tag]) for tag, model in models.items()
+            }
+
+        return self._builder.prepare_acquisition_function(models, datasets)
```
uri-granta: Like you were worried, this currently retraces every time: both when it gets a new point for a batch, and when it starts a new loop. To avoid this we need to implement an `update_acquisition_function` method that doesn't call self._builder.prepare_acquisition_function each time and which updates the fantasized data in the models rather than generating new models each time. This is actually quite complex, and will probably require at least the following steps:

1. `_fantasize_model` needs to store the fantasized data in Variables, like our GaussianProcessRegression wrapper
2. it needs to implement an `update_fantasized_data` method to update the fantasized data
3. `FantasizeAcquisitionFunction.prepare_acquisition_function` needs to store the fantasize models somewhere that `update_acquisition_function`  can access them (the cleanest is to define a new fantasize_acquisition_function class wrapper similar to expected_improvement, though there's probably a hacky shorter way)
4. we needs to implement an `FantasizeAcquisitionFunction.update_acquisition_function` that calculates the new fantasized_data like here and calls `update_fantasized_data` on the fantasized model Variables.

Given the complexity (I may have missed something too), I think it makes sense to do this in a separate PR.

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,247 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        return self._model.get_observation_noise()
+
+    def get_kernel(self) -> gpflow.kernels.Kernel:
+        return self._model.get_kernel()
+
+    def log(self) -> None:
+        return self._model.log()
+
+    def update(self, dataset: Dataset) -> None:
+        raise NotImplementedError
```
uri-granta: Maybe also add an error string (here and and for optimize) explaining why this isn't supported? (presumably since this isn't intended to model the data, just be used for the acquisition function, since if we'd really wanted to we could have always called update on the underlying models)

tests/unit/acquisition/function/test_greedy_batch.py
```diff
@@ -168,6 +176,137 @@ def test_lipschitz_penalizers_raises_for_invalid_pending_points_shape(
...
-        soft_local_penalizer(QuadraticMeanAndRBFKernel(), pending_points, lipshitz_constant, best)
+        soft_local_penalizer(QuadraticMeanAndRBFKernel(), pending_points, lipschitz_constant, best)
+
+
+def test_fantasized_expected_improvement_builder_raises_for_invalid_num_samples() -> None:
+    with pytest.raises(tf.errors.InvalidArgumentError):
+        FantasizeAcquisitionFunction(ExpectedImprovement().using("OBJECTIVE"), "notKB")
+
+
+def test_fantasized_expected_improvement_builder_raises_for_invalid_model() -> None:
```
uri-granta, line 189: Maybe add an integration test too?

docs/notebooks/batch_optimization.pct.py
```diff
@@ -204,6 +226,17 @@ def build_model(data):
...
 
+# %% [markdown]
+# then with `FantasizeAcquisitionFunction`...
+
+# %%
+kriging_believer_rule = EfficientGlobalOptimization(  # type: ignore
+    num_query_points=3, builder=kriging_believer_acq
+)
+kriging_believer_result = bo.optimize(
+    num_steps, initial_data, model, acquisition_rule=kriging_believer_rule
```
uri-granta, line 237: looks good (and since you're using the same num_steps you don't have to do anything to get the quickrun notebook behaviour in github)

uri-granta (commented):
Looks great! A long comment regarding retracing, though suggest we do that as a separate PR (I'm happy to volunteer).

tests/unit/acquisition/function/test_greedy_batch.py
vpicheny: if that's OK I'll leave it for another PR? there's no integration test for LP either so we might do both at the same time. The 2 notebooks act as integration tests for now

docs/notebooks/batch_optimization.pct.py
```diff
@@ -64,13 +64,13 @@ def build_model(data):
 
 # %% [markdown]
 # ## Batch acquisition functions.
-# To perform batch BO, we must define a batch acquisition function. Three batch acquisition functions supported in Trieste are `BatchMonteCarloExpectedImprovement`, `LocalPenalizationAcquisitionFunction` and `GIBBON`.
+# To perform batch BO, we must define a batch acquisition function. Four batch acquisition functions supported in Trieste are `BatchMonteCarloExpectedImprovement`, `LocalPenalizationAcquisitionFunction`, `FantasizeAcquisitionFunction` and `GIBBON`.
```
hstojic: title contains only the original three, perhaps remove all of them and leave only "Batch Bayesian Optimization"?

hstojic: we should probably add references for each of these approaches...

docs/notebooks/batch_optimization.pct.py
```diff
@@ -64,13 +64,13 @@ def build_model(data):
...
+# Although all these acquisition functions recommend batches of diverse query points, the batches are chosen in very different ways. `BatchMonteCarloExpectedImprovement` jointly allocates the batch of points as those with the largest expected improvement over our current best solution. In contrast, the `LocalPenalizationAcquisitionFunction` greedily builds the batch, sequentially adding the maximizers of the standard (non-batch) `ExpectedImprovement` function penalized around the current pending batch points. `FantasizeAcquisitionFunction` works similarly, but instead of penalizing the acquisition model, it iteratively updates the predictive equations after "fantasizing" obervations at the previously chosen query points. `GIBBON` also builds batches in a greedy manner but seeks batches that provide a large reduction in our uncertainty around the maximum value of the objective function.
 #
 # In practice, `BatchMonteCarloExpectedImprovement` can be expected to have superior performance for small batches (`batch_size`<5) but scales poorly for larger batches.
```
hstojic: lets add some guidelines for scaling of the other 3 approaches

docs/notebooks/batch_optimization.pct.py
```diff
@@ -155,6 +169,14 @@ def build_model(data):
...
+    points_chosen_by_kriging_believer[:, 0],
+    points_chosen_by_kriging_believer[:, 1],
+    color="blue",
+    lw=10,
+    label="Kriging \nBeliever",
+    marker="o",
+)
 plt.scatter(
     points_chosen_by_gibbon[:, 0],
     points_chosen_by_gibbon[:, 1],
```
hstojic, line 182: below at line 200:
` EfficientGlobalOptimization` -> `EfficientGlobalOptimization`

docs/notebooks/batch_optimization.pct.py
hstojic: @uri-granta is it safe to re-use the model object that was already used in a BO loop?

docs/notebooks/multi_objective_ehvi.pct.py
```diff
@@ -134,7 +134,7 @@ def build_stacked_independent_objectives_model(
 # We can now run the optimization loop
 
 # %%
-num_steps = 30
+num_steps = 1
```
hstojic: why only one step? for quick docs we do that automatically, for users we can leave more iterations no?

docs/notebooks/multi_objective_ehvi.pct.py
```diff
@@ -214,6 +214,69 @@ def log_hv(observations):
...
+# %%
+model = build_stacked_independent_objectives_model(initial_data, num_objective)
+
+from trieste.acquisition.function import FantasizeAcquisitionFunction
+
+batch_ehvi = FantasizeAcquisitionFunction(ExpectedHypervolumeImprovement())
+batch_rule: EfficientGlobalOptimization = EfficientGlobalOptimization(
+    builder=batch_ehvi, num_query_points=3
+)
+num_steps = 2
```
hstojic: should be 10 as written in the text, no?

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
         p = -5  # following experiments of :cite:`Alvi:2019`.
         penalization = ((pairwise_distances / (self._radius + self._scale)) ** p + 1) ** (1 / p)
         return tf.reduce_prod(penalization, axis=-1)
+
+
+class FantasizeAcquisitionFunction(GreedyAcquisitionFunctionBuilder[ProbabilisticModel]):
```
hstojic: we don't have `AcquisitionFunction` in the names of other acquisition functions, except for this one and LP, shall we remove those parts and make it consistent with other names?

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        return _fantasized_model(model, fantasized_data)
+
+
+class _fantasized_model(ProbabilisticModel):
+    """
+    Creates a new model from an existing one and additional data.
+    This new model posterior is conditioned on both current model data and the additional one.
+    """
+
+    def __init__(self, model: ProbabilisticModel, fantasized_data: Dataset):
```
hstojic, line 496:
```suggestion
    def __init__(self, model: FastUpdateModel, fantasized_data: Dataset):
```

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        """
+        :param model: a model, must be of :class:`~FantasizedGPRModel`
```
hstojic:
```suggestion
        :param model: a model, must be of :class:`~FastUpdateModel`
```

hstojic: not sure if we need to use :class:~trieste.models.interfacesFastUpdateModel. here for the doc building or Sphynx picks up the import, would be good to build the docs and check before merging in

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        :param fantasized_data: additional dataset to condition on
+        :raise NotImplementedError: If model is not of :class:`~FantasizedGPRModel`.
```
hstojic:
```suggestion
        :raise NotImplementedError: If model is not of :class:`~FastUpdateModel``.
```

trieste/acquisition/function/greedy.py
vpicheny: I actually don't want to pass those methods, because one shouldn't update or optimize an FantasizeModel

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            )
+
+        self._model = model
+        self._fantasized_data = fantasized_data
+
+    def predict(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
+        """
+        This function wraps conditional_predict_f. It cannot directly call
+        conditional_predict_f, since it does not accept query_points with rank > 2.
+        We use map_fn to allow leading dimensions for query_points.
```
hstojic, line 517: unclear what is `map_fn`

docs/notebooks/multi_objective_ehvi.pct.py
vpicheny: still wip

docs/notebooks/multi_objective_ehvi.pct.py
vpicheny: also still wip, will update later

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+        :param query_points: shape [...*, n, d]
```
hstojic: aren't we using N usually, not n? same for D instead of d?

trieste/acquisition/function/greedy_batch.py
vpicheny: can't do this, it breaks type checking

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    def optimize(self, dataset: Dataset) -> None:
+        raise NotImplementedError
+
+
+def _broadcast_predict(
+    query_points: TensorType, fun: Callable[[TensorType], tuple[TensorType, TensorType]]
+) -> tuple[TensorType, TensorType]:
+    """
+    Utility function that allows leading dimensions for query_points when
+    fun only accepts rank2 tensors. It works by flattening query_points into
```
hstojic:
```suggestion
    fun only accepts rank < 2 tensors. It works by flattening query_points into
```

docs/notebooks/batch_optimization.pct.py
vpicheny: probably for another PR... this one is getting really big already

vpicheny: besides we can't provide a lot of guidance

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+        :param query_points: shape [...*, n, d]
+        :return: mean, shape [...*, ..., n, L] and cov, shape [...*, ..., L, n],
+            where ... are the leading dimensions of fantasized_data
+        """
+
+        def fun(qp: TensorType) -> tuple[TensorType, TensorType]:
+            return self._model.conditional_predict_f(qp, self._fantasized_data)
+
+        return _broadcast_predict(query_points, fun)
```
hstojic, line 527: this broadcasting adds a lot of additional code (computational overhead as well?)... I don't know how to improve it, perhaps @uri-granta has an advice how to optimize it further?

trieste/acquisition/function/greedy_batch.py
```diff
@@ -16,17 +16,21 @@
 """
```
hstojic, line 16: had quite a few comments here, I guess they were removed when you renamed the file - did you resolve all of them?

trieste/acquisition/function/greedy_batch.py
hstojic:
```suggestion
class FantasizeAcquisitionFunction(GreedyAcquisitionFunctionBuilder[FastUpdateModel]):
```

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+    There are several ways to "fantasize" data: the "kriging believer" heuristic (KB, see
+    :cite:`ginsbourger2010kriging`) uses the mean of the model as observations.
+    "sample" uses samples from the model.
+    """
+
+    def __init__(
+        self,
+        base_acquisition_function_builder: Optional[
+            AcquisitionFunctionBuilder[ProbabilisticModel]
```
hstojic, line 382:
```suggestion
            AcquisitionFunctionBuilder[FastUpdateModel]
```

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            | SingleModelAcquisitionBuilder[ProbabilisticModel]
```
hstojic, line 383:
```suggestion
            | SingleModelAcquisitionBuilder[FastUpdateModel]
```

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,268 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        ] = None,
+        fantasize_method: str = "KB",
+    ):
+        """
+
+        :param base_acquisition_function_builder: The acquisition function builder to use.
+            Defaults to :class:`~trieste.acquisition.ExpectedImprovement`.
+        :param fantasize_method" one of "KB", "sample"
```
hstojic:
```suggestion
        :param fantasize_method: Following options are available: "KB" and "sample".
```

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,295 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        observations = tf.split(fantasized_data.observations, model._event_sizes, axis=-1)
+        fmods = []
+        for mod, obs, event_size in zip(model._models, observations, model._event_sizes):
+            fmods.append(
+                (
+                    _fantasized_model(mod, Dataset(fantasized_data.query_points, obs)),
+                    event_size,
+                )
+            )
+        return ModelStack(*fmods)
```
vpicheny, line 485: this line generates the following message at tox -e types:

error: Argument 1 to "ModelStack" has incompatible type "*List[Tuple[_fantasized_model, Any]]"; expected "Tuple[TrainableProbabilisticModel, int]"

I have no idea how to fix this neatly. @uri-granta ??

trieste/acquisition/function/greedy_batch.py
vpicheny: @uri-granta ?

trieste/acquisition/function/greedy_batch.py
vpicheny: Yes, this is a bit unfortunate. But in terms of efficiency, I don't think it's as efficient as automatic broadcasting...

trieste/acquisition/function/greedy_batch.py
hstojic: previously I suggested (lost comment on file renaming...) to use a class property that lists the methods, is that a good alternative?

trieste/acquisition/function/greedy.py
hstojic: then we overwrite those two methods, add 1 line with a super call and remove 6 unnecessary lines for the other 3 methods :)

docs/notebooks/multi_objective_ehvi.pct.py
vpicheny: back to normal values

docs/notebooks/multi_objective_ehvi.pct.py
vpicheny: back to normal values

trieste/acquisition/function/greedy_batch.py
vpicheny: unfortunately no - but the only ones pending were regarding adding CL, and how to deal with fantasize_method

tests/unit/acquisition/function/test_greedy_batch.py
hstojic: I guess that's fine but raise an issue about it so it doesn't fall through the cracks...

docs/notebooks/batch_optimization.pct.py
hstojic: I meant a one-sentence recommendation that they will work well or become too expensive for batches more than X

hstojic: or we explicitly say we cannot provide guidance for the other 3 - as long as users are aware its not meant to be used for 100s of points

trieste/acquisition/function/greedy_batch.py
vpicheny: sounds good to me! this PR is already massive

trieste/acquisition/function/greedy_batch.py
hstojic: I think if its consistently done throughout (in FantasizeAcquisitionFunction... as well) it should work, no?

vpicheny: prepare_acquisition_function expects a ProbabilisticModel for instance

vpicheny: FastUpdateModel is not a subtype of ProbabilisticModel

docs/notebooks/batch_optimization.pct.py
vpicheny: @henrymoss what would you want to see here?

henrymoss: LP scales the best, KG is expensive with large batches and MCEI is performs poorly  with large batches

tests/unit/acquisition/function/test_greedy_batch.py
uri-granta: yes that's fine (the notebook is enough proof for now that integration isn't totally broken)

trieste/acquisition/function/greedy_batch.py
uri-granta: not sure that's worth the effort as people are unlike to try and access the list programmatically. when we do upgrade to 3.8+ I might go back through the interfaces and make this and parameters have more precise types

docs/notebooks/batch_optimization.pct.py
uri-granta: we fit the initial model at the start of each new bo optimization loop so it should be fine, right?

trieste/acquisition/function/greedy_batch.py
uri-granta: It probably should be then (just like TrainableProbabilisticModel). Note that it'll still fine to inherit from both TrainableProbabilisticModel and FastUpdateModel. It'll also make it possible to write `AcquisitionFunctionBuilder[FastUpdateModel]` like @hstojic suggested.

trieste/acquisition/function/greedy_batch.py
uri-granta: I think it's complaining that `_fantasized_model` isn't a TrainableProbabilisticModel just a ProbabilisticModel, and ModelStack currently expects the former (so it can call update and optimize).

Not sure what the nicest solution is. Perhaps to split ModelStack into a `NonTrainableModelStack(ProbabilisticModel)` class that supports all models and a `ModelStack(NonTrainableModelStack, TrainableProbabilisticModel)` subclass that only supports trainable models (like the current version) and adds the update and optimize methods?

trieste/acquisition/function/greedy_batch.py
uri-granta: (alternatively you could just omit the update and optimize methods, as this class isn't inheriting from TrainableProbabilisticModel anyway)

trieste/acquisition/function/greedy_batch.py
vpicheny: @apaleyes suggested to write `FastUpdateModel` as it is. We said we would reconsider this design more globally later. Not sure what's best now.

trieste/acquisition/function/greedy_batch.py
vpicheny: Possibly `ModelStack` and `TrainableModelStack`?

The alternative would be to have dummy trainable methods for `fantasized_models` - a bit hacky but less code change?

I don't have a strong opinion either way...

trieste/acquisition/function/greedy_batch.py
vpicheny: after discussion with @uri-granta , this is much trickier than it looks, especially if we want to handle `ModelStack`s properly. We'll discuss this issue elsewhere.

trieste/acquisition/function/greedy_batch.py
vpicheny: went for splitting ModelStack

vpicheny:
Introduces two breaking changes:
- ModelStack is renamed TrainableModelStack
- file local_penalisation is renamed greedy_batch

trieste/acquisition/function/greedy.py
vpicheny: I'll stick to this version I think. I have removed the update and optimize methods though.

trieste/acquisition/function/greedy_batch.py
```diff
@@ -465,7 +468,7 @@ def _generate_fantasized_data(
 
 
 def _fantasize_model(
-    model: ProbabilisticModel, fantasized_data: Dataset
+    model: ProbabilisticModel | ModelStack, fantasized_data: Dataset
```
uri-granta: (isn't this addition superfluous?)

trieste/models/interfaces.py
```diff
@@ -293,6 +266,61 @@ def reparam_sampler(self, num_samples: int) -> ReparametrizationSampler:
...
+        The order of individual models specified at :meth:`__init__` determines the order of the
+        :class:`TrainableModelStack` output dimensions.
+
+        :param model_with_event_size: The first model, and the size of its output events.
+            **Note:** This is a separate parameter to ``models_with_event_sizes`` simply so that the
+            method signature requires at least one model. It is not treated specially.
+        :param \*models_with_event_sizes: The other models, and sizes of their output events.
+        """
+        super().__init__(model_with_event_size, *models_with_event_sizes)
+        self._models, self._event_sizes = zip(*(model_with_event_size,) + models_with_event_sizes)
```
uri-granta: Isn't this line superfluous (the super's init should already do this)?

trieste/models/interfaces.py
```diff
@@ -380,7 +408,7 @@ def get_trajectory(self) -> TrajectoryFunction:
         raise NotImplementedError
 
 
-class FastUpdateModel(ABC):
+class FastUpdateModel(ProbabilisticModel):
```
uri-granta: Was this change required in the end? If not might make sense to leave for now and address later?

uri-granta (approved):
Looks good. Just a couple of small comments.

hstojic:
> Introduces two breaking changes:
> 
> * ModelStack is renamed TrainableModelStack
> * file local_penalisation is renamed greedy_batch

LocalPenalizationAcquisitionFunction has been renamed to LocalPenalization

docs/notebooks/batch_optimization.pct.py
```diff
@@ -64,13 +64,13 @@ def build_model(data):
...
+# Note that all these acquisition functions have controllable parameters. In particular, `BatchMonteCarloExpectedImprovement` is computed using a Monte-Carlo method (so it requires a `sample_size`), but uses a reparametrisation trick to make it deterministic. The `LocalPenalization` has parameters controlling the degree of penalization that must be estimated from a random sample of `num_samples` model predictions (we recommend at least 1_000 for each search space dimension). Similarly, `GIBBON` requires a `grid_size` parameter that controls its approximation accuracy (which should also be larger than 1_000 for each search space dimension). `Fantasizer` requires a method for "fantasizing" the observations, which can be done by sampling from the GP posterior, using the GP posterior mean (a.k.a "kriging believer" heuristic, our default setup) or using a constant value (a.k.a. "constant liar heuristic)
```
hstojic:
```suggestion
# Note that all these acquisition functions have controllable parameters. In particular, `BatchMonteCarloExpectedImprovement` is computed using a Monte-Carlo method (so it requires a `sample_size`), but uses a reparametrisation trick to make it deterministic. The `LocalPenalization` has parameters controlling the degree of penalization that must be estimated from a random sample of `num_samples` model predictions (we recommend at least 1_000 for each search space dimension). Similarly, `GIBBON` requires a `grid_size` parameter that controls its approximation accuracy (which should also be larger than 1_000 for each search space dimension). `Fantasizer` requires a method for "fantasizing" the observations, which can be done by sampling from the GP posterior or by using the GP posterior mean (a.k.a "kriging believer" heuristic, our default setup).
```

docs/notebooks/multi_objective_ehvi.pct.py
```diff
@@ -214,6 +214,69 @@ def log_hv(observations):
...
+    builder=batch_ehvi, num_query_points=3
+)
+num_steps = 10
+bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+batch_result = bo.optimize(
+    num_steps, initial_data, model, acquisition_rule=batch_rule
+)
+
+# %% [markdown]
+# We can have a look at the results, as in the previous case.
```
hstojic: any comment about the results, differences etc?

docs/refs.bib
```diff
@@ -286,6 +286,24 @@ @incollection{chevalier2014corrected
...
+@incollection{ginsbourger2010kriging,
+  title={Kriging is well-suited to parallelize optimization},
+  author={Ginsbourger, David and Le Riche, Rodolphe and Carraro, Laurent},
+  booktitle={Computational intelligence in expensive optimization problems},
+  pages={131--162},
+  year={2010},
+  publisher={Springer}
+}
+
+@article{moss2021gibbon,
```
hstojic: there already is a ref for gibbon lines 163-168, can you update that one?

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,286 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    Builder of the acquisition function maker for greedily collecting batches.
+    Fantasizer allows us to perform batch Bayesian optimization with any
+    standard (non-batch) acquisition function.
+
+    Here, every time a query point is chosen by maximising an acquisition function,
+    its corresponding observation is "fantasized", and the models are conditioned further
+    on this new artificial data.
+
+    This implies that the models need to predict what their updated predictions would be given
+    new data, see :class:`~FantasizedGPRModel`. These equations are for instance in closed form
```
hstojic: should this be FastUpdateModel?

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,286 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            AcquisitionFunctionBuilder[ProbabilisticModel]
+            | SingleModelAcquisitionBuilder[ProbabilisticModel]
+        ] = None,
+        fantasize_method: str = "KB",
+    ):
+        """
+
+        :param base_acquisition_function_builder: The acquisition function builder to use.
+            Defaults to :class:`~trieste.acquisition.ExpectedImprovement`.
+        :param fantasize_method: the following options are available: "KB" and "sample".
```
hstojic:
```suggestion
        :param fantasize_method: The following options are available: "KB" and "sample". See class docs for more details.
```

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,286 @@ def __call__(self, x: TensorType) -> TensorType:
         p = -5  # following experiments of :cite:`Alvi:2019`.
         penalization = ((pairwise_distances / (self._radius + self._scale)) ** p + 1) ** (1 / p)
         return tf.reduce_prod(penalization, axis=-1)
+
+
+class Fantasizer(GreedyAcquisitionFunctionBuilder[ProbabilisticModel]):
```
hstojic, line 360: can you please raise an issue about adding integration tests in a future PR so we don't forget about it

docs/notebooks/asynchronous_greedy_multiprocessing.pct.py
```diff
@@ -83,7 +83,7 @@ def build_model(data):
 
 
 # these imports will be used later for optimization
-from trieste.acquisition import LocalPenalizationAcquisitionFunction
```
hstojic, line 86: did you try it out with Fantasizer? should work as well, right?

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,286 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        fantasized_obs = model.sample(pending_points, num_samples=1)[0]
+    else:
+        raise NotImplementedError(
+            f"fantasize_method must be KB or sample, " f"received {model.__repr__()}"
+        )
+
+    return Dataset(pending_points, fantasized_obs)
+
+
+def _fantasize_model(
```
hstojic: perhaps a name that is easier to distinguish from _fantasized_model, e.g. _generate_fantasized_model?
```suggestion
def _generate_fantasized_model(
```

trieste/acquisition/function/greedy_batch.py
```diff
@@ -351,3 +355,286 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            fantasized_data = {
+                tag: _generate_fantasized_data(
+                    fantasize_method=self._fantasize_method,
+                    model=model,
+                    pending_points=pending_points,
+                )
+                for tag, model in models.items()
+            }
+            new_models = {
+                tag: _fantasize_model(model, fantasized_data[tag]) for tag, model in models.items()
```
hstojic:
```suggestion
                tag: _generate_fantasized_model(model, fantasized_data[tag]) for tag, model in models.items()
```

tests/unit/acquisition/function/test_greedy_batch.py
```diff
@@ -0,0 +1,419 @@
...
+    ExpectedImprovement,
+    MinValueEntropySearch,
+    PenalizationFunction,
+    UpdatablePenalizationFunction,
+)
+from trieste.acquisition.function import NegativePredictiveMean, PredictiveVariance
+from trieste.acquisition.function.greedy_batch import (
+    Fantasizer,
+    LocalPenalization,
+    _fantasize_model,
```
hstojic:
```suggestion
    _generate_fantasized_model,
```

tests/unit/acquisition/function/test_greedy_batch.py
```diff
@@ -0,0 +1,419 @@
...
+    pending_points: TensorType,
+    penalizer: Callable[..., PenalizationFunction],
+) -> None:
+    best = tf.constant([0], dtype=tf.float64)
+    lipschitz_constant = tf.constant([1], dtype=tf.float64)
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        soft_local_penalizer(QuadraticMeanAndRBFKernel(), pending_points, lipschitz_constant, best)
+
+
+def test_fantasized_expected_improvement_builder_raises_for_invalid_num_samples() -> None:
```
hstojic:
```suggestion
def test_fantasized_expected_improvement_builder_raises_for_invalid_fantasize_method() -> None:
```

hstojic (approved):
Few final minor comments, after that good to go!

To follow up with:
1. Add integration tests
2. Test that it works with async BO

docs/notebooks/asynchronous_greedy_multiprocessing.pct.py
vpicheny: I haven't tried. Indeed it should work. As suggested I'll add this to the integration tests (next PR).

PR #429, opened 2021-11-23
t-IMSE criterion

vpicheny:
Implement the t-IMSE acquisition function for active learning.

This function only works with GPR and requires the conditional update predict method of [this PR](https://github.com/secondmind-labs/trieste/pull/426).


docs/refs.bib
```diff
@@ -266,3 +266,13 @@ @article{bect2012sequential
     year={2012},
     publisher={Springer}
 }
+
+@article{Picheny2010,
+  title={Adaptive Designs of Experiments for Accurate Approximation of Target Regions},
+  author={Picheny, Victor and Ginsbourger, David and Roustant, Olivier and Haftka, Raphael T and Kim, Nam-Ho},
+  journal={Journal of mechanical design},
```
hstojic: "Journal of Mechanical Design"

trieste/acquisition/function/entropy.py
```diff
@@ -540,11 +540,14 @@ def __call__(self, x: TensorType) -> TensorType:
         L = tf.linalg.cholesky(
             B + noise_variance * tf.eye(len(self._pending_points), dtype=B.dtype)
         )  # need predictive variance of observations
-        A = tf.expand_dims(
-            self._model.covariance_between_points(  # type: ignore
-                tf.squeeze(x, -2), self._pending_points
+        A = tf.squeeze(
```
hstojic: was this a bug? did you perhaps create a test that catches it?

trieste/acquisition/function/function.py
```diff
@@ -689,3 +689,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+class IntegratedVarianceReduction(SingleModelAcquisitionBuilder):
+    """
+    Builder for the reduction of the integral of the predicted variance over the search
+    space given a batch of query points.
+    """
+
+    def __init__(self, integration_points: TensorType, threshold: Sequence[float] = None) -> None:
+        """
+        :param integration_points: set of points to integrate the prediction variance over.
+        """
```
hstojic: we should do some basic checks on `integration_points` and `threshold` here

trieste/acquisition/function/function.py
```diff
@@ -689,3 +689,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        :param model: The model.
+        :param dataset: Unused.
+        """
+        return function  # no need to update anything
+
+
+class integrated_variance_reduction(AcquisitionFunctionClass):
+    """
+    The reduction of the average of the predicted variance over the integration points
+    (a.k.a. IMSE criterion). See :cite:`Picheny2010` for details.
```
hstojic: perhaps spell out what IMSE stands for?

trieste/acquisition/function/function.py
```diff
@@ -689,3 +689,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+    If no threshold is provided, the original variance is used. Otherwise,
```
hstojic: perhaps add when to use which option (briefly of course)?

trieste/acquisition/function/function.py
```diff
@@ -689,3 +689,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    def __init__(
+        self,
+        model: ProbabilisticModel,
+        integration_points: TensorType,
+        threshold: Sequence[float] = None,
+    ):
+        """
+        :param model: The model of the objective function.
+        :param integration_points: points over which to integrate the objective prediction variance
+        :param threshold:
```
hstojic: description missing

trieste/acquisition/function/function.py
```diff
@@ -689,3 +689,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            )
+
+        self._model = model
+        self._integration_points = integration_points
+        if threshold is None:
+            self._weights = tf.cast(1.0, integration_points.dtype)
+        else:
+            mean_old, var_old = self._model.predict(query_points=integration_points)
+            distr = tfp.distributions.Normal(mean_old, tf.sqrt(var_old))
+            self._weights = distr.cdf(tf.cast(threshold[1], mean_old.dtype)) - distr.cdf(
```
hstojic: this means threshold is either None or a range,shouldn't there be an option for a single threshold, a scalar?

trieste/acquisition/function/function.py
```diff
@@ -689,3 +689,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        additional_data = Dataset(x, tf.ones_like(x[..., 0:1]))
+
+        try:
+            mean, variance = self._model.conditional_predict_f(  # type: ignore
+                query_points=self._integration_points, additional_data=additional_data
+            )
+        except NotImplementedError:
+            raise ValueError(
+                """
+                PredictiveVariance only supports models with a conditional_predict_ft method.
```
hstojic: PredictiveVariance?

tests/unit/acquisition/function/test_function.py
```diff
@@ -635,3 +644,59 @@ def test_batch_monte_carlo_expected_improvement_updates_without_retracing() -> N
     assert batch_ei.__call__._get_tracing_count() == 1  # type: ignore
     npt.assert_allclose(batch_ei(xs), ei(xs), rtol=0.06)
     assert batch_ei.__call__._get_tracing_count() == 1  # type: ignore
+
+
+def test_integrated_variance_reduction() -> None:
```
hstojic: these are great, but we should also have some basic tests, like building successfully an acq fnc, output shapes etc, the usual battery that you can find for other acq functions

hstojic (commented):
looks quite, provided some quick comments (nothing major...)

trieste/acquisition/function/entropy.py
vpicheny: this change is a consequence of me updating the covariance_between_points method. Now it always returns a tensor with an extra dimension (which would always be 1 unless it's a multi-output GPR)

trieste/acquisition/function/function.py
vpicheny: added the option

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+            raise AttributeError(
+                """
+                Integrated variance reduction only supports models with a
+                conditional_predict_f method.
+                """
+            )
+
+        self._model = model
+
+        tf.debugging.assert_equal(
```
vpicheny, line 354: I placed these tests here rather than in `IntegratedVarianceReduction` because it made my life slightly easier. We haven' t been very consistent in trieste about this kind of tests (). Not sure if it matters? @apaleyes @hrvoje

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+        integration_points: TensorType,
+        threshold: Optional[Sequence[float]] = None,
+    ):
+        """
+        :param model: The model of the objective function.
+        :param integration_points: points over which to integrate the objective prediction variance
+        :param threshold: either None, or a sequence of 1 or 2 float values
+        :raise ValueError (or InvalidArgumentError): If ``threshold`` has more than 2 values.
+        """
+        if not hasattr(model, "conditional_predict_f"):
```
vpicheny, line 344: not sure if I should test this or if the model inherits from `FastUpdateModel`? @apaleyes

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +237,58 @@ def objective_function(x):
...
+
+# %%
+from trieste.acquisition.function import IntegratedVarianceReduction
+
+# Choose integration points uniformly over the design space
+integration_points = search_space.sample_halton(1000)
+acq_ivr = IntegratedVarianceReduction(integration_points=integration_points, threshold=[threshold,])
+
+# Set a batch size greater than 1
+rule_ivr = EfficientGlobalOptimization(builder=acq_ivr, num_query_points=5) # type: ignore
```
vpicheny: I don't know why I needed to add `type: ignore`, but otherwise `tox` complains: "error: Need type annotation for "rule_ivr""

and I have no idea why

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+    Builder for the reduction of the integral of the predicted variance over the search
+    space given a batch of query points.
+    """
+
+    def __init__(
+        self, integration_points: TensorType, threshold: Optional[Sequence[float]] = None
+    ) -> None:
+        """
+        :param integration_points: set of points to integrate the prediction variance over.
+        :raise ValueError (or InvalidArgumentError): If ``integration_points`` does not have
```
hstojic: this is raised in the acq fnc not here in the builder, so this shouldn't be here

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+
+
+class integrated_variance_reduction(AcquisitionFunctionClass):
+    """
+    The reduction of the average of the predicted variance over the integration points
+    (a.k.a. Integrated Means Square Error or IMSE criterion).
+    See :cite:`Picheny2010` for details.
+
+    If no threshold is provided, the goal is to learn a globally accurate model, and
+    the original variance is used. Otherwise, learning is 'targeted' towards regions
```
hstojic: unclear what original refers to here

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+    """
+    Builder for the reduction of the integral of the predicted variance over the search
+    space given a batch of query points.
+    """
+
+    def __init__(
+        self, integration_points: TensorType, threshold: Optional[Sequence[float]] = None
+    ) -> None:
+        """
+        :param integration_points: set of points to integrate the prediction variance over.
```
hstojic, line 262: missing the threshold parameter in the docs...

trieste/acquisition/function/active_learning.py
hstojic: we should keep this as inheritance is checked by mypy, users might not use mypy so we need to check this at runtime and provide a useful message

trieste/acquisition/function/active_learning.py
hstojic: yeah, we haven't been consistent about that (this design with builders and acq fncs is also inelegant to me, lots of duplication), personally I'm trying to put these in the builder as thats what is executed first, but it is fine here as well I think, as it is not executed for every call

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+            mean_old, var_old = self._model.predict(query_points=integration_points)
+            distr = tfp.distributions.Normal(mean_old, tf.sqrt(var_old))
+            self._weights = distr.cdf(threshold[1]) - distr.cdf(threshold[0])
+
+    @tf.function
+    def __call__(self, x: TensorType) -> TensorType:
+
+        additional_data = Dataset(x, tf.ones_like(x[..., 0:1]))
+
+        try:
```
hstojic: I think you don't need this check here as you do it above already?
(well, it could happen that a model has the method but not really implemented for some reason - we do that sometimes when interface forces you to have it even if it doesnt make sense for the model, but shouldn't be the case here I think)

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+    where the GP is close to particular values, and the variance is weighted by the
+    posterior GP pdf evaluated at the threshold (if a single value is given) or by the
+    probability that the GP posterior belongs to the interval between the 2 thresholds.
+    """
+
+    def __init__(
+        self,
+        model: ProbabilisticModel,
+        integration_points: TensorType,
+        threshold: Optional[Sequence[float]] = None,
```
hstojic: what about the epsilon parameter, its not useful? 
since we say we use the formulation from the paper, if we are dropping it, it would be good to mention this in the docstring above

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+    """
+    The reduction of the average of the predicted variance over the integration points
+    (a.k.a. Integrated Means Square Error or IMSE criterion).
+    See :cite:`Picheny2010` for details.
+
+    If no threshold is provided, the goal is to learn a globally accurate model, and
+    the original variance is used. Otherwise, learning is 'targeted' towards regions
+    where the GP is close to particular values, and the variance is weighted by the
+    posterior GP pdf evaluated at the threshold (if a single value is given) or by the
+    probability that the GP posterior belongs to the interval between the 2 thresholds.
```
hstojic: It would be good to mention that the function can do batches of points, a brief explanation about updated variance (models need to be able to do this...), and computational costs, recommended batch size etc

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+            )
+        except NotImplementedError:
+            raise ValueError(
+                """
+                integrated_variance_reduction only supports models with a conditional_predict_f
+                method.
+                """
+            )
+
+        return -tf.reduce_mean(variance * self._weights, axis=-2)
```
hstojic, line 420: why minus here? since timse is maximised same as Bichon, we shouldn't have minus, optimizer adds a minus

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,153 @@ def acquisition(x: TensorType) -> TensorType:
...
+            distr = tfp.distributions.Normal(mean_old, tf.sqrt(var_old))
+            self._weights = distr.cdf(threshold[1]) - distr.cdf(threshold[0])
+
+    @tf.function
+    def __call__(self, x: TensorType) -> TensorType:
+
+        additional_data = Dataset(x, tf.ones_like(x[..., 0:1]))
+
+        try:
+            mean, variance = self._model.conditional_predict_f(  # type: ignore
```
hstojic: since `mean` isn't used anywhere:
```suggestion
            _, variance = self._model.conditional_predict_f(  # type: ignore
```

hstojic: why `# type: ignore`?

vpicheny: not all `ProbablisticModels` have this method. We check this earlier anyway but this line makes PyCharm unhappy.

trieste/acquisition/function/active_learning.py
vpicheny: actually the criterion to maximise is the variance **reduction**, so it should write   
`tf.reduce_mean((old_variance - new_variance) * self._weights, axis=-2)`

but since `old_variance` just adds a constant term and incurs extra cost, I don't use it here. I'll explain this in the docstring.

tests/unit/acquisition/function/test_active_learning.py
```diff
@@ -235,3 +244,151 @@ def test_bichon_ranjan_criterion(threshold: float, at: tf.Tensor, alpha: float,
...
+
+
+@pytest.mark.parametrize("integration_points", [tf.zeros([0, 2]), tf.zeros([1, 2, 3])])
+def test_integrated_variance_reduction_raises_for_invalid_integration_points(
+    integration_points: tf.Tensor,
+) -> None:
+    threshold = [1.0, 2.0]
+    query_at = tf.zeros([1, 1, 1])
+
+    x = to_default_float(tf.constant(np.arange(1, 8).reshape(-1, 1)))
```
hstojic, line 308: `to_default_float` is getting deprecated, can you use tf.cast?

tests/unit/acquisition/function/test_active_learning.py
```diff
@@ -235,3 +244,151 @@ def test_bichon_ranjan_criterion(threshold: float, at: tf.Tensor, alpha: float,
...
+    _, pred_var6 = model6.predict(integration_points)
+
+    acq_noweight = IntegratedVarianceReduction(integration_points=integration_points)
+    acq = IntegratedVarianceReduction(threshold=[0.5, 0.8], integration_points=integration_points)
+
+    acq_function = acq.prepare_acquisition_function(model=model5, dataset=reduced_data)
+    acq_function_noweight = acq_noweight.prepare_acquisition_function(
+        model=model5, dataset=reduced_data
+    )
+    acq_values = -acq_function(tf.expand_dims(query_points, axis=-2))
```
hstojic, line 269: why minus here?

vpicheny: acquisition function returns minus new variance, I flip it so that everythin is a variance

docs/notebooks/feasible_sets.pct.py
```diff
@@ -218,10 +228,10 @@ def objective_function(x):
 # We can also examine what would happen if we would continue for many more active learning steps. One would expect that choices would be allocated closer and closer to the boundary, and uncertainty continuing to collapse. Indeed, on the figure below we observe exactly that. With 90 observations more the model is precisely representing the failure region boundary. It is somewhat difficult to see on the figure, but the most of the additional query points lie exactly on the threshold line.
```
hstojic:
```suggestion
# We can also examine what would happen if we would continue for many more active learning steps. One would expect that choices would be allocated closer and closer to the boundary, and uncertainty continuing to collapse. Indeed, on the figure below we observe exactly that. With 20 observations more the model is precisely representing the failure region boundary. Most of the additional query points lie close to the threshold line.
```

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,162 @@ def acquisition(x: TensorType) -> TensorType:
...
+
+    def __init__(
+        self,
+        model: ProbabilisticModel,
+        integration_points: TensorType,
+        threshold: Optional[Sequence[float]] = None,
+    ):
+        """
+        :param model: The model of the objective function.
+        :param integration_points: points over which to integrate the objective prediction variance
```
hstojic: perhaps we should add a recommendation for this, sample of points that cover the search space well enough?
```suggestion
        :param integration_points: Points over which to integrate the objective prediction variance.
```

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,162 @@ def acquisition(x: TensorType) -> TensorType:
...
+        :param threshold: either None, or a sequence of 1 or 2 float values
```
hstojic:
```suggestion
        :param threshold: Either None, or a sequence of 1 or 2 float values. See class docs for details.
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
 )
 
 
+# %% [markdown]
+# ## Active learning with Integrated Variance Reduction acquisition function
+#
+# An alternative to the `ExpectedFeasibility` function is called `IntegratedVarianceReduction`. This acquisition has the advantage of allowing batches, as we illustrate below. It is more expensive to compute than `ExpectedFeasibility`, as it writes as an integral over a set of input locations.
```
hstojic:
```suggestion
# An alternative to the `ExpectedFeasibility` acquisition function is called `IntegratedVarianceReduction`. This acquisition has the advantage of taking into account reduction of uncertainty in a region of the search space when choosing the next point to sample, instead of considering only the sampling point. This makes it more expensive to compute than `ExpectedFeasibility`, since it involves computing an integral over a set of integration points. This integration region is determined by the user, with the `integration_points` parameter. Another advantage is that `IntegratedVarianceReduction` can produce batches of points, which becomes useful when parallel evaluations are possible.
#
# Below we perform 10 active learning steps with `IntegratedVarianceReduction` acquisition function and same as above plot excursion probability of the final model.
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+final_model_ivr = result_ivr.try_get_final_model()
+dataset_ivr = result_ivr.try_get_final_dataset()
+query_points_ivr = dataset_ivr.query_points.numpy()
+
+plot_excursion_probability(
+    "Final probability of excursion", final_model_ivr, query_points_ivr
+)
+
+# %% [markdown]
+# One can also specify a range of thresholds rather than a single value. The resulting query points are likely to be more speard out than previously.
```
hstojic:
```suggestion
# One can also specify a range of thresholds rather than a single value. We can do this by specifying a range with a minimum and a maximum threshold, rather than a single threshold as the `threshold` parameter. The resulting query points are likely to be more spread out than previously, as now the whole region between the thresholds is aimed to be well estimated, rather than a single line.

# %%
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+acq_range = IntegratedVarianceReduction(
+    integration_points=integration_points, threshold=thresholds
+)
+rule_range = EfficientGlobalOptimization(builder=acq_range, num_query_points=3)  # type: ignore
+
+result_range = bo.optimize(num_steps, initial_data, initial_model, rule_range)
+
+# %% [markdown]
+# We can finally draw the probability that the GP value belongs to an interval rather than the probability that it exceeds a single value. We also compare to the probability obtained when optimising for a single value at the center of the range. As expected, the `IntegratedVarianceReduction` spreads query points a bit more, which leads to a sharper probability.
+
```
hstojic, line 291:
```suggestion

# %%
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+final_model_range = result_range.try_get_final_model()
+dataset_range = result_range.try_get_final_dataset()
+query_points_range = dataset_range.query_points.numpy()
+
+
+plot_excursion_probability(
+    "Probability of being in the range (IVE range)",
```
hstojic:
```suggestion
    "Probability of being in the range (IVR range of thresholds)",
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+    final_model_range,
+    query_points_range,
+    threshold=thresholds,
+)
+
+plot_excursion_probability(
+    "Probability of being in the range (IVE)",
```
hstojic:
```suggestion
    "Probability of being in the range (IVR single threshold)",
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+# Choose integration points uniformly over the design space
+integration_points = search_space.sample_halton(1000)
+acq_ivr = IntegratedVarianceReduction(
+    integration_points=integration_points,
+    threshold=[
+        threshold,
+    ],
+)
+
+# Set a batch size greater than 1
```
hstojic:
```suggestion
# Set a batch size greater than 1 with the 'num_query_points' parameter
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+thresholds = [50.0, 110.0]
+acq_range = IntegratedVarianceReduction(
+    integration_points=integration_points, threshold=thresholds
+)
+rule_range = EfficientGlobalOptimization(builder=acq_range, num_query_points=3)  # type: ignore
+
+result_range = bo.optimize(num_steps, initial_data, initial_model, rule_range)
+
+# %% [markdown]
+# We can finally draw the probability that the GP value belongs to an interval rather than the probability that it exceeds a single value. We also compare to the probability obtained when optimising for a single value at the center of the range. As expected, the `IntegratedVarianceReduction` spreads query points a bit more, which leads to a sharper probability.
```
hstojic:
```suggestion
# We can now illustrate the probability that a point in the search space belongs to the threshold interval rather than the probability that points exceed a single threshold. We compare probability maps obtained with the `IntegratedVarianceReduction` (IVR) when optimising for the threshold range and for the single threshold at the center of the range, as well as to a probability map for the `ExpectedFeasibility` function obtained with a single threshold. As expected, the `IntegratedVarianceReduction` with threshold range spreads query points a bit more, which leads to a sharper probability boundary.
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+# An alternative to the `ExpectedFeasibility` function is called `IntegratedVarianceReduction`. This acquisition has the advantage of allowing batches, as we illustrate below. It is more expensive to compute than `ExpectedFeasibility`, as it writes as an integral over a set of input locations.
+
+# %%
+from trieste.acquisition.function import IntegratedVarianceReduction
+
+# Choose integration points uniformly over the design space
+integration_points = search_space.sample_halton(1000)
+acq_ivr = IntegratedVarianceReduction(
+    integration_points=integration_points,
+    threshold=[
```
hstojic: having to use a list even for a single threshold is suboptimal from a user perspective, do you mind allowing for both a scalar and a list?

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+
+# %% [markdown]
+# One can also specify a range of thresholds rather than a single value. The resulting query points are likely to be more speard out than previously.
+thresholds = [50.0, 110.0]
+acq_range = IntegratedVarianceReduction(
+    integration_points=integration_points, threshold=thresholds
+)
+rule_range = EfficientGlobalOptimization(builder=acq_range, num_query_points=3)  # type: ignore
+
+result_range = bo.optimize(num_steps, initial_data, initial_model, rule_range)
```
hstojic:
```suggestion
model = build_model(initial_data)
result_range = bo.optimize(num_steps, initial_data, model, rule_range)
```

docs/notebooks/feasible_sets.pct.py
```diff
@@ -230,6 +240,77 @@ def objective_function(x):
...
+        threshold,
+    ],
+)
+
+# Set a batch size greater than 1
+rule_ivr = EfficientGlobalOptimization(builder=acq_ivr, num_query_points=3)
+bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+
+num_steps = 10
+result_ivr = bo.optimize(num_steps, initial_data, initial_model, rule_ivr)
```
hstojic:
```suggestion
model = build_model(initial_data)
result_ivr = bo.optimize(num_steps, initial_data, model, rule_ivr)
```

trieste/acquisition/function/active_learning.py
```diff
@@ -245,3 +245,162 @@ def acquisition(x: TensorType) -> TensorType:
...
+        if threshold is not None:
+            threshold = tf.cast(threshold, integration_points.dtype)
+
+            tf.debugging.assert_rank(
+                threshold,
+                1,
+                message=f"threshold should be a sequence "
+                f"or a rank 1 tensor, received {tf.shape(threshold)}",
+            )
+            tf.debugging.assert_less_equal(
```
hstojic: we should probably check that 2nd element > 1st?

vpicheny: I was considering leaving the option to learn outside a band (in that case just swapping the order of T1 and T2 would do the job), but maybe there's no usecase for that and it's error prone...

hstojic (approved):
Looks good now :)

PR #432, opened 2021-11-24
Limit notebooks to 80 characters

### LINKED ISSUE #423: Tutorial code snippets should be no more than 80 characters wide to avoid drag bars
Tutorial code snippets should be no more than 80 characters wide to avoid the ugly drag bars that we currently sometimes have.

We could enforce this by having a different black line-length limit for the docs.

uri-granta:


apaleyes:
Very nice! Just curious: how does this bypass the `[tool.black]` section in toml file, that explicitly says to ignore `pct.py` files?

hstojic (approved):
looks good

uri-granta:
> Just curious: how does this bypass the [tool.black] section in toml file, that explicitly says to ignore pct.py files?

Passing it explicitly as an argument takes precedence over the exclude setting (unless you use force-exclude, which we don't).

apaleyes (approved):

PR #433, opened 2021-11-24
Make AcquisitionFunctionBuilder generic on ProbabilisticModel

ChrisMorter:
This PR makes `AcquisitionFunctionBuilder` and `SingleModelAcquisitionFunctionBuilder` generic on the type of `ProbabilisticModel`.

Motivation for change:

In our project we have many different types of constraints, which are implemented as `AcquisitionFunction`s. Some of these constraints require particular types of model (i.e. particular subclasses of `ProbabilisticModel`). For example, a constraint on the quantile of a modelled quantity requires a model which can predict quantiles, however some of the our quantile models are speciailsed for predicting quantiles, and therefore don't predict other quantities, like the mean, and so can't be used in a constraint which requires a mean prediction. This PR enables mypy to enforce that the correct model types are used when constructing the `AcquistionFunction`. It should effectively amount to a no-op when it's valid to use any `ProbabilisticModel` with a given `AcquistionFunction`, which seems to be the case throughout the Trieste codebase.

### DIFF
--- a/trieste/acquisition/interface.py
+++ b/trieste/acquisition/interface.py
@@ -18,7 +18,7 @@ the utility of evaluating sets of candidate points.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Callable, Mapping, Optional
+from typing import Callable, Generic, Mapping, Optional, TypeVar
 
 from ..data import Dataset
 from ..models import ProbabilisticModel
@@ -49,13 +49,16 @@ class AcquisitionFunctionClass(ABC):
         """Call acquisition function."""
 
 
-class AcquisitionFunctionBuilder(ABC):
+T = TypeVar("T", bound=ProbabilisticModel)
+
+
+class AcquisitionFunctionBuilder(Generic[T], ABC):
     """An :class:`AcquisitionFunctionBuilder` builds and updates an acquisition function."""
 
     @abstractmethod
     def prepare_acquisition_function(
         self,
-        models: Mapping[str, ProbabilisticModel],
+        models: Mapping[str, T],
         datasets: Optional[Mapping[str, Dataset]] = None,
     ) -> AcquisitionFunction:
         """
@@ -70,7 +73,7 @@ class AcquisitionFunctionBuilder(ABC):
     def update_acquisition_function(
         self,
         function: AcquisitionFunction,
-        models: Mapping[str, ProbabilisticModel],
+        models: Mapping[str, T],
         datasets: Optional[Mapping[str, Dataset]] = None,
     ) -> AcquisitionFunction:
         """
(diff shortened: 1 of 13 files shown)


trieste/acquisition/interface.py
```diff
@@ -49,13 +49,16 @@ def __call__(self, x: TensorType) -> TensorType:
         """Call acquisition function."""
 
 
-class AcquisitionFunctionBuilder(ABC):
+T = TypeVar("T", bound=ProbabilisticModel)
+
+
+class AcquisitionFunctionBuilder(Generic[T], ABC):
```
hstojic, line 55: a bit of education please :)

what does adding this `Generic[T]` bit do for the class?

hstojic: "This PR enables mypy to enforce that the correct model types are used when constructing the AcquistionFunction." this would happen only in tests, right?

ChrisMorter: No problem :)

This basically makes it possible for classes inheriting from `AcquisitionFunctionBuilder` to specify that they require specific subclasses of `ProbabilisticModel` (i.e. rather than just the base `ProbabilisticModel` itself), and that that type should be consistent wherever you see the symbol `T`. 

For example, imagine if you have `SpecialProbabilisticModel`, which is a subclass that extends `ProbabilisticModel` with extra functionality required by `SpecialAcquisitionFunctionBuilder`. You want to make sure that nobody passes a standard `ProbabilisticModel` into `SpecialAcquisitionFunctionBuilder`'s `prepare_acquisition_function` (because it doesn't have the required extra functionality), by declaring `SpecialAcquisitionFunctionBuilder` as `SpecialAcquisitionFunctionBuilder[SpecialProbabilisticModel]` mypy can enforce that is true. If `AcquisitionFunctionBuilder` isn't generic, everything _seems_ ok (because everything is a `ProbabilisticModel`), until a runtime error occurs.

hstojic: thanks a lot, that's much clearer and quite useful, we were thinking of how to something along those lines

some follow up questions - how can mypy enforce that? only in tests we actually pass some models to acq functions, right? mypy wouldn't do anything for the end-user, if the user ignores the documentation where the types would be visible it would still be the runtime error

uri-granta (commented):
Looks great, thanks! For completeness though are you ok to also update GreedyAcquisitionFunctionBuilder and SingleModelGreedyAcquisitionBuilder?

trieste/acquisition/interface.py
uri-granta: If the end user uses mypy (which is always a good idea, though academic users probably won't) then it would catch it. Otherwise it would fail with a (possibly confusing) runtime error. I expect we'll continue to check the types at runtime for this reason (so we can give a clear error message) though it would be good to give precise types too — not only does it help people who use mypy, it also serves as better documentation for the acquisition functions.

avullo:
> Looks great, thanks! For completeness though are you ok to also update GreedyAcquisitionFunctionBuilder and SingleModelGreedyAcquisitionBuilder?

I've done the requested update.

hstojic:
to me it seems good, thanks @avullo for jumping in - @uri-granta you are probably the best person to approve this though, I don't feel comfortable having a final say  :)

uri-granta (approved):
Looks great, thank you!

PR #434, opened 2021-11-25
cleaning model test code and testing optimizer defaults

### LINKED ISSUE #425: test default optimizers
ModelConfig used to set a default optimizer when using config dictionaries, with new ModelRegistry we don't have these and rely on model wrappers providing defaults - its not clear whether defaults are properly tested, we need to double check that

hstojic:
5 main tasks:
1. gpflow model test code was unnecessary complex with the gpflow model factory in many tests where we end up testing only GPR, so I simplified the code in these instances and focused the tests on GPR
2. we were not testing that default optimizers are correctly implemented, now implemented additional tests for all models
3. we now explicitly check that scipy optimizer is used with the Optimizer wrapper and TF optimizers with the BatchOptimizer
4. configs were not using wrapper optimizer defaults, corrected now
5. for VGP we now restrict the user to scipy optimizer when `use_natgrads=False`, to help addressing #406


uri-granta (dismissed):
Looks great, thank you!

uri-granta:
> Looks great, thank you!

(sorry this was intended for a different PR! still need to look at this one...)

trieste/models/gpflux/models.py
```diff
@@ -38,49 +38,54 @@ class DeepGaussianProcess(GPfluxPredictor, TrainableProbabilisticModel):
...
-        if self.optimizer.minimize_args is None:
-            self._fit_args: Dict[str, Any] | None = dict(
-                {
-                    "verbose": 0,
-                    "epochs": 100,
-                    "batch_size": 100,
-                }
-            )
+        if not self.optimizer.minimize_args:
+            self._fit_args: Dict[str, Any] | None = {
```
uri-granta: (I always feel the following is more idiomatic, but no need to change)

```suggestion
            self._fit_args: Optional[Dict[str, Any]] = {
```

trieste/models/gpflow/utils.py
```diff
@@ -100,3 +101,33 @@ def squeeze_hyperparameters(
                 low = param.bijector.bijectors[0].shift
                 squeezed_param = tf.math.maximum(param, low + epsilon * tf.ones_like(param))
                 param.assign(squeezed_param)
+
+
+def check_optimizer(optimizer: BatchOptimizer | Optimizer) -> None:
+    """
+    Squeezes the parameters to be strictly inside their range defined by the Sigmoid,
```
uri-granta: (bad docstring)

trieste/models/config.py
```diff
@@ -21,93 +21,135 @@
...
 
+import gpflow
 import tensorflow as tf
 
 from .interfaces import TrainableProbabilisticModel
-from .optimizer import Optimizer
+from .optimizer import BatchOptimizer, Optimizer
 
 
 class ModelRegistry:
```
uri-granta, line 34: Is there a reason to make this one registry rather than two (ModelRegistry, OptimizerRegistry)? Separation of concerns would suggest splitting them.

trieste/models/config.py
```diff
@@ -21,93 +21,135 @@
...
+        Get a Trieste model optimizer wrapper for a given optimizer type.
 
-        :param model_type: The model type.
-        :return: The optimizer to be used with the model type.
+        :param optimizer_type: The model type.
+        :return: The optimizer wrapper to be used with the optimizer type.
         """
         try:
-            return cls._REGISTRY[model_type][1]
+            return cls._OPTIMIZER_TO_WRAPPER[optimizer_type]
```
uri-granta: (this comment applies equally well to the model registry)

Would it be better instead if we check if type is a _subclass_ of any of the registered types? That way we wouldn't have to explicitly register all the subclasses of tf.optimizers.Optimizer, and the registry wouldn't fail for trivial model subclasses (e.g. the ones we currently recommend people write to add new tensorboard logs).

In fact the best might be to loop over the `type.__mro__` types in order, and use the first match we find. That way we can define a more specific wrapper for a subclass if we want to, but don't have to if we don't.

trieste/models/config.py
```diff
@@ -21,93 +21,135 @@
...
+
+
+# We register supported optimizers and their wrappers for usage with ModelConfig.
+# GPflow's Scipy optimizer is not used in the batch mode and hence uses Optimizer wrapper
+ModelRegistry.register_optimizer(type(gpflow.optimizers.Scipy()), Optimizer)
+
+
+# TensorFlow optimizers are stochastic gradient descent variants and should be used
+# with BatchOptimizer wrapper
+for optimizer_type in tf.optimizers.Optimizer.__subclasses__():
```
uri-granta: (see comment above)

trieste/models/config.py
```diff
@@ -122,50 +164,65 @@ class ModelConfig:
...
     def _check_model_type(self) -> None:
         if type(self.model) in ModelRegistry.get_registered_models():
             return
-
         raise NotImplementedError(f"Not supported type {type(self.model)}")
 
+    def _check_optimizer_type(self) -> None:
+        if (
+            self.optimizer is not None
+            and type(self.optimizer) not in ModelRegistry.get_registered_optimizers()
```
uri-granta: (If you make the change I suggested then this check would need to change)

uri-granta (commented):
A few comments. Will look at the tests after you decide whether you agree or not!

trieste/models/config.py
hstojic: yeah, I was thinking about it, but these optimizers are for models so finally opted for a single registry, but if you think its better to separate them, we can do that

trieste/models/config.py
hstojic: yes, that would be an improvement, thanks for the suggestion!
what would be the best way to check if type is a subclass? using `issubclass` function?

not sure I understand the second part, you mean to do this when choosing what to return?

trieste/models/config.py
uri-granta: I'm not too fussed.

trieste/models/config.py
uri-granta: the `__mro__` property lists al the superclasses of a type in order (so starting with the type itself), so you can loop over them and check if any of them are registered.

```python
for optimizer_type in optimizer_type.__mro__:
    if optimizer_type in cls._OPTIMIZER_TO_WRAPPER:
        return cs._OPTIMIZER_TO_WRAPPER[optimizer_type]
else:
   raise ...
````

Comment (hstojic) [reply trieste/models/config.py]: changed now following your proposal!

Comment (hstojic) [reply trieste/models/config.py]: on some further reflection, I'm not sure configs can deal with this "the registry wouldn't fail for trivial model subclasses (e.g. the ones we currently recommend people write to add new tensorboard logs)" - with this modification we are checking GPR, SVGP and their superclasses, we never check for subclasses of our model wrappers, right?

that might be a bit of a gap in our configs (both old ones and this new version), if a same model can be used with multiple wrappers, configs would not know which one to use - so as soon as you want to use a non-default wrapper, you cannot use configs, or you have to overwrite the model-wrapper mapping in the registry

Comment (uri-granta) [reply trieste/models/config.py]: With this change we are checking all the supertypes of the optimizer types against the registered types. So if we pass in a GPR subclass then we would also check it against GPR.

File: trieste/models/config.py [hunk 7]
```diff
@@ -21,93 +21,136 @@
...
-        try:
-            return cls._REGISTRY[model_type][0]
-        except KeyError as e:
-            raise ValueError(
-                f"Model {type(model_type)} currently not supported. "
-                f"We only support the following model types: "
-                f"{list(k.__name__ for k in cls._REGISTRY.keys())}. "
-                f"You can register more via ModelRegistry.register_model."
-            ) from e
+        for _model_type in model_type.__mro__:
```
uri-granta, line 63: Worth a comment maybe?

```python
# Check against all the supertypes in order, from most specific to most general
```

trieste/models/config.py
```diff
@@ -122,50 +165,65 @@ class ModelConfig:
...
 
     def _check_model_type(self) -> None:
         if type(self.model) in ModelRegistry.get_registered_models():
             return
-
         raise NotImplementedError(f"Not supported type {type(self.model)}")
 
+    def _check_optimizer_type(self) -> None:
+        if self.optimizer is not None:
+            for optimizer_type in type(self.optimizer).__mro__:
```
uri-granta: Isn't it enough to call `ModelRegistry.get_optimizer_wrapper(optimizer_type)` here? (If you're concerned about the exception type, you could always catch and rethrow. Would be nice to keep the `__mro__` references to a minimum.)

tests/unit/models/gpflux/test_models.py
```diff
@@ -249,3 +254,39 @@ def scheduler(epoch: int, lr: float) -> float:
     model.optimize(dataset)
 
     npt.assert_allclose(model.model_keras.optimizer.lr.numpy(), init_lr, rtol=1e-6)
+
+
+def test_deep_gaussian_process_default_optimizer_is_correct(
```
uri-granta, line 259: Might be good to run the same test for a dummy subclass of DeepGaussianProcess too?

uri-granta (approved):
Looks good. Just a couple of small comments.

trieste/models/config.py
hstojic: you are right, thanks!

tests/unit/models/gpflux/test_models.py
hstojic: done

trieste/models/config.py
hstojic: right, but thats GPR as a GPflow object, this wouldn't work for subclasses of say GaussianProcessRegression wrapper, and these are the subclasses we care about, we would rarely create subclasses of GPR I think

uri-granta: yes you're right. if people want to build a ModelConfig with a wrapper subclass they would have to register it. though more likely they'll just construct the model wrapper themselves.

PR #435, opened 2021-11-25
Change random seed in constrained BO notebook

apaleyes:
Our recent commits don't pass full doc build with inequality constraints notebook failing with an error. Local testing suggests a different random seed in the notebook resolves the error. This PR updates the random seed.

Since quickdocs build doesn't catch this error, verified locally by reproducing the error and then running the full docs build.

### DIFF
--- a/docs/notebooks/inequality_constraints.pct.py
+++ b/docs/notebooks/inequality_constraints.pct.py
@@ -5,8 +5,8 @@
 import numpy as np
 import tensorflow as tf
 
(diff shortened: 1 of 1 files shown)


hstojic (approved):
sounds good, although its strange that this is the reason for the failure - can you please raise an issue about this to dig into this and quickdocs not catching the errors?

apaleyes:
Done: #436, #437

PR #438, opened 2021-11-25
Parallel acquisition function optimizer

henrymoss:
I have converted @jesnie's  parallel acquisition function optimizer for Trieste.

The resulting optimizer is **significantly** better than our old one with very little overhead. 

All the acquisition function evaluations required for each update step are done in parallel across the many optimization runs using tensorflow and so the greenlet (micro-thread) implementation basically adds no overheads but allows us to run many restarts in parallel.  I found no obvious slow down using 1000 parallel runs compared to our old implementation performing a single run.

- I have set the default number of parallel optimization runs to be  10 * d, but this could potentially set to be much larger.
- I have also added some unit tests  that check optimization performance on more challenging functions (rather than just the quadratic functions we had before). I found that the new optimizer using 1_000 initial points and 10 parallel multi-starts significantly outperform the old optimizer using 1_000_000 initial points and just a single optimization.
-  I have added unit tests to check that we use the correct number of multi-starts and recovery runs. These recovery runs are now also performed in parallel.
- I also fixed the random search optimizer to work with Box search spaces that have their bounds specified as floats.


henrymoss:
@hstojic, I had to only use a single restart for your active learning notebook. If we use multiple restarts and optimizer the acquisition function properly, we seem to try locations of the (batch) search space that break your batch predictive variance acquisition function, i.e. cholesky errors. You might want to look into this? EVen increasing jitter to 1e-2 doesnt fix this!

tests/unit/acquisition/test_optimizer.py
```diff
@@ -223,6 +342,29 @@ def test_continuous_optimizer(
...
+    [
+        (ackley_5, ACKLEY_5_MINIMIZER, ACKLEY_5_MINIMUM, ACKLEY_5_SEARCH_SPACE),
+        (hartmann_3, HARTMANN_3_MINIMIZER, HARTMANN_3_MINIMUM, HARTMANN_3_SEARCH_SPACE),
+        (hartmann_6, HARTMANN_6_MINIMIZER, HARTMANN_6_MINIMUM, HARTMANN_6_SEARCH_SPACE),
+    ],
+)
+def test_continuous_optimizer_on_toy_problems(
+    neg_function: Callable[[TensorType], TensorType],
+    expected_maximizer: TensorType,
+    expected_minimum: TensorType,
```
jesnie: It doesn't look like you use this?

trieste/acquisition/optimizer.py
```diff
@@ -83,7 +90,11 @@ def automatic_optimizer_selector(
 
     elif isinstance(space, (Box, TaggedProductSearchSpace)):
         num_samples = tf.minimum(NUM_SAMPLES_MIN, NUM_SAMPLES_DIM * tf.shape(space.lower)[-1])
```
jesnie: Unrelated to you changes, but it looks like there's something wrong here. It feels like this should either be:
```suggestion
        num_samples = tf.maximum(NUM_SAMPLES_MIN, NUM_SAMPLES_DIM * tf.shape(space.lower)[-1])
```
or
```suggestion
        num_samples = tf.minimum(NUM_SAMPLES_MAX, NUM_SAMPLES_DIM * tf.shape(space.lower)[-1])
```
???

trieste/acquisition/optimizer.py
```diff
@@ -129,27 +140,22 @@ def generate_continuous_optimizer(
 
     We advise the user to either use the default `NUM_SAMPLES_MIN` for `num_initial_samples`, or
     `NUM_SAMPLES_DIM` times the dimensionality of the search space, whichever is smaller.
```
jesnie: Again, either `NUM_SAMPLES_MIN` is poorly named, or you want to use `whichever is greater`?

trieste/acquisition/optimizer.py
```diff
@@ -129,27 +140,22 @@ def generate_continuous_optimizer(
...
+    Similarly, for `num_optimization_runs`, we recommend using `NUM_RUNS_DIM` times the
+    dimensionality of the search space.
 
-    This optimizer supports Scipy's L-BFGS-B optimizer (used via GPflow's Scipy optimizer wrapper),
-    which optimizes directly within and up to the bounds of the search space.
+    This optimizer supports Scipy's L-BFGS-B optimizer. We run `num_optimization_runs` separate
```
jesnie:
```suggestion
    This optimizer uses Scipy's L-BFGS-B optimizer. We run `num_optimization_runs` separate
```
??? It's not really optional, is it?

jesnie (approved):
Mostly LGTM.

I'm a but concerned about `tf.function`? For speed you probably want to compile your code? It doesn't look like you do that anywhere, and it's not clear to me how/where you plan to do it? Maybe update `test_random_search_optimizer_on_toy_problems` to compile it's function, and make sure everything works like you expect?

trieste/acquisition/optimizer.py
henrymoss: Good spot!

hstojic:
> @hstojic, I had to only use a single restart for your active learning notebook. If we use multiple restarts and optimizer the acquisition function properly, we seem to try locations of the (batch) search space that break your batch predictive variance acquisition function, i.e. cholesky errors. You might want to look into this? EVen increasing jitter to 1e-2 doesnt fix this!

@satrialoka implemented that, we should probably ping him to check this out - @satrialoka do you mind investigating this?

satrialoka:
> > @hstojic, I had to only use a single restart for your active learning notebook. If we use multiple restarts and optimizer the acquisition function properly, we seem to try locations of the (batch) search space that break your batch predictive variance acquisition function, i.e. cholesky errors. You might want to look into this? EVen increasing jitter to 1e-2 doesnt fix this!
> 
> @satrialoka implemented that, we should probably ping him to check this out - @satrialoka do you mind investigating this?

@hstojic  Sure!
I'll investigate this.

tests/unit/acquisition/test_optimizer.py
henrymoss: what do you mean? use what?

trieste/acquisition/optimizer.py
henrymoss: good spot

tests/unit/acquisition/test_optimizer.py
jesnie: > what do you mean? use what?

The parameter `expected_minimum` doesn't appear to be used below. I think you can remove it.

trieste/acquisition/optimizer.py
```diff
@@ -82,8 +89,12 @@ def automatic_optimizer_selector(
         return optimize_discrete(space, target_func)
 
     elif isinstance(space, (Box, TaggedProductSearchSpace)):
-        num_samples = tf.minimum(NUM_SAMPLES_MIN, NUM_SAMPLES_DIM * tf.shape(space.lower)[-1])
-        return generate_continuous_optimizer(num_samples)(space, target_func)
+        num_samples = tf.maximum(NUM_SAMPLES_MIN, NUM_SAMPLES_DIM * tf.shape(space.lower)[-1])
```
hstojic, line 92: yep, thats more sensible :)

trieste/acquisition/optimizer.py
```diff
@@ -82,8 +89,12 @@ def automatic_optimizer_selector(
...
+        num_runs = NUM_RUNS_DIM * tf.shape(space.lower)[-1]
```
hstojic, line 93: do we want to do something similar here `max(NUM_RUNS_MIN, NUM_RUNS_DIM * tf.shape(space.lower)[-1])`

then below in `generate_continuous_optimizer` we can also use `NUM_RUNS_MIN` as a default

trieste/acquisition/optimizer.py
```diff
@@ -117,8 +128,8 @@ def optimize_discrete(space: DiscreteSearchSpace, target_func: AcquisitionFuncti
 
 def generate_continuous_optimizer(
     num_initial_samples: int = NUM_SAMPLES_MIN,
-    num_optimization_runs: int = 1,
-    num_recovery_runs: int = 5,
+    num_optimization_runs: int = 10,
```
hstojic, line 131: 10 seems a bit much, I would expect 3-5 to be enough, did you test it in any way? - just trying to be careful before introducing a lot more computation by default

trieste/acquisition/optimizer.py
henrymoss: I think a min of 10 (i.e. 1 * 10) is fine for number of runs. Its just for the initial design where 1* 1000 wasn't enough and we needed to ensure > 5000.

trieste/acquisition/optimizer.py
```diff
@@ -197,68 +199,172 @@ def optimize_continuous(
         )  # [num_optimization_runs]
         initial_points = tf.gather(trial_search_space, top_k_indicies)  # [num_optimization_runs, D]
```
```suggestion
        initial_points = tf.gather(trial_search_space, top_k_indices)  # [num_optimization_runs, D]
```

trieste/acquisition/optimizer.py
henrymoss: Its hard to know how many are enough. I tested this change to see if it slowed anything down, and it didnt seem to! Happy to dial it back to 5 if you want?

trieste/acquisition/optimizer.py
```diff
@@ -197,68 +199,172 @@ def optimize_continuous(
...
+def _perform_parallel_continuous_optimization(
+    target_func: AcquisitionFunction,
+    space: SearchSpace,
+    starting_points: TensorType,
+    optimizer_args: dict[str, Any],
+) -> tuple[TensorType, bool]:
+    """
+    A function to perform parallel optimization of our acquisition functions
+    using Scipy. We perform L-BFGS-B starting from each of the locations contained
+    in `starting_points`.
```
hstojic: lets add that number of optimization runs is inferred from the shape of starting points

trieste/acquisition/optimizer.py
```diff
@@ -197,68 +199,172 @@ def optimize_continuous(
...
-                    successful_optimization = True
-                    break
-            if not successful_optimization:  # return error if still failed
-                raise FailedOptimizationError(
-                    f"""
+        chosen_point, successful_optimization = _perform_parallel_continuous_optimization(
+            target_func,
+            space,
+            initial_points,
+            optimizer_args,  # run num_optimization_runs in parallel
```
hstojic: why the comment here at `optimizer_args`?

trieste/acquisition/optimizer.py
```diff
@@ -197,68 +199,172 @@ def optimize_continuous(
...
+    (for each optimization step) using Tensorflow.
+
+    For :class:'TaggedProductSearchSpace' we only apply gradient updates to
+    its class:'Box' subspaces, fixing the discrete elements to the best values
+    found across the initial random search. To fix these discrete elements, we
+    optimize over a continuous class:'Box' relaxation of the discrete subspaces
+    which has equal upper and lower bounds, i.e. we specify an equality constraint
+    for this dimension in the scipy optimizer.
+
+    :param target_func:
```
hstojic: missing description?

trieste/acquisition/optimizer.py
```diff
@@ -197,68 +199,172 @@ def optimize_continuous(
...
+    :param space: The original search space.
+    :param starting_points: The points at which to begin our optimizations.
```
hstojic: remind here as well that this is used for inferring the number of optim runs

trieste/acquisition/optimizer.py
```diff
@@ -197,68 +199,172 @@ def optimize_continuous(
...
+                continue
+            child_results[i] = greenlet.switch(np_batch_y[i], np_batch_dy_dx[i, :])
+
+    best_run_id = np.argmax([-result.fun for result in child_results])  # identify best solution
+    np_chosen_point = child_results[best_run_id].x.reshape(1, -1)  # [1, D]
+    success = np.any(
+        [result.success for result in child_results]
+    )  # Check that at least one optimization was successful
+
+    return tf.constant(np_chosen_point, dtype=tf_dtype), success  # convert back to TF
```
hstojic: I'm thinking that some other metadata might be useful as well, e.g. which k was it that yielded the best solution, what were the values and points in all k runs, did we enter the recovery runs, optimization time etc - perhaps logging this to the tensorboard? 

I think it would be useful for monitoring, e.g. to spot whether you might need to decrease/increase the number of runs, or if you are consistently entering recovery runs that something might be off

trieste/acquisition/optimizer.py
```diff
@@ -197,68 +199,172 @@ def optimize_continuous(
...
+            nonlocal cache_dy_dx
+
+            if not (cache_x == x).all():
+                cache_x[:] = x  # Copy the value of `x`. DO NOT copy the reference.
+                # Send `x` to parent greenlet, which will evaluate all `x`s in a batch.
+                cache_y, cache_dy_dx = self.parent.switch(cache_x)
+
+            return cache_y, cache_dy_dx
+
+        return spo.minimize(
```
hstojic, line 363: how come we are switching to Scipy directly from GPflow's wrapper?
(or why were we using GPflow's wrapper in the first place)

trieste/acquisition/optimizer.py
hstojic: thats strange that it didn't slow things down - that might suggest additional run is not the bottleneck but the evaluations, perhaps its fine as it is

actually, more relevant default is the automatic selector one, not this one - in a 10D user would have 100 runs, perhaps necessary but that definitely seems a lot

tests/unit/acquisition/test_optimizer.py
```diff
@@ -116,6 +155,90 @@ def test_generate_continuous_optimizer_raises_with_invalid_init_params() -> None
...
+        == f"""
+                    Acquisition function optimization failed,
+                    even after {num_recovery_runs + num_optimization_runs} restarts.
+                    """
+    )
+
+
+@pytest.mark.parametrize("num_optimization_runs", [1, 10])
+@pytest.mark.parametrize("num_initial_samples", [1000, 5000])
+def test_optimize_continuous_uses_correctly_uses_init_params(
```
hstojic:
```suggestion
def test_optimize_continuous_correctly_uses_init_params(
```

tests/unit/acquisition/test_optimizer.py
```diff
@@ -116,6 +155,90 @@ def test_generate_continuous_optimizer_raises_with_invalid_init_params() -> None
...
+) -> None:
+
+    querying_initial_sample = True
+
+    def _target_fn(x: TensorType) -> TensorType:
+        nonlocal querying_initial_sample
+
+        if querying_initial_sample:  # check size of initial sample
+            assert tf.shape(x)[0] == num_initial_samples
+        else:  # check evaluations are in parallel with correct batch size
```
hstojic, line 187: is this condition ever entered?

tests/unit/acquisition/test_optimizer.py
```diff
@@ -116,6 +155,90 @@ def test_generate_continuous_optimizer_raises_with_invalid_init_params() -> None
...
+            assert tf.shape(x)[0] == num_optimization_runs
+
+        querying_initial_sample = False
+        return _quadratic_sum([0.5, 0.5])(x)
+
+    optimizer = generate_continuous_optimizer(num_initial_samples, num_optimization_runs)
+    optimizer(Box([-1], [1]), _target_fn)
+
+
+@pytest.mark.parametrize("failed_first_optimziation", [True, False])
```
hstojic:
```suggestion
@pytest.mark.parametrize("failed_first_optimiziation", [True, False])
```

trieste/acquisition/optimizer.py
henrymoss: Yep I agree. Will save for another PR! https://github.com/secondmind-labs/trieste/issues/440

trieste/acquisition/optimizer.py
henrymoss: we have essentially made our own wrapper that can do batching.

tests/unit/acquisition/test_optimizer.py
```diff
@@ -116,6 +155,90 @@ def test_generate_continuous_optimizer_raises_with_invalid_init_params() -> None
...
+            return _quadratic_sum([0.5, 0.5])  # return function that is easy to optimize
+
+        optimizer = generate_continuous_optimizer(
+            num_optimization_runs=1, num_recovery_runs=num_recovery_runs
+        )
+        if failed_first_optimziation and (num_recovery_runs == 0):
+            with pytest.raises(FailedOptimizationError):
+                optimizer(Box([-1], [1]), _target_fn)
+        else:
+            optimizer(Box([-1], [1]), _target_fn)
```
hstojic, line 235: so is there a difference between num_recovery_runs being 2 or 10, we don't check that correct number was really executed?

tests/unit/acquisition/test_optimizer.py
henrymoss: yep, we set`querying_initial_sample = False` in the line below

tests/unit/acquisition/test_optimizer.py
henrymoss: Higher up in that test. We assert that we use the correct number of recovery runs

hstojic (commented):
looks quite good to me

two main points
1. perhaps it has to be done, but scaling the number of runs with dimensionality dependence seems a bit much
2. it would be useful to send some optimization metadata to tensorboard

tests/unit/acquisition/test_optimizer.py
hstojic: ah indeed, missed that

hstojic (approved):
looks good now :)

setup.py
```diff
@@ -43,5 +43,6 @@
         "numpy",
         "tensorflow>=2.4",
         "tensorflow-probability>=0.12,<0.14.0",
+        "greenlet>1.1.0",
```
apaleyes: slightly surprising, this is usually using `>=` because you are relying on a certain version of the package, and up from there. why is this strictly greater then?

trieste/acquisition/optimizer.py
```diff
@@ -192,73 +194,182 @@ def optimize_continuous(
...
+
+    L-BFGS-B updates for each individual optimization are performed by
+    independent greenlets working with Numpy arrays, however, the evaluation
+    of our acquisition function (and its gradients) is calculated in parallel
+    (for each optimization step) using Tensorflow.
+
+    For :class:'TaggedProductSearchSpace' we only apply gradient updates to
+    its class:'Box' subspaces, fixing the discrete elements to the best values
+    found across the initial random search. To fix these discrete elements, we
+    optimize over a continuous class:'Box' relaxation of the discrete subspaces
```
```suggestion
    its :class:'Box' subspaces, fixing the discrete elements to the best values
    found across the initial random search. To fix these discrete elements, we
    optimize over a continuous :class:'Box' relaxation of the discrete subspaces
```

trieste/acquisition/optimizer.py
```diff
@@ -192,73 +194,182 @@ def optimize_continuous(
...
+
+        # Batch evaluate query `x`s from all children.
+        batch_x = tf.constant(np_batch_x, dtype=tf_dtype)  # [num_optimization_runs, d]
+        batch_y, batch_dy_dx = _objective_value_and_gradient(batch_x)
+        np_batch_y = batch_y.numpy().astype("float64")
+        np_batch_dy_dx = batch_dy_dx.numpy().astype("float64")
+
+        for i, greenlet in enumerate(child_greenlets):  # Feed `y` and `dy_dx` back to children.
+            if greenlet.dead:  # Allow for crashed greenlets
+                continue
```
apaleyes, line 328: Can they all die? What will we return in that case?

trieste/acquisition/optimizer.py
```diff
@@ -192,73 +194,182 @@ def optimize_continuous(
...
+    )
+
+    # Set up child greenlets
+    child_greenlets = [ScipyLbfgsBGreenlet() for _ in range(num_optimization_runs)]
+    child_results: List[Union[spo.OptimizeResult, np.ndarray]] = [
+        gr.switch(starting_points[i].numpy(), bounds[i], optimizer_args)
+        for i, gr in enumerate(child_greenlets)
+    ]
+
+    while True:
```
apaleyes, line 308: Is it guaranteed that this loop will always stop? It isn't immediately obvious, or I am just slow

trieste/acquisition/optimizer.py
```diff
@@ -128,28 +139,23 @@ def generate_continuous_optimizer(
...
     We advise the user to either use the default `NUM_SAMPLES_MIN` for `num_initial_samples`, or
-    `NUM_SAMPLES_DIM` times the dimensionality of the search space, whichever is smaller.
+    `NUM_SAMPLES_DIM` times the dimensionality of the search space, whichever is greater.
+    Similarly, for `num_optimization_runs`, we recommend using `NUM_RUNS_DIM` times the
+    dimensionality of the search space.
 
-    This optimizer supports Scipy's L-BFGS-B optimizer (used via GPflow's Scipy optimizer wrapper),
-    which optimizes directly within and up to the bounds of the search space.
+    This optimizer uses Scipy's L-BFGS-B optimizer. We run `num_optimization_runs` separate
+    optimizations in parallel, each starting from one of the top `num_optimization_runs` initial
```
apaleyes: What does "top" refer to here?

trieste/acquisition/optimizer.py
```diff
@@ -192,73 +194,182 @@ def optimize_continuous(
...
+
+        for i, greenlet in enumerate(child_greenlets):  # Feed `y` and `dy_dx` back to children.
+            if greenlet.dead:  # Allow for crashed greenlets
+                continue
+            child_results[i] = greenlet.switch(np_batch_y[i], np_batch_dy_dx[i, :])
+
+    return child_results
+
+
+class ScipyLbfgsBGreenlet(gr.greenlet):
```
apaleyes, line 334: Can greenlets fall with exceptions? Do they catch them all and show "deal == True"? If not, shall we be catching any outside this greenlet?

setup.py
henrymoss: okay have changed to >=

trieste/acquisition/optimizer.py
henrymoss: yep. eventually all the greenlets will do their maximum allowed number of steps

trieste/acquisition/optimizer.py
henrymoss: best. Have changed

icouckuy:
> @hstojic, I had to only use a single restart for your active learning notebook. If we use multiple restarts and optimizer the acquisition function properly, we seem to try locations of the (batch) search space that break your batch predictive variance acquisition function, i.e. cholesky errors. You might want to look into this? EVen increasing jitter to 1e-2 doesnt fix this!

Are you selecting duplicate points within one batch? Do you have an example batch where it fails? Which jitter value did you change?

henrymoss:
@icouckuy you can run the notebook but remove the `num_restarts=1` argument to the optimizer to get a concrete example

PR #441, opened 2021-11-29
integration tests for active learning

### LINKED ISSUE #439: Active learning acquisition functions are missing integration tests

hstojic:
active learning algos were missing integration tests, this would close #439

tests are a bit slow, which is to be expected I guess, since we want to be reasonably sure that we are really learning the function or the feasibility set we do need to run more iterations than usual

something to discuss: should we cover batch versions, async rules, trust region, tensorboard? (mirroring the optimization tests...)

### DIFF
--- a/tests/integration/test_active_learning.py
+++ b/tests/integration/test_active_learning.py
@@ -19,7 +19,6 @@ Integration tests for various forms of active learning implemented in Trieste.
 from __future__ import annotations
 
 import gpflow
-import numpy.testing as npt
 import pytest
 import tensorflow as tf
 import tensorflow_probability as tfp
@@ -78,7 +77,7 @@ def test_optimizer_learns_scaled_branin_function(
     # max absolute error needs to be bettter than this criterion
     test_query_points = search_space.sample_sobol(10000 * search_space.dimension)
     test_data = observer(test_query_points)
-    test_range = (tf.reduce_max(test_data.observations) - tf.reduce_min(test_data.observations))
+    test_range = tf.reduce_max(test_data.observations) - tf.reduce_min(test_data.observations)
     criterion = 0.01 * test_range
 
     # we expect a model with initial data to fail the criterion
@@ -110,7 +109,7 @@ def test_optimizer_learns_scaled_branin_function(
     [
         (50, EfficientGlobalOptimization(ExpectedFeasibility(80, delta=1)), 80),
         (50, EfficientGlobalOptimization(ExpectedFeasibility(80, delta=2)), 80),
-        (50, EfficientGlobalOptimization(ExpectedFeasibility(20, delta=1)), 20),
+        (100, EfficientGlobalOptimization(ExpectedFeasibility(20, delta=1)), 20),
     ],
 )
 def test_optimizer_learns_feasibility_set_of_thresholded_branin_function(
@@ -164,10 +163,10 @@ def test_optimizer_learns_feasibility_set_of_thresholded_branin_function(
         .optimize(num_steps, initial_data, model, acquisition_rule)
         .try_get_final_model()
     )
-    final_prob = _excursion_probability(threshold_test_points, final_model, threshold)
+    final_prob = _excursion_probability(test_query_points, final_model, threshold)
     final_accuracy = tf.reduce_sum(final_prob * (1 - final_prob))
 
-    assert initial_accuracy < final_accuracy
+    assert initial_accuracy > final_accuracy
     assert final_accuracy < criterion
 
 
--- a/tests/integration/test_active_learning.py
+++ b/tests/integration/test_active_learning.py
@@ -1,0 +1,150 @@
+# Copyright 2021 The Trieste Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Integration tests for various forms of active learning implemented in Trieste.
+"""
+
+from __future__ import annotations
+
+import gpflow
+import numpy.testing as npt
+import pytest
+import tensorflow as tf
+import tensorflow_probability as tfp
+
+from tests.util.misc import random_seed
+from trieste.acquisition.function import ExpectedFeasibility, PredictiveVariance
+from trieste.acquisition.rule import AcquisitionRule, EfficientGlobalOptimization
+from trieste.bayesian_optimizer import BayesianOptimizer
+from trieste.data import Dataset
+from trieste.models import TrainableProbabilisticModel
+from trieste.models.gpflow import GaussianProcessRegression
+from trieste.objectives import BRANIN_SEARCH_SPACE, branin, scaled_branin
+from trieste.objectives.utils import mk_observer
+from trieste.space import SearchSpace
+from trieste.types import TensorType
+
+
+@random_seed
+@pytest.mark.parametrize(
+    "num_steps, acquisition_rule",
+    [
+        (80, EfficientGlobalOptimization(PredictiveVariance())),
+    ],
+)
+def test_optimizer_learns_scaled_branin_function(
+    num_steps: int, acquisition_rule: AcquisitionRule[TensorType, SearchSpace]
+) -> None:
+    """
+    Ensure that the objective function is effecitvely learned, such that the final model
+    fits well and predictions are close to actual objective values.
+    """
+
+    search_space = BRANIN_SEARCH_SPACE
+
+    def build_model(data: Dataset) -> TrainableProbabilisticModel:
+        variance = tf.math.reduce_variance(data.observations)
+        kernel = gpflow.kernels.Matern52(variance=variance, lengthscales=[0.2, 0.2])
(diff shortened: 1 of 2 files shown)


tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,150 @@
...
+        BayesianOptimizer(observer, search_space)
+        .optimize(num_steps, initial_data, model, acquisition_rule)
+        .try_get_final_model()
+    )
+
+    test_query_points = search_space.sample_sobol(10000 * search_space.dimension)
+    test_data = observer(test_query_points)
+    predicted_means, _ = final_model.model.predict_f(test_query_points)  # type: ignore
+
+    npt.assert_allclose(predicted_means, test_data.observations, rtol=0.5)
```
vpicheny: just for checking: allclose would still use the default atol, which is quite restrictive, right? otherwise, this is very loose?

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,150 @@
...
+        (150, EfficientGlobalOptimization(ExpectedFeasibility(80, delta=1)), 80),
+        (150, EfficientGlobalOptimization(ExpectedFeasibility(80, delta=2)), 80),
+        (150, EfficientGlobalOptimization(ExpectedFeasibility(20, delta=1)), 20),
+    ],
+)
+def test_optimizer_learns_feasibility_set_of_thresholded_branin_function(
+    num_steps: int, acquisition_rule: AcquisitionRule[TensorType, SearchSpace], threshold: int
+) -> None:
+    """
+    Ensure that the objective function is effecitvely learned, such that the final model
```
vpicheny: typo

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+        kernel.lengthscales.prior = tfp.distributions.LogNormal(
+            tf.math.log(kernel.lengthscales), prior_scale
+        )
+        gpr = gpflow.models.GPR(data.astuple(), kernel, noise_variance=1e-5)
+        gpflow.set_trainable(gpr.likelihood, False)
+
+        return GaussianProcessRegression(gpr)
+
+    num_initial_points = 20
```
vpicheny: can we start with fewer? 20 points is already a lot, so it's possible that the `PredictiveVariance` doesn't do much.

Or we could check if the error decreases? this might be a bit tricky though...

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,150 @@
...
```
vpicheny: same as before: 20 seems really large. Try 6?

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,150 @@
...
+    model = build_model(initial_data)
+    final_model = (
+        BayesianOptimizer(observer, search_space)
+        .optimize(num_steps, initial_data, model, acquisition_rule)
+        .try_get_final_model()
+    )
+
+    test_points = search_space.sample_sobol(10000 * search_space.dimension)
+    prob = _excursion_probability(test_points, final_model, threshold)
+    accuracy = tf.reduce_sum(prob * (1 - prob))
```
vpicheny: use mean instead of sum? so it doesn't depend on the number of test points

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,150 @@
...
+
+    assert accuracy < 0.0001
```
vpicheny: this looks like an incredibly hard target to reach

tests/integration/test_active_learning.py
vpicheny: Also, this is not really consistent with the other test: the other tests a single average measure of accuracy, while this is a max. Max is fine, but maybe rewrite `crit = np.abs(predicted_means, test_data.observations)` and assert crit < something?

hstojic: yes, accuracy criteria is something to discuss, its difficult to say what to expect in terms of global accuracy with 100 observations, and we dont want to have tests that take ages - I'm open to suggestions here

the other test - you mean feasibility set? thats a different type of problem, but I guess we could take a mean here and be happy if its smth small?

tests/integration/test_active_learning.py
hstojic: sure, we can try starting with fewer points

I thought about doing the decrease in error test here but concluded that would be more suitable as a unit test

tests/integration/test_active_learning.py
hstojic: ok, will do

tests/integration/test_active_learning.py
hstojic: yeah, it seemed like it works but doesn't at the end, 1 works

tests/integration/test_active_learning.py
vpicheny: For this test, we probably want a criterion that would fail with `num_initial_points` and pass with more points, provided that they are not pathologically distributed (eg they should be roughly space-filling, not concentrated in one corner of the space). We can't do much more than that (eg we could see if we beat random search over replicated experiments, but we won't do that). But this is already a good test.

vpicheny: Max error is probably better than average error because it will reward more space-filling designs.

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,180 @@
...
+    test_range = (tf.reduce_max(test_data.observations) - tf.reduce_min(test_data.observations))
+    criterion = 0.01 * test_range
+
+    # we expect a model with initial data to fail the criterion
+    initial_model = build_model(initial_data)
+    initial_model.optimize(initial_data)
+    initial_predicted_means, _ = initial_model.model.predict_f(test_query_points)  # type: ignore
+    initial_accuracy = tf.reduce_max(tf.abs(initial_predicted_means - test_data.observations))
+
+    assert not initial_accuracy < criterion
```
hstojic, line 90: we test now that initial model fails and final model improves on this and satisfies the criterion

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,180 @@
...
+    initial_query_points = search_space.sample_halton(num_initial_points)
+    observer = mk_observer(scaled_branin)
+    initial_data = observer(initial_query_points)
+
+    # we set a performance criterion at 1% of the range
+    # max absolute error needs to be bettter than this criterion
+    test_query_points = search_space.sample_sobol(10000 * search_space.dimension)
+    test_data = observer(test_query_points)
+    test_range = (tf.reduce_max(test_data.observations) - tf.reduce_min(test_data.observations))
+    criterion = 0.01 * test_range
```
hstojic, line 82: max abs error shouldn't cross 1% of the range, that seems like a good global precision to me?

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,180 @@
...
+
+    num_initial_points = 6
+    initial_query_points = search_space.sample_halton(num_initial_points)
+    observer = mk_observer(branin)
+    initial_data = observer(initial_query_points)
+
+    # we set a performance criterion at 0.001 probability of required precision per point
+    n_test = 10000 * search_space.dimension
+    test_query_points = search_space.sample_sobol(n_test)
+    criterion = 0.001 * (1 - 0.001) * tf.cast(n_test, tf.float64)
```
hstojic: new criterion, 0.001/0.999 predicted probability per tested point seems like a reasonable accuracy to me, any thoughts?

tests/integration/test_active_learning.py
```diff
@@ -0,0 +1,180 @@
...
+        return GaussianProcessRegression(gpr)
+
+    num_initial_points = 6
+    initial_query_points = search_space.sample_halton(num_initial_points)
+    observer = mk_observer(branin)
+    initial_data = observer(initial_query_points)
+
+    # we set a performance criterion at 0.001 probability of required precision per point
+    n_test = 10000 * search_space.dimension
+    test_query_points = search_space.sample_sobol(n_test)
```
hstojic: I think we are more interested in model accuracy around the threshold (or perhaps the whole feasible region?), instead of getting test points from the whole search space, we should then collect test points around the boundary - we could do something like:
```
test_query_points = search_space.sample(100000)
test_data = observer(test_query_points)

mask = tf.reduce_all(tf.concat([test_data.observations > 80 - 5, test_data.observations < 80 + 5], axis = 1), axis=1)
threshold_test_points = tf.boolean_mask(test_query_points, mask)
assert threshold_test_points.shape[0] > 1000
```

vpicheny: I'm not sure I 100% agree, we also want our GP to be good enough to predict anywhere which side it's on with high probability. I think the random sample is fine.

We could also have 2 criteria, one 'global' as it is and one 'local' as you describe and we could check both. Not sure it's worth the extra effort

tests/integration/test_active_learning.py
vpicheny: I like this, it's principled now

vpicheny (approved):
great changes, looks good now!

tests/integration/test_active_learning.py
hstojic: hm, its sensitive to the volume of the search space occupied by the boundary, for the threshold of 20 boundary is much larger and then there are more points with smaller accuracy, so the 3rd test actually fails here :|

hstojic: perhaps a quick monte carlo estimate of the volume of the feasible space and make the criterion dependent on that as well?

hstojic: although its reasonable to expect to need to have more observations to achieve the same accuracy for a larger volume, so we could also simply increase number of steps for threshold 20

tests/integration/test_active_learning.py
hstojic: I went with the 2 criteria version...

PR #442, opened 2021-12-01
Fix slow running test failures

uri-granta:
Looks like there's some issue with the deecopy. I might investigate that separately to make sure that doesn't impact state tracking.


hstojic (approved):
looks ok to me

PR #443, opened 2021-12-03
Fast GPR/SGPR/SVGP predictions using Gpflow's posterior object

### LINKED ISSUE #351: Support for GPFlow fast predict
We should update our GPR models to use GPflow's recent(ish) update to make fast sequential predictions (i.e. cached Cholesky)

henrymoss:
I have changed our GPR/SGPR/SVGP model wrappers to use the posterior objects provided in the newest release of gpflow (we now require 2.3). These objects store computations that can be reused for sequential model predictions, thus making sequential model predictions (as performed when optimizing acq functions) substantially faster (sometimes by an order of magnitude).

Unfortunately, Gpflow does not currently provide a posterior object for VGP (although I see no good reason why). Therefore, I had to implement posterior support separately for our GPR and SVGP wrappers, rather than implementing it higher up in our class hierarchy (e.g at GPflowPredictor).

The only subtlety, it to ensure that we recalculate the posterior object every time we change the models. I believe this is just after an `update` or an `optimize`, but I might have missed something? 

I also updated @vpicheny's fantasy model work to make use of our new faster predictions. 

- [ ] Still need to update the`covariance_between_points` method to make use of this new functionality


henrymoss:
In one of the unit tests, I saw that making 10 sequential predictions from a GPR fit on only 10 datapoints is now 10x faster than before. This speed difference will only increase with larger datasets!

trieste/models/gpflow/models.py
```diff
@@ -52,6 +59,7 @@ def __init__(
         """
         super().__init__(optimizer)
         self._model = model
+        self._posterior = self._model.posterior()  # cache posterior for fast sequential predictions
```
alansaul: Note that there is a issue within GPflow that says that the posterior is broken for GPR (I don't know its merit but there is a PR to fix it so I guess it is indeed broken!)

https://github.com/GPflow/GPflow/issues/1733
https://github.com/GPflow/GPflow/pull/1734

I thought alpha was *supposed to be* equal to y, but I'm not an expert.

trieste/models/gpflow/models.py
```diff
@@ -206,6 +228,7 @@ def optimize(self, dataset: Dataset) -> None:
             )
 
         self.optimizer.optimize(self.model, dataset)
+        self._posterior = self.model.posterior()  # update cached posterior
```
alansaul: I *believe* it isn't necessary to recreate this object, and you should be able to do `self._posterior.update_cache()` or the same function with some other enum. Would need to read the docs properly though.

trieste/models/gpflow/models.py
```diff
@@ -501,6 +544,16 @@ def update(self, dataset: Dataset) -> None:
 
         num_data = dataset.query_points.shape[0]
         self.model.num_data = num_data
+        self._posterior = self.model.posterior()  # update cached posterior
```
alansaul: Same as above

trieste/models/gpflow/models.py
```diff
@@ -501,6 +544,16 @@ def update(self, dataset: Dataset) -> None:
...
+
+    def optimize(self, dataset: Dataset) -> None:
+        """
+        Optimize the model with the specified `dataset`.
+
+        :param dataset: The data with which to optimize the `model`.
+        """
+        self.optimizer.optimize(self.model, dataset)
+        self._posterior = self.model.posterior()  # update cached posterior
```
alansaul: Same as above

alansaul (approved):
LGTM as far as I can tell!

henrymoss:
This PR cannot work until https://github.com/GPflow/GPflow/issues/1733 is fixed in Gpflow

johnamcleod:
> This PR cannot work until [GPflow/GPflow#1733](https://github.com/GPflow/GPflow/issues/1733) is fixed in Gpflow

The fix for this issue has been merged into develop, and will be available in the next GPFlow release.

henrymoss:
Hi @johnamcleod. Hope you are doing well!

Thanks for this

tests/unit/models/gpflow/test_models.py
```diff
@@ -502,6 +503,82 @@ def test_gaussian_process_regression_optimize(
...
+
+    if isinstance(model, VariationalGaussianProcess):
+        pytest.skip("Cached predictions are not yet implemented for the VGP models.")
+
+    x_predict = np.linspace(0, 5, 2).reshape((-1, 1))
+    t_0 = time()
+    [model.predict(x_predict) for _ in range(10)]  # make sequential predictions
+    time_with_cache = time() - t_0
+    t_0 = time()
+    [model.model.predict_f(x_predict) for _ in range(10)]
```
apaleyes: That doesn't look right. Would be much nicer if Trieste user could have a choice of using or not using posterior object. Of course unless it is 100% certain users will never want the latter.

This seems like a job for a [decorator pattern](https://refactoring.guru/design-patterns/decorator/python/example).

henrymoss: I think we would always want to use them if they exist! Free  speeeed

docs/notebooks/constraints.txt
```diff
@@ -38,7 +38,7 @@ google-auth==1.34.0
 google-auth-oauthlib==0.4.5
 google-pasta==0.2.0
 googleapis-common-protos==1.53.0
-gpflow==2.2.1
+gpflow==2.3.0
```
hstojic: we need to use the develop version if we want this feature immediately, right?

trieste/models/gpflow/models.py
hstojic: that PR was merged into develop branch today :)

tests/unit/models/gpflow/test_models.py
hstojic: if there is really no downside to this, I also dont see a reason why not to make it a default

tests/unit/models/gpflow/test_models.py
```diff
@@ -568,6 +570,83 @@ def test_gaussian_process_regression_optimize(compile: bool) -> None:
     assert model.model.training_loss() < loss
 
 
+@random_seed
```
vpicheny, line 573: should we also test that the posterior objects update properly? or maybe since this is the new default it is already tested elsewhere?

trieste/models/gpflow/models.py
```diff
@@ -55,6 +55,13 @@ class GaussianProcessRegression(
     """
     A :class:`TrainableProbabilisticModel` wrapper for a GPflow :class:`~gpflow.models.GPR`
     or :class:`~gpflow.models.SGPR`.
+
+    As Bayesian optimization requires a large number of sequential predictions (i.e. when maximizing
+    acquisition functions), rather than calling the model directly at prediction time we instead
+    call the posterior objects built by these models. These posterior objects stores the
```
vpicheny: store

trieste/models/gpflow/models.py
```diff
@@ -557,6 +577,12 @@ def __init__(
 class SparseVariational(GPflowPredictor, TrainableProbabilisticModel):
     """
     A :class:`TrainableProbabilisticModel` wrapper for a GPflow :class:`~gpflow.models.SVGP`.
+
+    Similarly to our :class:`GaussianProcessRegression`, our :class:`~gpflow.models.SVGP` wrapper
+    directly calls the posterior objects built by these models at prediction
+    time. These posterior stores the pre-computed Gram matrices, which can be reused to allow faster
```
vpicheny: store

hstojic:
this can now be closed, right?

PR #448, opened 2021-12-07
random seed setting for testing

hstojic:
made some small changes to random_seed decorator to make it more robust and catch other important seeds that so far were not fixed - perhaps some stochasticity observed in tests previously was due to that


hstojic:
done by @uri-granta  in #450

PR #449, opened 2021-12-07
support for keras models part 1 - interface

hstojic:
support for neural nets with Keras comes in four parts:
1. interface (this PR)
2. some utilities
4. integration tests and notebook


trieste/models/keras/interface.py
```diff
@@ -0,0 +1,81 @@
...
+from typing import Optional
+
+import tensorflow as tf
+
+from ...types import TensorType
+from ..interfaces import ProbabilisticModel
+from ..optimizer import BatchOptimizer
+
+
+class KerasPredictor(ProbabilisticModel, tf.Module, ABC):
```
vpicheny: high level question: why not inherit from `TrainableProbabilisticModel`?

hstojic: not sure honestly, with GPflux and here we followed the setup with GPflow - I think the idea is that these high level model interfaces are not meant to provide optimize and update methods, model wrappers are supposed to do that, so they inherit from both KerasPredictor and TrainableProbabilisticModel

perhaps @uri-granta can shed more light on the design here?

trieste/models/keras/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+r"""
+This package contains the primary interface for deep neural network models. It also contains a
+number of :class:`TrainableProbabilisticModel` wrappers for Keras-based models. Note that currently
+copying/saving models is not supported, so in a Bayes Opt loop `track_state` should be set False.
+Note as well that `tf.keras.backend.set_floatx()` should be used to set the desired float type.
```
henrymoss: Is it important to be able to set the float type? This isnt something we offer for gpflow models etc

trieste/models/keras/interface.py
```diff
@@ -0,0 +1,82 @@
...
+from typing import Optional, Union
+
+import tensorflow as tf
+
+from ...types import TensorType
+from ..interfaces import ProbabilisticModel
+from ..optimizer import BatchOptimizer
+
+
+class NeuralNetworkPredictor(ProbabilisticModel, tf.Module, ABC):
```
henrymoss, line 27: Maybe this should be `KerasPredictor`, that seems to be a higher level name, i.e. we might have keras models that are not NNs?

trieste/models/keras/interface.py
```diff
@@ -0,0 +1,82 @@
...
+            """
+            NeuralNetworkPredictor class does not implement joint predictions. Acquisition
+            functions relying on it cannot be used with this class by default. Certain
+            types of neural networks might be able to generate joint predictions and
+            such subclasses should overwrite this method.
+            """
+        )
+
+    def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
+        return self.model.sample(query_points, num_samples)
```
henrymoss: Can you always sample from a Keras model?

henrymoss (approved):
LGTM. Just a couple minor comments

trieste/models/keras/interface.py
vpicheny: If we request an optimizer, then it should be a `TrainableProbabilisticModel`, right?

trieste/models/keras/interface.py
hstojic: good point, we can from deep ensembles or probabilistic networks, but not in general - I'll remove it here and leave it to the models to implement

trieste/models/keras/interface.py
hstojic: @henrymoss yes, we can have neural network models in TF that are not Keras, Keras is just a higher level framework that facilitates building them, but interface should be flexible enough that someone could use a model that is not Keras - many researchers would use this, as for developing new types of neural nets most likely you cannot do it with Keras

henrymoss: But, suppose we wanted to write a wrapper for a Keras model that wasnt a NN. Then it would be silly to use this class name?

hstojic: @vpicheny I think it makes sense that optimizer is defined here because all underlying models will need to have it so we might as well standardize it with an interface

if we would like to change these interfaces to inherit from TrainablePM instead, we should do it for GPflow and GPflux as well, I think that would be a job for a separate PR, is that fine?

trieste/models/keras/__init__.py
hstojic: Keras is using float32 by default, while trieste is on float64 - so far I haven't encountered issues as Sebastian did in GPflux, but I thought it would be prudent to instruct Keras to set it to use float64

perhaps I should make it a recommendation, as there are no errors as far as I can see

henrymoss: Sounds good

trieste/models/keras/interface.py
hstojic: > But, suppose we wanted to write a wrapper for a Keras model that wasnt a NN. Then it would be silly to use this class name?

Keras focuses on deep learning, I don't think it makes sense to use it for other types of models, in any case the idea here is to support neural network types of models, using Keras is just for convenience

perhaps this goes back to what we should call the supported model classes - GPs, DGPs and NNs rather than gpflow, gpflux and keras

vpicheny: you're right, better here to copy what we have for gpflow and solve this later.

vpicheny: On @henrymoss comment: I think the question was: could we have keras models that are not NNs? in that case, wouldn't KerasPredictor be a better name? So this interface ties with keras and not NNs. I don't see anything NN specific here, except perhaps the restriction on predict joint.

hstojic: @vpicheny there was another comment above that you might have missed?


> > But, suppose we wanted to write a wrapper for a Keras model that wasnt a NN. Then it would be silly to use this class name?
> 
> Keras focuses on deep learning, I don't think it makes sense to use it for other types of models, in any case the idea here is to support neural network types of models, using Keras is just for convenience
> 
> perhaps this goes back to what we should call the supported model classes - GPs, DGPs and NNs rather than gpflow, gpflux and keras

vpicheny (approved):

PR #450, opened 2021-12-08
Fix slow running tests

uri-granta:
Three changes to (hopefully) fix the slow running tests again:

1. Fix random_seed decorator so it also sets numpy and python seeds
2. Increase the number of steps for the ThompsonSamplerFromTrajectory branin test (@henrymoss: did you expect this to change with your PR?)
3. Ignore the occasional stray retracing in the mixed_space tests (would be nice to understand it, but as long as the test checks we don't retrace every loop then we're ok)

I've kicked off a manual run of the develop check, though we should also make sure to run test_bayesian_optimization.py locally with the `--runslow yes` argument whenever we change the branin tests that are being run.


hstojic (approved):
looks strangely familiar...

this will close #448

henrymoss (approved):
LGTM!

PR #451, opened 2021-12-10
Fantasize acquisition function retracing

### LINKED ISSUE #447: Avoid retracing with Fantasizer
The current Fantasizer retraces every time: both when it gets a new point for a batch, and when it starts a new loop. To avoid this we need to implement an update_acquisition_function method that doesn't call self._builder.prepare_acquisition_function each time and which updates the fantasized data in the models rather than generating new models each time. This is actually quite complex, and will probably require at least the following steps:

1)    _fantasize_model needs to store the fantasized data in Variables, like our GaussianProcessRegression wrapper
    it needs to implement an update_fantasized_data method to update the fantasized data
 2)   FantasizeAcquisitionFunction.prepare_acquisition_function needs to store the fantasize models somewhere that update_acquisition_function can access them (the cleanest is to define a new fantasize_acquisition_function class wrapper similar to expected_improvement, though there's probably a hacky shorter way)
  3)  we needs to implement an FantasizeAcquisitionFunction.update_acquisition_function that calculates the new fantasized_data like here and calls update_fantasized_data on the fantasized model Variables.

uri-granta:


tests/unit/acquisition/function/test_greedy_batch.py
```diff
@@ -188,7 +188,7 @@ def test_fantasized_expected_improvement_builder_raises_for_invalid_fantasize_me
 
 def test_fantasized_expected_improvement_builder_raises_for_invalid_model() -> None:
     data = {
-        "OBJECTIVE": Dataset(tf.zeros([3, 2], dtype=tf.float64), tf.ones([3, 2], dtype=tf.float64))
+        "OBJECTIVE": Dataset(tf.zeros([3, 2], dtype=tf.float64), tf.ones([3, 1], dtype=tf.float64))
```
vpicheny, line 191: wait - this should pass now?? the point of this test was to see if it catches data with a wrong observation shape.

tests/unit/acquisition/function/test_greedy_batch.py
```diff
@@ -231,7 +231,10 @@ def test_fantasize_with_kriging_believer_does_not_change_negative_predictive_mea
 
     builder = Fantasizer(NegativePredictiveMean())
     acq0 = builder.prepare_acquisition_function(models, data)
-    acq1 = builder.prepare_acquisition_function(models, data, pending_points)
+
+    acq1 = builder.update_acquisition_function(acq0, models, data, pending_points[:1])
+    up_acq1 = builder.update_acquisition_function(acq1, models, data, pending_points)
+    assert up_acq1 == acq1  # in-place updates
```
vpicheny: Not sure what's tested here. This test is a bit tricky because the `NegativePredictiveMean` does not change under Fantasize with Kriging believer. So this essentially checks that the equations are correct. With any other acquisition function acq1 and acq0 should differ. 

If you want to assess that `acq1 == up_acq1` in a general context, these lines should go in `test_fantasize_reduces_predictive_variance`. 

Sorry if I misunderstood something.

vpicheny (commented):
Should we also have a test that proves that Fantasize does not retrace?

tests/unit/acquisition/function/test_greedy_batch.py
uri-granta: I thought the point of the test was to check for an invalid model (QuadraticMeanAndRBFKernel, which isn't a FastUpdateModel). The wrong observation shape results in an InvalidArgumentError exception rather than a NotImplementedError exception.

uri-granta: (I've added another test for the observation shape)

tests/unit/acquisition/function/test_greedy_batch.py
uri-granta: I've moved this check to `test_fantasize_reduces_predictive_variance` and added extra checks to ensure we're not retracing.

tests/unit/acquisition/function/test_greedy_batch.py
vpicheny: my bad - and thanks for improving this!!

vpicheny (approved):
LGTM! this PR adds some complexity to the Fantasizer but it does seem necessary and it's overall quite readable still.

PR #452, opened 2021-12-10
support for keras models part 2 - utils

hstojic:
support for neural nets with Keras comes in four parts:

1. interface (done)
2. utilities (this PR)
4. trajectory sampler 
5. integration tests and notebook

### DIFF
--- a/tests/unit/models/keras/test_utils.py
+++ b/tests/unit/models/keras/test_utils.py
@@ -30,7 +30,8 @@ def test_get_tensor_spec_from_data_raises_for_incorrect_dataset() -> None:
 
 
 @pytest.mark.parametrize(
-    "query_point_shape, observation_shape", [([1], [1]), ([2], [1]), ([5], [1]), ([5], [2])]
+    "query_point_shape, observation_shape",
+    [([1], [1]), ([2], [1]), ([5], [1]), ([5], [2]), ([3, 2], [3, 1])],
 )
 def test_get_tensor_spec_from_data(
     query_point_shape: ShapeLike, observation_shape: ShapeLike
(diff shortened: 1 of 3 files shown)


trieste/models/keras/utils.py
```diff
@@ -0,0 +1,75 @@
...
+    :return: Tensor specification objects for the ``query_points`` and ``observations`` tensors.
+    :raise ValueError: If the dataset is not an instance of :class:`~trieste.data.Dataset`.
+    """
+    if not isinstance(dataset, Dataset):
+        raise ValueError(
+            f"This function works only on trieste.data.Dataset objects, however got"
+            f"{type(dataset)} which is incompatible."
+        )
+    input_tensor_spec = tf.TensorSpec(
+        shape=(dataset.query_points.shape[-1],),
```
vpicheny: why the last dimension only?

trieste/models/keras/utils.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+    tf.debugging.assert_positive(len(dataset), message="Dataset must not be empty.")
+
+    n_rows = dataset.observations.shape[0]
+
+    index_tensor = tf.random.uniform(
+        (n_rows,), maxval=n_rows, dtype=tf.dtypes.int32
+    )  # pylint: disable=all
+
+    observations = tf.gather(dataset.observations, index_tensor)  # pylint: disable=all
```
vpicheny: don't we need an `axis=-1` here? could be useful if we want this to work with rank > 2 tensors?

tests/unit/models/keras/test_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+    with pytest.raises(tf.errors.InvalidArgumentError):
+        sample_with_replacement(dataset)
+
+
+@random_seed
+def test_sample_with_replacement_seems_correct() -> None:
+
+    n_rows = 100
+    x = tf.constant(np.arange(0, n_rows, 1), shape=[n_rows, 1])
```
vpicheny: worth checking with x in dim higher than 1? (see other comment below)

vpicheny (approved):
LGTM - left a couple of minor comments

trieste/models/keras/utils.py
hstojic: size of the dataset is not needed for building neural nets, only dimensionality info is needed - I assume here we will only deal with usual matrices, thats why only the last dimension - I can make it more general

trieste/models/keras/utils.py
hstojic: yeah, I wasn't thinking rank > 2 cases would be relevant, I'll do that

tests/unit/models/keras/test_utils.py
hstojic: yep, will do

trieste/models/keras/utils.py
hstojic: done, but I'm not sure `Dataset` would currently support that, I think it has an incompatible check at https://github.com/secondmind-labs/trieste/blob/6b2bb0e73649debaac81157f0f9fdb8d3fdfef5b/trieste/data.py#L51

PR #454, opened 2021-12-16
Bump version to 0.9.0

uri-granta:


CITATION.cff
```diff
@@ -22,6 +22,6 @@ authors:
 - family-names: "Picheny"
   given-names: "Victor"
 title: "Trieste"
-version: 0.8.0
+version: 0.9.0
 date-released: 2021-08-20
```
hstojic: do we need to update the date as well?

hstojic (approved):
looks good, one small comment about the release date

CITATION.cff
uri-granta: Good spot!

PR #455, opened 2021-12-17
Force github to use 2.4.x

uri-granta:
Downgrade gpflux back to 0.2.3 (and revert the accompanying RFF changes).


hstojic (commented):
seems good to me, we just need to make sure we really want this constraint

henrymoss (approved):
LGTM

PR #456, opened 2021-12-17
OpenAI Gym Lunar Lander notebook

apaleyes:
No going away just yet!

This notebook shows how to use Trieste on something that isn't a BayesOpt benchmark function, and has nice visuals. Hopefully that piqued your interested enough to review this PR. Unfortunately you won't be able to see the videos while reviewing, but I've built the docs locally, and they render fine.

As always, the overall tone isn't super formal, please comment if you think the notebook shall mature!


CITATION.cff
```diff
@@ -19,6 +19,8 @@ authors:
   given-names: "Jixiang"
 - family-names: "Satrio"
   given-names: "Loka"
+- family-names: "Paleyes"
+  given-names: "Andrei"
```
apaleyes, line 23: If I may?

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,253 @@
...
+
+observer = trieste.objectives.utils.mk_observer(lander_objective)
+
+# %% [markdown]
+# ### Solving the optimization problem with Trieste
+#
+# Here we do normal steps required to solve an optimization problem with Trieste: generate some initial data, create a surrogate model, define an acquisition funciton and rule, and run the optimization. Optimization step may take a few minutes to complete.
+
+# %%
+num_initial_points = search_space.dimension
```
vpicheny: Usually we use more than this: since the number of an anisotropic stationary kernel has `d+1` parameters, we may want to have more observations than parameters. Not critical for this notebook though,

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,253 @@
...
+
+# %%
+N_RUNS = 10
+def lander_objective(x):
+    # for each point compute average reward over n_runs runs
+    all_rewards = []
+    for w in x.numpy():
+        rewards = [
+            demo_heuristic_lander(env, w)
+            for _ in range(N_RUNS)
```
vpicheny, line 172: Not clear why we use an average over several runs. We could achieve the same result with `N_RUNS=1`. "good on average over a certain number of runs" is actually wrong: BO is trying to maximise the expectation (over an infinite number of runs if you want). 
I guess the main reason is that a single value is too noisy to be directly useful, and the simulation time is small enough to make replicates attractive. Maybe cross-reference to batchBO as an alternative?

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,253 @@
...
+        ]
+        all_rewards.append(rewards)
+
+    rewards_tensor = tf.convert_to_tensor(all_rewards, dtype=tf.float64)
+    
+    # triste minimizes, and we want to maximize
+    return -1 * tf.reshape(tf.math.reduce_mean(rewards_tensor, axis=1), (-1, 1))
```
vpicheny, line 179: Not sure if worth mentioning here, but we could /should actually use the variance of the rewards_tensor to estimate the likelihood variance, since the noise is likely to be heteroscedastic.
Turns out it's not really easy to do this in GPflow, you need a custom VGP model as described here: 
https://gpflow.readthedocs.io/en/master/notebooks/advanced/varying_noise.html
but that would be the "proper" way to solve this problem.

docs/refs.bib
```diff
@@ -303,3 +303,20 @@ @article{wilson2018maximizing
...
+  url = {http://papers.nips.cc/paper/8788-scalable-global-optimization-via-local-bayesian-optimization.pdf},
+}
+
+@article{Moss2020BOSHBO,
+  title={BOSH: Bayesian Optimization by Sampling Hierarchically},
+  author={Henry B. Moss and David S. Leslie and Paul Rayson},
+  journal={ArXiv},
+  year={2020},
+  volume={abs/2007.00939}
+}
```
vpicheny: missing line here perhaps?

docs/notebooks/constraints.txt
```diff
@@ -1,56 +1,63 @@
 absl-py==0.15.0
-aiohttp==3.7.4.post0
+aiohttp==3.8.1
```
vpicheny, line 2: why so many changes? (out of curiosity)
is this generally OK? maybe a question for @uri-granta ?

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: Ok, we can fix that. How many would you suggest? d+1? 2*d?

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: `a single value is too noisy to be directly useful` - that's it, that's the reason we take an average. I really didn't want to go into complex modelling in this notebook, and after some experimentation that seemed like a way to produce a somewhat robust observation of a controller configuration.

I can rephrase if that's confusing.

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: Shall we leave a comment here? Or do you reckon we shall actually do it this way?

docs/refs.bib
apaleyes: indeed

docs/notebooks/constraints.txt
apaleyes: In our "requirements.txt" we don't fix most of the versions. So if between two runs of "generate_constraints" script our dependecies had minor releases (and many usually do), the script picks them up. That's as to "why".

It's not super convenient, I actually had a few errors that I fixed by rolling back some of the changes in this file. Perhaps fixing versions would be nicer. That's as to "is this OK".

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,253 @@
+# ---
```
hstojic: purely for consistency with other notebooks, lets remove this preamble

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,253 @@
...
+
+# %%
+import tensorflow as tf
+import numpy as np
+import trieste
+import gpflow
+
+
+import gym
+ENV_NAME = "LunarLander-v2"
```
hstojic: in other notebooks we haven't been using upper case for constants, so lets be consistent :)
(applies to various instances below as well...)

docs/notebooks/openai_gym_lunar_lander.pct.py
vpicheny: 2*d seems fine!

docs/notebooks/openai_gym_lunar_lander.pct.py
vpicheny: yes, rephrasing should do

docs/notebooks/openai_gym_lunar_lander.pct.py
vpicheny: Either way is fine. The VGP model should bring a bit of an extra performance, but it's probably 50 lines of extra code, so just leaving a comment that we could do better could suffice.

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,253 @@
...
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+# # Trieste meets OpenAI Gym
+#
+# This notebook demonstrates how to use Trieste to apply Bayesian optimization to a problem that is slightly more practical than classical optimization benchmarks shown used in other tutorials. We will use OpenAI Gym, which is a popular toolkit for reinforcement learning algorithms.
```
hstojic:
```suggestion
# This notebook demonstrates how to use Trieste to apply Bayesian optimization to a problem that is slightly more practical than classical optimization benchmarks shown used in other tutorials. We will use OpenAI Gym, which is a popular toolkit for reinforcement learning (RL) algorithms.
```

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,253 @@
...
+
+# %%
+for _ in range(10):
+    sample_w = search_space.sample(1).numpy()[0]
+    demo_heuristic_lander(env, sample_w, print_reward=True)
+
+# %% [markdown]
+# As you can see, most of the random sets of parameters result in a negative reward. So picking a value from this search space at random can result in a various unwanted behaviors. Here we show some examples of the landing not going according to plan. Each of these examples was created with a sample of the parameter values from the search space.
+#
+# **Warning:** all the videos in this notebook were pre-generated. Creating renders of OpenAI Gym environments requires various dependencies depending on software setups and operating systems, so we have chosen not to do it here in the interest of transferability of this notebook. For those interested in reproducing these videos, we have saved the input parameters and the code we used to generate them in the Trieste repository, in the folder next to this notebook. However because of the stochastic nature of the environment and the optimization described here, you results might differ slightly from those shown here.
```
hstojic:
```suggestion
# **Warning:** all the videos in this notebook were pre-generated. Creating renders of OpenAI Gym environments requires various dependencies depending on software setups and operating systems, so we have chosen not to do it here in the interest of transferability of this notebook. For those interested in reproducing these videos, we have saved the input parameters and the code we used to generate them in the Trieste repository, in the folder next to this notebook. However because of the stochastic nature of the environment and the optimization described here, your results might differ slightly from those shown here.
```

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,253 @@
...
+        a = 1
+    return a
+
+
+
+STEPS_LIMIT = 1000
+TIMEOUT_REWARD = -100
+
+
+def demo_heuristic_lander(env, w, print_reward=False):
```
hstojic, line 78: this is copied from Turbo repo as well? a bit unclear if its only the controller or this as welll...

hstojic (approved):
looks good overall, left some minor comments

1. not sure what is the best approach for those videos, its not great idea to push videos to version control, I would rather put a code in the notebook for generating and showing them, but I get that's not ideal as well...
2. recalling the talk by R. Calandra, it would be good to mention somewhere that this has been successfully used in robotics for problems where classical controllers are working fairly well but need to be optimized

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: 2*d it is then

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: rephrased

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: A comment is left in the `build_model` function

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: ok

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: Converted to lower case

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: someone's read this, yay!

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: no, this one isn't. i made the comment a bit clearer

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: Although I don't understad how the other notebooks don't have it. Jupytext seems to insert it by default when converting from ipynb. Am I the only one to create the notebooks as ipynb and then convert?

apaleyes:
Thanks, @hstojic ! To answer your comments:
1. There is a piece of code attached that generates these videos, so we have some transferability. They seem small enough not to cause much trouble in Git (the biggest is under 200Kb). As noted in the notebook, creating them on the fly in reproducible manner is a huge pain.
2. I am having hard time phrasing this idea in a way that would sound relevant in this notebook. If you could suggest something, please share and I will be happy to add it!

docs/notebooks/lunar_lander_videos/generate_video.py
```diff
@@ -0,0 +1,97 @@
...
+The video and several json files will be created in this folder, with some auto-generated names.
+
+Depending on your environment, there might be various dependecies you'd need to run this script.
+In general, these may include:
+apt install python-opengl
+apt install ffmpeg
+apt install xvfb
+apt install x111-utils
+pip install pyglet
+
```
henrymoss, line 14: Should we add these to the notebook requirements.txt?

docs/notebooks/lunar_lander_videos/generate_video.py
```diff
@@ -0,0 +1,97 @@
...
+##### timeout #####
+# seed = 155
+# w = np.array([0.06803627803169543, 0.4287189458093279, 0.476930399661873, 0.5592808413250296, 0.5573280433913701, 0.5095367359357519, 0.7429874662019844, 0.7249766383642469, 0.1320130664358321, 0.7567430455054414, 0.014051753581426185, 0.07791685682019334])
+
+##### out of bounds #####
+# seed = 5
+# w = np.array([0.9679939623340275, 0.2721022418748966, 0.24515670795541378, 0.8011176119748256, 0.13565253791220666, 0.7385592285062779, 0.3511027202815271, 0.44112350462209715, 0.02897150418914718, 0.8063915664159489, 0.21076948458335876, 0.8336549469213406])
+
+##### slam on the surface #####
+# seed = 351
```
henrymoss, line 90: how long did it take you to find this seed???

docs/notebooks/lunar_lander_videos/generate_video.py
apaleyes: Short answer: no, we shouldn't.

Long answer. These aren't required to run the notebook. People would need them if they want to generate their own videos, but I don't have a clue if this list is complete. In my env (Ubuntu WSL) it was enough, but perhaps Mac OS users will need a different setup. Besides, most of these aren't python libs, these are linux packages.

docs/notebooks/lunar_lander_videos/generate_video.py
apaleyes: I had a separate notebook where the lander source code was modified to return additional information of how exactly the landing went - if it crashed, went out of bounds, landed outside the helipad, etc. With this, and a bit of code for looping over seeds, not too long.

In fact I can't remember why this seed is so big, as this is a very common failure mode.

hstojic:
> Thanks, @hstojic ! To answer your comments:
> 
> 1. There is a piece of code attached that generates these videos, so we have some transferability. They seem small enough not to cause much trouble in Git (the biggest is under 200Kb). As noted in the notebook, creating them on the fly in reproducible manner is a huge pain.
> 2. I am having hard time phrasing this idea in a way that would sound relevant in this notebook. If you could suggest something, please share and I will be happy to add it!

All good

apaleyes:
@vpicheny @henrymoss are we good here? any more comments?

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+def build_model(data):
+    variance = tf.math.reduce_variance(data.observations)
+    kernel = gpflow.kernels.RBF(variance=variance)
+    gpr = gpflow.models.GPR(data.astuple(), kernel)
+    gpflow.set_trainable(gpr.likelihood, False)
+    # Since we are running multiple simulations per observation,
+    # it is possible to estimate likelihood variance from this data for each observation.
+    # This is possible to model with VGP, as described here:
+    # https://gpflow.readthedocs.io/en/master/notebooks/advanced/varying_noise.html\
+    # In the interest of brevity we have chosen not to do it in this notebook.
```
st--: This is a cool demo! I hope you don't mind me jumping in with a comment: I find the description above somewhat confusing, as it seems to throw together, on the one hand, whether your likelihood models homoskedastic noise (same variance everywhere, default in GPR) vs heteroskedastic noise (what you're alluding to with the VGP model), and on the other hand, whether you have the noise variance be trainable (default) or not (you explicitly set it to non-trainable in line 192).

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+
+np.random.seed(seed)
+tf.random.set_seed(seed)
+env.seed(seed)
+
+
+# %% [markdown]
+# ### Introduction
+#
+# Let's start by discussing the problem itself. In the Lunar Lander environment we are controlling a space module that needs to land on a Moon surface. The surface is piecewise linear and is generated randomly, but always has a flat landing pad in the middle (marked with flags on the renders). The module starts at the top with some random initial speed and direction. We are controlling three engines on the module: one on each side, and one at the bottom. At each step of the simulation we can choose to either fire one of the engines, or do nothing. The ultimate goal of the simulation is to land safely on the marked landing pad.
```
vpicheny, line 29: remove first sentence?

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+#
+# As usual in RL settings, the environment calculates reward points. Landing in the designated area gives the biggest reward, landing safely elsewhere is also rewarded but with less points. Crashing or flying off the screen results in big negative reward. Few points are also deducted along the way for firing up engines, thus motivating smaller usage of fuel. Additionally, to make the running time manageable, we are going to penalize simulations that take too long, but stopping them after a certain amount of steps and penalizing the reward.
```
vpicheny: but stopping => by stopping?

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+#
+# ### Optimization problem
+#
+# Now let's see how this task can be formulated as an optimization problem. We will be following an approach used by Turbo <cite data-cite="eriksson2019scalable"/> and BOSH <cite data-cite="Moss2020BOSHBO"/> papers. The environment comes with a heuristic controller that makes decisions based on the current position and velocity of the module. This controller can be parametrized by turning all numerical values in it into parameters. These parameters form our optimization search space. The objective is the same as in the original RL setup: maximize the reward. Therefore we will be using Trieste to learn how to land the module safely on the designated pad, without taking too much time and wasting too much fuel.
```
vpicheny: "turning all numerical values in it into parameters": not sure this is really clear. How about: "This controller can be tuned by modifying 12 of its internal numerical parameters."

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+#
+# The original code for the heuristic controller can be found in [OpenAI Gym GitHub repo](https://github.com/openai/gym/blob/master/gym/envs/box2d/lunar_lander.py). Here is the parametrized version, taken from the [repository](https://github.com/uber-research/TuRBO) of the Turbo paper:
+
+# %%
+# controller code is copied verbatim from https://github.com/uber-research/TuRBO
+def heuristic_Controller(s, w):
```
vpicheny, line 44: could be useful to add what `s` and `w` are, respectively, and their size?

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+        a = 1
+    return a
+
+
+
+steps_limit = 1000
+timeout_reward = -100
+
+
+def demo_heuristic_lander(env, w, print_reward=False):
```
vpicheny, line 78: change name to something descriptive? maybe add a short sentence explaining that this wraps the above to compute the reward, including a stop if too long?

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+            break
+
+    if print_reward:
+        print(f"Total reward: {total_reward}")
+
+    return total_reward
+
+
+# %% [markdown]
+# Original values for the various parameters take values between 0.05 and 1.0. We will set the search range for each parameter to be the same, from 0.0 to 1.2.
```
vpicheny: is this taken from TurBO? or it follows our own logic? are all of those parameters rescaled, so it makes sense to have them all within [0, 1.2]? 
Feels like this sentence says either too much or too little.

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+
+# %% [markdown]
+# #### Slamming on the surface
+# This is a very common failure mode in this environment - going too fast and slamming on the surface.
+
+# %%
+load_video("slam.mp4")
+
+# %% [markdown]
+# Finally, we need a way to deal with stochasticity of the simulation. To keep our modelling simple, let's aim at finding a controller that achieves highest average reward over a fixed number of runs. This gives us a simple way of implementing the Trieste observer that isn't too sensitive to the randomness of a single simulation. Also note that we reverse the reward sign: this is because we seek the maximum reward, and Trieste approaches optimization as minimization problem.
```
vpicheny: Sorry, still not happy with this sentence. How about:
"Since the reward is stochastic, our goal is to maximize its expectation (or rather here minimize minus its expectation since Trieste only deals with minimization problems). To decrease the observation noise, our observer returns the average of the reward over 10 runs."

docs/notebooks/openai_gym_lunar_lander.pct.py
vpicheny: agreed, how about:
"it is possible to account for variations in observation noise by using a different likelihood variance for each observation"

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+    # it is possible to estimate likelihood variance from this data for each observation.
+    # This is possible to model with VGP, as described here:
+    # https://gpflow.readthedocs.io/en/master/notebooks/advanced/varying_noise.html\
+    # In the interest of brevity we have chosen not to do it in this notebook.
+    return trieste.models.gpflow.GaussianProcessRegression(gpr)
+
+model = build_model(initial_data)
+
+# %%
+acq_fn = trieste.acquisition.function.ExpectedImprovement()
```
vpicheny: `EI` isn't great when noise is substantial, how about using `AEI` instead? or `MES` if we want to avertise @henrymoss fancy noisy version of it?

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+    
+    # triste minimizes, and we want to maximize
+    return -1 * tf.reshape(tf.math.reduce_mean(rewards_tensor, axis=1), (-1, 1))
+
+observer = trieste.objectives.utils.mk_observer(lander_objective)
+
+# %% [markdown]
+# ### Solving the optimization problem with Trieste
+#
+# Here we do normal steps required to solve an optimization problem with Trieste: generate some initial data, create a surrogate model, define an acquisition funciton and rule, and run the optimization. Optimization step may take a few minutes to complete.
```
vpicheny, line 186: Add a sentence here to describe your BO choices? e.g. `GPR` model and whatever acquisition function you use.

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+
+# %%
+N_OPTIMIZATION_STEPS = 200
+bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+result = bo.optimize(N_OPTIMIZATION_STEPS, initial_data, model, rule).final_result.unwrap()
+
+# %% [markdown]
+# ### Analyzing the results
+#
+# First, let's just plot a regular regret plot of our optimization, to ensure Trieste indeed found a better configuration of the controller. Remember that we flipped the sign of the reward.
```
vpicheny: Are we actually plotting the regret? this implies that we have access to the solution of the problem. Otherwise we're just plotting noisy observations of the expected reward.

docs/notebooks/openai_gym_lunar_lander.pct.py
```diff
@@ -0,0 +1,244 @@
...
+    result.dataset.observations.numpy(),
+    ax,
+    num_init=len(initial_data)
+)
+
+# %% [markdown]
+# We can also retrieve the best configuration found, and test it. We expect to see mostly large positive rewards here.
+
+# %%
+w_best = result.dataset.query_points[np.argmin(result.dataset.observations), :]
```
vpicheny: this makes me a bit sad... we should take the argmin of the GPR prediction at dataset.query_points. With a bit of luck they are the same!

docs/refs.bib
```diff
@@ -304,6 +304,23 @@ @article{wilson2018maximizing
...
+@inproceedings{eriksson2019scalable,
+  title = {Scalable Global Optimization via Local {Bayesian} Optimization},
+  author = {Eriksson, David and Pearce, Michael and Gardner, Jacob and Turner, Ryan D and Poloczek, Matthias},
+  booktitle = {Advances in Neural Information Processing Systems},
+  pages = {5496--5507},
+  year = {2019},
+  url = {http://papers.nips.cc/paper/8788-scalable-global-optimization-via-local-bayesian-optimization.pdf},
+}
+
+@article{Moss2020BOSHBO,
```
vpicheny, line 316: @henrymoss is this published? if so we should cite the conference / paper version, not arxiv

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: Why? Does it hurt?

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: good catch

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: updated

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: worth clarifying indeed

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: I think name is ok, but a description is certainly useful to add

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: I see your confusion, and tried to update the sentence. Let me know if that's not clear still.

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: sure

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: Not at all, thanks @st-- and @vpicheny for clarification

docs/refs.bib
apaleyes: this was a workshop paper, so I guess the answer is "no"?

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: I am a bit confused about the terminology here, so happy to adhere

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: added

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: using AEI, and seems to work fine on a few runs done so far

docs/notebooks/openai_gym_lunar_lander.pct.py
apaleyes: please don't be sad, Victor! they are the same indeed, so i've changed it. hope it makes you happy!

docs/notebooks/openai_gym_lunar_lander.pct.py
vpicheny: sorry, just the old habit of chasing the superfluous. Keep it if you like it, this is not an 8-page NeurIPS paper

apaleyes: will keep, thanks!

docs/notebooks/openai_gym_lunar_lander.pct.py
vpicheny: I can't see the update?

vpicheny (approved):
Cool notebook!

docs/notebooks/openai_gym_lunar_lander.pct.py
vpicheny: how about adding a sentence explaining our choice? e.g. "We choose here the query point that gives the best predictive expected reward according to our model".

docs/notebooks/openai_gym_lunar_lander.pct.py
vpicheny: how about adding a sentence explaining our choice? e.g. "We choose here the query point that gives the best predictive expected reward according to our model".

apaleyes: souds good!

PR #458, opened 2021-12-21
Support for vectorized acquisition functions

henrymoss:
We now support vectorizable batch acquisition functions, i.e. those for which acquisition function maximisation can be embarrassingly parallelised across a batch --- imagine a separate acquisition function for each batch element.  Therefore, I have included a `VectorizedAcquisitionFunction` and a `VectorizedAcquisitionFunctionBuilder`.

The desired use-case for this functionality is batch Thompson sampling for very large batches. However, we are still lacking a little functionality (see next PR), so I have implemented the parallel UCB from https://arxiv.org/pdf/2001.04833.pdf as a test case. Therefore, this PR also contains a new acquisition function that allows a nice integration test to check that everything is working as planned. 

I have updated all our acquisition functions to be able to maximize `VectorizedAcquisitionFunction` in parallel.

A subtle detail is that I have had to add a new attribute to search spaces. We can now store batch size as a attribute of search space. This is important as search spaces are how we have previously passed batch size info to our acq functions (i.e. via the optimzier). For traditional batch acq functions (which must be optimized in parallel), wejust pass the extended search space with B*d inputs, however, for  `VectorizedAcquisitionFunction` we need access to both B and d when optimising.

### DIFF
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -32,6 +32,7 @@ from trieste.acquisition import (
     Fantasizer,
     LocalPenalization,
     MinValueEntropySearch,
+    MultipleOptimismNegativeLowerConfidenceBound,
 )
 from trieste.acquisition.rule import (
     AcquisitionRule,
@@ -140,6 +141,15 @@ def OPTIMIZER_PARAMS() -> Tuple[
                         BRANIN_SEARCH_SPACE,
                     ).using(OBJECTIVE),
                     num_query_points=2,
+                ),
+            ),
+            (
+                10,
+                EfficientGlobalOptimization(
+                    MultipleOptimismNegativeLowerConfidenceBound(
+                        BRANIN_SEARCH_SPACE,
+                    ).using(OBJECTIVE),
+                    num_query_points=3,
                 ),
             ),
             (15, TrustRegion()),
--- a/tests/unit/acquisition/test_optimizer.py
+++ b/tests/unit/acquisition/test_optimizer.py
@@ -25,13 +25,20 @@ from trieste.acquisition.optimizer import (
     AcquisitionOptimizer,
     FailedOptimizationError,
     automatic_optimizer_selector,
-    batchify,
+    batchify_joint,
+    batchify_vectorize,
     generate_continuous_optimizer,
     generate_random_search_optimizer,
     get_bounds_of_box_relaxation_around_point,
     optimize_discrete,
 )
 from trieste.objectives import (
+    BRANIN_MINIMUM,
+    BRANIN_SEARCH_SPACE,
+    branin,
+    scaled_branin,
+    SCALED_BRANIN_MINIMUM,
+    SIMPLE_QUADRATIC_MINIMUM,
     ACKLEY_5_MINIMIZER,
     ACKLEY_5_SEARCH_SPACE,
     HARTMANN_3_MINIMIZER,
@@ -40,6 +47,7 @@ from trieste.objectives import (
     HARTMANN_6_SEARCH_SPACE,
     ackley_5,
     hartmann_3,
+    simple_quadratic,
     hartmann_6,
 )
 from trieste.space import Box, DiscreteSearchSpace, SearchSpace, TaggedProductSearchSpace
@@ -116,6 +124,10 @@ def test_discrete_and_random_optimizer_on_quadratic(
             npt.assert_allclose(maximizer, expected_maximizer, rtol=1e-4)
         else:
             npt.assert_allclose(maximizer, expected_maximizer, rtol=1e-1)
+
+
+
+
 
 
 @random_seed
@@ -377,6 +389,14 @@ def test_continuous_optimizer_on_toy_problems(
     npt.assert_allclose(maximizer, expected_maximizer, rtol=1e-1)
 
 
+
+
+
+
+
+
+
+
 @pytest.mark.parametrize(
     "search_space, point",
     [
@@ -456,10 +476,10 @@ def test_get_bounds_of_box_relaxation_around_point(
     npt.assert_array_equal(bounds.ub, upper)
 
 
-def test_optimize_batch_raises_with_invalid_batch_size() -> None:
+def test_batchify_joint_raises_with_invalid_batch_size() -> None:
     batch_size_one_optimizer = generate_continuous_optimizer()
     with pytest.raises(ValueError):
-        batchify(batch_size_one_optimizer, -5)
+        batchify_joint(batch_size_one_optimizer, -5)
 
 
 @random_seed
@@ -471,15 +491,106 @@ def test_optimize_batch_raises_with_invalid_batch_size() -> None:
         (Box([-1, -1, -1], [1, 1, 1]), _quadratic_sum([0.5, -0.5, 0.2]), ([[0.5, -0.5, 0.2]])),
     ],
 )
-def test_optimize_batch(
+def test_batchify_joint(
     search_space: Box, acquisition: AcquisitionFunction, maximizer: TensorType, batch_size: int
 ) -> None:
     batch_size_one_optimizer = generate_continuous_optimizer(num_optimization_runs=5)
-    batch_optimizer = batchify(batch_size_one_optimizer, batch_size)
+    batch_optimizer = batchify_joint(batch_size_one_optimizer, batch_size)
     points = batch_optimizer(search_space, acquisition)
     assert points.shape == [batch_size] + search_space.lower.shape
     for point in points:
         npt.assert_allclose(tf.expand_dims(point, 0), maximizer, rtol=2e-4)
+
+
+
+
+
+
+def test_batchify_vectorized_raises_with_invalid_batch_size() -> None:
+    batch_size_one_optimizer = generate_continuous_optimizer()
+    with pytest.raises(ValueError):
+        batchify_vectorize(batch_size_one_optimizer, -5)
+
+
+@random_seed
+@pytest.mark.parametrize("optimizer", [generate_random_search_optimizer(10_000), generate_continuous_optimizer()])
+def test_batchify_vectorized_for_random_and_continuous_optimizers_on_vectorized_quadratic(optimizer: AcquisitionOptimizer) -> None:
+    search_space = Box([-1, -2], [1.5, 2.5])
+    shifts = [[0.3, -0.4],[1.0, 4]]
+    expected_maximizers = [[0.3, -0.4], [1.0, 2.5]]
+    vectorized_batch_size=2
+
+    def vectorized_target(x: TensorType) -> TensorType: # [N, V, D] -> [N,V]
+        individual_func =  [_quadratic_sum(shifts[i])(x[:,i:i+1,:]) for i in range(vectorized_batch_size)]  
+        return tf.concat(individual_func, axis=-1)
+        
+    batched_optimizer =batchify_vectorize(optimizer, batch_size=vectorized_batch_size)
+    maximizers = batched_optimizer(search_space, vectorized_target)
+    npt.assert_allclose(maximizers, expected_maximizers, rtol=1e-1)
+    
+    maximizers = optimizer(search_space, vectorized_target, vectorized_batch_size=vectorized_batch_size)
+    npt.assert_allclose(maximizers, expected_maximizers, rtol=1e-1)
+
+def test_batchify_vectorized_for_discrete_optimizer_on_vectorized_quadratic() -> None:
+    search_space = DiscreteSearchSpace(tf.constant([[0.3, -0.4], [1.0, 2.5], [0.2,0.5], [0.5,2.0],[2.0,0.1]]))
+    shifts = [[0.3, -0.4],[1.0, 4]]
+    expected_maximizers = [[0.3, -0.4], [1.0, 2.5]]
+    vectorized_batch_size=2
+
+    def vectorized_target(x: TensorType) -> TensorType: # [N, V, D] -> [N,V]
+        individual_func =  [_quadratic_sum(shifts[i])(x[:,i:i+1,:]) for i in range(vectorized_batch_size)]  
+        return tf.concat(individual_func, axis=-1)
+        
+    batched_optimizer =batchify_vectorize(optimize_discrete, batch_size=vectorized_batch_size)
+    maximizers = batched_optimizer(search_space, vectorized_target)
+    npt.assert_allclose(maximizers, expected_maximizers, rtol=1e-1)
+    
+    maximizers = optimize_discrete(search_space, vectorized_target, vectorized_batch_size=vectorized_batch_size)
+    npt.assert_allclose(maximizers, expected_maximizers, rtol=1e-1)
+
+
+@random_seed
+@pytest.mark.parametrize("vectorization", [1,5])
+@pytest.mark.parametrize(
+    "neg_function, expected_maximizer, search_space",
+    [
+        (ackley_5, ACKLEY_5_MINIMIZER, ACKLEY_5_SEARCH_SPACE),
+        (hartmann_3, HARTMANN_3_MINIMIZER, HARTMANN_3_SEARCH_SPACE),
+        (hartmann_6, HARTMANN_6_MINIMIZER, HARTMANN_6_SEARCH_SPACE),
+    ],
+)
+def test_batchify_vectorized_for_continuous_optimizer_on_duplicated_toy_problems(
+    vectorization: int,
+    neg_function: Callable[[TensorType], TensorType],
+    expected_maximizer: TensorType,
+    search_space: Box,
+) -> None:
+    def target_function(x: TensorType) -> TensorType: # [N,V,D] -> [N, V]
+        individual_func =  [-1 * neg_function(x[:,i,:]) for i in range(vectorization)]  
+        return tf.concat(individual_func, axis=-1) # vectorize by repeating same function
+
+    optimizer = batchify_vectorize(generate_continuous_optimizer(num_initial_samples=1_000, num_optimization_runs=10), batch_size=vectorization)
+    maximizer = optimizer(search_space, target_function)
+    npt.assert_allclose(maximizer, tf.tile(expected_maximizer,[vectorization,1]), rtol=1e-1)
+
+
+@random_seed
+def test_batchify_vectorized_for_continuous_optimizer_on_vectorized_toy_problems() -> None:
+    search_space = BRANIN_SEARCH_SPACE
+    functions = [branin, scaled_branin, simple_quadratic]
+    expected_maximimums = [-BRANIN_MINIMUM, -SCALED_BRANIN_MINIMUM, -SIMPLE_QUADRATIC_MINIMUM]
+    vectorized_batch_size=3
+
+    def target_function(x: TensorType) -> TensorType: # [N,V,D] -> [N, V]
+        individual_func =  [-1 * functions[i](x[:,i,:]) for i in range(vectorized_batch_size)]  
+        return tf.concat(individual_func, axis=-1) # vectorize by concatenating three functions
+
+    optimizer = batchify_vectorize(generate_continuous_optimizer(num_initial_samples=1_000, num_optimization_runs=10), batch_size=vectorized_batch_size)
+    maximizer = optimizer(search_space, target_function)
+    npt.assert_allclose(target_function(maximizer[None,:,:]), tf.transpose(expected_maximimums), rtol=1e-5)
+
+
+
 
 
 @random_seed
@@ -513,3 +624,12 @@ def test_automatic_optimizer_selector(
     optimizer = automatic_optimizer_selector
     point = optimizer(search_space, acquisition)
     npt.assert_allclose(point, maximizer, rtol=2e-4)
+
+
+
+
+
+
+
+
+
--- a/tests/unit/acquisition/test_optimizer.py
+++ b/tests/unit/acquisition/test_optimizer.py
@@ -67,6 +67,22 @@ def test_generate_random_search_optimizer_raises_with_invalid_sample_size() -> N
         generate_random_search_optimizer(num_samples=-5)
 
 
+@pytest.mark.parametrize("batch_size", [0, -2])
+def test_optimize_discrete_raises_with_invalid_vectorized_batch_size(batch_size: int) -> None:
+    search_space = DiscreteSearchSpace(tf.constant([[-0.5], [0.2], [1.2], [1.7]]))
+    acq_fn = _quadratic_sum([1.0])
+    with pytest.raises(ValueError):
+        optimize_discrete(search_space, (acq_fn, batch_size))
+
+
+@pytest.mark.parametrize("batch_size", [0, -2])
+def test_random_optimizer_raises_with_invalid_vectorized_batch_size(batch_size: int) -> None:
+    search_space = Box([-1], [2])
+    acq_fn = _quadratic_sum([1.0])
+    with pytest.raises(ValueError):
+        generate_random_search_optimizer()(search_space, (acq_fn, batch_size))
+
+
 SP = TypeVar("SP", bound=SearchSpace)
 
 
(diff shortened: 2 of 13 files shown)


tests/integration/test_bayesian_optimization.py
```diff
@@ -142,6 +143,15 @@ def OPTIMIZER_PARAMS() -> Tuple[
                     num_query_points=2,
                 ),
             ),
+            (
+                10,
+                EfficientGlobalOptimization(
```
uri-granta, line 148: I assume you've tried running the slow version of this? (i.e. with --runslow yes)

Also FYI if your branch is on the trieste repo (rather than your own fork) then you can also manually kick off an alltests run on github.

trieste/space.py
```diff
@@ -34,6 +34,9 @@ class SearchSpace(ABC):
     A :class:`SearchSpace` represents the domain over which an objective function is optimized.
     """
 
+    def __init__(self) -> None:
+        self._vectorized_batch_size = 1  # TODO
```
uri-granta: TODO?

trieste/space.py
```diff
@@ -87,6 +90,15 @@ def __pow__(self: SP, other: int) -> SP:
         tf.debugging.assert_positive(other, message="Exponent must be strictly positive")
         return reduce(operator.mul, [self] * other)
 
+    @property
+    def vectorized_batch_size(self) -> int:
+        return self._vectorized_batch_size
+
+    def set_vectorized_batch_size(self, batch_size: int) -> None:
```
uri-granta: You should use a property setter here:

```suggestion
    @vectorized_batch_size.setter
    def vectorized_batch_size(self, batch_size: int) -> None:
```

You can then set it with normal-looking assignment `space.vectorized_batch_size = 3`.

trieste/space.py
```diff
@@ -566,3 +582,6 @@ def __mul__(self, other: SearchSpace) -> TaggedProductSearchSpace:
         :return: The Cartesian product of this search space with the ``other``.
         """
         return TaggedProductSearchSpace(spaces=[self, other])
+
+    def __deepcopy__(self, memo: dict[int, object]) -> TaggedProductSearchSpace:
```
uri-granta, line 571: Doesn't vectorized_batch_size break all these deepcopy overrides? Not having to copy spaces is predicated on them being immutable, which is no longer true. (Also, even if that weren't the case, not sure it's correct to have the override here, as someone could always define a weird subspace type that did require real copying.)

Not sure what the right solution is: this deecopy override is to avoid unnecessary copying during optimization, which may or may not be significant. I'll have a think.

trieste/acquisition/interface.py
```diff
@@ -37,6 +37,18 @@
 with a batch dimension, i.e. an input of shape `[..., 1, D]`.
 """
 
+VectorizedAcquisitionFunction = Callable[[TensorType], TensorType]
+"""
+Type alias for vectorized acquisition functions.
+
+An :const:`VectorizedAcquisitionFunction` differs from an :const:`AcquisitionFunction` only by its
```
uri-granta:
```suggestion
A :const:`VectorizedAcquisitionFunction` differs from an :const:`AcquisitionFunction` only by its
```

trieste/acquisition/interface.py
```diff
@@ -37,6 +37,18 @@
...
+output shape. An :const:`VectorizedAcquisitionFunction` maps a set of maps a set of `B` query points
```
uri-granta:
```suggestion
output shape. A :const:`VectorizedAcquisitionFunction` maps a set of `B` query points
```

trieste/acquisition/interface.py
```diff
@@ -37,6 +37,18 @@
...
+(each of dimension `D`) to a `B` values that describes how useful it would be evaluate each of these
```
uri-granta:
```suggestion
(each of dimension `D`) to `B` values that describe how useful it would be evaluate each of these
```

trieste/acquisition/interface.py
```diff
@@ -304,6 +316,114 @@ def update_acquisition_function(
         )
 
 
+class VectorizedAcquisitionFunctionBuilder(Generic[T], ABC):
```
uri-granta: So are these two classes identical to AcquisitionFunctionBuilder and SingleModelAcquisitionBuilder, apart from returning VectorizedAcquisitionFunction (which are really the same as AcquisitionFunction as the tensor shape isn't encoded in the type)? Is the only reason the type definition is required is so we know to call batchify_vectorize rather than batchify_joint?

If so could we get rid of VectorizedAcquisitionFunction altogether, and define VectorizedAcquisitionFunctionBuilder simply by:

```python
class VectorizedAcquisitionFunctionBuilder(AcquisitionFunctionBuilder):
    """
    An :class:`VectorizedAcquisitionFunctionBuilder` builds and updates a vectorized
    acquisition function. These differ from normal acquisition functions by their output shape:
    rather than returning a single value, they return one value per potential query point.
    Thus, they take input shape `[..., B, D]` and return shape `[..., B]`.
    """"
````

File: trieste/acquisition/function/function.py [hunk 9]
```diff
@@ -699,3 +702,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+                tf.cast(batch_size, dtype=x.dtype) + 1.0
+            )  # [B]
+            betas = normal.quantile(spread)  # [B]
+            scaled_betas = 5.0 * tf.cast(self._search_space_dim, dtype=x.dtype) * betas  # [B]
+            self._betas.assign(scaled_betas)  # [B]
+            self._initialized.assign(True)
+
+        mean, variance = self._model.predict(x)  # [..., B, 1]
+        mean, variance = tf.squeeze(mean, -1), tf.squeeze(variance, -1)
+        return -mean + tf.sqrt(variance) * self._betas  # [..., B]
```
uri-granta, line 808: Looks good!

uri-granta (commented):
Some initial talking points

trieste/acquisition/function/function.py
```diff
@@ -699,3 +702,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    def __call__(self, x: TensorType) -> TensorType:
+
+        batch_size = tf.shape(x)[-2]
+        tf.debugging.assert_positive(batch_size)
+
+        if self._initialized:  # check batch size hasnt changed during BO
+            tf.debugging.assert_equal(
+                batch_size,
+                tf.shape(self._betas)[0],
+                f"{type(self).__name__} requires a fixed batch size. Got batch size {batch_size}"
```
vpicheny, line 790: maybe also be explicit that beta shape doesn't match batch_size?

trieste/acquisition/function/function.py
vpicheny, line 808: missing a minus sign before `sqrt`, right?

trieste/acquisition/interface.py
vpicheny: Should we use (here and everywhere else) another letter for the vectorised dimension? It should be entirely possible to vectorise batch acquisition functions, i.e. you would obtain a batch of size `V*B` by optimising V batches of size B. Even if we never do this, I think it would avoid a lot of confusion.

trieste/acquisition/interface.py
```diff
@@ -37,6 +37,18 @@
...
 
+VectorizedAcquisitionFunction = Callable[[TensorType], TensorType]
+"""
+Type alias for vectorized acquisition functions.
+
+An :const:`VectorizedAcquisitionFunction` differs from an :const:`AcquisitionFunction` only by its
+output shape. An :const:`VectorizedAcquisitionFunction` maps a set of maps a set of `B` query points
+(each of dimension `D`) to a `B` values that describes how useful it would be evaluate each of these
+points (to our goal of optimizing the objective function). Thus, with leading dimensions, a
+:const:`VectorizedAcquisitionFunction` takes input shape `[..., B, D]` and returns shape `[..., B]`.
```
vpicheny: mmm wouldn't we want to have an input shape `[..., B, V, D]`? and return shape `[..., V]`

trieste/acquisition/rule.py
```diff
@@ -382,9 +404,9 @@ def __init__(
             builder = builder.using(OBJECTIVE)
 
         # even though we are only using batch acquisition functions
-        # there is no need to batchify the optimizer if our batch size is 1
+        # there is no need to batchify_joint the optimizer if our batch size is 1
```
vpicheny, line 407: why not the same for batchify_vectorize?

vpicheny (commented):
High level comment: I'm not sure I'm convinced by the design choice to let the search space carry the batch size. I think that this is very unintuitive, and also it defeats the purpose to have a acquisition rule, which precisely exists for this sort of reasons...

vpicheny:
I also wonder if we should make vectorise and batch compatible. It would be really nice if they were (better modularity, also much less confusing), but I appreciate the extra complexity involved.

henrymoss:
New version no longer stores vectorized batch size in the search space for the reasons discussed above. Instead we can now optionally pass the desired vectorization along with the acquisition function to the optimizer as a `Tuple[AcquisitionFunction, int]`

trieste/acquisition/optimizer.py
```diff
@@ -413,9 +529,16 @@ def batchify(
...
+    def optimizer(
+        search_space: SP, f: Union[AcquisitionFunction, Tuple[AcquisitionFunction, int]]
+    ) -> TensorType:
         expanded_search_space = search_space ** batch_size  # points have shape [B * D]
 
+        if isinstance(f, tuple):
+            raise ValueError(
+                "batchify_joint cannot be applied to a vectorized acquisition function"
+            )
+
```
uri-granta, line 547: The type error here is because f is being called inside a closure: by the time target_func_with_vectorized_inputs gets called f may have been assigned a new value. I think you can work around this by adding the following after the isinstance, and using `af` rather than `f` in target_func_with_vectorized_inputs:

```suggestion
af : AcquisitionFunction = f
```

trieste/acquisition/interface.py
```diff
@@ -304,6 +304,55 @@ def update_acquisition_function(
         )
 
 
+class VectorizedAcquisitionFunctionBuilder(AcquisitionFunctionBuilder[T]):
+    """
+    An :class:`VectorizedAcquisitionFunctionBuilder` builds and updates a vectorized
+    acquisition function These differ from normal acquisition functions only by their output shape:
+    rather than returng a single value, they return one value per potential query point.
```
hstojic:
```suggestion
    rather than returning a single value, they return one value per potential query point.
```

trieste/acquisition/function/function.py
```diff
@@ -699,3 +701,108 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        self._search_space_dim = search_space_dim
+
+        self._model = model
+        self._initialized = tf.Variable(False)  # Keep track of when we need to resample
+        self._betas = tf.Variable(tf.ones([0], dtype=tf.float64), shape=[None])  # [0] lazy init
+
+    @tf.function
+    def __call__(self, x: TensorType) -> TensorType:
+
+        batch_size = tf.shape(x)[-2]
```
hstojic, line 783: is -2 relying on x being rank=2? if so can we make it more flexible?

trieste/acquisition/optimizer.py
```diff
@@ -162,8 +186,8 @@ def generate_continuous_optimizer(
     if num_initial_samples <= 0:
         raise ValueError(f"num_initial_samples must be positive, got {num_initial_samples}")
 
-    if num_optimization_runs <= 0:
-        raise ValueError(f"num_optimization_runs must be positive, got {num_optimization_runs}")
+    if num_optimization_runs < 0:
+        raise ValueError(f"num_initial_samples must be positive, got {num_initial_samples}")
```
hstojic: num_initial_samples? you are testing here num_optimization_runs, right?

tests/unit/acquisition/test_optimizer.py
```diff
@@ -59,6 +67,22 @@ def test_generate_random_search_optimizer_raises_with_invalid_sample_size() -> N
...
+    with pytest.raises(ValueError):
+        optimize_discrete(search_space, (acq_fn, batch_size))
+
+
+@pytest.mark.parametrize("batch_size", [0, -2])
+def test_random_optimizer_raises_with_invalid_vectorized_batch_size(batch_size: int) -> None:
+    search_space = Box([-1], [2])
+    acq_fn = _quadratic_sum([1.0])
+    with pytest.raises(ValueError):
+        generate_random_search_optimizer()(search_space, (acq_fn, batch_size))
```
uri-granta, line 83: The mypy error here is actually a preexisting issue: generate_random_search_optimizer (added in #242) has a return type of `AcquisitionOptimizer[SP]` (which doesn't make sense). It should instead be `AcquisitionOptimizer[SearchSpace]`.

uri-granta: (you'll need to change the type of space on line 605 as well)

trieste/acquisition/interface.py
```diff
@@ -316,57 +304,24 @@ def update_acquisition_function(
...
-class SingleModelVectorizedAcquisitionBuilder(Generic[T], ABC):
+class SingleModelVectorizedAcquisitionBuilder(SingleModelAcquisitionBuilder[T]):
     """
-    Convenience acquisition function builder for a vectorized acquisition function (or component
+    Convenience acquisition function builder for vectorized acquisition functions (or component
     of a composite vectorized acquisition function) that requires only one model, dataset pair.
     """
 
-    def using(self, tag: str) -> VectorizedAcquisitionFunctionBuilder[T]:
+    def using(self, tag: str) -> AcquisitionFunctionBuilder[T]:
```
uri-granta, line 322: I don't understand this change. Don't we want `SingleModelVectorizedAcquisitionBuilder().using(OBJECTIVE)` to return a `VectorizedAcquisitionFunctionBuilder` rather than a `AcquisitionFunctionBuilder` (so that we recognise it as such in the rule?)

trieste/acquisition/rule.py
```diff
@@ -173,12 +173,12 @@ def __init__(
...
 
         if num_query_points > 1:  # need to build batches of points
-            if isinstance(builder, AcquisitionFunctionBuilder):
-                # optimize batch elements jointly
-                optimizer = batchify_joint(optimizer, num_query_points)
-            elif isinstance(builder, VectorizedAcquisitionFunctionBuilder):
+            if isinstance(builder, VectorizedAcquisitionFunctionBuilder):
                 # optimize batch elements independently
                 optimizer = batchify_vectorize(optimizer, num_query_points)
+            elif isinstance(builder, AcquisitionFunctionBuilder):
```
uri-granta, line 179: Do we need to also update the `AsynchronousOptimization` rule?

trieste/acquisition/optimizer.py
```diff
@@ -64,21 +63,20 @@ class FailedOptimizationError(Exception):
...
 ]
-
-
 """
 Type alias for a function that returns the single point that maximizes an acquisition function over
 a search space. For a search space with points of shape [D], and acquisition function with input
 shape [..., B, D] output shape [..., 1], the :const:`AcquisitionOptimizer` return shape should be
 [B, D].
+
+TODO
```
uri-granta: TODO

trieste/acquisition/optimizer.py
```diff
@@ -127,7 +125,14 @@ def optimize_discrete(
     :return: The **one** point in ``space`` that maximises ``target_func``, with shape [1, D].
     """
 
-    V = space.vectorized_batch_size
+    if isinstance(target_func, tuple):  # TODO
+        V = target_func[1]
```
```suggestion
        target_func, V = target_func
```

uri-granta (commented):
A few more comments, but looks almost done

trieste/acquisition/optimizer.py
```diff
@@ -98,32 +100,54 @@ def automatic_optimizer_selector(
...
+) -> TensorType:
     """
     An :const:`AcquisitionOptimizer` for :class:'DiscreteSearchSpace' spaces and
     batches of size of 1.
 
+    This function also support the maximization of vectorized target functions by evaluating
+    the points in the search space for each of the individual V functions making up target_func.
+
     :param space: The space of points over which to search, for points with shape [D].
     :param target_func: The function to maximise, with input shape [..., 1, D] and output shape
```
hstojic: parameter description needs to be updated, right? same for the cont and random function below

trieste/acquisition/optimizer.py
```diff
@@ -98,32 +100,54 @@ def automatic_optimizer_selector(
...
+    space: DiscreteSearchSpace,
+    target_func: Union[AcquisitionFunction, Tuple[AcquisitionFunction, int]],
+) -> TensorType:
     """
     An :const:`AcquisitionOptimizer` for :class:'DiscreteSearchSpace' spaces and
     batches of size of 1.
 
+    This function also support the maximization of vectorized target functions by evaluating
+    the points in the search space for each of the individual V functions making up target_func.
+
```
hstojic, line 125: it would be great to add a brief sentence on when are these vectorized target functions used

trieste/acquisition/optimizer.py
```diff
@@ -98,32 +100,54 @@ def automatic_optimizer_selector(
...
     :param space: The space of points over which to search, for points with shape [D].
     :param target_func: The function to maximise, with input shape [..., 1, D] and output shape
             [..., 1].
     :return: The **one** point in ``space`` that maximises ``target_func``, with shape [1, D].
     """
-    target_func_values = target_func(space.points[:, None, :])
+
```
hstojic, line 131: probably it would be useful to check before this that target_func is either an instance of AcquisitionFunction or a tuple of AcquisitionFunction and int

trieste/acquisition/optimizer.py
```diff
@@ -98,32 +100,54 @@ def automatic_optimizer_selector(
...
 
 
-def optimize_discrete(space: DiscreteSearchSpace, target_func: AcquisitionFunction) -> TensorType:
+def optimize_discrete(
+    space: DiscreteSearchSpace,
+    target_func: Union[AcquisitionFunction, Tuple[AcquisitionFunction, int]],
+) -> TensorType:
     """
     An :const:`AcquisitionOptimizer` for :class:'DiscreteSearchSpace' spaces and
     batches of size of 1.
```
hstojic: "batches of size 1" should probably be corrected in the light of the following paragraph

trieste/acquisition/optimizer.py
```diff
@@ -98,32 +100,54 @@ def automatic_optimizer_selector(
...
 
+    This function also support the maximization of vectorized target functions by evaluating
+    the points in the search space for each of the individual V functions making up target_func.
+
     :param space: The space of points over which to search, for points with shape [D].
     :param target_func: The function to maximise, with input shape [..., 1, D] and output shape
             [..., 1].
     :return: The **one** point in ``space`` that maximises ``target_func``, with shape [1, D].
```
hstojic: [V,D] right?

trieste/acquisition/optimizer.py
```diff
@@ -177,7 +201,8 @@ def generate_continuous_optimizer(
         raise ValueError(f"num_recovery_runs must be zero or greater, got {num_recovery_runs}")
 
     def optimize_continuous(
-        space: Box | TaggedProductSearchSpace, target_func: AcquisitionFunction
+        space: Box | TaggedProductSearchSpace,
+        target_func: Union[AcquisitionFunction, Tuple[AcquisitionFunction, int]],
     ) -> TensorType:
         """
         A gradient-based :const:`AcquisitionOptimizer` for :class:'Box'
```
hstojic, line 213: "batches of size of 1" need an update?

trieste/acquisition/optimizer.py
```diff
@@ -261,22 +345,39 @@ def _perform_parallel_continuous_optimization(
...
+        output shape [..., V].
     :param space: The original search space.
-    :param starting_points: The points at which to begin our optimizations. The
-        leading dimension of `starting_points` controls the number of individual
-        optimization runs.
+    :param starting_points: The points at which to begin our optimizations of shape
+        [num_optimization_runs, V, D]. The leading dimension of
+        `starting_points` controls the number of individual optimization runs
+        for each of the V target functions.
     :param optimizer_args: Keyword arguments to pass to the Scipy optimizer.
```
hstojic, line 363: what if somebody passes a different method, does it overwrites bfgs one set in `ScipyLbfgsBGreenlet` - perhaps worth pointing out method, jac and bounds cannot/shouldn't be changed

trieste/acquisition/optimizer.py
```diff
@@ -261,22 +345,39 @@ def _perform_parallel_continuous_optimization(
...
+        and the maximiser found my each of our optimziations.
     """
 
     tf_dtype = starting_points.dtype  # type for communication with Trieste
 
-    num_optimization_runs = tf.shape(starting_points)[0].numpy()
+    num_optimization_runs_per_function = tf.shape(starting_points)[0].numpy()
+
+    V = tf.shape(starting_points)[1].numpy()  # vectorized batch size
+    D = tf.shape(starting_points)[-1].numpy()  # search space dimension
```
hstojic, line 373: -1 would assume rank 2, what about stating 2:?

trieste/acquisition/optimizer.py
```diff
@@ -429,6 +552,39 @@ def target_func_with_vectorized_inputs(
...
+def batchify_vectorize(
+    batch_size_one_optimizer: AcquisitionOptimizer[SP],
+    batch_size: int,
+) -> AcquisitionOptimizer[SP]:
+    """
+    A wrapper around our :const:`AcquisitionOptimizer`s. This class wraps a
+    :const:`AcquisitionOptimizer` to allow it to optimize batch acquisition functions.
+
+    Unlike :meth:`batchify_joint`, :meth:`batchify_vectorize` is suitable
+    for :class:`VectorizedAcquisitionFunction`s who's individual batch element can be
```
uri-granta: FYI the trailing `s` after the inlined `VectorizedAcquisitionFunction` is causing the docs failure (see https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-markup-recognition-rules point 7). You could probably escape it as `\s`, but I think this is clearer:

```suggestion
    for a :class:`VectorizedAcquisitionFunction` whose individual batch element can be
```

trieste/acquisition/function/function.py
henrymoss: nope. We want points that have low mean and high variance?

trieste/acquisition/interface.py
henrymoss: done

trieste/acquisition/rule.py
henrymoss: Ive not added functionality for async BO

trieste/acquisition/function/function.py
henrymoss: just assuming the the batch dim is the penultimate one. This is assumed throughout our code base. Arbitrary leading dims would be fine :)

trieste/acquisition/rule.py
henrymoss: Not in this PR

trieste/acquisition/optimizer.py
henrymoss: good point

uri-granta (approved):
Looks good!

trieste/acquisition/optimizer.py
```diff
@@ -429,6 +552,39 @@ def target_func_with_vectorized_inputs(
...
 
+def batchify_vectorize(
+    batch_size_one_optimizer: AcquisitionOptimizer[SP],
+    batch_size: int,
+) -> AcquisitionOptimizer[SP]:
+    """
+    A wrapper around our :const:`AcquisitionOptimizer`s. This class wraps a
+    :const:`AcquisitionOptimizer` to allow it to optimize batch acquisition functions.
+
+    Unlike :meth:`batchify_joint`, :meth:`batchify_vectorize` is suitable
```
hstojic: I think meth should be func in case of standalone functions
```suggestion
    Unlike :func:`batchify_joint`, :func:`batchify_vectorize` is suitable
```

trieste/acquisition/optimizer.py
```diff
@@ -180,14 +185,14 @@ def generate_continuous_optimizer(
...
-        can be passed.
+        can be passed. Note that method, jac and bounds cannot/should not be changed.
     :return: The acquisition optimizer.
     """
     if num_initial_samples <= 0:
         raise ValueError(f"num_initial_samples must be positive, got {num_initial_samples}")
 
     if num_optimization_runs < 0:
-        raise ValueError(f"num_initial_samples must be positive, got {num_initial_samples}")
+        raise ValueError(f"num_optimization_runs must be positive, got {num_initial_samples}")
```
```suggestion
        raise ValueError(f"num_optimization_runs must be positive, got {num_optimization_runs}")
```

hstojic (approved):
looks like its in a good shape (with the caveat that I didn't manage to check optimizer tests), just some minor comments 

on a sidenote, `optimize_random` is identical to `optimize_discrete`, barring a single line - we could reduce the code by adding an argument `random` in `optimize_discrete` and use it in both places - I guess you would defer it to a separate PR?

PR #461, opened 2021-12-21
removed equal rank check in Dataset

hstojic:
Dataset currently checks that `query_points` and `observations` have the same rank, this shouldn't be enforced I think - e.g. you could have images as data that could have rank > 2 but observations could still be rank = 2. 

I have modified the check such that only leading dimension is checked for inequality (number of observations), this will make the Dataset more flexible while still performing some checks. Tests had to be modified as well.

### DIFF
--- a/tests/unit/test_data.py
+++ b/tests/unit/test_data.py
@@ -43,8 +43,7 @@ def test_dataset_raises_for_zero_dimensional_data(
     [
         ((1,), (2,)),
         ((2,), (1,)),
-        ((5, 6), (5, 4)),
-        ((5, 6), (4, 6)),
+        ((5, 6), (4,)),
         ((5, 6), (4, 4)),
     ],
 )
@@ -64,28 +63,11 @@ def test_dataset_raises_for_different_leading_shapes(
 @pytest.mark.parametrize(
     "query_points_shape, observations_shape",
     [
-        ((1, 2), (1,)),
-        ((1, 2), (1, 2, 3)),
-    ],
-)
-def test_dataset_raises_for_different_ranks(
-    query_points_shape: ShapeLike, observations_shape: ShapeLike
-) -> None:
-    query_points = tf.zeros(query_points_shape)
-    observations = tf.ones(observations_shape)
-
-    with pytest.raises(ValueError):
-        Dataset(query_points, observations)
-
-
-@pytest.mark.parametrize(
-    "query_points_shape, observations_shape",
-    [
         ((), ()),
         ((), (10,)),
         ((10,), (10,)),
         ((1, 2), (1,)),
-        ((1, 2), (1, 2, 3)),
+        ((1,), (1, 2)),
     ],
 )
 def test_dataset_raises_for_invalid_ranks(
--- a/tests/unit/test_observer.py
+++ b/tests/unit/test_observer.py
@@ -70,10 +70,8 @@ def test_filter_finite(query_points: tf.Tensor, expected: Dataset) -> None:
 @pytest.mark.parametrize(
     "qp_shape, obs_shape",
     [
-        ([3, 4], [3, 2]),  # observations not N x 1
-        ([3, 4], [4, 1]),  # different leading dims
-        ([3], [3, 1]),  # query_points missing a dimension
-        ([3, 4, 2], [3, 1]),  # query_points have too many dimensions
+        ([3, 4], [3, 2]),
+        ([3, 4], [3,]),
     ],
 )
 def test_filter_finite_raises_for_invalid_shapes(qp_shape: ShapeLike, obs_shape: ShapeLike) -> None:
--- a/trieste/data.py
+++ b/trieste/data.py
@@ -1,4 +1,4 @@
-# Copyright 2020 The Trieste Contributors
+# Copyright 2021 The Trieste Contributors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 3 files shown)


tests/unit/test_observer.py
```diff
@@ -70,10 +70,8 @@ def test_filter_finite(query_points: tf.Tensor, expected: Dataset) -> None:
 @pytest.mark.parametrize(
     "qp_shape, obs_shape",
     [
-        ([3, 4], [3, 2]),  # observations not N x 1
-        ([3, 4], [4, 1]),  # different leading dims
-        ([3], [3, 1]),  # query_points missing a dimension
-        ([3, 4, 2], [3, 1]),  # query_points have too many dimensions
```
hstojic, line 76: these were actually testing Dataset, not filter_finite, so removed them...

tests/unit/test_data.py
```diff
@@ -61,31 +60,14 @@ def test_dataset_raises_for_different_leading_shapes(
...
 
 
-@pytest.mark.parametrize(
-    "query_points_shape, observations_shape",
-    [
-        ((1, 2), (1,)),
-        ((1, 2), (1, 2, 3)),
-    ],
-)
-def test_dataset_raises_for_different_ranks(
```
uri-granta, line 71: Maybe update test_dataset_concatenation with an example that has different ranks (but the same leading dimension)?

uri-granta (approved):
Sounds reasonable to me

vpicheny:
Generally this goes in the right direction of more flexibility, but I wonder if we're doing things right? I have two concerns.

I like the image example - or actually exampleS, because I can see at least two canonical ones:
1- input is an image and output is a scalar or vector: observer is a mapping from  `[P x P]` to `[L]`
2- input is a vector and output is an image: observer is a mapping from `[D]` to `[P x P]`
(D = input dim, L = output dim, P = pixel)

Then a dataset would look like:
1- `X =  [N, P, P]`; `Y = [N, L]`
2- `X = [N, D]`; `Y = [N, P, P]`.

Note that `[N, P, P]` (as tested in this PR) is not a completely obvious convention (one could expect `[P, P, N]`). More importantly, in  `__post_init__` we check only the first dimension, but what if we have leading dimensions (e.g. [..., N, P, P])? Shouldn't we check that all leading dimensions are correct (not sure it's actually possible)?

My second concern is more general (but maybe I'm just not sufficiently aware of the code): until now there's no way to give datasets that are incompatible with `gpflow` models. Once we remove the current dataset check (which again I think is desirable), are we sure that things won't break silently somewhere?

hstojic:
Thanks @vpicheny. Very useful comments!

> Generally this goes in the right direction of more flexibility, but I wonder if we're doing things right? I have two concerns.
> 
> I like the image example - or actually exampleS, because I can see at least two canonical ones: 1- input is an image and output is a scalar or vector: observer is a mapping from `[P x P]` to `[L]` 2- input is a vector and output is an image: observer is a mapping from `[D]` to `[P x P]` (D = input dim, L = output dim, P = pixel)
> 
> Then a dataset would look like: 1- `X = `[N, P, P]`; `Y = [N, L]`2-`X = [N, D]`; `Y = `[N, P, P]`.

Your first case is what I had in mind, I thought about the second one as well but I don't know of any BO application like that (any reference in mind?) - acquisition functions would also need to be able to use PxP output and produce a single scalar score, right? do we have any that can do that?

> Note that `[N, P, P]` (as tested in this PR) is not a completely obvious convention (one could expect `[P, P, N]`). More importantly, in `__post_init__` we check only the first dimension, but what if we have leading dimensions (e.g. [..., N, P, P])? Shouldn't we check that all leading dimensions are correct (not sure it's actually possible)?

I would say its up to us to define whats the convention in trieste, we could state clearly that we expect first dimension to be N, no other leading dimensions ( I cannot imagine what they would be, perhaps batches? but we introduce these independently during training...) - I would say we take a practical approach and deal with those additional cases as they surface

> My second concern is more general (but maybe I'm just not sufficiently aware of the code): until now there's no way to give datasets that are incompatible with `gpflow` models. Once we remove the current dataset check (which again I think is desirable), are we sure that things won't break silently somewhere?

Tests should serve that purpose, they all pass at the moment, but that doesn't say that they are comprehensive enough to catch some of these breaks. We could also impose some additional checks on dataset shapes in gpflow wrappers if we are sure that gpflow cannot handle certain things. 

@uri-granta do you have any follow up thoughts on these comments?

uri-granta:
> @uri-granta do you have any follow up thoughts on these comments?

One question I should have asked originally is whether we have a concrete use case example we can write a test for? If not, there's an argument for postponing the change until we do have one.

Regarding how existing gpflow models cope with differently shaped Datasets, you could experiment by e.g. temporarily changing a notebook to confirm we get a sensible-looking error, and then maybe add a unit test to check for this?

hstojic:
> > @uri-granta do you have any follow up thoughts on these comments?
> 
> One question I should have asked originally is whether we have a concrete use case example we can write a test for? If not, there's an argument for postponing the change until we do have one.

I take your point, it would be good to have an integration test and pull it through a whole loop to see where things break, I suspect we would end up with having to make a lot of changes along the way. Personally I don't have a use-case in mind, but at the moment Dataset is preventing even experimenting with it, e.g. I wanted to allow keras models to have arbitrary inputs and outputs already now but couldn't. There is a case for making at least some changes now so we can add more flexibility in adding new code having these use cases in mind. Hence, I wouldn't postpone making at least this change.

> Regarding how existing gpflow models cope with differently shaped Datasets, you could experiment by e.g. temporarily changing a notebook to confirm we get a sensible-looking error, and then maybe add a unit test to check for this?

I'll check GPflow docs, perhaps we are overthinking this, users should know what they can use GPs for

vpicheny:
It would be good to get @henrymoss opinion on this - would it help the current ICE solution? even if the solution won't benefit from the change, would this have been a better way of dealing with sweeps?

Might also be good to get @vdutor  and / or @sebastianober 's opinions on this, since they've got experience with images and GPs.

hstojic:
it seems we should retire this feature until further notice, any objections @vpicheny?

hstojic:
ok, it seems this is a no go for now, when we get a use case we can think about this again - I'll close it for now

PR #463, opened 2021-12-31
Add better model typing checks to rules

uri-granta:
This is an attempt to improve our static type checking so that we can statically catch e.g. using models that don't implement covariance_between_points with GIBBON, or models that don't implement conditional_predict_f with Fantasizer. 

**Summary of changes**

1. Acquisition function builders were already generic in the model type, though this has now been marked covariant.
2. Rules are now also covariantly generic in the model type: e.g. `EfficientGlobalOptimization[Box, SupportsPredictJoint]` can only be passed models that support predict_joint, and can therefore use acquisition functions that require it.
3. BO and AstkTell loops can now statically check models if they are provided explicitly. Models provided via config (either ModelConfig or a dict) cannot be checked statically, so will could cause a runtime error.
4. I've started splitting `ProbabilisticModel` into a hierarchy of model mixins based on features: we currently have `SupportsPredictJoint`, `SupportsCovarianceBetweenPoints`, `TrainableModel`, `FastUpdateModel`, `SupportsGetKernel` and `SupportsGetObservationNoise`.
5. **I've had to turn `ProbablisticModel` and its mixin subclasses from ABCs into Protocols**. This is necessary since Python doesn't support intersection types (though [they're working on it](https://github.com/python/typing/issues/213)), meaning that we couldn't otherwise specify e.g. functions that require both `SupportsA` and `SupportsB`. Fortunately, Protocols with abstractmethods can be used just like ABCs to check that models implement all the required methods. The only side effect is that it would now be possible to use a model with a rule even if the model doesn't formally inherit from `ProbablisticModel`, just as long as it implements all the necessary methods (checked at compile time). I think this is fine.
6. `ModelStack` also needs to be generic in the type of models in the stack. Furthermore we define classes such as `TrainableModelStack` and `PredictJointModelStack` which are not only stacks of trainable and predict_joint models, but are also trainable and predict_joint models themselves.
7. `ReparametrizationSampler` and `TrajectorySampler` are likewise generic: for example BatchReparametrizationSampler only supports models with predict_joint.


trieste/bayesian_optimizer.py
```diff
@@ -206,8 +209,40 @@ def optimize(
...
+        fit_initial_model: bool = True,
+    ) -> OptimizationResult[S]:
+        ...
+
+    @overload
+    def optimize(
+        self,
+        num_steps: int,
+        datasets: Mapping[str, Dataset],
+        # there's no way to statically check config-based models
```
joelberkeley, line 240: `TypedDict`?

also, hi uri. This was a PR I considered adding back when. Glad to see it's happening

uri-granta: We may be able to do a bit better, but we'll always be limited since the type (and hence acquisition function compatibility) of a config-based model might only be inferable at runtime.

PS Feel free to leave lots more comments :-)

uri-granta:
Just spotted a worrying issue where the type checks seem to be passing incorrectly. Putting back into draft while I look into it.

trieste/bayesian_optimizer.py
```diff
@@ -156,15 +160,14 @@ class BayesianOptimizer(Generic[SP]):
...
-        self,
-        observer: Observer,
-        search_space: SP,
-    ):
+    def __init__(self, observer: Observer, search_space: SP):
         """
         :param observer: The observer of the objective function.
         :param search_space: The space over which to search. Must be a
             :class:`~trieste.space.SearchSpace`.
+        :param model_type: The model type required by the acquisition rules that will be used.
```
hstojic: not needed any more

trieste/bayesian_optimizer.py
hstojic: it would make sense to use typed dictionaries since they always have to have same arguments - would that make it possible to do type checks?

(we should probably use typed dicts anyway...)

thanks @joelberkeley

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -76,7 +76,11 @@ def __call__(self, x):
 
 num_steps = 15
 result, history = bo.optimize(
-    num_steps, initial_data, model, acquisition_rule
+    num_steps,
+    initial_data,
+    model,
+    acquisition_rule,
+    None,
```
hstojic: why do we need to pass None here?

trieste/models/gpflow/models.py
```diff
@@ -49,7 +55,32 @@
 )
 
 
-class GaussianProcessRegression(GPflowPredictor, TrainableProbabilisticModel, FastUpdateModel):
+class SupportsCovarianceBetweenPoints(SupportsPredictJoint):
```
hstojic: why here and not in the interfaces?

trieste/ask_tell_optimization.py
```diff
@@ -38,6 +39,9 @@
 SP = TypeVar("SP", bound=SearchSpace)
 """ Type variable bound to :class:`SearchSpace`. """
 
+M_contra = TypeVar("M_contra", bound=TrainableProbabilisticModel, contravariant=True)
```
hstojic, line 46: huh, no clue what is this for... can you recommend a good source to learn about these contravariants etc?

trieste/acquisition/rule.py
```diff
@@ -67,13 +70,15 @@ class AcquisitionRule(ABC, Generic[T_co, SP_contra]):
     (see e.g. :class:`TrustRegion`). Indeed, to use an :class:`AcquisitionRule` in the main
     :class:`~trieste.bayesian_optimizer.BayesianOptimizer` Bayesian optimization loop, the rule
     must return either a `TensorType` or `State`-ful `TensorType`.
+
+    An :class:`AcquisitionRule` can only be passed models of type :class:`M_contra`.
```
hstojic: this will be meaningful only to developers, can you add a brief explanation in more lay terms that ordinary users will understand?

trieste/acquisition/rule.py
```diff
@@ -67,13 +70,15 @@ class AcquisitionRule(ABC, Generic[T_co, SP_contra]):
...
     """
 
     @abstractmethod
     def acquire(
         self,
         search_space: SP_contra,
-        models: Mapping[str, ProbabilisticModel],
+        models: Mapping[str, M_contra],
```
hstojic, line 83: previous name was more descriptive, as a user it would be easier to understand - you can go to check the class etc, here we have additional step to figure out what is this M_contra

trieste/acquisition/rule.py
```diff
@@ -119,18 +124,41 @@ def acquire_single(
         )
 
 
-class EfficientGlobalOptimization(AcquisitionRule[TensorType, SP_contra]):
+class EfficientGlobalOptimization(AcquisitionRule[TensorType, SP_contra, M_contra]):
     """Implements the Efficient Global Optimization, or EGO, algorithm."""
```
hstojic, line 130: can we add some explanations about model types, SP_contra, M_contra...

trieste/models/interfaces.py
```diff
@@ -152,10 +144,94 @@ def optimize(self, dataset: Dataset) -> None:
         raise NotImplementedError
 
 
-class ModelStack(ProbabilisticModel):
+class SupportsPredictJoint(ProbabilisticModel):
```
hstojic: since most of these are meant to be used by acquisition functions in the BO loop should we make them children of `TrainableProbabilisticModel`?

trieste/models/interfaces.py
```diff
@@ -152,10 +144,94 @@ def optimize(self, dataset: Dataset) -> None:
...
+
+    def conditional_predict_y(
+        self, query_points: TensorType, additional_data: Dataset
+    ) -> tuple[TensorType, TensorType]:
+        """
+        Return the mean and variance of the independent marginal distributions at each point in
+        ``query_points`` for the observations, including noise contributions, given an additional
+        batch of (possibly fantasized) data.
+
+        Note that this is not supported by all models.
```
hstojic, line 240: we can drop this now, right?

trieste/models/interfaces.py
```diff
@@ -320,14 +368,63 @@ def optimize(self, dataset: Dataset) -> None:
...
+            the same order as they appear in :meth:`__init__`. If the wrapped models have predictive
+            distributions with event shapes [:math:`E_i`], the mean will have shape
+            [..., B, :math:`\sum_i E_i`], and the covariance shape
+            [..., :math:`\sum_i E_i`, B, B].
+        """
+        means, covs = zip(*[model.predict_joint(query_points) for model in self._models])
+        return tf.concat(means, axis=-1), tf.concat(covs, axis=-3)
+
+
+class TrainableSupportsPredictJoint(SupportsPredictJoint, TrainableProbabilisticModel):
```
hstojic: can we add a note for all of these empty classes for the PR in mypy on intersect development, once that is implemented we will be able to use intersect and all the classes directly where needed and remove these

hstojic (commented):
didn't check the tests but it looks better now

I think we can do a bit better in terms of readibility - introducing all these TypeVar's makes it more difficult to understand types and what should user provide as an input (e.g. M_contra is now more difficult to understand), more descriptive names and clarifications in docstrings will help here

I would also implement additional interfaces for `get_kernel` and `get_noise_observations` methods

joelberkeley:
@uri-granta not sure if you check slack, but I had some ideas about the model class hierarchy which might help simplify this. It would be easier to jump on a call than message

trieste/models/interfaces.py
hstojic: and perhaps raise an issue to remind us to keep an eye on this development?

trieste/models/gpflux/interface.py
```diff
@@ -20,11 +20,11 @@
 from gpflow.base import Module
 
 from ...types import TensorType
-from ..interfaces import ProbabilisticModel
+from ..interfaces import SupportsGetObservationNoise
 from ..optimizer import BatchOptimizer
 
 
-class GPfluxPredictor(ProbabilisticModel, ABC):
+class GPfluxPredictor(SupportsGetObservationNoise, ABC):
```
hstojic, line 27: predict_joint can be dropped now, right?

trieste/models/gpflow/sampler.py
```diff
@@ -240,16 +250,8 @@ def __init__(
         self._num_features = num_features  # m
         self._num_data = len(self._dataset.query_points)  # n
```
hstojic, line 263: perhaps we should add a runtime check that the model is instance of SupportsGetKernelObservationNoise, same as in acquisition functions?

trieste/models/gpflow/sampler.py
```diff
@@ -174,7 +178,13 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
         return mean[..., None, :, :] + tf.transpose(variance_contribution, new_order)
 
 
-class RandomFourierFeatureTrajectorySampler(TrajectorySampler):
+class SupportsGetKernelObservationNoise(SupportsGetKernel, SupportsGetObservationNoise, Protocol):
```
hstojic, line 187: I would count this as an interface and would move it to the interface file

trieste/models/gpflow/sampler.py
```diff
@@ -101,7 +105,7 @@ class BatchReparametrizationSampler(ReparametrizationSampler):
     form a continuous curve.
     """
 
-    def __init__(self, sample_size: int, model: ProbabilisticModel):
+    def __init__(self, sample_size: int, model: SupportsPredictJoint):
```
hstojic, line 108: perhaps we should add a runtime check that the model is instance of SupportsPredictJoint, same as in acquisition functions?

trieste/models/keras/interface.py
```diff
@@ -24,7 +24,7 @@
 from ..optimizer import KerasOptimizer
 
 
-class KerasPredictor(ProbabilisticModel, ABC):
+class KerasPredictor(ProbabilisticModel):
```
hstojic: predict_joint can be dropped now, right?

trieste/models/gpflow/sampler.py
uri-granta: The reason I put it here is to avoid polluting the interface with arbitrary combinations of features. The features themselves all belong there obviously, but the combinations are more to do with the classes that require them I think.

trieste/models/interfaces.py
```diff
@@ -84,17 +85,7 @@ def predict_y(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
...
-        Return the variance of observation noise.
-
-        Note that this is not supported by all models.
-
-        :return: The observation noise.
-        """
-        raise NotImplementedError(f"Model {self!r} does not provide scalar observation noise")
-
-    def reparam_sampler(self, num_samples: int) -> ReparametrizationSampler:
+    def reparam_sampler(self: T, num_samples: int) -> ReparametrizationSampler[T]:
```
hstojic, line 88: wht not separating these samplers into their own PM interfaces?

trieste/bayesian_optimizer.py
```diff
@@ -43,6 +44,9 @@
 SP = TypeVar("SP", bound=SearchSpace)
```
hstojic, line 44: more descriptive name, e.g. SearchSpaceType?
also some further explanation in the docstring for ordinary users, they will be a bit lost with this "type variable bound to"...

trieste/bayesian_optimizer.py
```diff
@@ -43,6 +44,9 @@
...
 """ Type variable bound to :class:`SearchSpace`. """
 
+M_contra = TypeVar("M_contra", bound=TrainableProbabilisticModel, contravariant=True)
```
hstojic, line 47: more descriptive name, e.g. TrainableProbabilisticModelType?
also some further explanation in the docstring for ordinary users, they will be a bit lost with this "type variable bound to"...

trieste/bayesian_optimizer.py
uri-granta: It's standard in Python to use initials for type variables: e.g. see https://www.python.org/dev/peps/pep-0008/#type-variable-names. Happy to discuss whether it makes sense to break this convention, but probably not as part of this PR?

trieste/bayesian_optimizer.py
uri-granta: (I think ordinary users won't even look at M_contra, just the docstring description of model_specs and the examples in the tutorials)

trieste/models/interfaces.py
uri-granta: PM? (also I think there's some useful refactoring work possible re samplers, but not as part of this PR)

trieste/acquisition/function/active_learning.py
```diff
@@ -26,12 +26,13 @@
 
 from ...data import Dataset
```
hstojic, line 27: `IntegratedVarianceReduction` acq function here requires conditional predictions and there is a method check that can be eliminated, model instance can be checked instead

hstojic (approved):
few smaller comments, but otherwise looks to be in a good shape

trieste/models/gpflux/interface.py
hstojic: ABC should be removed here no? (you did so for GPflowPredictor)

trieste/models/interfaces.py
hstojic: ProbabilisticModeling subclasses

trieste/models/gpflux/interface.py
uri-granta: Good spot. Might actually put ABC back in GPflowPredictor and KerasPredictor: it makes no practical difference (except to allow you to register a non-subclass as an "implementation") but makes it clearer that these are neither concrete model implementations nor protocols.

PR #466, opened 2022-01-10
Uri/fix mypy for trieste models

uri-granta:
Remove unnecessary model wrapper inheritance from tf.Module, which has the unfortunate side effect of restricting the type checking mypy can do for our model wrapper classes.

### DIFF
--- a/docs/notebooks/recovering_from_errors.pct.py
+++ b/docs/notebooks/recovering_from_errors.pct.py
@@ -76,7 +76,7 @@ bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
 num_steps = 15
 result, history = bo.optimize(
-    num_steps, initial_data, model, acquisition_rule
+    num_steps, initial_data, model, acquisition_rule, None
 ).astuple()
 
 # %% [markdown]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ strict_equality = true
 module = ["trieste.*", "tests.*"]
 disallow_any_generics = true
 disallow_untyped_defs = true
+disallow_subclassing_any = true
 
 [tool.pytest.ini_options]
 addopts = "--doctest-modules"
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -281,7 +281,7 @@ def _test_optimizer_finds_minimum(
         with tensorboard_writer(summary_writer):
 
             dataset = (
-                BayesianOptimizer(observer, search_space)
+                BayesianOptimizer(observer, search_space)  # type: ignore
                 .optimize(num_steps or 2, initial_data, model, acquisition_rule)
                 .try_get_final_dataset()
             )
--- a/tests/unit/models/gpflow/test_models.py
+++ b/tests/unit/models/gpflow/test_models.py
@@ -26,7 +26,7 @@ trieste model).
 from __future__ import annotations
 
 import unittest.mock
-from typing import Any
+from typing import Any, cast
 
 import gpflow
 import numpy as np
@@ -49,7 +49,7 @@ from tests.util.models.gpflow.models import (
 from tests.util.models.models import fnc_2sin_x_over_3, fnc_3x_plus_10
 from trieste.data import Dataset
 from trieste.logging import step_number, tensorboard_writer
-from trieste.models import TrajectorySampler
+from trieste.models import TrainableProbabilisticModel, TrajectorySampler
 from trieste.models.config import create_model
 from trieste.models.gpflow import (
     GaussianProcessRegression,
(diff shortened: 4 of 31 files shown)


docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -76,7 +76,7 @@ def __call__(self, x):
 
 num_steps = 15
 result, history = bo.optimize(
-    num_steps, initial_data, model, acquisition_rule
+    num_steps, initial_data, model, acquisition_rule, None
```
uri-granta, line 79: This is required to get a state-based return type that's consistent with the one on line 108. Also it's more explicit as we're using TrustRegion.

tests/integration/test_bayesian_optimization.py
```diff
@@ -281,7 +281,7 @@ def build_model(data: Dataset) -> GPflowPredictor:
         with tensorboard_writer(summary_writer):
 
             dataset = (
-                BayesianOptimizer(observer, search_space)
+                BayesianOptimizer(observer, search_space)  # type: ignore
```
uri-granta, line 284: I think this is due to the different types of acquisition rules (some with state some without) confusing mypy.

tests/unit/models/gpflow/test_models.py
```diff
@@ -763,14 +765,14 @@ def optimize(self, model: tf.Module, dataset: Dataset) -> None:
         vgp_matern_model(x_observed[:10], y_observed[:10]), optimizer=optimizer, use_natgrads=True
     )
 
-    old_num_trainable_params = len(model.trainable_variables)
+    old_num_trainable_params = len(model.model.trainable_variables)
```
uri-granta, line 768: We're no longer a tf.Module

tests/unit/models/gpflow/test_models.py
```diff
@@ -89,7 +89,9 @@ def test_gpflow_wrappers_update(gpflow_interface_factory: ModelFactoryType) -> N
 
     x_new = tf.concat([x, tf.constant([[10.0], [11.0]], dtype=gpflow.default_float())], 0)
     new_data = Dataset(x_new, fnc_3x_plus_10(x_new))
-    model.update(new_data)
+    # Would be nice if ModelFactoryType could return an intersection type of
+    # GPflowPredictor and TrainableProbabilisticModel but this isn't possible
+    cast(TrainableProbabilisticModel, model).update(new_data)
```
uri-granta, line 94: Might go back to this if I figure out how to handle intersection types for the model type checking PR.

tests/unit/models/gpflow/test_utils.py
```diff
@@ -37,50 +38,6 @@
...
-
-
-class _NestedModuleWithBijector(tf.Module):
-    def __init__(self) -> None:
-        super().__init__()
-        self.inst = _ModuleWithBijector()
-        self.bijector = tfp.bijectors.Sigmoid(0.0, 1.0)
-
-
-def test_module_deepcopy_is_copyable() -> None:
```
hstojic, line 55: we should keep a test that checks if the model can be deepcopied, right?

hstojic (commented):
seems ok to me - but it seems prudent to keep some test that we can do a deepcopy of our models?

tests/unit/models/gpflow/test_utils.py
uri-granta: This test is checking a sepcific issue with copying tf.Modules, not whether models are deepcopyable. The test we want to keep is `test_gaussian_process_deep_copyable`.

uri-granta:
> seems ok to me - but it seems prudent to keep some test that we can do a deepcopy of our models?

We have individual tests for the model types (test_gaussian_process_deep_copyable, test_gpflux_predictor_deepcopy_raises_not_implemented, test_keras_predictor_deepcopy_raises_not_implemented). The test removed here was checking a specific issue with tf.Modules not with models in general.

pyproject.toml
```diff
@@ -29,6 +29,7 @@ strict_equality = true
 module = ["trieste.*", "tests.*"]
 disallow_any_generics = true
 disallow_untyped_defs = true
+disallow_subclassing_any = true
```
hstojic, line 32: what does this do?

tests/unit/models/gpflow/test_models.py
```diff
@@ -616,7 +618,7 @@ def test_sparse_variational_model_attribute() -> None:
 
 
 def test_sparse_variational_model_num_data_mixin_supports_subclasses() -> None:
-    class SVGPSubclass(SVGP):
+    class SVGPSubclass(SVGP):  # type: ignore[misc]
```
hstojic, line 621: this is now required bcs of new mypy setting in the toml file?

tests/unit/models/gpflow/test_utils.py
hstojic: ah, right, there are more deepcopy tests below

pyproject.toml
uri-granta: See https://mypy.readthedocs.io/en/stable/command_line.html#disallow-dynamic-typing:

> This flag reports an error whenever a class subclasses a value of type Any. This may occur when the base class is imported from a module that doesn’t exist (when using --ignore-missing-imports) or is ignored due to --follow-imports=skip or a # type: ignore comment on the import statement.
> 
> Since the module is silenced, the imported class is given a type of Any. By default mypy will assume that the subclass correctly inherited the base class even though that may not actually be the case. This flag makes mypy raise an error instead.

tests/unit/models/gpflow/test_models.py
uri-granta: Yes. SVGP has no type hints so as far as mypy is concerned SVGPSubclass is just `Any`. The type ignore is us telling it that we're ok with that.

pyproject.toml
hstojic: so bcs tf.Module was Any mypy would report it as an error?

hstojic (approved):
looks good

pyproject.toml
uri-granta: That's right (and I checked that it did)

PR #467, opened 2022-01-11
support for keras models part 4 - trajectory sampler

hstojic:
support for neural nets with Keras comes in five parts:

1. interface (done)
2. utilities (done)
4. trajectory sampler - this PR
5. integration tests and notebook


trieste/models/gpflow/sampler.py
```diff
@@ -220,8 +220,6 @@ def __init__(
         :param sample_size: The desired number of samples.
         :param model: The model to sample from.
         :param dataset: The data from the observer. Must be populated.
-        :sample_min_value: If True then sample from the minimum value of the function,
-            else sample the function's minimiser.
         :param num_features: The number of features used to approximate the kernel. We use a default
```
uri-granta: (I'm disappointed this didn't give a docs build warning. I'll jot down a note to investigate.)

trieste/models/interfaces.py
```diff
@@ -475,3 +475,49 @@ def conditional_predict_y(
         raise NotImplementedError(
             f"Model {self!r} does not support predicting observations, just the latent function"
         )
+
+
+class EnsembleModel(ProbabilisticModel):
```
uri-granta: If my PR goes in first, then this will want to be marked as a Protocol

```suggestion
class EnsembleModel(ProbabilisticMode, Protocoll):
```

trieste/models/keras/models.py
```diff
@@ -107,6 +108,11 @@ def __init__(
         self._model = model
         self._bootstrap = bootstrap
 
+        self._indices = tf.Variable(
+            0, name="sampling_indices", dtype=tf.int32, shape=tf.TensorShape(None)
+        )
+        # self._resample_indices(5)
```
uri-granta: ?

trieste/models/keras/sampler.py
```diff
@@ -0,0 +1,60 @@
...
+
+from __future__ import annotations
+
+from typing import cast
+
+from ...types import TensorType
+from ..interfaces import EnsembleModel, TrajectoryFunction, TrajectorySampler
+
+
+class EnsembleTrajectorySampler(TrajectorySampler):
```
uri-granta: If my PR goes in first, this will need to marked as generic:

```suggestion
class EnsembleTrajectorySampler(TrajectorySampler[EnsembleModel]):
```

trieste/models/keras/sampler.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_trajectory(self) -> TrajectoryFunction:
+        """
+        Generate an approximate function draw (trajectory) by randomly choosing a network from
+        the ensemble and using its predicted means as a trajectory.
+
+        :return: A trajectory function representing an approximate trajectory from the ensemble
+            model, taking an input of shape `[N, D]` and returning shape `[N, 1]`
+        """
+        network_index = cast(EnsembleModel, self._model).sample_index(1)[0]
```
uri-granta: (and if you do, you won't need the cast any more, here or below)
```suggestion
        network_index = self._model.sample_index(1)[0]
```

uri-granta (commented):
Looks good!

hstojic:
Thanks @uri-granta, addressed your comments - I'm sure yours will go in first :)

trieste/models/keras/sampler.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def get_trajectory(self) -> TrajectoryFunction:
+        """
+        Generate an approximate function draw (trajectory) by randomly choosing a network from
+        the ensemble and using its predicted means as a trajectory.
+
+        :return: A trajectory function representing an approximate trajectory from the ensemble
+            model, taking an input of shape `[N, D]` and returning shape `[N, 1]`
+        """
+        network_index = self._model.sample_index(1)[0]
```
hstojic, line 61: potential limitation here is the number of networks in the ensemble, thats giving us the number of approximations available, thoughts?

an alternative might be to sample from the network's output distribution rather than use the mean, it would be more costly, but not as much as sampling from a GP, so perhaps good enough?

uri-granta (approved):
Looks good! (apart from the broken tests and formatting, which I've hopefully fixed)

trieste/models/keras/sampler.py
```diff
@@ -0,0 +1,116 @@
...
+        :return: The new resampled trajectory function.
+        """
+        tf.debugging.Assert(isinstance(trajectory, ensemble_trajectory), [])
+        trajectory.resample()  # type: ignore
+        return trajectory
+
+
+class ensemble_trajectory(TrajectoryFunctionClass):
+    """
+    Generate an approximate function draw (trajectory) by randomly choosing a network from
```
hstojic, line 83: potential limitation here is the number of networks in the ensemble, thats giving us the number of approximations available

an alternative might be to sample from the network's output distribution rather than use the mean, it would be more costly, but not as much as sampling from a GP, so perhaps good enough? or leave it as an option?

@henrymoss and @vpicheny any thoughts?

vpicheny: I think that we should stick to the simple solution, which would be to check that you have more models than query points and throw an error if not?

The other case seems interesting but much more on the research side: at least for now, we have no evidence that this is a good idea and it's not that natural, so maybe we don't want to support it in the short term?

hstojic: I think you mean more than batch size not query points?
not sure how we can check this here, perhaps if I allow batch variant (I restrict it below but actually nothing prevents us to return a batch directly here...) I can test that, but at the moment nothing is using this in a batch mode - e.g. `ThompsonSamplerFromTrajectory` is just taking a single sample... perhaps we should have a `size` argument in `get_trajectory`?

good enough for the second point, I'll test that one first...

henrymoss: So, I have a draft where we pass a` num_samples` arg to `get_trajectory` as this is important when using our parallel acq optimiser to efficiently optimize multiple trajectories in parallel.  I'm not ready to merge this one yet. But feel free to add in this argument (making the default value=1 and making my RFF sampler raise an error for now when >1) ?

hstojic: ah, great to hear that, I'll merge this in now and revise it after you are done with yours

PR #468, opened 2022-01-11
support for keras models part 5 - integration tests

hstojic:
support for neural nets with Keras comes in five parts:

1. interface (done)
2. utilities (done)
3. model and architectures (deep ensembles only - done)
4. trajectory sampler (under review but independent)
5. integration tests and notebook - this PR

(also simplified BO integration tests)


docs/notebooks/util/plotting_plotly.py
```diff
@@ -160,6 +161,47 @@ def plot_gp_plotly(
     return fig
 
 
+def plot_de_plotly(
```
vpicheny: Why do you need a specific function for DE? you should be able to use the gp one?

If not, docstring needs updating

hstojic: gp model one uses predict_f method, that's why...

while we are at it, this whole plotting facility needs a rehaul, its poorly documented and can be made more reusable between the models - we could make it its own package? not a high priority thing, but nice visualization tools would be a plus for users...

uri-granta: I'd be wary of trying to make the visualisation functions and officially published thing (either part of trieste or a separate package). Making generic datavis tools is very tricky and a massive timesink. That said, I agree we should rehaul our notebook helper utilities.

docs/notebooks/quickrun/deep_ensembles.yaml
```diff
@@ -0,0 +1,4 @@
+replace:
+  - { from: "num_steps = \\d+", to: "num_steps = 2" }
+  - { from: "grid_size = \\d+", to: "grid_size = 10" }
+  - { from: "num_samples = \\d+", to: "grid_size = 1" }
```
uri-granta: ?

```suggestion
  - { from: "num_samples = \\d+", to: "num_samples = 1" }
```

trieste/models/keras/models.py
```diff
@@ -56,10 +56,9 @@ class DeepEnsemble(KerasPredictor, TrainableProbabilisticModel):
     can be supplied, as long as it has a Gaussian distribution as a final layer and follows the
     :class:`~trieste.models.keras.KerasEnsembleNetwork` interface.
 
-    Note that setting up the model with dictionary configs is not currently possible.
-
-    When using the model we recommend to set `tf.keras.backend.set_floatx(tf.float64)` for
-    alignment with the Trieste toolbox.
+    Note that currently we do not support setting up the model with dictionary configs and saving
+    the model during Bayesian optimization loop (``track_state`` argument in
+    :meth:`~trieste.bayesian_optimizer.BayesianOptimizer.optimize` method should be set to `False`).
```
uri-granta, line 61: Maybe also update the deepcopy methods for KerasPredictor and GPfluxPredictor so that they mention running with track_state=False?

tests/integration/test_bayesian_optimization.py
```diff
@@ -217,11 +221,75 @@ def test_optimizer_with_vgp_model(use_natgrads: bool) -> None:
...
+
+
+@random_seed
+@pytest.mark.slow
+@pytest.mark.parametrize(
+    "num_steps, acquisition_rule",
+    [
+        (80, EfficientGlobalOptimization()),
+        (30, DiscreteThompsonSampling(500, 3)),
+        # (10, DiscreteThompsonSampling(500, 3, thompson_sampler=ThompsonSamplerFromTrajectory())),
```
uri-granta, line 277: ?

tests/integration/test_bayesian_optimization.py
```diff
@@ -232,46 +300,103 @@ def _test_optimizer_finds_minimum(
...
+        rtol_level = 0.005
+    else:
+        search_space = SIMPLE_QUADRATIC_SEARCH_SPACE
+        minimizers = SIMPLE_QUADRATIC_MINIMIZER
+        minima = SIMPLE_QUADRATIC_MINIMUM
+        rtol_level = 0.05
+
+    if model_type in ["GPR", "VGP", "SVGP"]:
+
+        def build_model(data: Dataset) -> GPflowPredictor:  # type: ignore
```
uri-granta, line 336: Do talk to me if you hit typing issues after you merge in my recent PR.

tests/integration/models/keras/test_predictions.py
```diff
@@ -0,0 +1,51 @@
...
+        keras_ensemble,
+        KerasOptimizer(optimizer, negative_log_likelihood, fit_args),
+        False,
+    )
+    model.optimize(example_data)
+
+    predicted_means, _ = model.predict(example_data.query_points)
+    mad = tf.reduce_mean(tf.abs(predicted_means - example_data.observations))
+
+    assert mad < 2
```
uri-granta: (what does mad stand for, and where does the 2 come from?)

uri-granta (commented):
Looks good overall (apart from the notebook which I haven't looked at yet, but will do next)

tests/integration/test_bayesian_optimization.py
hstojic: this waits for the keras trajectory sampling PR

tests/integration/models/keras/test_predictions.py
hstojic: mean absolute deviation, and 2 is just a precision that seemed to be good enough

docs/notebooks/util/plotting_plotly.py
hstojic: ok, I have refactored these particular functions now, this triggered some changes in the notebooks, you can probably ignore those

tests/integration/test_bayesian_optimization.py
hstojic: all seems good after your merge - I didn't introduce those new build functions now, better left for a separate PR and doing it properly

docs/notebooks/deep_ensembles.pct.py
```diff
@@ -126,40 +125,29 @@ def build_cubic_model(data: Dataset) -> DeepEnsemble:
...
-plt.fill_between(
-    test_points,
-    tf.squeeze(y_hat_minus_3sd),
-    tf.squeeze(y_hat_plus_3sd),
-    color="gray",
-    alpha=0.5,
-    label="$\mu -/+ 3SD$",
-)
+plt.plot(test_points, y_hat, color='gray', label='model $\mu$')
+plt.fill_between(test_points, tf.squeeze(y_hat_minus_3sd), tf.squeeze(y_hat_plus_3sd), color='gray', alpha=0.5, label='$\mu -/+ 3SD$')
```
uri-granta: How do you format the notebooks? If you just run `black` it should ignore them. If you run `task format` it should format them to width 80 (which looks better in the docs as there's no scrolling).

trieste/models/gpflux/interface.py
```diff
@@ -88,4 +88,12 @@ def get_observation_noise(self) -> TensorType:
         return noise_variance
 
     def __deepcopy__(self, memo: dict[int, object]) -> GPfluxPredictor:
-        raise NotImplementedError("`deepcopy` not yet supported for `GPfluxPredictor`")
+        raise NotImplementedError(
+            """
+            GPfluxPredictor does not support deepcopy at the moment. For this reason,
+            ``track_state`` argument in when calling
```
uri-granta: "in when calling"?

uri-granta (approved):

docs/notebooks/deep_ensembles.pct.py
uri-granta: (assuming you run this from the root directory)

hstojic: oops, yes, I usually run only black :)

PR #469, opened 2022-01-12
Victor/gpfluxmonitoring

vpicheny:
This is a very short PR that will enable monitoring the training of gpflux models. Incidentally this also seems to have a positive effect on training.

In short, with this PR we keep track of the number of epochs across the BO iterations, and set the parameter `initial_epoch` to this number whenever we call keras.fit.

This PR will be followed by another one illustrating monitoring gpflux model training with tensorboard.

### DIFF
--- a/trieste/models/gpflux/models.py
+++ b/trieste/models/gpflux/models.py
@@ -14,7 +14,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict
 
 import tensorflow as tf
 from gpflow.inducing_variables import InducingPoints
@@ -145,6 +145,11 @@ class DeepGaussianProcess(GPfluxPredictor, TrainableProbabilisticModel):
         :param dataset: The data with which to optimize the `model`.
         """
         fit_args = dict(self._fit_args)
+
+        # Tell optimizer how many epochs have been used before: the optimizer will "continue"
+        # optimization across multiple BO iterations rather than start fresh at each iteration.
+        # This allows us to monitor training across iterations.
+
         if "epochs" in fit_args:
             fit_args["epochs"] = fit_args["epochs"] + self.absolute_epochs
 
@@ -155,6 +160,7 @@ class DeepGaussianProcess(GPfluxPredictor, TrainableProbabilisticModel):
         )
 
         self.absolute_epochs = self.absolute_epochs + len(hist.history["loss"])
+
         # Reset lr in case there was an lr schedule: a schedule will have change the learning rate,
         # so that the next time we call `optimize` the starting learning rate would be different.
         # Therefore we make sure the learning rate is set back to its initial value.
--- a/trieste/models/gpflux/models.py
+++ b/trieste/models/gpflux/models.py
@@ -56,8 +56,8 @@ class DeepGaussianProcess(GPfluxPredictor, TrainableProbabilisticModel):
             https://keras.io/api/models/model_training_apis/#fit-method for a list of possible
             arguments.
         :param continuous_optimisation: if True (default), the optimizer will keep track of the
-        number of epochs across BO iterations and use this number as initial_epoch. This is
-        essential to allow monitoring of model training across BO iterations.
+            number of epochs across BO iterations and use this number as initial_epoch. This is
+            essential to allow monitoring of model training across BO iterations.
         """
         for layer in model.f_layers:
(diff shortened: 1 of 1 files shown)


trieste/models/gpflux/models.py
```diff
@@ -143,10 +144,17 @@ def optimize(self, dataset: Dataset) -> None:
         Optimize the model with the specified `dataset`.
         :param dataset: The data with which to optimize the `model`.
         """
-        self.model_keras.fit(
-            {"inputs": dataset.query_points, "targets": dataset.observations}, **self._fit_args
+        fit_args = self._fit_args
```
uri-granta: This will currently also override the entry in the original self._fit_args, which I don't think is what you want?

```suggestion
        fit_args = dict(self._fit_args)
```

trieste/models/gpflux/models.py
```diff
@@ -143,10 +144,17 @@ def optimize(self, dataset: Dataset) -> None:
...
+        if fit_args is not None and "epochs" in fit_args:
```
uri-granta: I know that that `self._fit_args` is described as `Optional[Dict[str, Any]] ` in DeepGaussianProcess but that isn't right: it's always a dictionary (otherwise **fit_args would fail). If you fix the type declaration you can simplify this:

```suggestion
        if "epochs" in fit_args:
```

trieste/models/gpflux/models.py
```diff
@@ -143,10 +144,17 @@ def optimize(self, dataset: Dataset) -> None:
...
+            fit_args["epochs"] = fit_args["epochs"] + self.absolute_epochs
+
+        hist = self.model_keras.fit(
+            {"inputs": dataset.query_points, "targets": dataset.observations},
+            **fit_args,
+            initial_epoch=self.absolute_epochs,
         )
 
+        self.absolute_epochs = self.absolute_epochs + len(hist.history["loss"])
```
uri-granta: maybe add a comment explaining what's happening here?

uri-granta (approved):
Looks good (bar some minor comments)

trieste/models/gpflux/models.py
vpicheny: not sure how I should change this, just remove optional?
`self._fit_args: Dict[str, Any] = {...`

trieste/models/gpflux/models.py
hstojic: this would make it a default behaviour, right?
I don't think we should do that, we should leave it as optional through an argument at init

vpicheny: why not? if anything, it improves performance...

hstojic: its a behaviour I would not expect and there is now way to switch it off, right?
we can leave it as a default but allow for turning it off, how does that sound?

vpicheny: sounds good

vpicheny: The only way I see to do that is with an extra boolean argument at init(). Is this OK? 
I'm not sure it's worth the extra API clutter.

vpicheny: @sebastianober any opinion?

hstojic: I think another argument in the init is fine, we don't have that many arguments in any case

trieste/models/gpflux/models.py
```diff
@@ -143,10 +144,23 @@ def optimize(self, dataset: Dataset) -> None:
...
-        self.model_keras.fit(
-            {"inputs": dataset.query_points, "targets": dataset.observations}, **self._fit_args
+        fit_args = dict(self._fit_args)
+
+        # Tell optimizer how many epochs have been used before: the optimizer will "continue"
+        # optimization across multiple BO iterations rather than start fresh at each iteration.
+        # This allows us to monitor training across iterations.
+
+        if "epochs" in fit_args:
+            fit_args["epochs"] = fit_args["epochs"] + self.absolute_epochs
```
hstojic: this would increase the number of epochs in each iteration, if you dont use any stopping rule in callbacks that could quickly become a long training regime, no?

vpicheny: actually, we always use a constant number of epochs. You start at `self.absolute_epochs` and pursue for `self._fit_args["epochs"]`.

Here `self._fit_args["epochs"]` never changes.

trieste/models/gpflux/models.py
vpicheny: done

PR #470, opened 2022-01-13
Improve ask-tell tensorboard logging

uri-granta:


vpicheny (commented):
will `ask` log automatically the rule? if not, we should probably make it happen, so that there is no difference between the standard optimizer and ask/tell?

uri-granta:
> Will ask automatically log the rule

Yes, the rules get logged as part of acquire, which is called by ask. The only thing I've just realised BO logs and AskTell doesn't is the best _new_ observation, but I'll push a change to add that.

vpicheny (approved):

PR #474, opened 2022-01-25
Fix MRO on Python 3.8+

### LINKED ISSUE #473: develop branch seems broken on Python 3.8
**Describe the bug**
Importing Trieste fails on Python 3.8 in clean venv

**To reproduce**
Following steps shall be executed separately for Python 3.7 and 3.8. `X` below stands for 7 or 8.
1. clone fresh `trieste` and `cd` into it
2. `python3.X -m venv .venv3X`
3. `source .venv3X/bin/activate`
4. `pip install -U pip`
5. `pip install .`
6. `python -c "import trieste"`

**Expected behaviour**
In Python 3.7 the above code works. In 3.8 it fails with an error: 
```
...
  File "/home/ubuntu/projects/trieste/secondmind_trieste/trieste/trieste/models/interfaces.py", line 253, in <module>
    class ModelStack(Generic[T], ProbabilisticModel):
  File "/usr/lib/python3.8/abc.py", line 85, in __new__
    cls = super().__new__(mcls, name, bases, namespace, **kwargs)
TypeError: Cannot create a consistent method resolution
order (MRO) for bases Generic, ProbabilisticModel
```

**System information**
 - OS: Ubuntu 20.04 LTS
 - Python version: 3.7 and 3.8
 - Trieste version: tip of develop

uri-granta:
The recent Protocol changes seem to have broken trieste on Python 3.8+. This fixes that.

The issue seems to be connected to MRO and the Protocol definition in typing_extensions (which we use since we're officially on 3.7). The rest of the changes are unrelated formatting changes (which should probably be caught by the format tests, I'll investigate).

### DIFF
--- a/trieste/models/interfaces.py
+++ b/trieste/models/interfaces.py
@@ -250,7 +250,7 @@ class FastUpdateModel(ProbabilisticModel, Protocol):
         )
 
 
-class ModelStack(Generic[T], ProbabilisticModel):
+class ModelStack(ProbabilisticModel, Generic[T]):
     r"""
     A :class:`ModelStack` is a wrapper around a number of :class:`ProbabilisticModel`\ s of type
     :class:`T`. It combines the outputs of each model for predictions and sampling.
--- a/trieste/models/keras/architectures.py
+++ b/trieste/models/keras/architectures.py
@@ -25,7 +25,6 @@ from typing import Any, Sequence, Union
 import numpy as np
 import tensorflow as tf
 import tensorflow_probability as tfp
-from tensorflow_probability.python.distributions import distribution as tfd
 
 from ...data import Dataset
 from .utils import get_tensor_spec_from_data
@@ -221,7 +220,7 @@ class GaussianNetwork(KerasEnsembleNetwork):
 
         distribution = dist_layer(
             self.flattened_output_shape,
-            tfd.Distribution.mean,
+            lambda s: s.mean(),
             name=self.output_layer_name,
         )(parameter_layer)
 
--- a/trieste/models/keras/architectures.py
+++ b/trieste/models/keras/architectures.py
@@ -25,6 +25,7 @@ from typing import Any, Sequence, Union
 import numpy as np
 import tensorflow as tf
 import tensorflow_probability as tfp
+from tensorflow_probability.python.distributions import distribution as tfd
 
 from ...data import Dataset
 from .utils import get_tensor_spec_from_data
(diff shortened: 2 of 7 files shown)


trieste/models/interfaces.py
```diff
@@ -250,7 +250,7 @@ def conditional_predict_y(
         )
 
 
-class ModelStack(Generic[T], ProbabilisticModel):
+class ModelStack(ProbabilisticModel, Generic[T]):
```
apaleyes, line 253: Is that it? Nice fix! Can you explain a bit why the order matters here?

trieste/models/keras/architectures.py
```diff
@@ -220,7 +221,7 @@ def _gen_output_layer(self, input_tensor: tf.Tensor) -> tf.Tensor:
 
         distribution = dist_layer(
             self.flattened_output_shape,
-            lambda s: s.mean(),
+            tfd.Distribution.mean,
```
apaleyes: that doesn't look like a formatting fix

apaleyes (commented):
Thanks for the quick fix!

About the official version support, quote from the README: "Trieste supports Python 3.7 onwards". I suppose we shall either remove "onwards" bit and replace it by "only". Or add a quick build action to run unit tests on 3.8 and maybe 3.9 too

trieste/models/keras/architectures.py
uri-granta: You're right. That was an unrelated fix. I'll revert it and check it in separately.

trieste/models/interfaces.py
uri-granta: No idea! But it seems to work (and doesn't break anything on 3.7). It's probably some weird interaction between Protocol and Generic.

uri-granta: Looks like Protocol (unlike ABC) inherits from Generic, meaning we need to put it first to ensure monotonicity ("a class always appears before its ancestor"). See https://en.wikipedia.org/wiki/C3_linearization for more details if you're interested.

apaleyes: Thanks for the reference! Apparently this wasn't the case in 3.7, hence no error. Also, worth noting that `Protocol` was moved from extensions to `typing` package in 3.8.

apaleyes (approved):

hstojic (approved):
looks good

trieste/models/interfaces.py
uri-granta: Yes. We're only using typing_extensions because we support 3.7.

uri-granta:
I'll create an issue for testing that trieste installs/runs with other versions (3.8, 3.9, 3.10). Though given that it's probably a fair amount of work, and will slow down testing (or releases, if we do it just for them), I'm not sure how highly we should prioritise it. Worth noting that we also only test with one version of tensorflow, and it's fairly normal to assume backwards compatibility of future releases.

PR #475, opened 2022-01-25
Super duper thompson sampler

henrymoss:
We can now do continuous Thompson sampling with Trieste!

By continuous, I mean optimizing sample draws (trajectories) using L-BFG-B rather than just choosing the best points over a discrete set (as currently performed by our `DiscreteThompsonSamplingRule`).

In order to avoid retracing I added some update/resample logic to our RFF trajectory sampler, inspired by our `AcquisitionFunctionBuilders`. I also changed the expected input shape of trajectories so that they behave more like acquisition functions and so can be optimized using our optimizers.

In this PR, we sequentially maximize trajectory samples using our `GreedyAcquisitionFuncton` logic. 

Future PRs will add (in order)

1. A Matheron's rule trajectory sampler for our VGP and SVGP models to allow continuous Thompson sampling with these models.
2. Added support for parallel continuous Thompson sampling, i.e maximising all the trajectories in parallel using our newish parallel acquisition optimiser.


trieste/models/interfaces.py
```diff
@@ -503,9 +530,38 @@ def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self._model!r})"
 
     @abstractmethod
-    def get_trajectory(self) -> TrajectoryFunction:
+    def get_negative_trajectory(self) -> TrajectoryFunction:
```
uri-granta: Maybe also mention here why we return the negative trajectory? (Also is it really cleaner to have to remember to negate all the trajectory implementations rather than negate it before we pass it to the optimizer?)

trieste/acquisition/sampler.py
```diff
@@ -222,12 +213,15 @@ def sample(self, model: ProbabilisticModel, sample_size: int, at: TensorType) ->
...
-            sampled_trajectory = trajectory_sampler.get_trajectory()
-            evaluated_trajectory = sampled_trajectory(at)  # [N, 1]
+            sampled_trajectory = trajectory_sampler.get_negative_trajectory()
+            expanded_at = tf.expand_dims(at, -2)  # [N, 1, D]
+            evaluated_trajectory = sampled_trajectory(expanded_at)  # [N, 1]
             if self._sample_min_value:
-                sample = tf.reduce_min(evaluated_trajectory, keepdims=True)  # [1, 1]
+                sample = -1.0 * tf.reduce_max(
+                    evaluated_trajectory, keepdims=True
+                )  # [1, 1] TODO say neg
```
uri-granta: (TODO)

trieste/models/gpflow/models.py
```diff
@@ -277,8 +286,17 @@ def trajectory_sampler(self) -> TrajectorySampler[GaussianProcessRegression]:
...
+
+        return RandomFourierFeatureTrajectorySampler(self, self._num_rff_features)
+
+    def get_internal_data(self) -> Dataset:
+        """
+        Return the model's training data.
+
+        :return: The model's training data.
+        """
         models_data = Dataset(self.model.data[0].value(), self.model.data[1].value())
```
uri-granta: (note sure there's anything gained from setting to a variable)

```suggestion
        return Dataset(self.model.data[0].value(), self.model.data[1].value())
```

trieste/models/gpflow/sampler.py
```diff
@@ -114,7 +115,7 @@ def __init__(self, sample_size: int, model: SupportsPredictJoint):
         super().__init__(sample_size, model)
         if not isinstance(model, SupportsPredictJoint):
             raise NotImplementedError(
-                f"RandomFourierFeatureTrajectorySampler only works with models that support "
+                f"BatchReparametrizationSampler only works with models that support "
```
uri-granta, line 118: Good spot!

trieste/models/gpflow/sampler.py
```diff
@@ -346,10 +353,119 @@ def get_trajectory(self) -> TrajectoryFunction:
             process, taking an input of shape `[N, D]` and returning shape `[N, 1]`
         """
 
-        theta_sample = self._theta_posterior.sample(1)  # [1, m]
+        data_dtype = self._model.get_internal_data().query_points.dtype
+
+        self._feature_functions = RandomFourierFeaturesCosine(
```
uri-granta, line 357: This will presumably break if we need to downgrade develop to gpflux 0.2.3 so that we support tf 2.4 (which looks plausible, at least in the short term, given the internal discussions that have kicked off)?

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -0,0 +1,94 @@
...
+    ) -> TrajectoryFunction:
+        """
+        :param model: The model.
+        :param dataset: The data from the observer (not used).
+        :param pending_points: The points already in the current batch (not used).
+        :return: A trajectory sampled from the model.
+        """
+
+        try:
+            self._trajectory_sampler = model.trajectory_sampler()
```
uri-granta, line 60: guess you don't fancy defining a `HasTrajectorySampler` protocol (and ideally also `HasReparamSampler`)?

tests/integration/test_bayesian_optimization.py
```diff
@@ -180,6 +181,13 @@ def OPTIMIZER_PARAMS() -> Tuple[
                     num_query_points=3,
                 ),
             ),
+            (
+                15,
+                EfficientGlobalOptimization(
+                    GreedyContinuousThompsonSampling(),
```
uri-granta, line 187: Have you checked that alltests pass (either running locally or kicking off a manual job on GitHub)?

uri-granta (commented):
Looks great! Just some minor comments.

trieste/models/gpflow/sampler.py
henrymoss: We would have to change our existing use of `Random...` if we need to downgrade anyway.

trieste/acquisition/function/continuous_thompson_sampling.py
henrymoss: Leaving it for another PR

tests/integration/test_bayesian_optimization.py
henrymoss: yeep

trieste/models/interfaces.py
henrymoss: Yep changed this back to non negative

uri-granta (approved):

Issue #347, opened 2021-09-13
Numpy-related type errors are hidden by the absence of numpy stubs


apaleyes:
**Describe the bug**
There is a bunch of type errors in the code base that mypy runs never reveal because Trieste is relying on numpy 1.19.x which isn't typed yet. If you run mypy with numpy >= 1.20, 46 new type errors surface.

Ideal fix would be to move to newer version of numpy. The errors will then fail the builds and force us to fix them. Unfortunately numpy 1.19.x seems to be forced by [Tensorflow](https://github.com/tensorflow/tensorflow/blob/07dd16b3a09e0cfa007d9694251dd6b2cbdf36ad/tensorflow/lite/tools/pip_package/setup_with_binary.py) and using one numpy version for type checks while using older one for everything else feels weird.

We could also consider installing numpy stubs separately via [numpy-stubs](https://github.com/numpy/numpy-stubs) or [data-science-types](https://github.com/predictive-analytics-lab/data-science-types), but both of these projects are archived, so it's probably best not to rely on them.

**To reproduce**
We will be using tox here.

1. Start with a new fresh clone of Trieste, make sure there is no `.tox` cache. Your venv should be running python3.7 and only have tox installed. This is how GitHub Action for type checking is setup.
2. Add numpy and matplotlib to tox requirements:
`echo -e "numpy>=1.20\nmatplotlib" >> common_build/types/requirements.txt`
3. run type checks with tox
`tox -e types`
4. 46 errors show up:  [mypy_errors.txt](https://github.com/secondmind-labs/trieste/files/7153689/mypy_errors.txt)

**Expected behaviour**
No errors

**System information**
 - OS: Ubuntu 18
 - Python version: 3.7.5
 - Trieste version: git clone

uri-granta:
Due to TF version restriction, it might be a little while before we upgrade. Looking at the errors there doesn't seem to be anything worrying. By far the most common error is using a TensorType result (e.g. from sample) as if it was necessarily a tf.Tensor (which it is in practice, but that's not what the type says). I've raised a PR to discuss a possible fix.

uri-granta:
Looking at this again, as we'll need to fix these typing issues at some point to upgrade tf and numpy.

[...]

PR #477, opened 2022-01-25
Use to_numpy on TensorType

uri-granta:
As noted in https://github.com/secondmind-labs/trieste/issues/347, we currently don't properly check numpy-related types since numpy 1.19 (required by TensorFlow 2.4 and 2.5) doesn't yet have type stubs . Looking at the typing issues caught in numpy 1.20, it seems most are the result of us using a TensorType result from e.g. a sample call as if it must be a tf.Tensor (which it will be in practice, but is not what the type says). We have two options on how to fix this:

1. Continue to return TensorType but call trieste.utils.to_numpy rather than calling .numpy()
2. Continue to accept TensorType for our inputs but only return tf.Tensors.

This PR shows what option 1 looks like in the code and in one of the notebooks. If it looks ok then I'll make it for the other notebooks too.

### DIFF
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -125,8 +125,10 @@ dataset = result.try_get_final_dataset()
 # We can now get the best point found by the optimizer. Note this isn't necessarily the point that was last evaluated.
 
 # %%
-query_points = dataset.query_points.numpy()
-observations = dataset.observations.numpy()
+from trieste.utils import to_numpy
+
+query_points = to_numpy(dataset.query_points)
+observations = to_numpy(dataset.observations)
 
 arg_min_idx = tf.squeeze(tf.argmin(observations, axis=0))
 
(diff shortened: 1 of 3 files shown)


hstojic:
why do we actually need anything other than `tf.Tensors`?

hstojic (commented):
I would have thought we have more of these?!
(we have some more in plotting utils in notebook folder...)

uri-granta:
> why do we actually need anything other than tf.Tensors?

Good question. I'll see if I can get the type checker to tell me which interfaces (if any) we're actually using with ndarrays rather than tf.Tensors. Also worth seeing what the others think.

> I would have thought we have more of these?!

Yes, there are lots more. I just want to get feedback before I make more changes!

hstojic:
> > why do we actually need anything other than tf.Tensors?
> 
> Good question. I'll see if I can get the type checker to tell me which interfaces (if any) we're actually using with ndarrays rather than tf.Tensors. Also worth seeing what the others think.

mostly we use `TensorType` with data, in these cases everything has to pass through `Dataset` so if we convert everything to a Tensor there we should be fine with `tf.Tensor` - one other usage that we probably have is as parameters/arguments, there it might be useful to allow both numpy and tensors

> > I would have thought we have more of these?!
> 
> Yes, there are lots more. I just want to get feedback before I make more changes!

Ok


Overall, to me switching to `tf.Tensor` might be the best solution, and after that your 2)

uri-granta:
> Overall, to me switching to `tf.Tensor` might be the best solution, and after that your 2)

PR #478, opened 2022-01-26
making gpflux builder consistent with gpflow builders

hstojic:
gpflux builder was not really consistent with new gpflow build functions (setting some defaults, moved to builders.py etc), this PR corrects that

along the way I made some other smaller changes
- updated integration tests to use this build function
- using KerasOptimizer instead of BatchOptimizer
- made some adjustments to Keras defaults in its build function

@sebastianober your input with the default values would be very welcome :)


trieste/models/gpflux/builders.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def build_vanilla_deep_gp(
+    data: Dataset,
+    num_layers: int = 2,
+    num_inducing_points: Optional[int] = None,
+    search_space: Optional[SearchSpace] = None,
+    inner_layer_sqrt_factor: float = 1e-5,
+    likelihood_variance: float = 1e-2,
+    trainable_likelihood: bool = True,
```
hstojic: @sebastianober we recently created similar build functions for gpflow models and we tried to give sensible default values for everything - so I'm making this function more consistent with those, e.g. using data to set the kernel variance and mean function and so on, please see builders.py file in gpflow, but all the values here would welcome your input

trieste/models/gpflux/models.py
```diff
@@ -24,9 +22,9 @@
 from ...data import Dataset
 from ...types import TensorType
 from ..interfaces import TrainableProbabilisticModel
-from ..optimizer import BatchOptimizer
+from ..optimizer import KerasOptimizer
 from .interface import GPfluxPredictor
-from .utils import sample_dgp
+from .sampler import sample_dgp
```
uri-granta, line 27: Not due to your change, but the samplers in sampler.py (previously utils.py) should really have docstrings.

tests/unit/models/keras/test_builders.py
```diff
@@ -12,53 +12,44 @@
...
-    assert isinstance(network_built.layers[0], tf.keras.layers.InputLayer)
-    assert len(network_built.layers[1:-2]) == num_hidden_layers
-    assert isinstance(network_built.layers[-1], tfp.layers.DistributionLambda)
+    assert keras_ensemble.ensemble_size == ensemble_size
+    assert len(keras_ensemble.model.layers) == num_hidden_layers * ensemble_size + 3 * ensemble_size
+    if independent_normal:
+        assert isinstance(keras_ensemble.model.layers[-1], tfp.layers.IndependentNormal)
+    else:
+        assert isinstance(keras_ensemble.model.layers[-1], tfp.layers.MultivariateNormalTriL)
+    if num_hidden_layers > 0:
```
uri-granta, line 52: his check shouldn't be necessary, right, given your previous assert about the number of layers?
```suggestion
```

uri-granta (approved):
Looks good architecturally!

tests/integration/models/gpflux/test_predictions.py
```diff
@@ -35,19 +36,28 @@ def test_dgp_model_close_to_actuals(depth: int, keras_float: None) -> None:
...
 
-    model, _ = trieste_deep_gaussian_process(
-        query_points=example_data.query_points,
-        depth=depth,
-        num_inducing=num_inducing,
-        learning_rate=0.01,
-        batch_size=batch_size,
-        epochs=epochs,
-        fix_noise=True,
+    def scheduler(epoch: int, lr: float) -> float:
```
vpicheny: I think there's a "reduce on plateau" strategy already implemented in Keras, should we use it instead of that one?

tests/integration/models/gpflux/test_predictions.py
```diff
@@ -65,31 +75,37 @@ def test_dgp_model_close_to_simple_implementation(depth: int, keras_float: None)
 
     example_data = hartmann_6_dataset(dataset_size)
 
+    # optimization settings
```
vpicheny: feels slightly annoying that we need to redefine an optimizer every time..

trieste/models/gpflux/models.py
```diff
@@ -77,14 +75,12 @@ def __init__(
 
         self.original_lr = self.optimizer.optimizer.lr.numpy()
 
-        if not self.optimizer.minimize_args:
-            self._fit_args: Dict[str, Any] = {
+        if not self.optimizer.fit_args:
+            self.optimizer.fit_args = {
```
vpicheny, line 86: Would be good to have a LR scheduler here? I'm actually unsure what optimizer is used by default, is it adam? I think adam + ReduceOnPlateau should be the default

tests/integration/models/gpflux/test_predictions.py
hstojic: true, its already in Keras... I thought @sebastianober had a reason for using this one

on a sidenote, in my experience with deep ensembles this was resulting in worse performance, but haven't done systematic experiments

tests/integration/models/gpflux/test_predictions.py
hstojic: yes, perhaps we should redo the default optimizer settings

trieste/models/gpflux/models.py
hstojic: its adam, re ReduceOnPlateau I would wait for @sebastianober opinion

tests/unit/models/keras/test_builders.py
hstojic: its needed after all as one of the inputs for this argument is 0, see line 28:
```
@pytest.mark.parametrize("num_hidden_layers", [0, 1, 3])
```

trieste/models/optimizer.py
```diff
@@ -154,23 +154,25 @@ def train_fn() -> None:
...
-
     fit_args: dict[str, Any] = field(default_factory=lambda: {})
     """
-    The keyword arguments to pass to the `fit` method of a :class:`~tf.keras.Model` instance.
+    The keyword arguments to pass to the ``fit`` method of a :class:`~tf.keras.Model` instance.
     See https://keras.io/api/models/model_training_apis/#fit-method for a list of possible
     arguments in the dictionary.
     """
 
+    loss: Optional[
```
uri-granta, line 168: You need to update the existing calls to KerasOptimizer in test.utils.models.keras.models and and test.unit.modles.keras.test_models (and maybe elsewhere too)

tests/integration/models/gpflux/test_predictions.py
sebastianober: I don't have much experience with the reduce on plateau callback, but it could well be a good idea here. The scheduler I have here is quite basic but from experience it works ok, but could certainly be improved

hstojic: ok, we will switch to Keras one after we get some experience with it

vpicheny: I think ReduceOnPlateau is @vdutor 's default

tests/integration/models/gpflux/test_predictions.py
sebastianober: It should be possible to create a utility function that does this for us?

trieste/models/gpflux/builders.py
```diff
@@ -0,0 +1,124 @@
...
+    :param num_inducing_points: Number of inducing points to use in each layer. If left unset
+        (default) it will be set to the number of query points in ``data``.
+    :param search_space: Search space for performing Bayesian optimization. Used for initialization
+        of inducing locations if ``num_inducing_points`` is larger than the amount of data.
+    :param inner_layer_sqrt_factor: A multiplicative factor used to rescale hidden layers. See
+        :class:`~gpflux.architectures.Config` for details.
+    :param likelihood_variance: Initial noise variance in the likelihood function. See
+        :class:`~gpflux.architectures.Config` for details.
+    :return: A :class:`~gpflux.models.DeepGP` model with sensible default settings.
+    """
```
sebastianober, line 100: Should we include a `raise` for invalid values?

tests/integration/models/gpflux/test_predictions.py
sebastianober: Ah I was not aware of this, I'll have to have a look!

sebastianober (approved):
Everything looks good on the GPflux side

trieste/models/gpflux/interface.py
```diff
@@ -21,25 +21,29 @@
...
+        :param optimizer: The optimizer wrapper containing the optimizer with which to train the
+            model and arguments for the wrapper and the optimizer. The optimizer must
+            be an instance of a :class:`~tf.optimizers.Optimizer`. Defaults to
+            :class:`~tf.optimizers.Adam` optimizer with default parameters.
         """
         super().__init__()
 
         if optimizer is None:
-            optimizer = BatchOptimizer(tf.optimizers.Adam())
+            optimizer = KerasOptimizer(tf.optimizers.Adam())
```
sebastianober:
```suggestion
            optimizer = KerasOptimizer(tf.optimizers.Adam(0.01))
```

trieste/models/gpflux/models.py
```diff
@@ -77,14 +75,12 @@ def __init__(
...
                 "verbose": 0,
                 "epochs": 100,
```
sebastianober:
```suggestion
                "epochs": 400,
```

tests/integration/models/gpflux/test_predictions.py
```diff
@@ -12,84 +12,33 @@
...
+    model = DeepGaussianProcess(dgp)
     model.optimize(example_data)
     predicted_means, _ = model.predict(example_data.query_points)
 
     np.testing.assert_allclose(predicted_means, example_data.observations, atol=0.2, rtol=0.2)
-
-
-@pytest.mark.slow
-@random_seed
-def test_dgp_model_close_to_simple_implementation(depth: int, keras_float: None) -> None:
```
sebastianober, line 59: Why was this one deleted?

tests/unit/models/gpflux/test_models.py
```diff
@@ -136,8 +136,7 @@ def test_deep_gaussian_process_optimize_with_defaults(
     y_observed = fnc_2sin_x_over_3(x_observed)
     data = x_observed, y_observed
     dataset = Dataset(*data)
-    optimizer = BatchOptimizer(tf.optimizers.Adam())
-    model = DeepGaussianProcess(two_layer_model(x_observed), optimizer)
+    model = DeepGaussianProcess(two_layer_model(x_observed))
```
sebastianober, line 139: Is this change because you're using the default `fit_args` so you don't need to pass the optimizer?

sebastianober:
Happy with GPflux things, LGTM :)

tests/integration/models/gpflux/test_predictions.py
hstojic: in test_builders we already test that the structure is the same, so I thought this one is an overkill

tests/unit/models/gpflux/test_models.py
hstojic: yes

PR #479, opened 2022-01-27
A model sampler protocols

uri-granta:
Split off reparam_sampler and trajectory_sampler model methods into separate protocols.

### DIFF
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -81,6 +81,7 @@ from trieste.types import State, TensorType
 # Optimizer parameters for testing against the branin function.
 # We also use these for a quicker test against a simple quadratic function
 # (regenerating is necessary as some of the acquisition rules are stateful).
+# The various   # type: ignore[arg-type]  are for rules that are only supported by GPR models.
 def OPTIMIZER_PARAMS() -> Tuple[
     str,
     List[
@@ -108,7 +109,7 @@ def OPTIMIZER_PARAMS() -> Tuple[
             (
                 22,
                 EfficientGlobalOptimization(
-                    MinValueEntropySearch(
+                    MinValueEntropySearch(  # type: ignore[arg-type]
                         BRANIN_SEARCH_SPACE,
                         min_value_sampler=ThompsonSamplerFromTrajectory(sample_min_value=True),
                     ).using(OBJECTIVE)
@@ -143,7 +144,7 @@ def OPTIMIZER_PARAMS() -> Tuple[
             (
                 10,
                 EfficientGlobalOptimization(
-                    GIBBON(  # type: ignore[arg-type]  # (only supported by GPR models)
+                    GIBBON(  # type: ignore[arg-type]
                         BRANIN_SEARCH_SPACE,
                     ).using(OBJECTIVE),
                     num_query_points=2,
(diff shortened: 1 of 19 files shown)


trieste/acquisition/function/entropy.py
```diff
@@ -41,7 +41,10 @@
 CLAMP_LB = 1e-8
 
 
-class MinValueEntropySearch(SingleModelAcquisitionBuilder[ProbabilisticModel]):
+T = TypeVar("T", bound=ProbabilisticModel, contravariant=True)
```
uri-granta, line 44: In a separate PR I'd like to go over all the type variables in trieste and decide on more user friendly names (but still consistent and short)!

trieste/acquisition/rule.py
```diff
@@ -442,7 +443,9 @@ def __init__(
             )
 
         if builder is None:
-            builder = BatchMonteCarloExpectedImprovement(10_000)
+            builder = cast(
+                SingleModelAcquisitionBuilder[M_contra], BatchMonteCarloExpectedImprovement(10_000)
```
uri-granta, line 447: Not sure why we need to tell mypy this (I though it should be able to infer it from the first overload).

tests/integration/test_bayesian_optimization.py
```diff
@@ -81,6 +81,7 @@
 # Optimizer parameters for testing against the branin function.
 # We also use these for a quicker test against a simple quadratic function
 # (regenerating is necessary as some of the acquisition rules are stateful).
+# The various   # type: ignore[arg-type]  are for rules that are only supported by GPR models.
```
henrymoss, line 84: why do we need this?

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -55,18 +55,15 @@ def prepare_acquisition_function(
         :param pending_points: The points already in the current batch (not used).
         :return: A trajectory sampled from the model.
         """
-
-        try:
-            self._trajectory_sampler = model.trajectory_sampler()
-            function = self._trajectory_sampler.get_trajectory(negate=True)
-        except (NotImplementedError):
+        if not isinstance(model, HasTrajectorySampler):
```
henrymoss, line 57: looks a bit cleaner :)

trieste/acquisition/function/entropy.py
```diff
@@ -198,7 +221,17 @@ class SupportsCovarianceObservationNoise(
...
 
-class GIBBON(SingleModelGreedyAcquisitionBuilder[SupportsCovarianceObservationNoise]):
+class SupportsCovarianceObservationNoiseTrajectory(
+    HasTrajectorySampler, SupportsCovarianceObservationNoise, Protocol
+):
+    """A model that supports covariance_between_points and get_observation_noise, and also
+    has an associated trajectory sampler."""
+
+
+GIBBONT = TypeVar("GIBBONT", bound=SupportsCovarianceObservationNoise, contravariant=True)
```
henrymoss, line 231: whats going on here?

trieste/acquisition/rule.py
```diff
@@ -702,7 +705,10 @@ def state_func(
         return state_func
 
 
-class DiscreteThompsonSampling(AcquisitionRule[TensorType, SearchSpace, ProbabilisticModel]):
+T = TypeVar("T", bound=ProbabilisticModel, contravariant=True)
```
henrymoss, line 708: Please could you explain this to me

henrymoss (approved):
LGTM. Just a couple of questions to improve my knowledge

tests/integration/test_bayesian_optimization.py
uri-granta: _test_optimizer_finds_minimum expects an acquisition_rule that can support any GPflowPredictor model, since it's sometimes run with VGP or SVGP models. However, some of the rules only support GPR.  At some point I might try to figure out how to refactor it without making the test look messier, but I think this is fine for now.

trieste/acquisition/function/entropy.py
uri-granta: It probably needs a better type variable (and I'm planning on another PR where I rename al the typevars so that they're a bit more intuitively named when they appear in method argument types) , but this is saying that the GIBBON acquisition function can only handle models that

1. support both covariance_between_points and get_observation_noise
2. are supported by the min_value_sampler

In particular, if you pass in min_value_sampler=ThompsonSamplerFromTrajectory, then the model must also support trajectory_sampler

trieste/acquisition/rule.py
uri-granta: This is similar to your previous question (but a bit simpler): DiscreteThompsonSampling can support any model, as long as the thomspon sampler supports it. The default sampler (ExactThompsonSampler) supports all ProbabilisticModels, but ThompsonSamplerFromTrajectory for example requires the modles to provide a trajectory sampler.

tests/integration/test_bayesian_optimization.py
henrymoss: got it

trieste/acquisition/function/entropy.py
henrymoss: I see

PR #480, opened 2022-01-27
Uri/remove negate argument

uri-granta:
Remove the negate argument from get_trajectory by having GreedyContinuousThompsonSampling update the callable trajectory to negate it.

### DIFF
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -333,8 +333,7 @@ def test_rff_trajectory_sampler_returns_deterministic_trajectory() -> None:
     npt.assert_allclose(trajectory_eval_1, trajectory_eval_2)
 
 
-@pytest.mark.parametrize("negate", [True, False])
-def test_rff_trajectory_sampler_can_return_negative_trajectory(negate: bool) -> None:
+def test_rff_trajectory_sampler_can_return_negative_trajectory() -> None:
     x_range = tf.linspace(2.0, 3.0, 5)
     x_range = tf.cast(x_range, dtype=tf.float64)
     xs = tf.reshape(tf.stack(tf.meshgrid(x_range, x_range, indexing="ij"), axis=-1), (-1, 2))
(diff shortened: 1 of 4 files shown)


hstojic (approved):
looks good to me

tests/unit/models/gpflow/test_sampler.py
```diff
@@ -333,8 +333,7 @@ def test_rff_trajectory_sampler_returns_deterministic_trajectory() -> None:
     npt.assert_allclose(trajectory_eval_1, trajectory_eval_2)
 
 
-@pytest.mark.parametrize("negate", [True, False])
-def test_rff_trajectory_sampler_can_return_negative_trajectory(negate: bool) -> None:
+def test_rff_trajectory_sampler_can_return_negative_trajectory() -> None:
```
henrymoss: This is no longer testing that it can return negative trajectory. I would jsut delete the test

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -67,6 +68,22 @@ def prepare_acquisition_function(
...
+                    return -1.0 * super().__call__(x)
+
+            function.__class__ = NegatedTrajectory
+        else:
+
+            def negated_trajectory(x: TensorType) -> TensorType:
+                return -1.0 * function(x)
+
+            function = negated_trajectory
+
```
henrymoss, line 86: Very clever!

tests/unit/models/gpflow/test_sampler.py
```diff
@@ -348,15 +347,12 @@ def test_rff_trajectory_sampler_can_return_negative_trajectory(negate: bool) ->
     )  # need a gpflow kernel object for random feature decompositions
 
     sampler = RandomFourierFeatureTrajectorySampler(model, num_features=100)
```
hstojic: this is not executed by the new continuous TS so I dont see why would trajectory be negated, by default it should be positive?

henrymoss (approved):
LGTM

tests/unit/models/gpflow/test_sampler.py
hstojic: essentially this does not test that a negative trajectory can be returned - this should probably be moved to the cont TS  test file and tested there properly

hstojic (commented):
Ok I think I have hit the approve to soon, that test doesn't look good anymore to me (see comments above)

tests/unit/models/gpflow/test_sampler.py
uri-granta: As @henrymoss noted, this test no longer doing anything useful, so I've removed it.

PR #482, opened 2022-01-27
Rename TypeVars to something more novice-friendly

uri-granta:
To make the function signatures less confusing for beginners, I've renamed nearly all of TypeVars to something more descriptive:

* `SearchSpaceType` (a type var bound to SearchSpace)
* `ProbabilisticModelType` (a type var bound to ProbabilisticModels)
* `TrainableProbabilisticModelType` (a type var bound to TrainableProbabilisticModel)
* `GIBBONModelType` (a type var bound to SupportsCovarianceObservationNoise used by GIBBON)
* `ResultType` (unbound type var used by AcquisitionRule)
* `StateType` (unbound type var used by Records and the BO loop)

Note that I've not included whether the types are contravariant in the type name as is sometimes done (e.g. `ProbabilisticModelType_contra`) as that makes them unweildy and is meaningless to beginners. However, I make sure to mention this in the type var docstrings.

### DIFF
--- a/trieste/acquisition/function/entropy.py
+++ b/trieste/acquisition/function/entropy.py
@@ -32,6 +32,7 @@ from ..interface import (
     AcquisitionFunction,
     AcquisitionFunctionClass,
     PenalizationFunction,
+    ProbabilisticModelType,
     SingleModelAcquisitionBuilder,
     SingleModelGreedyAcquisitionBuilder,
     UpdatablePenalizationFunction,
@@ -39,12 +40,6 @@ from ..interface import (
 from ..sampler import ExactThompsonSampler, ThompsonSampler
 
 CLAMP_LB = 1e-8
-
-
-ProbabilisticModelType = TypeVar(
-    "ProbabilisticModelType", bound=ProbabilisticModel, contravariant=True
-)
-""" Contravariant type variable bound to :class:`~trieste.models.ProbabilisticModel`. """
 
 
 class MinValueEntropySearch(SingleModelAcquisitionBuilder[ProbabilisticModelType]):
--- a/trieste/acquisition/function/function.py
+++ b/trieste/acquisition/function/function.py
@@ -36,7 +36,9 @@ from ..interface import (
     SingleModelVectorizedAcquisitionBuilder,
 )
 
-M_contra = TypeVar("M_contra", bound=ProbabilisticModel, contravariant=True)
+ProbabilisticModelType = TypeVar(
+    "ProbabilisticModelType", bound=ProbabilisticModel, contravariant=True
+)
 
 
 class ExpectedImprovement(SingleModelAcquisitionBuilder[ProbabilisticModel]):
@@ -422,7 +424,7 @@ def probability_of_feasibility(
     return acquisition
 
 
-class ExpectedConstrainedImprovement(AcquisitionFunctionBuilder[M_contra]):
+class ExpectedConstrainedImprovement(AcquisitionFunctionBuilder[ProbabilisticModelType]):
     """
     Builder for the *expected constrained improvement* acquisition function defined in
     :cite:`gardner14`. The acquisition function computes the expected improvement from the best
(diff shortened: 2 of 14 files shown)


trieste/acquisition/function/function.py
```diff
@@ -36,7 +36,10 @@
     SingleModelVectorizedAcquisitionBuilder,
 )
 
-M_contra = TypeVar("M_contra", bound=ProbabilisticModel, contravariant=True)
+ProbabilisticModelType = TypeVar(
+    "ProbabilisticModelType", bound=ProbabilisticModel, contravariant=True
+)
+""" Contravariant type variable bound to :class:`~trieste.models.ProbabilisticModel`. """
```
henrymoss: Could they all be defined somewhere else once?

trieste/acquisition/function/entropy.py
```diff
@@ -228,10 +231,13 @@ class SupportsCovarianceObservationNoiseTrajectory(
     has an associated trajectory sampler."""
 
 
-GIBBONT = TypeVar("GIBBONT", bound=SupportsCovarianceObservationNoise, contravariant=True)
+GIBBONModelType = TypeVar(
+    "GIBBONModelType", bound=SupportsCovarianceObservationNoise, contravariant=True
+)
+""" Type variable bound to :class:`~trieste.models.SupportsCovarianceObservationNoise`. """
```
henrymoss, line 232: Why do we not have one of these in the cts Thompson sampling acq? That needs the model to have a sampler

uri-granta: In general we don't need these just to specify that you need a specific feature: e.g. GreedyContinuousThompsonSampling is defined as a SingleModelGreedyAcquisitionBuilder[HasTrajectorySampler]. You only need type variables if the features you require depend on one of the input arguments.

For example, a rule that gets passed an acqusition function builder that only supports models with predict_joint also only supports models with predict_joint. Or an acquisition function (such as GIBBON) that gets passed a sampler that only supports models with trajectory functions also only supports models with trajectory functions.

trieste/acquisition/function/function.py
uri-granta: Good suggestion. Done.

trieste/models/interfaces.py
```diff
@@ -25,7 +25,10 @@
 from ..types import TensorType
 from ..utils import DEFAULTS
 
-T = TypeVar("T", bound="ProbabilisticModel", contravariant=True)
+ProbabilisticModelType = TypeVar(
```
hstojic, line 28: should this be exported in __init__

trieste/utils/__init__.py
```diff
@@ -14,7 +14,7 @@
 """ This package contains library utilities. """
 from typing import Callable, Mapping, TypeVar
 
-from .misc import DEFAULTS, Err, Ok, Result, T_co, jit, shapes_equal, to_numpy
+from .misc import DEFAULTS, Err, Ok, Result, ResultType, jit, shapes_equal, to_numpy
 
 K = TypeVar("K")
```
hstojic: a small digression: why is map_values defined in __init__ and not in e.g. misc.py?

hstojic (commented):
much better :)
I would only add some lay man description in the docstrings of these type vars, right now its developer oriented description

trieste/utils/__init__.py
joelberkeley: i wanted to put everything in \_\_init\_\_.py but never got round to it. Given there's nothing else in `utils`, it now may as well be in a single utils.py file (no package)

trieste/acquisition/function/entropy.py
```diff
@@ -69,11 +67,11 @@ def __init__(
 
     @overload
     def __init__(
-        self: "MinValueEntropySearch[T]",
+        self: "MinValueEntropySearch[ProbabilisticModelType]",
```
joelberkeley, line 70: curious why you've typed `self`

trieste/utils/__init__.py
uri-granta: No idea! I'll move it.

uri-granta: Ah ok. I'll migrate it in a separate PR.

henrymoss (approved):

PR #483, opened 2022-01-29
using gpflow build fncs in integration tests 

hstojic:


tests/integration/test_bayesian_optimization.py
```diff
@@ -104,54 +99,54 @@ def OPTIMIZER_PARAMS() -> Tuple[
...
-            (
-                10,
-                EfficientGlobalOptimization(
-                    GIBBON(  # type: ignore[arg-type]
-                        BRANIN_SEARCH_SPACE,
-                    ).using(OBJECTIVE),
-                    num_query_points=2,
-                ),
-            ),
+            # (20, EfficientGlobalOptimization()),
```
uri-granta: Uncomment.

Also, I assume the AugmentedExpectedImprovement test wasn't always passing so you upped the iterations? Either way, remember to manually kick off an deploy-checks run before merge.

tests/integration/test_bayesian_optimization.py
```diff
@@ -221,7 +216,11 @@ def test_bayesian_optimizer_with_gpr_finds_minima_of_simple_quadratic(
 ) -> None:
     # for speed reasons we sometimes test with a simple quadratic defined on the same search space
     # branin; currently assume that every rule should be able to solve this in 5 steps
-    _test_optimizer_finds_minimum(min(num_steps, 5), acquisition_rule)
+    _test_optimizer_finds_minimum(
+        min(num_steps, 5),
+        acquisition_rule,
+        # model_args={"num_kernel_samples": 100}
```
uri-granta: ?

tests/integration/test_bayesian_optimization.py
```diff
@@ -336,12 +335,11 @@ def _test_optimizer_finds_minimum(
         State[TensorType, AsynchronousRuleState | TrustRegion.State], Box, GPflowPredictor
     ],
     optimize_branin: bool = False,
-    model_type: str = "GPR",  # in Python 3.8+ this could be Literal["GPR", "VGP", "SVGP"]
-    model_args: Optional[Mapping[str, Any]] = None,
+    model_type: str = "GPR",  # in Python 3.8+ a Literal["GPR", "VGP", "SVGP", "DGP", "DE"]?
+    model_args: Mapping[str, Any] = {},
```
uri-granta: You might get a warning here for having a mutable default argument. And because Python doesn't have a built in frozendict class it's probably best to leave it as an optional defaulting to None.

tests/integration/test_bayesian_optimization.py
```diff
@@ -354,104 +352,81 @@ def _test_optimizer_finds_minimum(
...
-                )
-                gpflow.utilities.set_trainable(svgp.likelihood, False)
-                return SparseVariational(svgp, **model_args)
+    if model_type == "GPR":
+        if "LocalPenalization" in acquisition_rule.__repr__():
+            likelihood_variance = 1e-3
+        else:
+            likelihood_variance = 1e-5
+        gpr = build_gpr(initial_data, search_space, likelihood_variance=likelihood_variance)
+        model = GaussianProcessRegression(gpr, **model_args)
```
uri-granta, line 371: If you mark this as a GPflowPredictor then you shouldn't need the type ignores below (except for the Keras model) plus it corresponds to the model type we claim in GPR_OPTIMIZER_PARAMS

```suggestion
        model: GPflowPredictor = GaussianProcessRegression(gpr, **model_args)
```

That said the test is currently a bit of a typing mess, as there are already some rules that only work with GPR rather than all GPflowPredictors. And adding Keras made it worse, as many more of the rules don't work with that. I might refactor it in a separate PR.

tests/integration/test_constrained_bayesian_optimization.py
```diff
@@ -26,17 +25,16 @@
...
 from trieste.types import TensorType
-from trieste.utils import map_values
 
 
 @random_seed
 @pytest.mark.parametrize(
     "num_steps, acquisition_function_builder",
     [
-        (10, ExpectedConstrainedImprovement),
+        (20, ExpectedConstrainedImprovement),
```
uri-granta: :-(

tests/integration/test_mixed_space_bayesian_optimization.py
```diff
@@ -55,16 +52,16 @@
             ]
         ],
         [
-            (15, EfficientGlobalOptimization()),
+            (24, EfficientGlobalOptimization()),
```
uri-granta: How are the test run times affected by this PR (both alltests and quicktests)?

uri-granta (commented):
Nice. Just a few comments.

tests/integration/test_bayesian_optimization.py
hstojic: eh, didnt save the file so these comments were a leftover - yes, had to up the iterations

tests/integration/test_bayesian_optimization.py
hstojic: I didn't get any warning, that's why I kept it?

tests/integration/test_mixed_space_bayesian_optimization.py
hstojic: some increase, but not that much, seemed worthwhile the simplification of the code

tests/integration/test_bayesian_optimization.py
hstojic: yeah, its a bit of a mess in terms of typing, but also doesnt make sense to duplicate whole BO loop part between the models...

hstojic: that suggestion raises some other typing issues (in optimize on line 425...), so perhaps we keep it as is for now?

tests/integration/test_bayesian_optimization.py
uri-granta: I think it would still be better to keep the immutable default as it's both more Pythonic and less error prone.

hstojic: ok, done!

tests/integration/test_mixed_space_bayesian_optimization.py
hstojic: it seems like here running time is almost not affected at all, 5s more or less

uri-granta (approved):
Happy for you to merge once the integration tests are passing

Issue #476, opened 2022-01-25
Test with multiple versions of Python


uri-granta:
We should consider have some tests run with multiple versions of Python (and possibly Tensorflow too), to catch bugs like https://github.com/secondmind-labs/trieste/issues/473.

Pros:
- catch backwards compatibility issues

Cons:
- more infrastructure
- more time to run tests (even if we only run them at release time)

apaleyes:
Is it that much work? Say, to verify new version of Python all that's needed is few more lines in the action yaml:
```
  tests-3.8:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - uses: actions/setup-python@v2
      with:
        python-version: 3.8
    - run: pip install tox
    - run: tox -e tests
```

If we add this to develop checks, it won't even make the build longer, because the `alltests` action will take longer that this one, and they'll very likely start simultaneously. Happy to turn this into quick PR if you agree!

PR #484, opened 2022-01-31
Support gpflux 0.2.3 and pin GH to tf 2.4

uri-granta:
* Support TF 2.4 by using using RandomFourierFeatures rather than RandomFourierFeaturesCosine if gpflux 0.2.3 is installed.
* Add an example test that depends on the version of gpflux so people know how to write them in future
* Pin GitHub to TF 2.4 for now.

Note that this increases the priority of https://github.com/secondmind-labs/trieste/issues/476, as any tests specific to TF 2.5+ would currently not be run on GH.


hstojic (approved):
looks ok to me

PR #488, opened 2022-02-02
Henry/wallclock

henrymoss:
As requested by @apaleyes  and @vpicheny , I have added in some wallclock timing to our Tensorboard logging. 

Its very self explanatory.


trieste/bayesian_optimizer.py
```diff
@@ -444,25 +445,33 @@ def optimize(
...
 
         for step in range(num_steps):
             set_step_number(step)
             try:
-                if track_state:
-                    models_copy = copy.deepcopy(models)
-                    acquisition_state_copy = copy.deepcopy(acquisition_state)
-                    history.append(Record(datasets, models_copy, acquisition_state_copy))
+
+                t_0_total = time()  # start wallclock for total BO step time
```
uri-granta: Use time.perf_counter instead as it's (1) more accurate and (2) doesn't depend on the system clock so can't go backwards if somebody e.g. changes the system to summer time during execution!

```suggestion
                t_0_total = perf_counter()  # start wallclock for total BO step time
```

trieste/bayesian_optimizer.py
```diff
@@ -444,25 +445,33 @@ def optimize(
...
                     for tag, model in models.items():
                         dataset = datasets[tag]
                         model.update(dataset)
                         model.optimize(dataset)
+                    initial_model_fitting_time = time() - t_0_model
 
+                if track_state:
+                    models_copy = copy.deepcopy(models)
+                    acquisition_state_copy = copy.deepcopy(acquisition_state)
+                    history.append(Record(datasets, models_copy, acquisition_state_copy))
```
uri-granta, line 457: Out of curiosity, why did you move the state tracking to after fitting the initial model?

trieste/bayesian_optimizer.py
```diff
@@ -444,25 +445,33 @@ def optimize(
...
             try:
-                if track_state:
-                    models_copy = copy.deepcopy(models)
-                    acquisition_state_copy = copy.deepcopy(acquisition_state)
-                    history.append(Record(datasets, models_copy, acquisition_state_copy))
+
+                t_0_total = time()  # start wallclock for total BO step time
 
                 if step == 0 and fit_initial_model:
+                    t_0_model = time()  # start wallclock for initial model fitting
```
uri-granta: You can make these timing calculations nicer (and less error prone) by defining a timing context manager in trieste.utils:

```python
from time import perf_counter
from contextlib import contextmanager

@contextmanager
def timeit() -> float:
    t_0 = perf_counter()
    yield lambda: perf_counter() - t_0
```

which will let you simplify the code to:

```python
with timeit() as initial_model_fitting_time:
        ...
```

You can nest timeits so the total timer will work too.

trieste/bayesian_optimizer.py
```diff
@@ -500,6 +515,21 @@ def optimize(
                                 np.min(tagged_output[tag].observations),
                                 step=step,
                             )
+                        tf.summary.scalar(
+                            "Wallclock.total",
```
uri-granta: Is "Wallclock" a standard phrase? Also I think our other scalars use lower case for the titles (unless they start with a class name or an objective):

```suggestion
                            "wallclock.total",
```

tests/unit/test_logging.py
```diff
@@ -75,3 +91,103 @@ def test_step_number(step: int) -> None:
...
+    ordered_scalar_names = [
+        "A.observation.best_overall",
+        "A.observation.best_new",
+        "Wallclock.total",
+        "Wallclock.query_point_generation",
+        "Wallclock.model_fitting",
+    ]
+    N = len(ordered_scalar_names)
+
+    for step in range(steps):  # iterate over saved scalars
```
uri-granta: This is fine, if a bit confusing and error prone, plus it doesn't check that the mocked_summary_scalar.call_args_list doesn't contain superfluous entries. You should be able to simplify it something like:

```python
for call_arg, (step, scalar_name) in zip_longest(
    mocked_summary_scalar.call_args_list,
    product(range(steps), ordered_scalar_names
):
    assert call_arg[0][0] == scalar_name
    assert call_arg[-1]["step"] == step
    assert isinstance(call_arg[0][1], float)

File: tests/unit/test_logging.py [hunk 6]
```diff
@@ -75,3 +91,103 @@ def test_step_number(step: int) -> None:
...
+            sleep(model_fit_time)
+
+    class _FixedAcquisitionRuleWithWaiting(FixedAcquisitionRule):
+        def acquire(
+            self,
+            search_space: SearchSpace,
+            models: Mapping[str, ProbabilisticModel],
+            datasets: Optional[Mapping[str, Dataset]] = None,
+        ) -> TensorType:
+            """
```
uri-granta: I don't think the docstring adds much!

tests/unit/test_logging.py
```diff
@@ -75,3 +91,103 @@ def test_step_number(step: int) -> None:
...
+        )
+
+    for scalar in mocked_summary_scalar.call_args_list:
+        if scalar[0][0][:9] == "Wallclock":
+            assert scalar[0][1] > 0  # want positive wallclock times
+
+    # check logs of first BO step
+    if fit_initial_model:  # logging should count two model fits
+        npt.assert_allclose(
+            mocked_summary_scalar.call_args_list[2][0][1], 2 * model_fit_time + acq_time, rtol=0.1
```
uri-granta: Isn't this going to be very fragile? If anyone changes the scalars we log then it'll break, and the fix won't be obvious.

Since this test is only trying to check the values are correct, why not loop over the call args and check the values programmatically? In fact you should be able to just change the loop on line 166 to say something like this instead: (and maybe increase steps to 3 just for good luck)

```python
other_scalars = 0

for scalar in mocked_summary_scalar.call_args_list:
    name = scalar[0][0]
    value = scalar[0][1]
    step = scalar[-1]["step"]
    if name.startswith("Wallclock"):
        assert value > 0  # want positive wallclock times
    if name == "Wallclock.total":
        npt.assert_allclose(value, model_fit_time + acq_time + (model_fit_time if fit_initial_model and step == 0 else 0, rtol=0.1)
    elif name == "Wallclock.query_point_generation"):
        npt.assert_allclose(value, acq_time, rtol=0.1)
    elif name == "Wallclock.model_fitting"):
        npt.assert_allclose(value, model_fit_time + (model_fit_time if fit_initial_model and step == 0 else 0, rtol=0.1)
    else:
        other_scalars += 1

# check that we processed all the wallclocks we were expecting
assert len(mocked_summary_scalar.call_args_list) == other_scalars + 3 * steps
```

uri-granta (approved):
Looks good. Various comments but happy for you to merge once you've addressed them.

vpicheny (commented):
not sure I can contribute much technically to this PR, but should we consider reflecting this functionality in a notebook somehow?

trieste/bayesian_optimizer.py
henrymoss: Sorry, will switch back! I moved this when I initially put timings in our Result object (which was a silly thing to try)

trieste/bayesian_optimizer.py
henrymoss: Much nicer!

trieste/bayesian_optimizer.py
henrymoss: @vpicheny is wallclock a standard phrase?

henrymoss:
> not sure I can contribute much technically to this PR, but should we consider reflecting this functionality in a notebook somehow?

Mentioned it in the logging notebook

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -44,8 +44,7 @@
 # %% [markdown]
 # We can now also load the TensorBoard extension, though at this point there will not be any data to dispay.
 
-# %% magic_args="echo Loading TensorBoard..." language="script"
-#
+# %%
```
uri-granta: Did you mean to revert this line? It's there to prevent from tensorboard from actually being loaded in the autogenerated docs (we use screenshots instead).

tests/unit/utils/test_misc.py
```diff
@@ -81,3 +82,10 @@ def test_err() -> None:
 
     assert Err(ValueError()).is_ok is False
     assert Err(ValueError()).is_err is True
+
+
+def test_timeit() -> None:
```
uri-granta: Maybe extend this a tiny bit to check that nested timers work?

trieste/utils/misc.py
```diff
@@ -210,3 +212,18 @@ def map_values(f: Callable[[U], V], mapping: Mapping[K, U]) -> Mapping[K, V]:
         applying ``f`` to each value in ``mapping``.
     """
     return {k: f(u) for k, u in mapping.items()}
+
+
+class timeit:
```
apaleyes: There is a module with the same name in the standard lib: https://docs.python.org/3/library/timeit.html . I'd suggest to pick a different name to avoid confusions

uri-granta: Sorry to be a pain, but now that were' implementing timeit as a class, we should really name it TimeIt instead!

Also there should be a class docstring explaining what this is for and how to use it.

uri-granta (commented):
Looks much better. Just a couple more comments (sorry).

trieste/bayesian_optimizer.py
```diff
@@ -500,6 +506,26 @@ def optimize(
                                 np.min(tagged_output[tag].observations),
                                 step=step,
                             )
+                        tf.summary.scalar(
+                            "wallclock.total",
```
apaleyes: total sounds like overall time, e.g. time from the start of the optimization and up until this point. Maybe "step_time"?

trieste/bayesian_optimizer.py
```diff
@@ -444,45 +444,51 @@ def optimize(
...
-                        dataset = datasets[tag]
-                        model.update(dataset)
-                        model.optimize(dataset)
+                with timeit() as total_wallclock_timer:
+                    if step == 0 and fit_initial_model:
+                        with timeit() as initial_model_fitting_timer:
+                            for tag, model in models.items():
+                                dataset = datasets[tag]
+                                model.update(dataset)
+                                model.optimize(dataset)
```
apaleyes: That looks like way too many levels of nesting, but i don't have a good alternative. @uri-granta any ideas?

apaleyes (approved):
That was lightning fast, thanks Henry! Only nitpicking at this point

trieste/bayesian_optimizer.py
uri-granta: I think it's fine (though I used to write LISP code so maybe I'm not scared of nesting as much as I should be)

trieste/utils/misc.py
henrymoss: okay changed it to Timer

trieste/bayesian_optimizer.py
henrymoss: yep

tests/unit/utils/test_misc.py
henrymoss: done

trieste/bayesian_optimizer.py
apaleyes: this method overall looks pretty scary to me. half-way through it's hard to figure out at which level you are

uri-granta (approved):
LGTM

PR #490, opened 2022-02-03
Test with Python 3.9 and TF 2.5

### LINKED ISSUE #476: Test with multiple versions of Python

uri-granta:


tox.ini
```diff
@@ -24,29 +24,36 @@ extend-exclude = docs/notebooks/,.venv
 
 [testenv]
 basepython = python3.7
-deps = pip~=20.3
+deps = pip
```
hstojic, line 27: will this potentially mean an old pip version will be used?

tox.ini
```diff
@@ -24,29 +24,36 @@ extend-exclude = docs/notebooks/,.venv
...
 whitelist_externals = bash
 setenv =
     CUDA_VISIBLE_DEVICES =
 commands =
+    # type checking
     types: pip install -r common_build/types/requirements.txt -c common_build/types/constraints.txt
-    types: mypy
+    types: mypy {posargs}
```
hstojic, line 34: where from are these `{posargs}` coming from?

hstojic (approved):
seems ok to me, just some minor comments

tox.ini
uri-granta: This should use the latest version (which we were explicitly doing anyway in most of the commands via a pip install).

tox.ini
uri-granta: They're passed in on the command line. Hence to run the slow tests we can now run "tox -e tests -- --runslow yes" instead of having a whole different tox job. This also lets you run just one test or type check just one file if you want to.

PR #491, opened 2022-02-04
Covariance between points method for our variational GPflow models

henrymoss:
I have implemented (and tested) support for `covariance_between_points` for our SVGP and GVP models.

This functionality will allow us to use GIBBON with these models.

The implementation works for VGP and both whitened and non-whitened SVGP models.


tests/unit/models/gpflow/test_models.py
```diff
@@ -934,3 +901,72 @@ def test_gaussian_process_regression_conditional_predict_f_sample() -> None:
         sample_cov = tfp.stats.covariance(samples[i, :, :, 0], sample_axis=0)
         np.testing.assert_allclose(sample_mean, predj_meani, atol=1e-2, rtol=1e-2)
         np.testing.assert_allclose(sample_cov, predj_covi[0], atol=1e-2, rtol=1e-2)
+
+
+def test_gpflow_models_pairwise_covariance(gpflow_interface_factory: ModelFactoryType) -> None:
```
vpicheny, line 930: maybe this deserves more tests with eg leading dimensions for qp1 ? also, not sure if all this works well when we have more than 1 latent

henrymoss: things you conveniently ignored when you wrote yours ;)

uri-granta (approved):
Looks good. Just one question: is it now possible to remove the # type: ignore[arg-type] from the GIBBON test in test_bayesian_optimization? (if not, why not? if so, why isn't complaining about a superfluous type ignore?)

trieste/models/gpflow/models.py
```diff
@@ -871,3 +895,181 @@ def get_internal_data(self) -> Dataset:
...
+
+
+def compute_kernel_blocks(
+    kernel: gpflow.kernels.Kernel,
+    inducing_variable: InducingVariables,
+    query_points_1: TensorType,
+    query_points_2: TensorType,
+    num_latent: int,
+) -> tuple[TensorType, TensorType, TensorType, TensorType]:
+
```
henrymoss, line 1035: @vpicheny, What would a good doc-string be for this function? Its very hard to understand

tests/unit/models/gpflow/test_models.py
vpicheny: this is fixed here and for GPR now, yay!

tests/unit/models/gpflow/test_models.py
```diff
@@ -939,3 +906,104 @@ def test_gaussian_process_regression_conditional_predict_f_sample() -> None:
...
+    actual_covariance = model.covariance_between_points(  # type: ignore
+        query_points_1, query_points_2
+    )
+
+    np.testing.assert_allclose(expected_covariance, actual_covariance[0], atol=1e-4)
+
+
+@random_seed
+@pytest.mark.parametrize("whiten", [True, False])
+@pytest.mark.parametrize(
```
vpicheny, line 959: @uri-granta is this written well enough? (I wrote this test so don't worry you can't upset Henry) 

See also the `two_output_svgp_model` function

trieste/models/gpflow/models.py
vpicheny: Return the covariance between inducing variables and query_points_1, inducing variables and query_points_2, query_points_1 and query_points_2 and the covariance matrix of inducing_variables.

The calculations are conducted differently depending on the type of kernel (single output, separate independent multi-output or shared independent multi-output) and inducing variables (simple set, SharedIndependent or SeparateIndependent) to always return similar shapes.

vpicheny: feel free to improve on this text!!!

vpicheny: maybe it misses something to say that this is for each latent GP independently

trieste/models/gpflow/models.py
```diff
@@ -871,3 +895,181 @@ def get_internal_data(self) -> Dataset:
...
+    )
+    return cov
+
+
+def compute_kernel_blocks(
+    kernel: gpflow.kernels.Kernel,
+    inducing_variable: InducingVariables,
+    query_points_1: TensorType,
+    query_points_2: TensorType,
+    num_latent: int,
```
vpicheny, line 1019: num_latent is only used when we use a single kernel for a multioutput model (hence a SharedIndependent kernel, but not declared as such) - don't know if it's worth explaining

trieste/models/gpflow/models.py
```diff
@@ -871,3 +895,181 @@ def get_internal_data(self) -> Dataset:
...
+
+    if len(tf.shape(K)) == 2:
+        K = tf.repeat(tf.expand_dims(K, -3), num_latent, axis=-3)
+        Kx1 = tf.repeat(tf.expand_dims(Kx1, -3), num_latent, axis=-3)
+        Kx2 = tf.repeat(tf.expand_dims(Kx2, -3), num_latent, axis=-3)
+        K12 = tf.repeat(tf.expand_dims(K12, -3), num_latent, axis=-3)
+    elif len(tf.shape(K)) > 3:
+        raise NotImplementedError(
+            "Covariance between points is not supported " "for kernels of type " f"{type(kernel)}."
+        )
```
vpicheny, line 1095: I forgot to add the assert_shapes, possibly worth it?

trieste/models/gpflow/models.py
```diff
@@ -871,3 +895,181 @@ def get_internal_data(self) -> Dataset:
...
+        K12 = kernel(query_points_1, query_points_2)  # [..., A, B]
+
+    if len(tf.shape(K)) == 2:
+        K = tf.repeat(tf.expand_dims(K, -3), num_latent, axis=-3)
+        Kx1 = tf.repeat(tf.expand_dims(Kx1, -3), num_latent, axis=-3)
+        Kx2 = tf.repeat(tf.expand_dims(Kx2, -3), num_latent, axis=-3)
+        K12 = tf.repeat(tf.expand_dims(K12, -3), num_latent, axis=-3)
+    elif len(tf.shape(K)) > 3:
+        raise NotImplementedError(
+            "Covariance between points is not supported " "for kernels of type " f"{type(kernel)}."
```
vpicheny, line 1094: we should probably remove this now? I'm not sure sure this a case for which the shape is larger than 3. But we could throw this error when asserting shapes instead, see below

trieste/models/gpflow/models.py
```diff
@@ -871,3 +895,181 @@ def get_internal_data(self) -> Dataset:
...
+                inducing_points, query_points_1, full_cov=True, full_output_cov=False
+            )  # [..., L, M, A]
+            Kx2 = kernel(
+                inducing_points, query_points_2, full_cov=True, full_output_cov=False
+            )  # [L, M, B]
+            K12 = kernel(
+                query_points_1, query_points_2, full_cov=True, full_output_cov=False
+            )  # [..., L, A, B]
+    else:
+        K = kernel(inducing_points)  # [M, M]
```
vpicheny, line 1068: add comment here to explain that this is the single output case?

trieste/models/gpflow/models.py
vpicheny: if we want to explain in more details:

- case shared inducings + shared kernel: K etc. will be a rank 2 tensor (single kernel applied to a single set of inducings). We then add the latent dimension after computing the matrices
- case shared inducings + separate kernel: K etc. will be directly a rank 3 tensor (the kernel has the latent dimension)
- case separate inducings + shared kernels: we apply the kernel to a list of inducing points, then concatenate to retrieve the latent dimension
- case separate inducings + separate kernels: we apply each kernel to its corresponding set of inducing points, then concatenate.

trieste/models/gpflow/models.py
henrymoss: What do you mean?

vpicheny: `tf.debugging.assert_shapes([(K, [num_latent, "N", "D"]),])`

etc.

vpicheny: Kx1 should be [num_latent, query_points_1.shape], etc.

henrymoss: done

PR #492, opened 2022-02-07
GPR with 2 output dims

vpicheny:
Slight modification of the GPR's `covariance_between_points` to work with multiple outputs.

The test now passes with 1 and 2 output dimensions.


henrymoss (approved):
LGTM

uri-granta (approved):
Looks good! (and the shape comments make things much easier to read)

PR #493, opened 2022-02-08
random sampling

hstojic:
seemed useful to have a random sampling acq rule as a baseline


vpicheny (approved):
neat

tests/unit/acquisition/test_rule.py
```diff
@@ -191,6 +192,25 @@ def test_discrete_thompson_sampling_acquire_returns_correct_shape(
...
+
+@pytest.mark.parametrize("num_query_points", [1, 10, 50])
+def test_random_sampling_acquire_returns_correct_shape(num_query_points: int) -> None:
+    search_space = Box([-2.2, -1.0], [1.3, 3.3])
+    rule = RandomSampling(num_query_points)
+    dataset = Dataset(tf.zeros([1, 2], dtype=tf.float64), tf.zeros([1, 1], dtype=tf.float64))
+    model = QuadraticMeanAndRBFKernelWithSamplers(
+        dataset=dataset, noise_variance=tf.constant(1.0, dtype=tf.float64)
+    )
+    query_points = rule.acquire(search_space, model)
```
hstojic: I get typing errors here that I'm not sure how to deal with, @uri-granta?
```
tests/unit/acquisition/test_rule.py:204: error: Need type annotation for "rule"  [var-annotated]
tests/unit/acquisition/test_rule.py:209: error: Argument 2 to "acquire" of "RandomSampling" has incompatible type "QuadraticMeanAndRBFKernelWithSamplers"; expected "Mapping[str, Any]"  [arg-type]
```

henrymoss (commented):
LGTM. It might be worth adding some of the tensorboard logging that @uri-granta and I added for the EGO rule (i.e. chosen points and timings)?

hstojic:
> LGTM. It might be worth adding some of the tensorboard logging that @uri-granta and I added for the EGO rule (i.e. chosen points and timings)?

fair enough, I'll see about adding that

hstojic:
> > LGTM. It might be worth adding some of the tensorboard logging that @uri-granta and I added for the EGO rule (i.e. chosen points and timings)?
> 
> fair enough, I'll see about adding that

@henrymoss as far as I can see in EGO we add only this to the tensorboard:
```
tf.summary.scalar("EGO.acquisition_function.maximum_found", value)
```
which is an acquisition function value, and here it doesn't make sense, there is no acq function, we just sample from the search space?

tests/unit/acquisition/test_rule.py
uri-granta: You're should be calling acquire_single not acquire as you're passing in just a model, not a mapping of models (it happens to work because RandomSampling ignores that argument anyway). ~~Not sure why the first error is happening though.~~ (see other comment about the first error)

trieste/acquisition/rule.py
```diff
@@ -718,6 +718,49 @@ def state_func(
         return state_func
 
 
+class RandomSampling(AcquisitionRule[TensorType, SearchSpace, ProbabilisticModelType]):
```
uri-granta: The other type error is because unlike DiscreteThompsonSampling this doesn't need to be generic on the supported model type: it supports any ProbabilisticModel. (DiscreteThompsonSampling only supports models that are supported by the thompson_sampler that's used.)

```suggestion
class RandomSampling(AcquisitionRule[TensorType, SearchSpace, ProbabilisticModel]):
```

PR #494, opened 2022-02-08
Fix develop_checks/tests_latest

uri-granta:


tests/integration/test_active_learning.py
```diff
@@ -49,7 +49,7 @@
     [
         (50, EfficientGlobalOptimization[SearchSpace, SupportsPredictJoint](PredictiveVariance())),
         (
-            70,
+            80,
```
hstojic: I think this one is a bit stochastic, sometimes fails sometimes passes

uri-granta: You may be right. It did pass for me just now, but I might investigate relaxing the test slightly to make it consistent.

hstojic (approved):
thats fine, good to go

PR #496, opened 2022-02-08
simplified optimizer code

### LINKED ISSUE #462: remove redundancy in `optimize_random`
`optimize_random` is identical to `optimize_discrete`, barring a single line - we could reduce the code by adding an argument random in `optimize_discrete` and use it in both places

hstojic:
I have noticed some redundancy in optimizer code between discrete and random optimizer

this will close #462


henrymoss (approved):
Nice spot!

uri-granta (approved):
Looks good, thanks!

PR #497, opened 2022-02-08
Support batching acquisition function calls

### LINKED ISSUE #495: PredictiveVariance scales poorly
`PredictiveVariance` is not scaling well enough, e.g. already with datasets of medium size, about 200 points in 6 dim hartmann problem my computer runs out of memory and process gets killed

I think `predict_joint` call in `PredictiveVariance` is the culprit, single call is fine for the data sizes we deal with here, you really need to increase the data size a lot to hit the memory issue, but I suspect I get it even for smaller datasets because our acquisition optimizer calls it in parallel

I'm not sure whether there is any solution to this, perhaps it simply cannot be used with a lot of data and we put a warning in the docs, some other ideas:
1. warning as above + @uri-granta idea to put the acquisition optimizer in a serial mode (looking into this right now, @henrymoss?), so the memory issue is hit later on + do it in the notebook as well
2. for batch=1, call `predict` and just do max var, there is no need for joint call here, right (@vpicheny @satrialoka)? that would make the function much more efficient at least in this case 

any other ideas?

uri-granta:


docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -124,6 +124,24 @@ def __call__(self, x):
     )
     plot_bo_points(data.query_points.numpy(), ax[0, 0], 5, arg_min_idx)
 
+# %% [markdown]
+# ## Out of memory errors
+#
+# At the moment, trieste's `OptimizationResult` doesn't allow you to recover from Out Of Memory errors, which normally result in the Python process shutting down. One possible cause of memory errors is performing a gradient-based optimization over a large dataset. To work around this, you can specify that evaluations of the acquisition function be batchified: this splits them (on the first dimension) into batches of a given size, then stitches them back together. To do this, you need to provide an explicit batchified optimizer and specify a desired batch size.
+
```
henrymoss: One possible cause of memory errors is trying to evaluate an acquisition function over a large dataset, e.g. when initializing our gradient-based optimizers.

trieste/acquisition/batch.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    A wrapper around an :const:`AcquisitionFunction` to split its input into batches.
+    Splits `x` into batches along the first dimension, calls `fn` on each batch, and then stitches
+    the results back together, so that it looks like `fn` was called with all of `x` in one batch.
+    :param fn: Acquisition function to call with batches of data.
+    :param batch_size: Call fn with tensors of at most this size.
+    :returns Batched acquisition function.
+    """
+    if batch_size <= 0:
+        raise ValueError(f"batch_size must be positive, got {batch_size}")
```
henrymoss: I kinda think using the word batchify is confusing here as that means a very specific thing in our acq optimizers, i.e. batchify is to do with batches of query points. This PR is all about batching computations. Any thoughts on terminology @hstojic or @vpicheny ?

henrymoss: Perhaps `throttled_acquisition_function`

trieste/acquisition/batch.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @functools.wraps(fn)
+    def wrapper(x: TensorType) -> TensorType:
+        x = tf.convert_to_tensor(x)
+
+        # this currently assumes leading dimension of x is batch dimension.
```
henrymoss: again, batch dimension is what we call the -2 dimension in Trieste. This naming is a bit confusing

henrymoss (approved):
This is great. I just have a couple of concerns about naming

henrymoss:
So is it possible to have our `OptimizationResult` store that the crash was due to OOM?

trieste/acquisition/batch.py
henrymoss: @vpicheny agreed that this is a bit confusing

uri-granta:
> So is it possible to have our `OptimizationResult` store that the crash was due to OOM?

An OOM will typically crash the entire process and not allow us to access the result at all. This makes having track_state support serialising state to disk (which we don't currently do) even more important as that would at let us reload the computation that was successfully completed.

trieste/acquisition/batch.py
uri-granta: :+1: If there aren't any other suggestions I'll rename it to `throttle_acquisition_function` and `throttle_acquisition_function_calls`. Shall I continue to cal the batch size `batch_size` or shall I rename that to `throttle_batch_size`?

uri-granta: Also any suggestions for the file name? It didn't obviously fit anywhere else. `throttle.py`? `memory.py`?

henrymoss: maybe `memory_throttled` or `split` ? Any suggestions @hstojic or @vpicheny ?

henrymoss:
> 

Can we still have an informative error message if the whole process crashes?

trieste/acquisition/batch.py
vpicheny: `hashed` or `sliced`? 

We should indeed NOT call anything else `batch_size`, this would be so confusing.

docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -124,6 +124,24 @@ def __call__(self, x):
...
+# %%
+from trieste.acquisition.optimizer import (
+    automatic_optimizer_selector,
+    batchify_acquisition_function_calls,
+)
+from trieste.acquisition.rule import EfficientGlobalOptimization
+
+optimizer = batchify_acquisition_function_calls(
+    automatic_optimizer_selector, batch_size=10_000
```
hstojic: to make it more realistic perhaps something smaller than 10_000? conceivably you would use something like 5-10 rather than 10000 if you have memory issues, no?

trieste/acquisition/optimizer.py
```diff
@@ -595,6 +596,36 @@ def optimizer(
     return optimizer
 
 
+def batchify_acquisition_function_calls(
```
hstojic: why not putting this on in batch.py as well?

trieste/acquisition/batch.py
hstojic: yes, probably its best to avoid batch terms since it has very specific meaning in BO

I don't like throttle, but other than that I'm somewhat indifferent, I can add few more to the mix if you really need more :) `split_compute` and `split_optimize`

docs/notebooks/recovering_from_errors.pct.py
henrymoss: I think that is the correct order of size. 1_000 predictions is only a few hundred Mbs of memory. We are getting into issues when trying 1_000_000 evals

hstojic (approved):
looks quite good!

could you run a realistic memory test with it as well?
you can use PredictiveVariance integration test in test_active_learning by increasing the initial data size to something like 200-300

trieste/acquisition/batch.py
henrymoss: I think `split_compute` is probs best.

uri-granta: I don't know what you're suggesting. We need names for:

1. the functions that convert acquisition functions and optimizers (currently `batchify_acquisition_function` and `batchify_acquisition_function_calls`)
2. the argument name for how big each batched call to the function should be (currently `batch_size`)
3. the name of the file where the acquisition function convertor is defined (currently `batch.py`)

What are your suggestions for these?

docs/notebooks/recovering_from_errors.pct.py
uri-granta: Yeah, batches of 5-10 would be very inefficient. Remember this is the overall size of the tensor, not the leading dimension size.

trieste/acquisition/batch.py
vpicheny: batch_size -> split_size?

docs/notebooks/recovering_from_errors.pct.py
hstojic: ah, I completely misunderstood it then, I thought this was batching optimization runs, rather than predictions...much clearer now :)

trieste/acquisition/batch.py
hstojic: 1. `split_acquisition_function_evaluations` and `split_acquisition_function_evaluation_calls`
2. `split_size`
3. these are good candidates for a `util.py` file, we probably have some other candidates in acqusition package, otherwise I would be tempted to leave it in the `optimizer.py`

uri-granta:
> Can we still have an informative error message if the whole process crashes?

There are (I think) four ways an out of memory error can express itself. If you're lucky, the library that tried to allocate too much memory notices and raises a MemoryError exception. I've extended our generic error log in BayesianOptimizer.optimize to give some relevant info if this happens.

The other three ways are: grinding to a halt due to disk swapping, segfaulting due to a library not handling the memory error properly, and being killed by the OS for using too much memory or taking too long. In those cases there's nothing we can do beyond mentioning memory errors in the documentation, as any code to print a useful message will not be run.

uri-granta:
> could you run a realistic memory test with it as well? you can use PredictiveVariance integration test in test_active_learning by increasing the initial data size to something like 200-300

Just tried that and it did reduce the maximum memory usage a fair bit (while still completing).

PR #499, opened 2022-02-09
Remove superfluous super calls

uri-granta:
Remove unnecessary (and in some cases — e.g. when the parent is a protocol — incorrect) calls to `super().__init__`. In general you should only call an explicitly defined init method in a super class.


hstojic (approved):
looks good

PR #500, opened 2022-02-10
Actually parallel "parallel Thompson sampling"

henrymoss:
I have added a new rule for parallel batch continuous Thompson sampling. Unlike our `GreedyContinuousThompsonSampling` builder which builds batches sequentially, the new one builds batches in parallel using our vectorized acq function optimizer. 

I also included a nice little helper function for flattening and expanding leading dims of tensors courtesy of @vdutor . It was super useful for this PR.

Ive added plenty of tests for the new functionality, as well as adding some for  `GreedyContinuousThompsonSampling`  which seemed not to be tested? Woops


trieste/acquisition/function/__init__.py
```diff
@@ -22,7 +22,10 @@
     integrated_variance_reduction,
     predictive_variance,
 )
-from .continuous_thompson_sampling import GreedyContinuousThompsonSampling
+from .continuous_thompson_sampling import (
```
hstojic, line 25: I wonder whether "parametrized" is a better description than "continuous"...

trieste/models/interfaces.py
```diff
@@ -584,9 +584,9 @@ def reset_sampler(self) -> None:
 Type alias for trajectory functions. These have essentially the same behavior as an
 :const:`AcquisitionFunction` but have additional sampling properties.
 
-An :const:`TrajectoryFunction` evaluates a particular sample at a set of `N` query
-points (each of dimension `D`) i.e. takes input of shape `[N, 1, D]` and returns
-shape `[N, 1]`.
+An :const:`TrajectoryFunction` evaluates `B` samples, each across different sets of `N` query
```
hstojic:
```suggestion
An :const:`TrajectoryFunction` evaluates a batch of `B` samples, each across different sets of `N` query
```

trieste/utils/misc.py
```diff
@@ -16,7 +16,7 @@
 from abc import ABC, abstractmethod
 from time import perf_counter
 from types import TracebackType
-from typing import Any, Callable, Generic, Mapping, NoReturn, Optional, Type, TypeVar
+from typing import Any, Callable, Generic, Mapping, NoReturn, Optional, Tuple, Type, TypeVar
```
hstojic, line 19: you shouldn't need Tuple, with `from __future__ import annotations` above you can use `tuple` instead
```suggestion
from typing import Any, Callable, Generic, Mapping, NoReturn, Optional, Type, TypeVar
```

trieste/utils/misc.py
```diff
@@ -235,3 +235,21 @@ def __exit__(
     ) -> None:
         self.end = perf_counter()
         self.time = self.end - self.start
+
+
+def flatten_leading_dims(x: TensorType) -> Tuple[TensorType, Callable[[TensorType], TensorType]]:
```
hstojic, line 240:
```suggestion
def flatten_leading_dims(x: TensorType) -> tuple[TensorType, Callable[[TensorType], TensorType]]:
```

trieste/utils/misc.py
```diff
@@ -235,3 +235,21 @@ def __exit__(
...
+    """
+    Flattens the leading dimensions of `x`, and returns a function that can be used to restore them.
```
hstojic:
```suggestion
    Flattens the leading dimensions of ``x``, and returns a function that can be used to restore them.
```

trieste/models/gpflow/sampler.py
```diff
@@ -449,24 +452,47 @@ def __init__(
         """
         self._feature_functions = feature_functions
         self._weight_distribution = weight_distribution
-        self._theta_sample = tf.Variable(self._weight_distribution.sample(1))  # sample weights
+        self._initialized = tf.Variable(False)
```
hstojic, line 455: perhaps move this to the interface, similar to what we have for `ReparametrizationSampler`?

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -82,34 +82,86 @@ def update_acquisition_function(
...
+        new_function = self._trajectory_sampler.update_trajectory(function)
 
-            function = negated_trajectory
+        if new_function is not function:
+            function = negate_trajectory_function(new_function)
 
         return function
+
+
+def negate_trajectory_function(function: TrajectoryFunction) -> TrajectoryFunction:
```
hstojic, line 151: seems like this should be a private function with leading _, no?

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -82,34 +82,86 @@ def update_acquisition_function(
...
+class ParallelContinuousThompsonSampling(
+    SingleModelVectorizedAcquisitionBuilder[HasTrajectorySampler]
+):
+    r"""
+    Acquisition function builder for performing parallel continuous Thompson sampling.
+
+    This builder provides broadly the same behavior as our :class:`GreedyContinuousThompsonSampler`
+    however optimizes trajectory samples in parallel rather than sequentially.
+    Consequently, :class:`ParallelContinuousThompsonSampling` can choose query points faster
+    than  :class:`GreedyContinuousThompsonSampler` however it has much larger memory usage.
```
hstojic, line 104: perhaps point to the new "throttled" stuff added by Uri to deal with memory?

hstojic (approved):
looks good, few minor comments

trieste/acquisition/function/continuous_thompson_sampling.py
henrymoss: yep good spot

tests/integration/test_bayesian_optimization.py
```diff
@@ -188,6 +189,13 @@ def GPR_OPTIMIZER_PARAMS() -> Tuple[
                     num_query_points=5,
                 ),
             ),
+            (
+                10,
+                EfficientGlobalOptimization(
+                    ParallelContinuousThompsonSampling(),  # type: ignore[arg-type]
```
uri-granta, line 195: Did you remember to manually kick off the develop-checks job to make sure the slowtests pass?

trieste/models/gpflow/sampler.py
uri-granta: Probably fine as is, since:

1. TrajectoryFunctionClass isn't actually the interface, TrajectoryFunction is
2. We're not implementing a reset_trajectory method like in ReparametrizationSampler

trieste/models/gpflow/sampler.py
```diff
@@ -449,24 +452,47 @@ def __init__(
...
+            by calling the get_trajectory method of the trajectory sampler.
+            """,
         )
-        x = tf.squeeze(x, -2)  # [N, d]
-        feature_evaluations = self._feature_functions(x)  # [N, m]
-        return tf.matmul(feature_evaluations, self._theta_sample, transpose_b=True)
+
+        flat_x, unflatten = flatten_leading_dims(x)  # [N*B, d]
+        flattened_feature_evaluations = self._feature_functions(flat_x)  # [N*B, m]
+        feature_evaluations = unflatten(flattened_feature_evaluations)  # [N, B, m]
```
uri-granta, line 486: cute

trieste/utils/misc.py
uri-granta: Maybe an extra few words explain why this is useful?

```suggestion
    Flattens the leading dimensions of `x`, and returns a function that can be used to restore them
    (typically after first manipulating the flattened tensor).
```

trieste/utils/misc.py
```diff
@@ -235,3 +235,21 @@ def __exit__(
...
+    """
+    x_batched_shape = tf.shape(x)
+    batch_shape = x_batched_shape[:-1]
+    input_shape = x_batched_shape[-1:]
+    x_flat_shape = tf.concat([[-1], input_shape], axis=0)
+
+    def unflatten(y: TensorType) -> TensorType:
+        y_flat_shape = tf.shape(y)
```
uri-granta, line 253: Should this have some sort of shape assert before it tries unflattening?

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -82,34 +82,86 @@ def update_acquisition_function(
...
+def negate_trajectory_function(function: TrajectoryFunction) -> TrajectoryFunction:
+    """
+    Return the negative of trajectories so that our acquisition optimizers (which are
+    all maximizers) can be used to extract the minimizers of trajectories.
+
+    We negate the trajectory function object's call method but otherwise leave it alone,
+    as it may have e.g. update and resample methods.
+    """
+
+    class NegatedTrajectory(type(function)):  # type: ignore[misc]
```
uri-granta: Did you intentionally remove the `if isinstance(function, TrajectoryFunctionClass): ... else: ...` logic? Your current function won't work with a user-defined function (which is a valid TrajectoryFunction), only a class.

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -82,34 +82,86 @@ def update_acquisition_function(
             for the current step. Defaults to ``True``.
         :return: A new trajectory sampled from the model.
         """
-        tf.debugging.Assert(self._trajectory_sampler is not None, [])
+        tf.debugging.Assert(isinstance(function, TrajectoryFunctionClass), [])
```
uri-granta: How do you know this? Trajectory samplers returned by get_trajectory are defined as TrajectoryFunction (any function-like), not necessarily TrajectoryFunctionClass.

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -82,34 +82,86 @@ def update_acquisition_function(
...
 
         return function
 
-    def _negate_trajectory_function(self, function: TrajectoryFunction) -> TrajectoryFunction:
 
-        if isinstance(function, TrajectoryFunctionClass):
-            # we want to negate the trajectory function object but otherwise leave it alone,
-            # as it may have e.g. update and resample methods
+class ParallelContinuousThompsonSampling(
+    SingleModelVectorizedAcquisitionBuilder[HasTrajectorySampler]
```
uri-granta, line 96: So is this class identical to GreedyContinuousThompsonSampling apart from the fact that it inherits from SingleModelVectorizedAcquisitionBuilder rather than SingleModelGreedyAcquisitionBuilder? If so it might be nicer to define a ContinuousThompsonsSampling mixin with the prepare, update (and possibly negate) methods, and then just have:

```python
class ParallelContinuousThompsonSampling(
    SingleModelVectorizedAcquisitionBuilder[HasTrajectorySampler],
    ContinuousThompsonSampling,
)
    r""" [docstring] """
```

tests/unit/utils/test_misc.py
```diff
@@ -99,3 +99,10 @@ def test_Timer_with_nesting() -> None:
             sleep(sleep_time)
     npt.assert_allclose(timer_1.time, 2.0 * sleep_time, rtol=0.01)
     npt.assert_allclose(timer_2.time, 1.0 * sleep_time, rtol=0.01)
+
+
+def test_flatten_leading_dims() -> None:
```
uri-granta, line 104: This should be a bit more thorough (e.g. check dimensions). Right now the following function would pass the test!

```
def flatten_leading_dims(x): return x, lambda x: x
```

Also, maybe a few more tests? e.g.

1. basic errors on flattening
2. check that you can unflatten a modified array with the correct shape
3. basic errors on unflattening

tests/unit/acquisition/function/test_function.py
```diff
@@ -704,6 +704,3 @@ def test_multiple_optimism_negative_confidence_bound_raises_for_changing_batch_s
     with pytest.raises(tf.errors.InvalidArgumentError):
         query_at = tf.reshape(tf.linspace([[-10]], [[10]], 100), [5, 10, 2])
         acq_fn(query_at)
-
-
-# check builder update with different acq
```
uri-granta, line 709: Good spot :-)

uri-granta (commented):
A few comments/questions

tests/integration/test_bayesian_optimization.py
henrymoss: yeh boi

trieste/acquisition/function/continuous_thompson_sampling.py
henrymoss: will add back!

trieste/acquisition/function/continuous_thompson_sampling.py
henrymoss: Good point will fix

trieste/acquisition/function/continuous_thompson_sampling.py
henrymoss: the update methods are very different

PR #506, opened 2022-02-15
Use library stubs with mypy

uri-granta:
Turns out the types job on github didn't install libraries, so was missing various errors. Fix that (and also upgrade to gpflow 2.3, which includes a typing fix that was exposed).


hstojic (approved):
lgtm

PR #507, opened 2022-02-16
Start fixing typing issues hidden by lack of numpy stubs

uri-granta:
Fixing a few typing errors currently hidden by the lack of numpy stubs. We'll want to fix these before upgrade to a more recent numpy version.

The main remaining issue is our careless use of `TensorType = Union[np.ndarray, tf.Tensor]`, where we often assume that inputs and outputs are `tf.Tensors` and don't bother handling the `np.ndarray` case. One possible fix is to drop support for ndarrays by redefining `TensorType = tf.Tensor`, and then (if we need to) add it back in some of the higher level interfaces.


hstojic (approved):
looks good to me

as discussed previously to me it seems like we should switch to tf.Tensor, its probably a rare case that we deal with numpy arrays

PR #510, opened 2022-02-18
Fix broken slow test

henrymoss:
My PR yesterday broke two of the slow integration tests (they seemed to work on my machine!).

I have fixed the two failing tests by :

1. adding in some more optimization steps to the new SVGP + TS test to make it more robust
2. Removing the discrete TS with trajecotry sampler test. This one has always been super flakey and is just a rubbish version of the two new cts TS.  Changing the number of BO steps and  discretization size allow you to pass on either the branin or quadratic tests, but not reliably (across machines) on both.


hstojic (approved):
looks good

PR #512, opened 2022-02-18
Fix copy-paste typos

apaleyes:
Found two typos worth fixing. One is very minor, while the other was actually hiding a test that didn't run


henrymoss (approved):
LGTM. Both good spots!

I actually corrected the first one already in my pending PR. But feel free to merge here

PR #513, opened 2022-02-21
Fix SearchSpace.dimension typing

### LINKED ISSUE #509: typing error not caught?
@sebastianober noted that in search spaces dimension is commonly defined as 
```
@property
    def dimension(self) -> int:
        """The number of inputs in this search space."""
        return self._dimension
```
now, even though the output is actually a tensor, as  it is always computed as 
```
```
no typing error is reported by mypy?

uri-granta:
As noted in https://github.com/secondmind-labs/trieste/issues/509, the typing of `SearchSpace.dimension` is currently wrong. However, the absence of typing stubs for tensorflow means mypy doesn't catch this.


hstojic (approved):
looks good, thanks Uri :)

sebastianober (approved):
Thanks Uri, LGTM :)

Issue #401, opened 2021-11-02
Improve monitoring support


uri-granta:
Since https://github.com/secondmind-labs/trieste/pull/370 we have basics support for monitoring scalar metrics using Tensorboard. It would be good to:

* add more default scalar metrics
* make customising metrics simpler, both adding new ones and disabling existing ones
* log experiment configuration details (including toolbox version) to the same directory
* investigate any other non-scalar metrics (histograms?) or logs we may wish to monitor
* investigate visualising computation graph

henrymoss:
We would also like textual logging with Tensorboard (https://www.tensorflow.org/tensorboard/text_summaries)

For example, https://github.com/secondmind-labs/trieste/issues/440

vpicheny:
When we do greedy batch acquisitions, we log `EGO.acquisition_function.maximum_found` for each point in the batch, which makes the tensorboard cluttered with rather useless info. 

A simple fix could be to group these by name so they appear as a single line? Alternatively, we could decide to not log this quantity for batches by default, or only if the batch size is smaller than, say, 5.

vpicheny:
We currently don't log the query points. And we only log the best observation of the batch (which is convenient because it becomes a scalar, but then we don't have the full information). This is possibly the most important stuff, so we definitely need to do something.

For batch-one strategies, we may want to record the query points (as a series of scalars, grouped under "query point" name). 

For batches, we could record each component still, provided that we can group them neatly to avoid having a very large tensorboard. We might want to cap the number of plots? e.g. if `input_dim * num_query_points > 20` (I put 20 a bit randomly, depends on how it looks perhaps) we don't log those values.

A nice way of representing query points is using `pairplots`. We could consider logging these, with say different colours for the past observations and the new batch. Maybe this incurs some overhead, in that case we should make it optional?

Lastly, for large batches (say, >10) the way to go could be histograms, for each coordinate of the query points and for the new observations, see the image attached. 

![image](https://user-images.githubusercontent.com/9933795/154260745-9c4a886d-210f-4ed8-b398-781cdcfafc92.png)

vpicheny:
minor point: `OBJECTIVE.observation.best_new` is not straightforward to understand, replace by `(...).best_in_batch`?

hstojic:
I would also add some metadata, e.g. run info like num initial points, budget size, batch size, model and acq rule/function used with all the arguments

when we have several runs and want to compare them its easy to lose sense on which exps was done with what variant

[...]

PR #514, opened 2022-02-22
Improve tensorboard monitoring

uri-granta:
Some initial improvements, based on suggestions at https://github.com/secondmind-labs/trieste/issues/401:

1. Group related plots together (wallclock, observations, etc)
2. Plot query points, either as one scalar plot per dimension (if batch=1) or one histogram per dimension (if batch>1)
3. Rename best_new to best_in_batch
4. Rename .i to [i] when plotting specific elements
4. Add wallclock plotting (plus the changes above) to ask_tell


trieste/ask_tell_optimization.py
```diff
@@ -368,15 +368,30 @@ def ask(self) -> TensorType:
...
 
         if callable(points_or_stateful):
             self._acquisition_state, query_points = points_or_stateful(self._acquisition_state)
         else:
             query_points = points_or_stateful
 
+        summary_writer = get_tensorboard_writer()
+        if summary_writer:
+            with summary_writer.as_default(step=get_step_number()):
+                if tf.rank(query_points) == 2:
```
henrymoss, line 384: why wouldnt this be true?

trieste/bayesian_optimizer.py
```diff
@@ -492,39 +492,45 @@ def optimize(
...
-                                np.min(datasets[tag].observations),
-                                step=step,
+                            tf.summary.histogram(
+                                f"{tag}.observation/new_observations",
+                                tagged_output[tag].observations,
                             )
                             tf.summary.scalar(
-                                f"{tag}.observation.best_new",
+                                f"{tag}.observation/best_in_batch",
                                 np.min(tagged_output[tag].observations),
```
henrymoss, line 505: It doesnt really make sense to define the best like this when observations are noisy.  Perhaps its better to name this "best_obeservation_in_batch" just to be clear

henrymoss (approved):
LGTM

trieste/ask_tell_optimization.py
uri-granta: Because the type system doesn't tell me so :-)

Seriously: does it _have_ to be true (i.e. do we assume/assert it anywhere) or is just always true in practice?

trieste/bayesian_optimizer.py
```diff
@@ -492,39 +492,45 @@ def optimize(
...
-                    with summary_writer.as_default():
+                    with summary_writer.as_default(step=step):
                         for tag in datasets:
                             with tf.name_scope(f"{tag}.model"):
                                 models[tag].log()
-                            tf.summary.scalar(
-                                f"{tag}.observation.best_overall",
-                                np.min(datasets[tag].observations),
-                                step=step,
+                            tf.summary.histogram(
```
hstojic, line 499: histogram makes sense only for batches, no?

hstojic (commented):
looks ok to me

what I found needed was recording some info about the run, like size of the data at the beginning (i.e. num initial points), budget aka number of steps, and information about the model and acq rule - this can all be filed as text, can serve as a useful metadata when comparing several runs

trieste/bayesian_optimizer.py
uri-granta: I mean it still works for batch size one (though it doesn't provide any more information that the best_in_batch scalar plot).

uri-granta:
> what I found needed was recording some info about the run, like size of the data at the beginning (i.e. num initial points), budget aka number of steps, and information about the model and acq rule - this can all be filed as text, can serve as a useful metadata when comparing several runs

good suggestion, though i'd rather add text summaries in a separate pr

PR #515, opened 2022-02-23
More TensorBoard monitoring improvements

uri-granta:
- Added the ability to filter out tensorboard summaries. By default we filter out any whose name (at any scope) starts with "_".
- Add query point pairplot (disabled by default, and only generated if seaborn is present)
- Add metadata text summary
- Update tutorial


trieste/bayesian_optimizer.py
```diff
@@ -22,15 +22,22 @@
...
+from typing import Dict, Generic, Optional, TypeVar, cast, overload
 
+import absl
 import numpy as np
 import tensorflow as tf
-from absl import logging
 
+try:
+    import pandas as pd
+    import seaborn as sns
```
hstojic, line 33: so we are introducing (optional) dependencies for this?

uri-granta: Yes. And if you enable any summaries that require seaborn but don't have it installed you get a warning message telling you this.

hstojic: is it possible to have these in setup.py as optional?

trieste/bayesian_optimizer.py
```diff
@@ -446,9 +453,27 @@ def optimize(
...
+        summary_writer = logging.get_tensorboard_writer()
+        if summary_writer:
+            with summary_writer.as_default(step=0):
+                logging.text(
+                    "metadata",
+                    f"Number of steps: `{num_steps}`\n\n"
+                    f"Number of initial points: "
+                    f"`{dict((k, len(v)) for k,v in datasets.items())}`\n\n"
+                    f"Acquisition rule:\n\n    {acquisition_rule}\n\n"
+                    f"Models:\n\n    {models}",
```
hstojic: another would be batch size, perhaps some info about the observer as well (e.g. dimensionality, not sure what else can be inferred from the observer, there is no function name there)

trieste/logging.py
```diff
@@ -72,7 +89,9 @@ def set_step_number(step_number: int) -> None:
 
 def get_step_number() -> int:
     """
-    Set an optimization step number to use for logging purposes.
+    Get the optimization step number used for logging purposes.
+
+    :return: current step number.
```
hstojic, line 94:
```suggestion
    :return: Current step number.
```

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -62,16 +62,24 @@
...
+# ## Deciding what to log
+#
+# When monitoring is enabled, Trieste decides what information is interesting enough to log. This includes objective and acquisition function values, query points, and some model parameters. Some data summaries, such as query point pairplots, are not generated by default as they can affect running time. You can control which summaries to generate and display by overriding the default summary filter. By default this shows any summary apart from those whose name starts with an underscore character.
+
+# %%
+trieste.logging.set_summary_filter(lambda name: True)  # enabe all summaries
+
+
 # %% [markdown]
 # ## Logging additional model parameters
```
hstojic, line 82: perhaps make one if these experiments with a batch size larger than 1, to illustrate that functionality
we should also mention that some of the summaries require Pandas and seabourn libraries

hstojic (commented):
some small comments, but otherwise looks ok
filtering is a bit fuzzy to me though to be honest...

trieste/bayesian_optimizer.py
hstojic: what I want to say, its hard to see this dependency unless you run into this error

uri-granta: Good suggestion. I already knew poetry supported this, but looks like setup.py does too via something like this. I'll add it and check it works.

```python
    extras_require = { "plotting": ["seaborn"] }
```

hstojic (approved):
looks good!

PR #516, opened 2022-02-25
Support ExpectedFeasibility with LocalPenalization

### LINKED ISSUE #503: ExpectedFeasibility cannot be used with LocalPenalization
As pointed out by @vpicheny this doesn't work out of the box because EF can return negative values while LP only works with positive acquisition functions (such as EI). The simple hack is to transform EF using log(1 + exp(x)).

I think this transformation should not affect the acq function, we would only add a small computational overhead, so I would vote for modifying the EF

uri-granta:
See https://github.com/secondmind-labs/trieste/issues/503

### DIFF
--- a/tests/integration/test_active_learning.py
+++ b/tests/integration/test_active_learning.py
@@ -115,7 +115,7 @@ def test_optimizer_learns_scaled_branin_function(
 
 
 @random_seed
-# @pytest.mark.slow
+@pytest.mark.slow
 @pytest.mark.parametrize(
     "num_steps, acquisition_rule, threshold",
     [
--- a/tests/integration/test_active_learning.py
+++ b/tests/integration/test_active_learning.py
@@ -26,12 +26,14 @@ import tensorflow as tf
 import tensorflow_probability as tfp
 
 from tests.util.misc import random_seed
+from trieste.acquisition import LocalPenalization
 from trieste.acquisition.function import (
     BayesianActiveLearningByDisagreement,
     ExpectedFeasibility,
     IntegratedVarianceReduction,
     PredictiveVariance,
 )
+from trieste.acquisition.function.function import MakePositive
 from trieste.acquisition.rule import AcquisitionRule, EfficientGlobalOptimization
 from trieste.bayesian_optimizer import BayesianOptimizer
 from trieste.data import Dataset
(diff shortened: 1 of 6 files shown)


tests/integration/test_active_learning.py
```diff
@@ -113,28 +115,59 @@ def test_optimizer_learns_scaled_branin_function(
 
 
 @random_seed
-@pytest.mark.slow
+# @pytest.mark.slow
```
hstojic: ?

tests/unit/acquisition/function/test_function.py
```diff
@@ -704,3 +706,29 @@ def test_multiple_optimism_negative_confidence_bound_raises_for_changing_batch_s
     with pytest.raises(tf.errors.InvalidArgumentError):
         query_at = tf.reshape(tf.linspace([[-10]], [[10]], 100), [5, 10, 2])
         acq_fn(query_at)
+
+
+@pytest.mark.parametrize("in_place_update", [False, True])
```
hstojic, line 1052: whats this in place about?

trieste/acquisition/function/function.py
```diff
@@ -798,3 +798,65 @@ def __call__(self, x: TensorType) -> TensorType:
         mean, variance = self._model.predict(x)  # [..., B, 1]
         mean, variance = tf.squeeze(mean, -1), tf.squeeze(variance, -1)
         return -mean + tf.sqrt(variance) * self._betas  # [..., B]
+
+
+class MakePositive(SingleModelAcquisitionBuilder[ProbabilisticModelType]):
+    r"""
+    Converts an acquisition function builder into one that only returns positive values, via
+    :math:`x \mapsto \log(1 + \exp(x))`.
```
hstojic, line 1086:
```suggestion
    :math:`x \mapsto \log(1 + \exp(x))`.
    
    This is sometimes a useful transformation, for example to convert
    non-batch acquisition functions into batch acquisition functions with
    local penalization requires acquisition functions that return positive
    values. 
```

trieste/acquisition/function/function.py
```diff
@@ -798,3 +798,65 @@ def __call__(self, x: TensorType) -> TensorType:
         mean, variance = self._model.predict(x)  # [..., B, 1]
         mean, variance = tf.squeeze(mean, -1), tf.squeeze(variance, -1)
         return -mean + tf.sqrt(variance) * self._betas  # [..., B]
+
+
+class MakePositive(SingleModelAcquisitionBuilder[ProbabilisticModelType]):
```
hstojic, line 1083: this class should be imported in `__init__` files in this dir and the parent one, right?

hstojic (approved):
looks great, few small comments

tests/integration/test_active_learning.py
uri-granta: oops

tests/unit/acquisition/function/test_function.py
uri-granta: This checks that MakePositive's update method works both when the underlying update method updates the acquisition function "in place" (when we don't have to regenerate the wrapper) and when the underlying update method creates a new acquisition function (when we do).

PR #517, opened 2022-02-26
try_get_optimal_point helper method

### LINKED ISSUE #453: Can we make it easier to get the best point from the optimizer?

As a user I am often only really interested in the single point where my functions has the optimal value. However, when I run `BayesianOptimizer.optimize` it returns an `OptimizationResult`, which can give me the entire dataset, or the final models, and I have to sort through those myself to find the optimal point.

Can we add methods to `OptimizationResult` for getting the optimal/best point directly?

Something like:

```python
result = bo.optimize(num_steps, initial_data, model)
optimal_x = result.try_get_optimal_point()
```

N/A

N/A

**Additional context**
N/A

uri-granta:
Add a helper method for getting the best point from the optimizer.

### DIFF
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -125,19 +125,19 @@ dataset = result.try_get_final_dataset()
 # We can now get the best point found by the optimizer. Note this isn't necessarily the point that was last evaluated.
 
 # %%
-query_points = dataset.query_points.numpy()
-observations = dataset.observations.numpy()
+query_point, observation, arg_min_idx = result.try_get_optimal_point()
 
-arg_min_idx = tf.squeeze(tf.argmin(observations, axis=0))
-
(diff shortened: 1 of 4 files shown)


docs/notebooks/expected_improvement.pct.py
```diff
@@ -125,20 +125,20 @@ def build_model(data):
 # We can now get the best point found by the optimizer. Note this isn't necessarily the point that was last evaluated.
 
 # %%
-query_points = dataset.query_points.numpy()
-observations = dataset.observations.numpy()
-
-arg_min_idx = tf.squeeze(tf.argmin(observations, axis=0))
+query_point, observation, arg_min_idx = result.try_get_optimal_point()
```
hstojic, line 128: this can probably be used in other notebooks as well, can we propagate it there as well?

hstojic (approved):
looks good to me

PR #518, opened 2022-02-26
Define mulitplication on mixed spaces and equality between spaces

### LINKED ISSUE #369: Allow default __mul__ behaviour of disrete and continuous to provide a TaggedProductSpace
Allow default __mul__ behaviour of disrete and continuous to provide a TaggedProductSpace

### LINKED ISSUE #388: equality between search space objects
Add functionality for defining equality between different search spaces (i.e. same type, same bounds for Boxes, same points for discrete, same subspaces for tagged).

uri-granta:
- Allow mulitplication between arbitrary search spaces. Spaces of the same type will always create a new space of that type. Spaces of different types will create a tagged product space instead.
- Define equality between search spaces. Spaces compare as equal if they're of the same type and have the same limits/points/tags/etc. **Note that discrete spaces are equal even if their points are in a different order.**


henrymoss (approved):
LGTM

PR #519, opened 2022-02-28
HIPPO: HIghly Parallelizable Pareto Optimization

apaleyes:
Welcome Hippo (and thanks Henry for suggesting the name)! It is a penalization method for collecting cheap batches in multi-objective optimization setting. The core idea is to penalize a point by its distance to already collected batch. The details shall be clear in comments and docstrings, so please flag if they aren't!

We have a pretty good idea of Hippo's performance, see this link for some benchmarking (MO LP was the working title): https://github.com/apaleyes/trieste/blob/mo-lp/docs/notebooks/mo_lp/plot_results.ipynb 

Implementation detail. Since we want Hippo to work with both EHVI and ECHVI, it isn't implemented as a `SingleModel...` builder, instead we ask the user to provide the tag to be penalized, and use `OBJECTIVE` by default.


tests/integration/test_multi_objective_bayesian_optimization.py
```diff
@@ -19,6 +19,7 @@
 from trieste.acquisition import (
     BatchMonteCarloExpectedHypervolumeImprovement,
     ExpectedHypervolumeImprovement,
+    HippoPenalizationAcquisitionFunction,
```
henrymoss: I would capitalize HIPPO, its an acronym

tests/integration/test_multi_objective_bayesian_optimization.py
```diff
@@ -72,6 +73,16 @@
...
         ),
+        pytest.param(
+            10,
+            EfficientGlobalOptimization(
+                HippoPenalizationAcquisitionFunction(),
+                num_query_points=4,
+                optimizer=generate_continuous_optimizer(num_initial_samples=500),
+            ),
+            -3.2095,
+            id="qehvi_vlmop2_q_4",
```
henrymoss: give it a better id>?

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
...
             self._expected_improvement_fn.update(_partition_bounds)  # type: ignore
+
+
+class HippoPenalizationAcquisitionFunction(GreedyAcquisitionFunctionBuilder[ProbabilisticModel]):
+    r"""
+    HIPPO: HIghly Parallelizable Pareto Optimization
+
+    Builder of the acquisition function for greedily collecting batches by HIPPO
+    penalization in multi-objective optimization. The resulting acquistion function
+    takes in a set of pending points and returns a base acquisition function
```
henrymoss: a base "multi-objective" acquisition

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
                 isinstance(self._expected_improvement_fn, expected_hv_improvement), []
             )
             self._expected_improvement_fn.update(_partition_bounds)  # type: ignore
+
+
+class HippoPenalizationAcquisitionFunction(GreedyAcquisitionFunctionBuilder[ProbabilisticModel]):
+    r"""
+    HIPPO: HIghly Parallelizable Pareto Optimization
+
+    Builder of the acquisition function for greedily collecting batches by HIPPO
```
henrymoss, line 393: by penalizing batch points by their distance in objective space?

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
...
+    """
+
+    def __init__(
+        self,
+        objective_tag: str = OBJECTIVE,
+        base_acquisition_function_builder: AcquisitionFunctionBuilder[ProbabilisticModel]
+        | None = None,
+    ):
+        """
+        Initializes the HIPPO penalization function.
```
henrymoss: HIPPO acquisition function

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
...
+    penalized around those points.
+
+    Penalization is applied to the acquisition function multiplicatively. However, to
+    improve numerical stability, we perform additive penalization in a log space.
+    """
+
+    def __init__(
+        self,
+        objective_tag: str = OBJECTIVE,
+        base_acquisition_function_builder: AcquisitionFunctionBuilder[ProbabilisticModel]
```
henrymoss: This needs to be a multi-objective acquisition function. I wonder if there is a way to specify this?

henrymoss: In fact, as we only have EHVI, perhaps we should just hard code it and not let the user choose?

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
                 isinstance(self._expected_improvement_fn, expected_hv_improvement), []
             )
             self._expected_improvement_fn.update(_partition_bounds)  # type: ignore
+
+
+class HippoPenalizationAcquisitionFunction(GreedyAcquisitionFunctionBuilder[ProbabilisticModel]):
```
henrymoss: Is there anyway of inheriting from the local penalization builder? It looks very similar to yours

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
...
+                models, datasets
+            )
+        else:
+            self._base_acquisition_function = self._base_builder.update_acquisition_function(
+                self._base_acquisition_function, models, datasets
+            )
+        return self._base_acquisition_function
+
+
+class hippo_penalizer:
```
henrymoss, line 545: Can this inherit from UpdateablePenalizationFunction?

tests/unit/acquisition/multi_objective/test_function.py
```diff
@@ -617,3 +622,119 @@ def prepare_acquisition_function(
...
+@random_seed
+@pytest.mark.parametrize(
+    "base_builder",
+    [
+        ExpectedHypervolumeImprovement().using(""),
+        ExpectedConstrainedHypervolumeImprovement("", _Certainty(), 0.0),
+    ],
+)
+def test_hippo_penalized_acquisitions_combine_base_and_penalization_correctly(
+    base_builder: AcquisitionFunctionBuilder[ProbabilisticModel],
```
henrymoss, line 722: Could you add a test to check retracing? We have some for other acquisition functions

henrymoss (approved):
LGTM, a few comments

tests/integration/test_multi_objective_bayesian_optimization.py
uri-granta: I would also drop the AcquisitionFunction: we don't use it elsewhere (e.g. ExpectedHypervolumeImprovement, GIBBON).

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
                 isinstance(self._expected_improvement_fn, expected_hv_improvement), []
             )
             self._expected_improvement_fn.update(_partition_bounds)  # type: ignore
+
+
+class HippoPenalizationAcquisitionFunction(GreedyAcquisitionFunctionBuilder[ProbabilisticModel]):
+    r"""
+    HIPPO: HIghly Parallelizable Pareto Optimization
```
uri-granta, line 391: Inspired name! Maybe @henrymoss can come up with a backronym for trieste?

tests/integration/test_multi_objective_bayesian_optimization.py
apaleyes: Consider how readable are these:

```
HippoPenalizationAcquisitionFunction
HIPPOPenalizationAcquisitionFunction
```

personally I prefer the former, but maybe it's just me? Aloternatively we can just use `HIPPO` similarly to `GIBBON`

tests/integration/test_multi_objective_bayesian_optimization.py
apaleyes: will do! although not sure what these ids are supposed to do

trieste/acquisition/function/multi_objective.py
apaleyes: We also want to support ECHVI, do we not?

trieste/acquisition/function/multi_objective.py
apaleyes: Alas! `UpdateablePenalizationFunction` makes some LP-specific references, namely

        lipschitz_constant: TensorType,
        eta: TensorType,

trieste/acquisition/function/multi_objective.py
uri-granta: ExpectedHypervolumeImprovement, BatchMonteCarloExpectedHypervolumeImprovement and ExpectedConstrainedHypervolumeImprovement, right?

t's not something that can currently be inferred automatically. We could explicitly tag them with a dummy MultiObjectiveAcquisitionFunction protocol if we wanted to, and then be able to use that here, but not sure it's worth it.

trieste/acquisition/function/multi_objective.py
apaleyes: Not directly, because it has some bits we don't want, like penalizer choice (we don't have a choice) and sampling (which we don't do). But we could maybe consider moving their common bits out?

trieste/acquisition/function/multi_objective.py
apaleyes: sure

trieste/acquisition/function/multi_objective.py
apaleyes: will add

trieste/acquisition/function/multi_objective.py
apaleyes: ok

tests/unit/acquisition/multi_objective/test_function.py
apaleyes: will look into adding one for it

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
...
+        Initializes the HIPPO penalization function.
+
+        :param objective_tag: The tag for the objective data and model.
+        :param base_acquisition_function_builder: Base acquisition function to be
+            penalized. Defaults to Expected Hypervolume Improvement, also supports
+            its constrained version.
+        """
+        self._objective_tag = objective_tag
+        if base_acquisition_function_builder is None:
+            self._base_builder: AcquisitionFunctionBuilder[
```
uri-granta, line 420: simpler like this? (or with `ProbabilisticModelType` if generic)

```suggestion
            self._base_builder = ExpectedHypervolumeImprovement[ProbabilisticModel]().using(self._objective_tag)
```

trieste/acquisition/function/multi_objective.py
uri-granta: I expect you want to make this generic in the type of supported models. That way you can pass in a base builder that doesn't support all model types and still use this for those types of model.

```suggestion
class HippoPenalizationAcquisitionFunction(GreedyAcquisitionFunctionBuilder[ProbabilisticModelType]):
```

trieste/acquisition/function/multi_objective.py
uri-granta: Also if you make this generic, then:

```suggestion
        base_acquisition_function_builder: AcquisitionFunctionBuilder[ProbabilisticModelType]
```

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
...
+                ProbabilisticModel
+            ] = ExpectedHypervolumeImprovement().using(self._objective_tag)
+        else:
+            self._base_builder = base_acquisition_function_builder
+
+        if isinstance(self._base_builder, SingleModelAcquisitionBuilder):
```
uri-granta: The parameter typing currently forbids this. If you want to allow it you'll need to add it to the Union. (You'll probablty also need to make this part of the if-else above, so that self._base_builder has a consistent type. Feel free to ask me if you're confused.

trieste/acquisition/function/multi_objective.py
apaleyes: Btw, if we don't want ECHVI, the code and usage becomes much easier! But from our discussion of the problems to run HIPPO on i was under the impression that we do

trieste/acquisition/function/multi_objective.py
```diff
@@ -379,3 +384,245 @@ def _update_expected_improvement_fn(
...
+        tf.debugging.assert_positive(
+            len(datasets[self._objective_tag]),
+            message=f"{self._objective_tag} dataset must be populated.",
+        )
+        tf.debugging.Assert(self._base_acquisition_function is not None, [])
+
+        if new_optimization_step:
+            self._update_base_acquisition_function(models, datasets)
+
+        if pending_points is None or len(pending_points) == 0:
```
uri-granta, line 495: As @henrymoss noted, this does share a fair amount of code with LocalPenalization. Annoyingly, making this generic in the type will make it a little bit more difficult to just inherit from LocalPenalization. Feel free to prioritise inheriting, and I can always make both classes generic in a separate PR.

tests/integration/test_multi_objective_bayesian_optimization.py
henrymoss: @uri-granta ?

trieste/acquisition/function/multi_objective.py
henrymoss: Yeah you are right. Perhaps only let us pass one of those two bases?

trieste/acquisition/function/multi_objective.py
henrymoss: @uri-granta ?

trieste/acquisition/function/multi_objective.py
henrymoss: ofc. Ignore me

uri-granta: That would allow you to use hippo_penalizer directly with LocalPenalization. Does that make sense? If so, what's the difference between that and using HippoAcqusuisitionFunction?

Note also that the update method for UpdateablePenalizationFunction has lipschitz_constant and eta parameters that we don't need here, but we could always just ignore those.

uri-granta (commented):
Some initial comments/questions. The later ones sometimes affect the earlier ones so please read them all before you make any changes (sorry!).

trieste/acquisition/function/multi_objective.py
apaleyes: Great question! We never considered HIPPO for single objective use case, but perhaps we should, @henrymoss ?

The main differences now are handling of multiple models and multi-objective defaults.

Ignoring parameters is something I would like to avoid if possible. It's very confusing to have a parameter that isn't used. There are precedents in Trieste of parameters being unused and thus documented like that, so it's not the end of the world if we do it here too. But I'd rather not

trieste/acquisition/function/multi_objective.py
apaleyes: BatchMonteCarloExpectedHypervolumeImprovement we don't want. I agree adding mock interfaces here isn't worth it

trieste/acquisition/function/multi_objective.py
apaleyes: thanks!

trieste/acquisition/function/multi_objective.py
apaleyes: no-no, seems clear, thanks

trieste/acquisition/function/multi_objective.py
apaleyes: I don't think we shall inherit at all, because "is a" semantics doesn't seem to apply here. For instance, at the moment we consider HIPPO multi-objective only, while LP is single objective only.

I am more inclined to have a common base which both inherit. But as you say typing might make it very difficult to have such a base. THoughts?

trieste/acquisition/function/multi_objective.py
henrymoss: HIPPO doesnt make any sense for single objective. I think you should keep this separate

tests/integration/test_multi_objective_bayesian_optimization.py
apaleyes: Just HIPPO it is

tests/integration/test_multi_objective_bayesian_optimization.py
apaleyes: fixed now

trieste/acquisition/function/multi_objective.py
apaleyes: I made this change, but the other types stayed as `ProbabilisticModel`, and I cannot fully claim understanding why it would't work with `ProbabilisticModelType` the whole way. Scanning other Trieste code didn't bring any clarity. Can you enlighten me please, where which type is supposed to go?

tests/unit/acquisition/multi_objective/test_function.py
apaleyes: added

trieste/acquisition/function/multi_objective.py
uri-granta: Think of `ProbabilisticModelType` as meaning "a specific type of model", while `ProbabilisticModel` means "any model". Hence making HIPPO a `GreedyAcquisitionFunctionBuilder[ProbabilisticModelType]` is saying it might only support specific model types. Doing this only makes sense if you also change the base_acquisition_function_builder to be `AcquisitionFunctionBuilder[ProbabilisticModelType]`. That allows you to pass in a base builder that only supports certain model types, as long as you only use HIPPO with those model types.

At the very least you'll want to change all mentions of `ProbabilsiticModel` in HIPPO (e.g. in prepare_acquisition_function) to `ProbabilsticModelType`. If you're getting typing errors feel free to message me and I can take a look.

trieste/acquisition/function/multi_objective.py
uri-granta: Actually the main thing that makes it difficult to have a base class is the fact that LocalPenalization is a SingleModelGreedyAcquisitionBuilder while HIPPO is a GreedyAcquisitionFunctionBuilder. I would leave this for now. We can always clean it up later.

apaleyes: Yup, didn't touch it yet, and won't, thanks!

trieste/acquisition/function/multi_objective.py
apaleyes: I am still not 100% clear with this difference. "a specific type of model" - which type is it? That's only known at the time when the user declares a variable of type HIPPO, and then they will replace the generic with a concrete type. That's happening the same way regardless of which generic type we used, no?

In other words, as a user I can write:
```
hippo: HIPPO[TrainableProbabilisticModel] = build_model()
```
regardless of whether HIPPO was declared with `ProbabilisticModelType` or `ProbabilisticModel`. If so, what's the difference between the two?

All that said, we shouldn't restrict HIPPO to any particular type of models, since the goal here is to make it general. Does this mean we shall use `ProbabilisticModel` throughout?

apaleyes: Upon further thinking, I think I get it, @uri-granta. This code would be valid when it should not be?
```
class HIPPO[ProbabilisticModel]:
    self.base_builder: AcquisitionBuilder[ProbabilisticModel]
    
    def __init__(self, base_builder):
        self.base_builder = base_builder

ehvi = EHVI[B]()
hippo = HIPPO[A](envi)
```

whereas if we were to declare HIPPO like that
```
class HIPPO[ProbabilisticModelType]
```

the above code would indeed be caught by mypy as invalid, right? Because the types of base builder and hippo don't match, and `ProbabilisticModelType` controls for it.

Is that understanding correct? If so, I certainly agree we shall use `ProbabilisticModelType` to ensure penalization and base functions are typed the same way

uri-granta: If we don't use `ProbabilisticModelType` then mypy should complain if we try to pass in a base builder like `AugmentedExpectedImprovement` or `PredictiveVariance`, since those builders don't support all `ProbabilisticModel`s.

Would you like me to have a go at updating this so you can see what I mean?

apaleyes: Got it!

We are only interested in multi-objective non-batch builders, which support any models at the moment (EHVI and ECHVI). But clearly we don't want to be restrictive in case later on we have some multi-objective builders that only support a subset of models. I guess it's worth to use `ProbabilisticModelType` here.

Thanks for the offer. I'll give it another go first, watch out for the update (likely tomorrow).

apaleyes: updated

henrymoss:
Okay, so is this ready to go?

apaleyes:
Hopefully! But I'd like @uri-granta to approve before merging, just to make sure the typing is all correct

uri-granta (approved):
Looks good!

PR #521, opened 2022-02-28
Fix remaining numpy mypy issues

### LINKED ISSUE #347: Numpy-related type errors are hidden by the absence of numpy stubs

uri-granta:
This fixes the remaining issues exposed by the numpy stubs introduce in 1.21 and adds a workflow to test this.


tests/unit/acquisition/test_utils.py
```diff
@@ -41,7 +42,7 @@
     ],
 )
 def test_split_acquisition_function(
-    f: AcquisitionFunction, x: np.ndarray, split_size: int, expected_batches: int
+    f: AcquisitionFunction, x: np.ndarray[Any, Any], split_size: int, expected_batches: int
```
hstojic: so here we explicitly state the type of objects in arrays, similar as with tuple, List etc?

hstojic (approved):
looks good to me

tests/unit/acquisition/test_utils.py
uri-granta: Yes: the parameters refer to shape and dtype (though shape isn't properly supported at the moment, and we currently don't handle dtypes well enough to use that part). There's a cleaner shorthand in numpy.typing called `npt.NDArray` but that doens't work with older versions of numpy.

tests/unit/models/conftest.py
```diff
@@ -95,9 +95,8 @@ def _two_layer_model_fixture(request: Any) -> Callable[[TensorType], DeepGP]:
 
 
 # Teardown fixture to set keras floatx to float64 then return it to previous value at test finish
-# pytest uses yield in a funny way, so we use type ignore
-@pytest.fixture(name="keras_float")  # type: ignore
-def _keras_float() -> None:
+@pytest.fixture(name="keras_float")
```
hstojic, line 98: should we then move this fixture up to `tests/conftest.py` since we have it now in multiple places?

uri-granta: :+1: but planning to tidy up the conftests in a separate PR

PR #522, opened 2022-03-01
Update to gpflow 2.4.0

uri-granta:
Update to using gpflow 2.4.0 (which will allow us to use posterior objects to speed up acquisition).


hstojic (approved):
looks good

uri-granta:
> looks good

Wait until you see if the tests pass! :-)

PR #523, opened 2022-03-02
Use GPFlow posterior objects for great good

uri-granta:
Use GPflow posterior objects to speed up acquisition.

There remains some (partly pre-existing) work to ensure deserialised models are consistent with regards data shape and now posterior cache. This will be done as part of the upcoming work on serialisation.


tests/unit/models/gpflow/test_models.py
```diff
@@ -403,6 +403,7 @@ def test_gaussian_process_regression_trajectory_sampler_has_correct_samples(
         gpr_model(x, _3x_plus_gaussian_noise(x)), use_decoupled_sampler=use_decoupled_sampler
     )
     model.model.likelihood.variance.assign(1e-3)
+    model._posterior.update_cache()
```
henrymoss: Its a little worrying that we have to do this here!

henrymoss: Perhaps it would be better to not calculate the posterior object until we make the first prediction?  Then we can do whatever model init stuff we want

trieste/models/gpflow/models.py
```diff
@@ -147,6 +150,16 @@ def _ensure_variable_model_data(self) -> None:
...
+    def predict(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
+        return self._posterior.predict_f(query_points)
+
+    def predict_joint(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
+        return self._posterior.predict_f(query_points, full_cov=True)
+
+    def predict_y(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
+        f_mean, f_var = self.predict(query_points)
+        return self.model.likelihood.predict_mean_and_var(f_mean, f_var)
+
```
hstojic: can we move this to GPflowPredictor? of course if all models are using the posterior

trieste/models/gpflow/models.py
```diff
@@ -508,93 +523,6 @@ def conditional_predict_y(
         return self.model.likelihood.predict_mean_and_var(f_mean, f_var)
 
 
-class NumDataPropertyMixin:
```
hstojic, line 511: nice that we can get rid of this :)

tests/unit/models/gpflow/test_models.py
hstojic: not sure what you mean, do it all or at this specific spot? it doesn't seem like its avoidable

hstojic (approved):
looks good to me

tests/unit/models/gpflow/test_models.py
uri-granta: Good suggestion. Done.

trieste/models/gpflow/models.py
uri-granta: Not all the models do, but still a good suggestion. Done.

tests/unit/models/gpflow/test_models.py
uri-granta: Ok, it turns out that tensorflow compilation stops us from doing this. I'll have a think if there's another way.

trieste/models/gpflow/models.py
uri-granta: Actually turns out that we need to initialise the posteriors at the end of initialisation so can't delegate it to the parent class, which makes refactoring a bit trickier. Might leave as is for now and take a look in a separate PR.

uri-granta: I've had a second go at refactoring it which hopefully won't break all the tests!

tests/unit/models/gpflow/test_models.py
uri-granta: For now I've made `update_posterior_cache` a method and clarified the docstring slightly. Not sure if there's much else we can do, as the object need to exist when we compile the wrapper after initialisation. We could decide to always regenerate the cache the first time we call predict, but that feels wasteful. @jesnie : any ideas?

uri-granta: As an aside, it's worth remembering that you only need to update the cache if you modified the model after creating the model wrapper. You can play with model init stuff all you want as long as you wrap it afterwards.

trieste/models/gpflow/models.py
```diff
@@ -74,6 +65,14 @@ class GaussianProcessRegression(
...
     A :class:`TrainableProbabilisticModel` wrapper for a GPflow :class:`~gpflow.models.GPR`
     or :class:`~gpflow.models.SGPR`.
+
+    As Bayesian optimization requires a large number of sequential predictions (i.e. when maximizing
+    acquisition functions), rather than calling the model directly at prediction time we instead
+    call the posterior objects built by these models. These posterior objects store the
+    pre-computed Gram matrices, which can be reused to allow faster subsequent predictions. However,
+    note that these posterior objects need to be updated whenever the underlying model is changed
+    by calling :meth:`update_posterior_cache` (this
+    happens automatically after calls to :meth:`update` or :math:`optimize`).
```
jesnie, line 75: Maybe move this to the `GPflowPredictor` class?

trieste/models/gpflow/models.py
```diff
@@ -655,18 +581,17 @@ def __init__(
...
-            self._model._num_data = tf.Variable(model.num_data, trainable=False, dtype=tf.float64)
-            self._model.__class__ = SVGPWrapper
+    def _ensure_variable_model_data(self) -> None:
+        # GPflow stores the data in Tensors. However, since we want to be able to update the data
+        # without having to retrace the acquisition functions, put it in Variables instead.
+        # Data has to be stored in variables with dynamic shape to allow for changes
+        # Sometimes, for instance after serialization-deserialization, the shape can be overridden
+        # Thus here we ensure data is stored in dynamic shape Variables
+        if not is_variable(self._model.num_data):
+            self._model.num_data = tf.Variable(self._model.num_data, trainable=False)
```
jesnie, line 594: Do you actually need this? I think I came to the conclusion it wasn't used anywhere? Or is it used outside `Trieste`?

tests/unit/models/gpflow/test_models.py
jesnie: > For now I've made `update_posterior_cache` a method and clarified the docstring slightly. Not sure if there's much else we can do, as the object need to exist when we compile the wrapper after initialisation. We could decide to always regenerate the cache the first time we call predict, but that feels wasteful. @jesnie : any ideas?

Not really. I have been thinking the last couple of weeks, that it would be cool with some kind of framework that dynamically decide what to recompute when variables change, but I don't think there really is a good solution. For the foreseeable future I think we have to manually update stuff whenever a variable is changed.

trieste/models/gpflow/models.py
uri-granta: We update `num_data` in `SparseVariational.update` so need to ensure it's in a variable, no?

jesnie: > We update `num_data` in `SparseVariational.update` so need to ensure it's in a variable, no?

But why do you update it? Do you ever read it? Can you not just leave it be?

uri-granta: Good question. I vaguely remember it being used internally, but looking at the code it doesn't seem to make sense to override it: if we want it to vary with the data size we would just leave it as None.

I'll remove this and see if anything breaks :-)

tests/unit/models/gpflow/test_models.py
henrymoss: So whats the situation with this change?

uri-granta: My two comments above

1. if you modify the model after wrapping it up you have to call update_posterior_cache
2. this is mentioned in the docstring for the wrapper
3. really you shouldn't be modifying the model manually after wrapping it, just before

PR #527, opened 2022-03-09
New TS notebook

henrymoss:
Finally a notebook to show off our scalable Thompson sampling (i.e. sparse GP, decoupled sampling and inducing point management).

Basically just an implementation of our S-GP-TS paper.

### DIFF
--- a/docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
+++ b/docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
@@ -1,0 +1,122 @@
+# -*- coding: utf-8 -*-
+# %% [markdown]
+# # Scalable Thompson Sampling using Sparse Gaussian Process Models
+
+# %% [markdown]
+# In our other [Thompson sampling notebook](thompson_sampling.pct.py) we demonstrate how to perform batch optimization using a traditional implementation of Thompson sampling that samples exactly from an underlying Gaussian Process surrogate model. Unfortunately, this approach incurs a large computational overhead that scales polynomially with the optimization budget and so cannot be applied to settings with larger optimization budgets, e.g. those where large batches (>>10) of points can be collected.
+#
+# Luckily, Trieste also supports a scalable formulation of Thompson sampling. In particular, by coupling a sparse Gaussian process surrogate model with an approximate sampling scheme we can apply Thompson sampling over very large  optimization budgets and batch sizes, i.e. for those settings where standard BO methods are infeasbible. The implementation in this notebook follows [our paper](https://arxiv.org/pdf/2006.05356.pdf).
+
+# %%
+import numpy as np
+import tensorflow as tf
+
+np.random.seed(1793)
+tf.random.set_seed(1793)
+
+# %% [markdown]
+# ## Define the problem and model
+#
+# We'll use a continuous bounded search space, and build a noisy observer. We then collect an initial design of 15 random points.
+
+# %%
+import trieste
+from trieste.objectives import hartmann_6, HARTMANN_6_MINIMUM, HARTMANN_6_SEARCH_SPACE
+from trieste.types import TensorType
+search_space = HARTMANN_6_SEARCH_SPACE
+
+def noisy_hartmann_6(x:TensorType) -> TensorType: # contaminate observations with Gaussian noise
+    return hartmann_6(x)+ tf.random.normal([len(x),1], 0, 0.01, tf.float64) 
+
+num_initial_data_points = 15
+initial_query_points = search_space.sample(num_initial_data_points)
+observer = trieste.objectives.utils.mk_observer(noisy_hartmann_6)
+initial_data = observer(initial_query_points)
+
+# %% [markdown]
+# We'll use a sparse Gaussian process regression to model the function, as implemented in GPflow. The GPflow models cannot be used directly in our Bayesian optimization routines, so we build a GPflow's `SVGP` model using Trieste's convenient model build function `build_svgp` and pass it to the `SparseVariational` wrapper. Note that we also define a `KMeansInducingPointSelector` selector, i.e. we reallocate the 50 inducing points of our `SVGP` model at the start of each BO step to be the centroids of a k-means clustering of the observations. As the optimization progresses, observations are likely to be concentrated in the optimal regions, so clustering provides “targeted” inducing points for BO.
+
+# %%
+from trieste.models.gpflow import SparseVariational, build_svgp, KMeansInducingPointSelector
+from trieste.models.optimizer import BatchOptimizer
+
+gpflow_model = build_svgp(
+    initial_data, 
+    search_space, 
+    likelihood_variance=0.01, 
+    num_inducing_points=50
+)
+
+inducing_point_selector = KMeansInducingPointSelector(search_space)
+
+model = SparseVariational(
+    gpflow_model, 
+    num_rff_features=1_000, 
+    inducing_point_selector=inducing_point_selector,
+    optimizer = BatchOptimizer(tf.optimizers.Adam(0.1), max_iter=100, batch_size =50, compile=True)
+)
+
+
+# %% [markdown]
+# ## Create the Thompson sampling acquisition rule
+#
+# Thompson sampling chooses query points as the minimizers of random samples from the model of our objective
+# function. 
+#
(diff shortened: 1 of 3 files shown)


henrymoss:
[ts_notebook.pdf](https://github.com/secondmind-labs/trieste/files/8214574/ts_notebook.pdf)

Here is a pdf if you cant be bothered to run, but wanna see the pic

docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
```diff
@@ -0,0 +1,122 @@
+# -*- coding: utf-8 -*-
+# %% [markdown]
+# # Scalable Thompson Sampling using Sparse Gaussian Process Models
+
+# %% [markdown]
+# In our other [Thompson sampling notebook](thompson_sampling.pct.py) we demonstrate how to perform batch optimization using a traditional implementation of Thompson sampling that samples exactly from an underlying Gaussian Process surrogate model. Unfortunately, this approach incurs a large computational overhead that scales polynomially with the optimization budget and so cannot be applied to settings with larger optimization budgets, e.g. those where large batches (>>10) of points can be collected.
+#
+# Luckily, Trieste also supports a scalable formulation of Thompson sampling. In particular, by coupling a sparse Gaussian process surrogate model with an approximate sampling scheme we can apply Thompson sampling over very large  optimization budgets and batch sizes, i.e. for those settings where standard BO methods are infeasbible. The implementation in this notebook follows [our paper](https://arxiv.org/pdf/2006.05356.pdf).
```
apaleyes: Why not cite the paper same way we refer to all other publications in the docs?

docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
```diff
@@ -0,0 +1,122 @@
...
+# We'll use a continuous bounded search space, and build a noisy observer. We then collect an initial design of 15 random points.
+
+# %%
+import trieste
+from trieste.objectives import hartmann_6, HARTMANN_6_MINIMUM, HARTMANN_6_SEARCH_SPACE
+from trieste.types import TensorType
+search_space = HARTMANN_6_SEARCH_SPACE
+
+def noisy_hartmann_6(x:TensorType) -> TensorType: # contaminate observations with Gaussian noise
+    return hartmann_6(x)+ tf.random.normal([len(x),1], 0, 0.01, tf.float64)
```
apaleyes:
```suggestion
    return hartmann_6(x) + tf.random.normal([len(x),1], 0, 0.01, tf.float64) 
```

docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
```diff
@@ -0,0 +1,122 @@
...
+    num_inducing_points=50
+)
+
+inducing_point_selector = KMeansInducingPointSelector(search_space)
+
+model = SparseVariational(
+    gpflow_model, 
+    num_rff_features=1_000, 
+    inducing_point_selector=inducing_point_selector,
+    optimizer = BatchOptimizer(tf.optimizers.Adam(0.1), max_iter=100, batch_size =50, compile=True)
```
apaleyes:
```suggestion
    optimizer = BatchOptimizer(tf.optimizers.Adam(0.1), max_iter=100, batch_size=50, compile=True)
```

docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
```diff
@@ -0,0 +1,122 @@
...
+
+# %% [markdown]
+# ## Create the Thompson sampling acquisition rule
+#
+# Thompson sampling chooses query points as the minimizers of random samples from the model of our objective
+# function. 
+#
+# Using a [decoupled sampling scheme](https://arxiv.org/abs/2002.09309), we can build approximate samples from our sparse GP surrogate model at low cost. As we can cheaply evaluate the values and gradients of these approximate samples at any point across the search space, our acquisition function optimizers can be used to find the minimizers of the samples across the whole search space. We can increase the quality of these approximate samples at the expense of computational cost by increasing `num_rff_features` (as specified when defining our model above).
+#
+# We either build batches element by element with `GreedyContinuousThompsonSampling` or allocate a whole batch at once with `ParallelContinuousThompsonSampling`. The later is faster but has a much higher memory usage. Memory usage can be controlled using `split_acquisition_function_calls`, which limits the number of individual evaluations that can be made in parallel (in our case we set this to 100_000).
```
apaleyes:
```suggestion
# We either build batches element by element with `GreedyContinuousThompsonSampling` or allocate a whole batch at once with `ParallelContinuousThompsonSampling`. The latter is faster but has a much higher memory usage. Memory usage can be controlled using `split_acquisition_function_calls` utility, which limits the number of individual evaluations that can be made in parallel (in our case we set this to 100_000).
```

docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
```diff
@@ -0,0 +1,122 @@
...
+
+# %%
+from trieste.acquisition.rule import EfficientGlobalOptimization
+from trieste.acquisition import ParallelContinuousThompsonSampling
+from trieste.acquisition.optimizer import automatic_optimizer_selector
+from trieste.acquisition.optimizer import automatic_optimizer_selector
```
apaleyes: double import. isort would be handy in this section!

apaleyes (commented):
Looks good, and the explanations seem clear. But I have two general comments:
* The final plot clearly shows how unnecessary big the batch size is for this problem. Is there a maybe harder problem we could use for demonstration here?
* How long does this notebook run? Are there any build-time optimizations needed, say batch size of 10 instead of 100?

docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
henrymoss: In notebooks we dont use the cite thing I think. Not sure if it would work? I think this is cleaner for notebooks

apaleyes: I think we do. Have a look this paragraph of the GIBBON notebook: https://secondmind-labs.github.io/trieste/notebooks/batch_optimization.html#Batch-acquisition-functions.

henrymoss: In notebooks we dont use the cite thing I think. Not sure if it would work? I think this is cleaner for notebooks

hstojic: yep, we do use it

docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
```diff
@@ -0,0 +1,122 @@
...
+num_steps = 5
+result = bo.optimize(
+    num_steps, initial_data, model, acq_rule, track_state=False
+)
+dataset = result.try_get_final_dataset()
+
+# %% [markdown]
+# ## Visualising the result
+#
+# By plotting the regret achieved by the algorithm, we see the convergence over five BO steps after the initial design (denoted by a vertical line and cross). Each batch of evaluations getting (dots) achieves better objective values than its predecessor.
```
hstojic: "getting" is perhaps unnecessary?

docs/notebooks/scalable_thompson_sampling_using_sparse_gaussian_processes.pct.py
```diff
@@ -0,0 +1,122 @@
...
+import numpy as np
+import tensorflow as tf
+
+np.random.seed(1793)
+tf.random.set_seed(1793)
+
+# %% [markdown]
+# ## Define the problem and model
+#
+# We'll use a continuous bounded search space, and build a noisy observer. We then collect an initial design of 15 random points.
```
hstojic: can you add a sentence on why a noisy function is needed?

hstojic (approved):
looks great! just a few minor comments...

apaleyes (approved):
Still wonder how illustrative this example actually is, but that's probably not critical. Otherwise looks good!

henrymoss:
> Still wonder how illustrative this example actually is, but that's probably not critical. Otherwise looks good!

Ive boosted the noise level up a lot, so it does require a large number of evals. Looks much better now

PR #528, opened 2022-03-09
Compiled model optimizers for gpflow models

henrymoss:
The default optimizers for our gpflow models now have `compile=True`. Not sure why we didn't do this before, as it speeds model fitting up a lot!

### DIFF
--- a/tests/unit/models/gpflow/test_models.py
+++ b/tests/unit/models/gpflow/test_models.py
@@ -167,7 +167,7 @@ def test_gpflow_wrappers_ref_optimize(gpflow_interface_factory: ModelFactoryType
             ),
         )
         gpflow.optimizers.Scipy().minimize(
-            reference_model.training_loss_closure(compile=False),
+            reference_model.training_loss_closure(compile=True),
             reference_model.trainable_variables,
         )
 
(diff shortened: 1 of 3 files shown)


hstojic (commented):
if we are convinced its a better default lets simply change it in `Optimizer` definition, which at the moment is `False` (in `trieste/models/optimizer.py`)

do @joelberkeley  or @awav know why this was left as `False`, perhaps there is a good reason?

joelberkeley:
hi hrvoje, don't know, don't have much of an opinion

henrymoss:
> 

I think the default should be False for general models. If you implement your own thing, then it might not work. I don't know if it works for the NN stuff for example.

hstojic (approved):
DGP and Keras do not use `Optimizer` so there is no current danger for that - but ok, fair enough

awav:
- Ideally, there should be no hardcoded compile options here
- `Optimiser` has a compile option. The whole `minimize` will be compiled if user set Optimizer's compile option to True or it was set to True by default
- It is better to compile an outer function that consists of other smaller functions, rather than compiling smaller functions first and then the outer function. In this case, `minimise` is the outer function that contains `loss` function which is the sub-function. That was the reason for setting compile to `False`, as it will be part of larger compile anyway. This is subtle, but False was the right thing to do.

awav:
@henrymoss, @hstojic ^^^

PR #531, opened 2022-03-09
sgpr wrapper

### LINKED ISSUE #329: Incorrect covariance_between_points for SGPR
**Describe the bug**
In trieste/models/model_interfaces.py, the `covariance_between_points` method of the `GaussianProcessRegression` class uses the full GP covariance, even when the model is SGPR. 

**To reproduce**
N/A

**Expected behaviour**
When SGPR is used as the internal model, `covariance_between_points` should return the covariance computed using the sparse approximation in SGPR (http://proceedings.mlr.press/v5/titsias09a.html).

**System information**
N/A

**Additional information**
The predictive covariance is written in closed form in Eq. 4 of http://proceedings.mlr.press/v139/artemev21a.html. A reference implementation can be found in `predict_f` in the GPflow implementation: https://gpflow.readthedocs.io/en/master/_modules/gpflow/models/sgpr.html#SGPR

hstojic:
current status
- inducing point management seems to work fine, with the caveat that I didn't look into details of different types of inducing points (SharedIndependentInducingVariables etc) - I assumed these are the same for both SVGP and SGPR in GPflow, so I simply reused these parts of code from SVGP - @henrymoss can you perhaps confirm this?
- I have reused the existing covariance function that we use in SVGP and VGP via the `compute_qu` method in SGPR model which gives us `q_sqrt`, but I still want to check math in the paper that this is the best way forward
- the only thing that doesn't work at the moment is trajectory sampling, I get some weird errors, potentially stemming from shape of `q_sqrt` that we get from `compute_qu` 

note that I did a bit of reshuffling of tests, it was a mess so I grouped them wrt the wrapper they are testing, ignore those tests if you are looking at tests, only sgpr ones are relevant
(I was itching to divide the tests into 5 different files: test_models_shared and then for each wrapper)

### DIFF
--- a/tests/unit/models/gpflow/test_models.py
+++ b/tests/unit/models/gpflow/test_models.py
@@ -280,6 +280,71 @@ def test_gpflow_models_raise_for_pairwise_covariance_for_invalid_query_points(
         model.covariance_between_points(data[0], tf.expand_dims(data[0], axis=0))  # type: ignore
 
 
+@random_seed
+@pytest.mark.parametrize("after_model_optimize", [True, False])
+@pytest.mark.parametrize("after_model_update", [True, False])
+def test_gpflow_models_cached_predictions_correct(
+    after_model_optimize: bool,
+    after_model_update: bool,
+    gpflow_interface_factory: ModelFactoryType,
+) -> None:
+    x = np.linspace(0, 5, 10).reshape((-1, 1))
+    y = fnc_2sin_x_over_3(x)
+    data = x, y
+    dataset = Dataset(*data)
+    model, _ = gpflow_interface_factory(x, y)
+
+    if after_model_optimize:
+        model._optimizer = BatchOptimizer(tf.optimizers.SGD(), max_iter=1)
+        model.optimize(dataset)
+
+    if after_model_update:
+        new_x = np.linspace(0, 5, 3).reshape((-1, 1))
+        new_y = fnc_2sin_x_over_3(new_x)
+        new_dataset = Dataset(new_x, new_y)
+        cast(TrainableProbabilisticModel, model).update(new_dataset)
+
+    x_predict = np.linspace(0, 5, 2).reshape((-1, 1))
+
+    # get cached predictions
+    cached_fmean, cached_fvar = model.predict(x_predict)
+    cached_joint_mean, cached_joint_var = model.predict_joint(x_predict)
+    cached_ymean, cached_yvar = model.predict_y(x_predict)
+
+    # get reference (slow) predictions from underlying model
+    reference_fmean, reference_fvar = model.model.predict_f(x_predict)
+    reference_joint_mean, reference_joint_var = model.model.predict_f(x_predict, full_cov=True)
+    reference_ymean, reference_yvar = model.model.predict_y(x_predict)
+
+    npt.assert_allclose(cached_fmean, reference_fmean)
+    npt.assert_allclose(cached_ymean, reference_ymean)
+    npt.assert_allclose(cached_joint_mean, reference_joint_mean)
+    npt.assert_allclose(cached_fvar, reference_fvar, atol=1e-5)
+    npt.assert_allclose(cached_yvar, reference_yvar, atol=1e-5)
+    npt.assert_allclose(cached_joint_var, reference_joint_var, atol=1e-5)
+    npt.assert_allclose(cached_yvar - model.get_observation_noise(), cached_fvar, atol=5e-5)
+
+
+@random_seed
+def test_gpflow_models_cached_predictions_faster(
+    gpflow_interface_factory: ModelFactoryType,
+) -> None:
+    x = np.linspace(0, 10, 10).reshape((-1, 1))
+    y = fnc_2sin_x_over_3(x)
+    model, _ = gpflow_interface_factory(x, y)
+    n_calls = 100
+
+    x_predict = np.linspace(0, 5, 2).reshape((-1, 1))
+    t_0 = time()
+    [model.predict(x_predict) for _ in range(n_calls)]
+    time_with_cache = time() - t_0
+    t_0 = time()
+    [model.model.predict_f(x_predict) for _ in range(n_calls)]
+    time_without_cache = time() - t_0
+
+    npt.assert_array_less(time_with_cache, time_without_cache)
+
+
 def test_gaussian_process_regression_raises_for_invalid_init() -> None:
     x_np = np.arange(5, dtype=np.float64).reshape(-1, 1)
     x = tf.convert_to_tensor(x_np, x_np.dtype)
@@ -536,71 +601,6 @@ def test_gaussian_process_regression_trajectory_sampler_has_correct_samples(
     npt.assert_allclose(sample_variance[:3], true_variance[:3], rtol=0.15)
 
 
-@random_seed
-@pytest.mark.parametrize("after_model_optimize", [True, False])
-@pytest.mark.parametrize("after_model_update", [True, False])
-def test_gaussian_process_cached_predictions_correct(
-    after_model_optimize: bool,
-    after_model_update: bool,
-    gpflow_interface_factory: ModelFactoryType,
-) -> None:
-    x = np.linspace(0, 5, 10).reshape((-1, 1))
-    y = fnc_2sin_x_over_3(x)
-    data = x, y
-    dataset = Dataset(*data)
-    model, _ = gpflow_interface_factory(x, y)
-
-    if after_model_optimize:
-        model._optimizer = BatchOptimizer(tf.optimizers.SGD(), max_iter=1)
-        model.optimize(dataset)
-
-    if after_model_update:
-        new_x = np.linspace(0, 5, 3).reshape((-1, 1))
-        new_y = fnc_2sin_x_over_3(new_x)
-        new_dataset = Dataset(new_x, new_y)
-        cast(TrainableProbabilisticModel, model).update(new_dataset)
-
-    x_predict = np.linspace(0, 5, 2).reshape((-1, 1))
-
-    # get cached predictions
-    cached_fmean, cached_fvar = model.predict(x_predict)
-    cached_joint_mean, cached_joint_var = model.predict_joint(x_predict)
-    cached_ymean, cached_yvar = model.predict_y(x_predict)
-
-    # get reference (slow) predictions from underlying model
-    reference_fmean, reference_fvar = model.model.predict_f(x_predict)
-    reference_joint_mean, reference_joint_var = model.model.predict_f(x_predict, full_cov=True)
-    reference_ymean, reference_yvar = model.model.predict_y(x_predict)
-
-    npt.assert_allclose(cached_fmean, reference_fmean)
-    npt.assert_allclose(cached_ymean, reference_ymean)
-    npt.assert_allclose(cached_joint_mean, reference_joint_mean)
-    npt.assert_allclose(cached_fvar, reference_fvar, atol=1e-5)
-    npt.assert_allclose(cached_yvar, reference_yvar, atol=1e-5)
-    npt.assert_allclose(cached_joint_var, reference_joint_var, atol=1e-5)
-    npt.assert_allclose(cached_yvar - model.get_observation_noise(), cached_fvar, atol=5e-5)
-
-
-@random_seed
-def test_gaussian_process_cached_predictions_faster(
-    gpflow_interface_factory: ModelFactoryType,
-) -> None:
-    x = np.linspace(0, 10, 10).reshape((-1, 1))
-    y = fnc_2sin_x_over_3(x)
-    model, _ = gpflow_interface_factory(x, y)
-    n_calls = 100
-
-    x_predict = np.linspace(0, 5, 2).reshape((-1, 1))
-    t_0 = time()
-    [model.predict(x_predict) for _ in range(n_calls)]
-    time_with_cache = time() - t_0
-    t_0 = time()
-    [model.model.predict_f(x_predict) for _ in range(n_calls)]
-    time_without_cache = time() - t_0
-
-    npt.assert_array_less(time_with_cache, time_without_cache)
-
-
 def test_gpr_config_builds_and_default_optimizer_is_correct() -> None:
     data = mock_data()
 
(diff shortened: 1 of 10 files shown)


tests/unit/models/gpflow/test_models.py
```diff
@@ -432,46 +596,336 @@ def test_gaussian_process_regression_trajectory_sampler_has_correct_samples(
...
+
+        modeli = GaussianProcessRegression(gpr_model(xi, yi))
+        predj_meani, predj_covi = modeli.predict_joint(query_points)
+        sample_mean = tf.reduce_mean(samples[i], axis=0)
+        sample_cov = tfp.stats.covariance(samples[i, :, :, 0], sample_axis=0)
+        np.testing.assert_allclose(sample_mean, predj_meani, atol=1e-2, rtol=1e-2)
+        np.testing.assert_allclose(sample_cov, predj_covi[0], atol=1e-2, rtol=1e-2)
+
+
+def test_sparse_gaussian_process_regression_raises_for_invalid_init() -> None:
```
hstojic, line 714: from here until line 920 is the new part, I have mostly reused tests from gpr and svgp, I didn't include some of the svgp tests on inducing points that required a two-output model - @henrymoss do you think it would be important to add those?

trieste/models/gpflow/models.py
```diff
@@ -94,7 +93,7 @@ def __init__(
             kernel when calling :meth:`trajectory_sampler`. We use a default of 1000 as it
             typically perfoms well for a wide range of kernels. Note that very smooth
             kernels (e.g. RBF) can be well-approximated with fewer features.
-        :param used_decoupled_sampler: If True use a decoupled random Fourier feature sampler, else
+        :param use_decoupled_sampler: If True use a decoupled random Fourier feature sampler, else
```
hstojic, line 96: its strange that these errors are not caught when building docs, @uri-granta we noticed this before, did you perhaps implement something to catch them?

trieste/models/gpflow/models.py
```diff
@@ -120,10 +119,14 @@ def __init__(
 
     def __repr__(self) -> str:
         """"""
-        return f"GaussianProcessRegression({self.model!r}, {self.optimizer!r})"
+        return (
+            f"GaussianProcessRegression({self.model!r}, {self.optimizer!r},"
+            f"{self._num_kernel_samples!r}, {self._num_rff_features!r},"
+            f"{self._use_decoupled_sampler!r})"
+        )
```
hstojic, line 126: we missed to update repr I think, we want to add all the arguments, right?

tests/unit/models/gpflow/test_models.py
henrymoss: No leave the two outputs for SVGP

trieste/models/gpflow/models.py
henrymoss: might as well

trieste/models/gpflow/models.py
```diff
@@ -507,13 +496,291 @@ def conditional_predict_y(
...
+
+    @property
+    def inducing_point_selector(
+        self,
+    ) -> Optional[InducingPointSelector[SparseGaussianProcessRegression]]:
+        return self._inducing_point_selector
+
+    def predict_y(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
+        f_mean, f_var = self.predict(query_points)
+        return self.model.likelihood.predict_mean_and_var(f_mean, f_var)
```
henrymoss, line 605: no cacheing?

trieste/models/gpflow/models.py
```diff
@@ -507,13 +496,291 @@ def conditional_predict_y(
...
+            inducing_points = [
+                cast(TensorType, inducing_variable).Z
+                for inducing_variable in inducing_variable.inducing_variables
+            ]  # list of L [M, D] tensors
+        else:
+            inducing_points = inducing_variable.Z  # [M, D]
+
+        q_mu, q_var = self.model.compute_qu()
+        q_sqrt = tf.linalg.cholesky(q_var)
+        whiten = False  # GPflow's SGPR model does not use the whitened representation
```
henrymoss: Are you sure of this?

henrymoss (commented):
Looks good! Is it worth adding some integration tests? Perhaps just copy the SVGP one I made?

trieste/models/gpflow/models.py
```diff
@@ -507,13 +496,291 @@ def conditional_predict_y(
...
+                f" shape {self.model.inducing_variable.Z.shape} of existing query points."
+                f" Trailing dimensions must match."
+            )
+
+        if dataset.observations.shape[-1] != q_mu.shape[-1]:
+            raise ValueError(
+                f"Shape {dataset.observations.shape} of new observations is incompatible with"
+                f" shape {self.model.q_mu.shape} of existing observations. Trailing"
+                f" dimensions must match."
+            )
```
hstojic, line 674: @henrymoss I have copied this part from SVGP, but I suspect this was needed there as data is not an argument of the model and provided only at optimisation and update time?

since we already do a check of the data above and q_mu is computed with respect to that data we should be fine without this check?

henrymoss: This is @uri-granta 's area?

trieste/models/gpflow/models.py
hstojic: since I'm currently reusing `_covariance_between_points_for_variational_models` function for computing the convariance I'm computing the `q` objects here, using the `compute_qu` - thanks @sebastianober for the suggestion!

trieste/models/gpflow/models.py
hstojic: done automatically by GPflowPredictor now in `predict` call, that's why we need to reimplement `predict_y` method here

trieste/models/gpflow/models.py
hstojic: nope, that needs checking, though this does pass the tests so a bit encouraging

trieste/models/gpflow/models.py
uri-granta: Yeah it's weird, but it doesn't seem to (in the docs it ends up showing both the correctly spelled parameter name and the misspelled docstring one). There's probably some setting that we need to enable. Fancy creating an issue?

trieste/models/gpflow/models.py
uri-granta: It is?!

My vote would be to leave it in, as it's still a valid and cheap check on the new observations.

tests/unit/models/gpflow/test_models.py
```diff
@@ -432,46 +596,336 @@ def test_gaussian_process_regression_trajectory_sampler_has_correct_samples(
...
+    optimizer = Optimizer(gpflow.optimizers.Scipy(), compile=compile)
+    model = SparseGaussianProcessRegression(sgpr_model(x_observed, y_observed), optimizer=optimizer)
+    loss = model.model.training_loss()
+    model.optimize(dataset)
+
+    assert model.model.training_loss() < loss
+
+
+@random_seed
+@pytest.mark.skip
```
uri-granta: Is the skip intentional?

tests/unit/models/gpflow/test_models.py
```diff
@@ -176,6 +197,149 @@ def test_gpflow_wrappers_ref_optimize(gpflow_interface_factory: ModelFactoryType
...
+    assert mocked_summary_scalar.call_args_list[1][0][1].numpy() == 1
+
+
+@random_seed
+def test_gpflow_models_pairwise_covariance(gpflow_interface_factory: ModelFactoryType) -> None:
+    x = tf.constant(np.arange(1, 5).reshape(-1, 1), dtype=gpflow.default_float())  # shape: [4, 1]
+    y = fnc_3x_plus_10(x)
+    model, _ = gpflow_interface_factory(x, y)
+
+    query_points_1 = tf.concat([0.5 * x, 0.5 * x], 0)  # shape: [8, 1]
```
uri-granta, line 259: what was the "for speed just update q_sqrt rather than optimize" bit that you removed?

tests/unit/models/gpflow/test_models.py
```diff
@@ -1269,25 +1477,3 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
     actual_covariance = model.covariance_between_points(query_points_1, query_points_2)
 
     np.testing.assert_allclose(expected_covariance, actual_covariance, atol=1e-4)
-
-
-@random_seed
-def test_sparse_variational_raises_for_pairwise_covariance_for_invalid_query_points(
```
uri-granta, line 1288: (what a badly named test!)

uri-granta (approved):
LGTM!

trieste/models/gpflow/models.py
hstojic: perhaps, but it will be at the bottom of my to do list :)

tests/unit/models/gpflow/test_models.py
hstojic: yep, for now, until I figure out why trajectory sampler is not working

tests/unit/models/gpflow/test_models.py
hstojic: SVGP test was failing, but without which should still be correct (and not much slower) the test passes

hstojic: perhaps warrants a closer look

trieste/models/gpflow/sampler.py
```diff
@@ -512,22 +519,31 @@ def __init__(
...
+        :raise NotImplementedError: If the model is not of valid type, or we try to use the
+            sampler with a model that has more than one latent GP.
         """
+        if model.model.num_latent_gps > 1:
+            raise NotImplementedError(
+                f"""
+                DecoupledTrajectorySampler does not currently support models with multiple latent
+                GPs, however received a model with {model.model.num_latent_gps} latent GPs.
+                """
+            )
```
hstojic: added a check that we are not getting a model with multiple latent gps, currently unsupported, added a line to the docstring
@henrymoss looks good?

same thing added to both types of trajectory samplers, and added a test checking that it raises...

trieste/models/gpflow/sampler.py
```diff
@@ -631,18 +663,19 @@ def __init__(
...
             )
 
         self._kernel = model.get_kernel()
         self._n_components = n_components
         super().__init__(self._kernel, self._n_components, dtype=tf.float64)
 
-        if isinstance(model, SupportsGetInternalData):
-            dummy_X = model.get_internal_data().query_points[0:1, :]
-        else:
+        if isinstance(model, SupportsGetInducingVariables):
```
hstojic, line 654: I was running into an error with the SGPR wrapper, SGPR wrapper is both `FeatureDecompositionInducingPointModel` and `FeatureDecompositionInternalDataModel`, which with some methods here means that either query points or inducing points are used. Now, the the order of checking what kind of model we have was not consistent and for this reason I was getting errors, I made it consistent across the file now and `SupportsGetInducingVariables` is checked, which means that the model that satisfies both will be seen as inducing variables one `FeatureDecompositionInducingPointModel` - for SGPR wrapper that is a correct choice. @henrymoss and @uri-granta does this solution looks good?
This implicit default might be a fragile thing, do you have a better idea how to do it perhaps?

trieste/models/gpflow/models.py
```diff
@@ -507,13 +496,308 @@ def conditional_predict_y(
...
+            # gpflow says inducing_variable might be a ndarray; it won't
+            cast(TensorType, self.model.inducing_variable.inducing_variable).Z.assign(
+                new_inducing_points
+            )  # [M, D]
+        else:
+            self.model.inducing_variable.Z.assign(new_inducing_points)  # [M, D]
+
+    def get_inducing_variables(
+        self,
+    ) -> Tuple[Union[TensorType, list[TensorType]], TensorType, TensorType, bool]:
```
hstojic, line 728: this return type for `SupportsGetInducingVariables` makes sense for SVGP and VGP but not really for SGPR, should we create another interface for q objects since SGPR does not have these objects?
@henrymoss and @uri-granta any thoughts?

trieste/models/gpflow/models.py
```diff
@@ -507,13 +496,308 @@ def conditional_predict_y(
...
+            inducing_points = cast(TensorType, inducing_variable.inducing_variable).Z  # [M, D]
+        elif isinstance(inducing_variable, SeparateIndependentInducingVariables):
+            inducing_points = [
+                cast(TensorType, inducing_variable).Z
+                for inducing_variable in inducing_variable.inducing_variables
+            ]  # list of L [M, D] tensors
+        else:
+            inducing_points = inducing_variable.Z  # [M, D]
+
+        q_mu, q_var = self.model.compute_qu()
```
hstojic, line 764: `compute_mu` seems like it doesn't return shape that is consistent with the SVGP model (in SVGP q objects also depend on q_diag argument...), not a big problem at the moment but it also doesn't return them correctly in case of multiple latent gps - anyone noticed these isses? @sebastianober and @jesnie perhaps?

tests/unit/models/gpflow/test_models.py
```diff
@@ -432,46 +597,339 @@ def test_gaussian_process_regression_trajectory_sampler_has_correct_samples(
...
+    sample_variance = tf.math.reduce_variance(samples, axis=1, keepdims=True)
+
+    true_mean, true_variance = model.predict(x_predict)
+
+    # test predictions approx correct away from data
+    npt.assert_allclose(sample_mean[3:] + 1.0, true_mean[3:] + 1.0, rtol=0.2)
+    npt.assert_allclose(sample_variance[3:], true_variance[3:], rtol=0.5)
+
+    # test predictions almost correct at data
+    npt.assert_allclose(sample_mean[:3] + 1.0, true_mean[:3] + 1.0, rtol=0.1)
```
hstojic: for some reason had to relax the precision here and increase the number of samples above, not sure why - @henrymoss any thoughts?

hstojic:
> Looks good! Is it worth adding some integration tests? Perhaps just copy the SVGP one I made?

good catch!

trieste/models/gpflow/models.py
uri-granta: Can you clarify? Should SGPR models work with any function that requires SupportsGetInducingVariables models? (At the moment that's just DecoupledTrajectorySampler I think, and that already supports SGPR as a FeatureDecompositionInternalDataModel.) If so, what happens when those functions want a q object? If not, what is actually going to call get_inducing_variables on an SGPR?

tests/unit/models/gpflow/test_models.py
henrymoss: that looks  bit worrying! What was it before?

hstojic: for `test_sparse_variational_trajectory_sampler_has_correct_samples` we had  `num_samples = 100` and `rtol` here was `0.01`

hstojic: although I don't do this part in SGPR since it doesnt have q objects:
```
model.model.q_sqrt.assign(tf.expand_dims(tf.eye(len(x)) * tf.math.sqrt(1e-5), 0))
model.model.q_mu.assign(y)
model.update_posterior_cache()
```

henrymoss: Is it the sample mean or sample variance test that has been relaxed?

hstojic: fixed it, all good

PR #532, opened 2022-03-10
Fix slowtests broken by gpflow posterior pr

uri-granta:
Partly fix failing slowtests.

1. handle depickling models with posterior caches
2. temporary(?) workaround for the fact that posterior predict_f can sometimes return negative variances

A few of the active learning integration tests are still failing due to being numerically very unstable. Will look at those in a separate PR.

### DIFF
--- a/trieste/models/gpflow/interface.py
+++ b/trieste/models/gpflow/interface.py
@@ -66,6 +66,12 @@ class GPflowPredictor(
         """
         self._posterior = self.model.posterior(PrecomputeCacheType.VARIABLE)
 
+    def __setstate__(self, state: dict) -> None:
+        # when unpickling we may need to regenerate the posterior cache
+        self.__dict__.update(state)
+        if self._posterior is not None:
+            self.create_posterior_cache()
+
     def update_posterior_cache(self) -> None:
         """Update the posterior cache. This needs to be called whenever the underlying model
         is changed."""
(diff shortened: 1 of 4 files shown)


trieste/models/gpflow/interface.py
```diff
@@ -78,10 +84,18 @@ def model(self) -> GPModel:
         """The underlying GPflow model."""
 
     def predict(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
-        return (self._posterior or self.model).predict_f(query_points)
+        mean, cov = (self._posterior or self.model).predict_f(query_points)
+        # posterior predict has been observed to return negative values; stop it
+        if self._posterior is not None:
+            cov = tf.clip_by_value(cov, 1e-12, cov.dtype.max)
```
henrymoss, line 90: Is someone looking into this? Looks a bit dodge but fine as a fix if temporary .

henrymoss (approved):
LGTM

trieste/models/gpflow/interface.py
uri-granta: I'm raising a gpflow issue, but I think this will require a researcher to investigate. Are you volunteering? :-)

Also, next time you should probably wait until the tests pass before ticking as you missed an oopsie: I was making all the covariances in predict_joint positive too! I've fixed it so it only updates the variances. (Alternatively I could skip changing skip_joint as it doesn't currently make any tests fail, but it feels bad to knowingly return incorrect values.)

uri-granta: PS I'll add the gpflow issue number to the comments

Update: https://github.com/GPflow/GPflow/issues/1813

hstojic (approved):
looks good until its investigated on the GPflow side

PR #533, opened 2022-03-10
Raise error if use SVGP model with q_diag=True

henrymoss:
Our `SparseVariational` does not support gpflow SVGP with q_diag=True.  In a future PR, I will add this functionality. For now I just make sure we raise a useful error.


hstojic (commented):
this looks good, but is the case that only trajectory sampling depends on this or other methods are affected?

henrymoss:
I dont think the update will work, or the pairwise cov

hstojic (approved):
in that case, good to go

PR #535, opened 2022-03-21
Add support for saving tracked state

uri-granta:
Playing around with different ways to save/load models, wwrappers and records.

# :1st_place_medal: Pickling (or dilling)

Pros: easy to use and customize, recreates original objects. Cons:

1. Not portable: you can only load models pickled on a system with the same library version
2. Not safe: unpickling can run arbitrary code
3. Less flexible than deepcopy: by default it only handles functions and classes that are top level entities and which can therefore be pickled by name. However, these are picklable using dill, as are lambdas.

# tf.saved_model

Pros: portable(?) and safe(?). Cons: haven't got it to work yet to a point where we can reuse the reloaded model for further computation.

1. Doesn't recreate original object: you just get a tf.Module containing any tf.Variable attributes, tf.function-decorated methods (wth input signatures) and sub-tf.Modules.
2. Struggles exporting methods with **kwargs arguments: e.g. predict_f's full_cov.
3. Struggles exporting methods that don't return tensors: e.g. posterior
4. ...

# gpflow.utilities.read_values/multiple_assign 

GPFlow-specific way of getting a dictionary of hyperparameters that can be easily serialised, and then assigned back to another model of the same type. Not yet got working. Cons:

1. Gpflow-specific.
2. Requires being able to build a blank model to assign to, which will require saving some additional metadata and a reasonable amount more infrastructure.
3. Not sure how portable this is between library versions.


trieste/bayesian_optimizer.py
```diff
@@ -18,13 +18,15 @@
 
 from __future__ import annotations
 
-import copy
```
hstojic, line 21: we need similar thing for ask-tell right?

trieste/bayesian_optimizer.py
```diff
@@ -18,13 +18,15 @@
...
+import datetime
 import traceback
 from collections.abc import Mapping
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Dict, Generic, Optional, TypeVar, cast, overload
 
 import absl
+import dill
```
hstojic, line 29: we need the same thing for ask-tell right?

trieste/bayesian_optimizer.py
```diff
@@ -99,14 +126,14 @@ class OptimizationResult(Generic[StateType]):
     exception.
     """
 
-    history: list[Record[StateType]]
+    history: list[FrozenRecord[StateType]]
```
hstojic: could there be any issues like write protections? we should perhaps test it and raise errors if filesystem is not accessible? 

not sure if there is a merit to support both in memory and on disk and leave to the user to choose one?

hstojic (commented):
to me this looks like a much better solution - I'm thinking whether its worthwhile allowing both in memory and on disk systems

we need the same thing for ask-tell right?

trieste/bayesian_optimizer.py
uri-granta: (Yes, this definitely needs error handling. And I think it probably _does_ make sense to allow both in-memory and on-disk tracking, though it will make the interfaces more complicated.)

trieste/bayesian_optimizer.py
```diff
@@ -89,7 +89,7 @@ def model(self) -> TrainableProbabilisticModel:
             raise ValueError(f"Expected a single dataset, found {len(self.datasets)}")
```
hstojic: dataset? shouldn't this be model?

trieste/bayesian_optimizer.py
```diff
@@ -449,6 +492,8 @@ def optimize(
             :class:`Record`.
         :param track_state: If `True`, this method saves the optimization state at the start of each
             step. Models and acquisition state are copied using `copy.deepcopy`.
+        :param track_path: If set, the optimization state is saved to disk at this path,
```
hstojic, line 520: I would add a note in the class docstring as well, expanding a bit, e.g. indicating when this would be desirable

trieste/bayesian_optimizer.py
```diff
@@ -204,12 +231,23 @@ def try_get_final_model(self) -> TrainableProbabilisticModel:
     @property
     def loaded_history(self) -> list[Record[StateType]]:
         """The history of the optimization process loaded into memory."""
-        return [record.load() for record in self.history]
+        return [record if isinstance(record, Record) else record.load() for record in self.history]
+
+    def save_result(self, path: Path) -> None:
```
hstojic: if files with same names are present this will overwrite them, right? we should probably mention in it in the docs

hstojic (commented):
looks good to me, few minor comments
we should add explanations of these options in some notebook...

trieste/bayesian_optimizer.py
```diff
@@ -84,7 +86,57 @@ def model(self) -> TrainableProbabilisticModel:
...
+    Only open frozen records generated on the same system.
+    """
+
+    path: Path
+    """ The path to the pickled Record. """
+
+    def load(self) -> Record[StateType]:
+        """Load the record into memory."""
+        with open(self.path, "rb") as f:
+            return dill.load(f)
```
henrymoss, line 116: Whats dill? is it the same as pickle?

trieste/bayesian_optimizer.py
```diff
@@ -99,14 +151,16 @@ class OptimizationResult(Generic[StateType]):
...
     """
 
-    history: list[Record[StateType]]
+    history: list[Record[StateType] | FrozenRecord[StateType]]
     r"""
     The history of the :class:`Record`\ s from each step of the optimization process. These
-    :class:`Record`\ s are created at the *start* of each loop, and as such will never include the
-    :attr:`final_result`.
+    :class:`Record`\ s are created at the *start* of each loop, and as such will never
+    include the :attr:`final_result`. The records may be either in memory or on disk.
```
henrymoss, line 160: Im missing something here. How do you specify if will be saved to disk? I guess we dont always want this to happen? Or maybe we do?

uri-granta: You specify it in the track_path argument to optimize (as it's BayesianOptimizer that actually generates these). The default is still to track in memory.

trieste/bayesian_optimizer.py
uri-granta: Yes, dill is an extension of pickle that supports more things out of the box (e.g. lambdas).

trieste/bayesian_optimizer.py
uri-granta: I've updated the tutorials to explain this.

trieste/ask_tell_optimization.py
```diff
@@ -303,7 +303,7 @@ def __repr__(self) -> str:
     @classmethod
     def from_record(
```
hstojic, line 304: shouldn't we modify `to_record` or `to_result` method to accept a path and can save the results?

hstojic: also, with those two methods we have a serious issue for gpflux and keras, we cannot obtain results with ask-tell interface, right?

uri-granta: 1. `to_record` returns a `Record` which can be saved simply by calling `save` on it. Similarly `to_result`.
2. I don't understand the question. Ask tell works exactly like it did before. We already weren't able to call `to_record` or `to_result` with gpflux/keras as it requires deepcopying the model. And saving to disk isn't enabled by default anyway. But an ask tell loop can still be executed.

uri-granta: Hoping to take another look at deepcopying keras models soon.

hstojic: 1) OK, perhaps adding a sentence or two about saving output from `to_result` in the ask-tell notebook? we should also update that notebook to use the new saving methods
2) I wasn't aware it wasn't possible. With `BayesianOptimizer`, you can still obtain `OptimizationResult` with gpflux or keras model even if they cannot use state tracking, while here we cannot obtain `OptimizationResult` at all - why is deepcopying the models necessary here but not in the `BayesianOptimizer` case?

trieste/bayesian_optimizer.py
```diff
@@ -493,9 +590,16 @@ def optimize(
...
-                    history.append(Record(datasets, models_copy, acquisition_state_copy))
+                    if track_path is None:
+                        datasets_copy = copy.deepcopy(datasets)
+                        models_copy = copy.deepcopy(models)
+                        acquisition_state_copy = copy.deepcopy(acquisition_state)
+                        history.append(Record(datasets_copy, models_copy, acquisition_state_copy))
+                    else:
+                        track_path = Path(track_path)
+                        record = Record(datasets, models, acquisition_state)
+                        record_path = track_path / f"step.{step:0{len(str(num_steps-1))}d}.pickle"
```
hstojic: perhaps avoiding hardcoding here, filename "step" could be defined as a constant at the top of the file to make it changeable and can then be reused in other methods

trieste/bayesian_optimizer.py
```diff
@@ -597,9 +701,15 @@ def optimize(
                         "\nusing split_acquisition_function or split_acquisition_function_calls.",
                         output_stream=absl.logging.ERROR,
                     )
-                return OptimizationResult(Err(error), history)
+                result = OptimizationResult(Err(error), history)
+                if track_state and track_path is not None:
+                    result.save_result(Path(track_path) / "result.pickle")
```
hstojic: hardcoded filename, same comment as above

trieste/bayesian_optimizer.py
```diff
@@ -597,9 +701,15 @@ def optimize(
...
+                return result
 
         tf.print("Optimization completed without errors", output_stream=absl.logging.INFO)
 
         record = Record(datasets, models, acquisition_state)
-        return OptimizationResult(Ok(record), history)
+        result = OptimizationResult(Ok(record), history)
+        if track_state and track_path is not None:
+            result.save_result(Path(track_path) / "result.pickle")
```
hstojic: hardcoded filename, same comment as above

hstojic (commented):
mostly looks good to me, few minor comments and one discussion related to ask-tell interface

trieste/ask_tell_optimization.py
uri-granta: `BayesianOptimizer` provides an `OptimizationResult` containing the final model state after the loop is finished. If `AskTellOptimizer` returned the uncopied model state then it would be modified as soon as you run another `tell`.

It might be useful to have a models property so you can more cleanly inspect the models during optimization (so `ask_tell.models` rather than `ask_tell._models`) but that's definitely another PR.

hstojic: `models` property is a good idea (another that I would find useful is `num_query_points` in acq rules) - I agree, a separate PR for these...

hstojic: I'm not sure what is the solution but its still a problem that we cannot get `OptimizationResult` if we use a model that doesn't have deepcopy method

I think it might be fine for trieste, eventually we will sort out deepcopying for the models we support, but for any researcher using trieste and implementing their own models its unlikely they will sort out deepcopying as well and its not good if they cannot get basic things like optimization result

perhaps we could make model an optional return in the `OptimizationResult`? combined with model property and some caveats in the docs that might be good enough

uri-granta: (We could even add a `copy: bool = True` argument to `to_result`, so people can get an optimization result that uses the non-copied model/data/state if they want.)

hstojic (approved):
looks good now

PR #536, opened 2022-03-24
Tensorboard handling of multi-dimensional observations

uri-granta:


docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -78,6 +78,11 @@
 trieste.logging.set_summary_filter(lambda name: True)  # enabe all summaries
 
 
+# %% [markdown]
+# Furthermore, some summaries such as pairplots also require special plotting tools, which can be installed by running `pip install trieste[plotting]`. A warning message will be printed if you try to generate a summary without the necessary tools.
```
vpicheny, line 82: is this the only way to install it?

uri-granta: No. You can also just install seaborn with `pip install seaborn` but if we ever require other plotting libraries it'll be easy to extend what trieste[plotting] installs.

trieste/bayesian_optimizer.py
```diff
@@ -466,26 +468,14 @@ def optimize(
...
-                    f"Search Space: `{self._search_space}`\n\n"
-                    f"Acquisition rule:\n\n    {acquisition_rule}\n\n"
-                    f"Models:\n\n    {models}",
-                )
-            if logging.include_summary("query_points/_pairplot") and not (pd and sns):
-                tf.print(
-                    "\nPairplot TensorBoard summaries require seaborn to be installed."
-                    "\nOne way to do this is to install 'trieste[plotting]'.",
-                    output_stream=absl.logging.INFO,
+                self._write_summary_init(
```
hstojic, line 477: good choice to move it to a method

trieste/bayesian_optimizer.py
```diff
@@ -532,52 +522,17 @@ def optimize(
...
-
-                        logging.scalar("wallclock/step", total_step_wallclock_timer.time)
-                        logging.scalar(
-                            "wallclock/query_point_generation",
-                            query_point_generation_timer.time,
-                        )
-                        logging.scalar(
-                            "wallclock/model_fitting",
-                            model_fitting_timer.time + initial_model_fitting_timer.time,
+                        self._write_summary_step(
```
hstojic, line 525: as well

trieste/bayesian_optimizer.py
```diff
@@ -603,3 +558,173 @@ def optimize(
 
         record = Record(datasets, models, acquisition_state)
         return OptimizationResult(Ok(record), history)
+
+    def _write_summary_init(
```
hstojic, line 562: in both methods self is not really used, we could easily move them to some utils file?

uri-granta: It might make sense to refactor bits of these out if and when we try to reuse them elsewhere, though I expect there'll remain some bits that are specific to BayesianOptimizer. Either way, don't think it should be part of this PR.

hstojic: ok, fair enough

hstojic (approved):
looks good

PR #540, opened 2022-03-29
Uri/saving keras models

uri-granta:
Note: wait until https://github.com/secondmind-labs/trieste/pull/535 is merged before reviewing.


trieste/models/keras/models.py
```diff
@@ -332,3 +334,37 @@ def optimize(self, dataset: Dataset) -> None:
 
         x, y = self.prepare_dataset(dataset)
         self.model.fit(x=x, y=y, **self.optimizer.fit_args)
+
+    def __getstate__(self) -> dict[str, Any]:
```
hstojic, line 338: a bit of education please :)
when are these getstate and setstate methods called?

hstojic: when deepcopy is called these will be automatically called?

trieste/models/keras/architectures.py
```diff
@@ -142,6 +158,35 @@ def connect_layers(self) -> tuple[tf.Tensor, tf.Tensor]:
         raise NotImplementedError
 
 
+class MultivariateNormalTriL(tfp.layers.MultivariateNormalTriL):  # type: ignore[misc]
```
hstojic, line 161: so IndependentNormal works out of the box?

trieste/models/keras/architectures.py
```diff
@@ -142,6 +158,35 @@ def connect_layers(self) -> tuple[tf.Tensor, tf.Tensor]:
...
+    """Fixed version of tfp.layers.MultivariateNormalTriL that handles saving."""
```
hstojic, line 162: perhaps raise an issue in TFP about this, they might fix this with time?

trieste/models/gpflux/interface.py
```diff
@@ -88,14 +88,3 @@ def get_observation_noise(self) -> TensorType:
             raise NotImplementedError(f"Model {self!r} does not have scalar observation noise")
 
         return noise_variance
-
-    def __deepcopy__(self, memo: dict[int, object]) -> GPfluxPredictor:
```
hstojic, line 92: GPflux now works? some recent changes on GPflux side?

hstojic (commented):
looks great, couple of minor questions

trieste/models/keras/models.py
uri-granta: Yes. The getstate and setstate methods let you control pickling, and deepcopy defaults to pickling if there is no explicit deepcopy override. For a motivating example see https://docs.python.org/3/library/pickle.html#pickle-state

trieste/models/keras/architectures.py
uri-granta: yes

trieste/models/gpflux/interface.py
uri-granta: No GPflux still fails, but I've moved the exception with the useful advice about using track_state=False to the optimize method itself, so it'll catch any model that fails to deepcopy, not just our ones.

trieste/models/keras/architectures.py
hstojic: can you add a link to the issue so we keep track of it? or perhaps some dummy test that will be triggered?

trieste/models/gpflux/interface.py
hstojic: ah, right, that's indeed better

trieste/models/keras/architectures.py
hstojic: that would probably give us a boilerplate test to check whether a layer can be deepcopied as well

uri-granta: (I don't understand. Some tfp layers are savable, others aren't. MNTL wasn't savable so I fixed it so it is.)

hstojic: it seems useful to have a test that when deepcopying the original MNTL we get a failure - if after upgrading to some future TFP version this will have been resolved, such test would fail and alert us to remove our customization, no?

uri-granta: gotcha

uri-granta: added

hstojic (approved):
great :)

PR #541, opened 2022-04-04
Deep ensemble trajectory sampler update

hstojic:
with the parallel continuous thompson sampling trajectory samplers in gpflow were modified substantially, so now keras and gpflux trajectory samplers need to be updated as well, here is the one for deep ensembles


tests/unit/models/keras/test_sampler.py
```diff
@@ -18,71 +18,221 @@
...
+    eval_1 = trajectory(test_data)
+    eval_2 = trajectory(test_data)
+
+    npt.assert_allclose(eval_1, eval_2)
 
 
-def test_ensemble_trajectory_sampler_returns_deterministic_trajectory() -> None:
+@pytest.mark.skip
+@random_seed
+def test_ensemble_trajectory_sampler_samples_are_distinct_for_new_instances(
```
hstojic, line 88: so, this doesnt pass in the graph mode, but does in the eager mode, probably doesn't matter as we don't use it that way, what we care about is that resampling results in new trajectories

tests/unit/models/keras/test_sampler.py
```diff
@@ -18,71 +18,221 @@
...
+    test_data = tf.expand_dims(test_data, -2)  # [N, 1, d]
+    test_data = tf.tile(test_data, [1, 2, 1])  # [N, 2, D]
+
+    model, _, _ = trieste_deep_ensemble_model(example_data, _ENSEMBLE_SIZE * 3)
+
+    sampler1 = EnsembleTrajectorySampler(model, use_samples=use_samples)
+    trajectory1 = sampler1.get_trajectory()
+    evals_11 = trajectory1(test_data)
+
+    sampler2 = EnsembleTrajectorySampler(model, use_samples=use_samples)
```
hstojic: if I try to do this without creating another sampler old trajectory is changed as well, perhaps we can live with this but its not ideal (@uri-granta said it effectively means that sampler is changed when we change trajectory, not a desired behaviour...)

tests/unit/models/keras/test_sampler.py
```diff
@@ -18,71 +18,221 @@
...
+    evals_22 = trajectory22(test_data)
+    evals_12 = trajectory1(test_data)
+
+    npt.assert_array_less(1e-1, tf.reduce_max(tf.abs(evals_22 - evals_21)))
+    npt.assert_allclose(evals_11, evals_21)
+    npt.assert_allclose(evals_11, evals_12)
+
+
+@random_seed
+def test_ensemble_trajectory_sampler_new_trajectories_diverge(
```
hstojic: even though initial draws seem to be the same, if we resample, both have different seeds/indices and result in different trajectories

tests/unit/models/keras/test_sampler.py
```diff
@@ -100,3 +250,73 @@ def test_ensemble_trajectory_sampler_resample_provides_new_samples_without_retra
...
+        )
+        npt.assert_array_less(
+            0.01, tf.reduce_max(tf.abs(eval_before - eval_after))
+        )  # two samples should be different
+
+    assert trajectory.__call__._get_tracing_count() == 1  # type: ignore
+
+
+# @pytest.mark.skip
+@random_seed
```
hstojic, line 299: this fails for sampling variant :(
it seems like sampling, even if the same seed is used, also depend on the input...

also, trajectory via sampling is not smooth (means are way smoother), despite using the same seed for each point it samples from a different part of the distribution
- in other words sampling is a no go :(
(strangely, it does work reasonably well in the HV project...)

trieste/models/interfaces.py
```diff
@@ -287,6 +291,17 @@ def predict_ensemble(self, query_points: TensorType) -> tuple[TensorType, Tensor
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def ensemble_distributions(self, query_points: TensorType) -> tuple[tfd.Distribution, ...]:
```
hstojic: although this resolves typing issues its not great as this is somewhat specific to DeepEnsembles, I wouldn't generally expect ensembles to have individual probabilistic models... @uri-granta perhaps we should dispense with this interface until we get more ensemble models and move the sampler to models.py?

trieste/models/keras/sampler.py
```diff
@@ -80,29 +102,98 @@ def resample_trajectory(self, trajectory: TrajectoryFunction) -> TrajectoryFunct
...
+        if self._use_samples:
+            indices = tf.random.shuffle(tf.range(self._ensemble_size * self._sample_size))
+            self._indices.assign(indices[: self._batch_size])  # [B]
+            self._seeds.assign(
+                tf.random.uniform(
+                    shape=(self._ensemble_size, 2), minval=1, maxval=999999999, dtype=tf.int32
+                )
+            )  # [E, 2]
+        else:
+            self._indices.assign(self._model.sample_index(self._batch_size))  # [B]
```
hstojic: I'm thinking that probably it doesn't make sense to sample with replacement unless batch is larger than number of networks, it only reduces diversity when batch is smaller or equal

trieste/models/interfaces.py
```diff
@@ -281,38 +282,14 @@ def ensemble_size(self) -> int:
         raise NotImplementedError
 
     @abstractmethod
-    def sample_index(self, size: int) -> TensorType:
+    def ensemble_outputs(self, query_points: TensorType) -> tuple[Any, ...]:
```
uri-granta: I don't think this makes sense as an interface, since you can't usefully do anything with the result unless you know the concrete type. Using `Any` just hides the problem: e.g. it lets you call `mean` in `DeepEnsembleTrajectorySampler`, but  that would crash i you passed in any other type of `EnsembleModel` (which despite the name, `DeepEnsembleTrajectorySampler` still claims to support).

My recommendation for now would be one of the following:

1. change this to `def ensemble_distributions(self, query_points: TensorType) -> tuple[tfd.Distribution, ...]:` and have the interface be overly specific for now (at least until we have another use case)
2. move it to a `DeepEnsembleModel` subinterface instead, which DeepEnsembleTrajectorySampler can depend on
3. remove it from the interfaces altogether, and have DeepEnsembleTrajectorySampler check that the method is defined at runtime

trieste/models/keras/architectures.py
```diff
@@ -44,14 +44,27 @@ def __init__(
...
+            are not the same.
         """
+        input_shapes, output_shapes = [], []
         for index, network in enumerate(networks):
-            if not isinstance(network, KerasEnsembleNetwork):
-                raise ValueError(
-                    f"Individual networks must be an instance of KerasEnsembleNetwork, "
-                    f"received {type(network)} instead."
-                )
             networks[index].network_name = f"model_{index}_"
```
uri-granta: Is `networks[index]` (here and below) not just `network`?

trieste/models/keras/architectures.py
```diff
@@ -44,14 +44,27 @@ def __init__(
...
+            input_shapes.append(networks[index].input_tensor_spec.shape)
+            output_shapes.append(networks[index].output_tensor_spec.shape)
+
+        if not all(x == input_shapes[0] for x in input_shapes):
```
uri-granta, line 62: Should probably add an explicit `if not networks:` check somewhere to catch an empty sequence.

trieste/models/keras/models.py
```diff
@@ -89,6 +98,12 @@ def __init__(
             arguments.
         :param bootstrap: Sample with replacement data for training each network in the ensemble.
             By default set to `False`.
+        :param diversify: Whether to use quantiles from final probabilistic layer as
+            trajectories or mean predictions when calling :meth:`trajectory_sampler`. Quantiles are
```
uri-granta:
```suggestion
            trajectories instead of mean predictions when calling :meth:`trajectory_sampler`. Quantiles are
```

tests/unit/models/keras/test_utils.py
```diff
@@ -102,3 +106,28 @@ def test_sample_with_replacement_seems_correct(rank: int) -> None:
...
     assert (tf.reduce_mean(mean) - tf.reduce_mean(x)) < 1
     assert tf.math.abs(tf.math.reduce_std(mean) - tf.math.reduce_std(x) / 10.0) < 0.1
+
+
+@pytest.mark.parametrize("size", [2, 10])
+@pytest.mark.parametrize("num_samples", [0, 1, 10])
+def test_sample_model_index_call_shape(size: int, num_samples: int) -> None:
+    indices = sample_model_index(size, num_samples)
+
+    assert indices.shape == (num_samples,)
```
uri-granta, line 116: You could probably add some more tests for sample_model_index (e.g. check that the first size indices are unique)

tests/unit/models/keras/test_sampler.py
```diff
@@ -14,75 +14,222 @@
...
 
 
 @random_seed
 def test_ensemble_trajectory_sampler_samples_are_distinct_for_new_instances() -> None:
+    """
+    If seeds are not fixed instantiating a new sampler should give us different trajectories.
+    We use `diversify=True` here because same network could be accidentally sampled in the
+    mean mode and since we cannot fix the seed here we could get stochastic failure here.
+    """
+    diversify = True
```
uri-granta: So is this test currently failing?

Also if you wanted to test it for means, you could just instantiate n samplers and check that the trajectories aren't all the same, for some n where (1/ensemble_size)^n < 0.01% or something.

tests/integration/test_bayesian_optimization.py
```diff
@@ -361,23 +355,86 @@ def test_bayesian_optimizer_with_deep_ensemble_finds_minima_of_scaled_branin(
     )
 
 
+@random_seed
```
uri-granta: Have you manually run the develop-checks job for your branch to see whether this passes?

uri-granta (commented):
Looks good! Just a couple of questions.

tests/unit/models/keras/test_sampler.py
hstojic: actually, with the modified sample_model_index now with means it should work without issues

hstojic: for some reason when I run just this test or `test_sampler` file it runs without an error, but when running the whole keras folder it reports an error, no idea what is going on...

trieste/models/interfaces.py
hstojic: opted for moving this to `keras/interface.py` file and making it specific, thanks for the suggestions!

tests/unit/models/keras/test_sampler.py
```diff
@@ -108,7 +105,7 @@ def test_ensemble_trajectory_sampler_samples_are_distinct_for_new_instances() ->
 
     eval_1 = trajectory1(test_data)
     eval_2 = trajectory2(test_data)
-
+    # breakpoint()
```
uri-granta: (remove)

tests/unit/models/keras/test_sampler.py
uri-granta: That's weird. You could maybe try commenting out other tests to see whether it's being affected by previous test state somehow. Or change the ensemble size and run again in case it's just bad luck with the RNG?

uri-granta (approved):
Looks good! Happy to help you look at the failing test if you want.

PR #544, opened 2022-04-06
Tensorboard logging fixes

### LINKED ISSUE #538: GPflowPredictor log method cannot handle complex kernels
If a model has a complex kernel e.g. a sum or a product of kernels, then `log` method raises an error as such kernels do not have `variance` or `lengthscales` attributes

@uri-granta fancy a look at this? could we perhaps try to log this and if it encounters a complex kernel not recording anything?

uri-granta:


trieste/models/gpflow/interface.py
```diff
@@ -142,13 +142,15 @@ def log(self) -> None:
...
             with summary_writer.as_default(step=logging.get_step_number()):
-                logging.scalar("kernel.variance", self.get_kernel().variance)
-                lengthscales = self.get_kernel().lengthscales
-                if tf.rank(lengthscales) == 0:
-                    logging.scalar("kernel.lengthscale", lengthscales)
-                elif tf.rank(lengthscales) == 1:
-                    for i, lengthscale in enumerate(lengthscales):
-                        logging.scalar(f"kernel.lengthscale[{i}]", lengthscale)
+                kernel = self.get_kernel()
+                if isinstance(kernel, gpflow.kernels.Stationary):
```
hstojic: hard to say if this would be a good enough check, I don't know GPflow well enough - perhaps add a test for this?

it might be good enough to support only the basic case, but I think we probably can do something more generic, like fetching all trainable variables? I think they also come with names so we can still make it descriptive enough, no?

hstojic (commented):
this would be a bug fix for the kernel parameters, but i suspect we can come up with generic readout of kernel parameters and log all of them - that might work for complex kernels as well

other changes look good

trieste/models/gpflow/interface.py
uri-granta: The check should be sufficient: gpflow.kernels.Stationary is where the lengthscales and variance parameters are set.

Makes sense to add a test. I'll see how much we can extract generically, but it might be tricky inferring types, shapes, etc. If it's too much work this may have to do for now (though I can add it to a WIBNI issue).

trieste/models/gpflow/interface.py
```diff
@@ -142,13 +143,15 @@ def log(self) -> None:
...
             with summary_writer.as_default(step=logging.get_step_number()):
-                logging.scalar("kernel.variance", self.get_kernel().variance)
-                lengthscales = self.get_kernel().lengthscales
-                if tf.rank(lengthscales) == 0:
-                    logging.scalar("kernel.lengthscale", lengthscales)
-                elif tf.rank(lengthscales) == 1:
-                    for i, lengthscale in enumerate(lengthscales):
-                        logging.scalar(f"kernel.lengthscale[{i}]", lengthscale)
+                kernel = self.get_kernel()
+                components = _merge_leaf_components(leaf_components(kernel))
```
hstojic, line 147: wasnt aware of this, great that you have worked it out!

hstojic (approved):
much better now, great work :)

PR #547, opened 2022-04-13
Clean up ParallelContinuousThompsonSampling and slightly improve testing

uri-granta:
Slightly improve test coverage, and make the code a little bit clearer and more robust: it now passes the original function to `update_trajectory` rather than the negated function (currently this doesn't make a difference as either (1) they're the same, because we updated the function in place, or (2) `update_trajectory` ignores the input and creates a new function from scratch, but this would be better if `update_trajectory` tried to use the passed in function in some unexpected way).


hstojic (approved):
looks fine to me

PR #548, opened 2022-04-19
Fix slow DGP alltest

uri-granta:
The DGP test is a bit fragile at the moment. It converges fine in alltests_latest, but not in alltests. Bumping the number of steps doesn't seem to help, but bumping the number of initial points does. Would be good to replace this at some point with a more stable test.


hstojic (approved):
looks good

PR #549, opened 2022-05-03
Batched decoupled Thompson sampling for DGP model

sebastianober:
Tests and docs currently WIP


trieste/models/gpflux/sampler.py
```diff
@@ -191,3 +209,353 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+    for cheap approximate trajectory samples, as opposed to exact trajectory sampling, which scales
+    cubically in the number of query points.
+
+    Note that we do not currently support deep GP models with multiple outputs.
+    """
+
+    def __init__(
+        self,
+        model: GPfluxPredictor,
+        num_features: int = 1000,
```
hstojic, line 139: this parameter should be available in the init of the `DeepGaussianProcess` model, right? same as in GPflow models

trieste/models/gpflux/sampler.py
```diff
@@ -191,3 +209,353 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+
+        :param trajectory: The trajectory function to be updated and resampled.
+        :return: The updated and resampled trajectory function.
+        :raise InvalidArgumentError: If ``trajectory`` is not a
+            :class:`dgp_feature_decomposition_trajectory`
+        """
+
+        tf.debugging.Assert(isinstance(trajectory, dgp_feature_decomposition_trajectory), [])
+
+        cast(dgp_feature_decomposition_trajectory, trajectory).update()
```
hstojic, line 196: should there be a `resample()` call after the `update()` call?

tests/unit/models/gpflux/test_sampler.py
```diff
@@ -157,3 +172,370 @@ def test_deep_gaussian_process_sampler_samples_are_distinct_for_new_instances(
...
+    batch_size = 5
+
+    _, model = _build_dataset_and_train_deep_gp(two_layer_model)
+
+    sampler = DeepGaussianProcessDecoupledTrajectorySampler(model)
+
+    trajectory = sampler.get_trajectory()
+    xs = tf.random.uniform([num_evals, 2], minval=-10.0, maxval=10.0, dtype=tf.float64)  # [N, D]
+    xs_with_dummy_batch_dim = tf.expand_dims(xs, -2)  # [N, 1, D]
+    xs_with_full_batch_dim = tf.tile(xs_with_dummy_batch_dim, [1, batch_size, 1])  # [N, B, D]
```
hstojic: you use this snippet in many tests here, perhaps turn it into a private function somewhere at the top and call everywhere?

hstojic (commented):
great job! 
I have few minor comments, pretty much good to go...

trieste/models/gpflux/sampler.py
sebastianober: No, it's fine - this calls `update()` in `dgp_feature_decomposition_trajectory`, which then calls `update()` in each layer, which makes sure weights are resampled.

hstojic (approved):
good to go now!

PR #550, opened 2022-05-04
Update to gpflow 2.5.1

uri-granta:


trieste/models/gpflow/models.py
```diff
@@ -952,6 +952,7 @@ def update(self, dataset: Dataset) -> None:
             )
 
         num_data = dataset.query_points.shape[0]
+        assert self.model.num_data is not None
```
hstojic, line 955: whats this bit about?

hstojic (approved):
looks good, I had only one small question...

trieste/models/gpflow/models.py
uri-granta: num_data is optional for svgp, though the call to `self._ensure_variable_model_data` ensures that we always have a variable to assign to here. The assert makes the type checker happy that this is the case.

PR #551, opened 2022-05-04
List number of available devices in metadata

uri-granta:


hstojic (approved):
this will just log available devices, it won't show any usage stats, right?

uri-granta:
> this will just log available devices, it won't show any usage stats, right?

Yes, just the available devices. Logging memory usage for gpus etc is best left to the tensorboard Profiler.

PR #553, opened 2022-05-09
Avoid recompiling training loss closure

uri-granta:
Avoid regenerating a new closure every BO step by compiling and saving a single function.

Also avoid recompiling acquisition functions combined using a Product or other reducer.


hstojic (approved):
seems fine, but this code is beyond my skills, I'll trust you know what you are doing :)

PR #554, opened 2022-05-10
Consistency with GPflux/GPflow reparameterization samplers

sebastianober:
Ensuring consistency with shapes between MCEI/MCAE and GPflux/GPflow reparameterization samplers.

### DIFF
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -108,14 +108,6 @@ def GPR_OPTIMIZER_PARAMS() -> Tuple[str, List[ParameterSet]]:
                 id="MonteCarloExpectedImprovement",
             ),
             pytest.param(
-                20,
-                EfficientGlobalOptimization(
-                    MonteCarloAugmentedExpectedImprovement(int(1e3)).using(OBJECTIVE),
-                    generate_continuous_optimizer(100),
-                ),
-                id="MonteCarloAugmentedExpectedImprovement",
-            ),
-            pytest.param(
                 24,
                 EfficientGlobalOptimization(
                     MinValueEntropySearch(
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -33,9 +33,12 @@ from trieste.acquisition import (
     GreedyContinuousThompsonSampling,
     LocalPenalization,
     MinValueEntropySearch,
+    MonteCarloAugmentedExpectedImprovement,
+    MonteCarloExpectedImprovement,
     MultipleOptimismNegativeLowerConfidenceBound,
     ParallelContinuousThompsonSampling,
 )
+from trieste.acquisition.optimizer import generate_continuous_optimizer
 from trieste.acquisition.rule import (
     AcquisitionRule,
     AsynchronousGreedy,
@@ -95,6 +98,22 @@ def GPR_OPTIMIZER_PARAMS() -> Tuple[str, List[ParameterSet]]:
                 30,
                 EfficientGlobalOptimization(AugmentedExpectedImprovement().using(OBJECTIVE)),
                 id="AugmentedExpectedImprovement",
+            ),
+            pytest.param(
+                20,
+                EfficientGlobalOptimization(
+                    MonteCarloExpectedImprovement(int(1e3)).using(OBJECTIVE),
+                    generate_continuous_optimizer(100),
+                ),
+                id="MonteCarloExpectedImprovement",
+            ),
+            pytest.param(
+                20,
+                EfficientGlobalOptimization(
+                    MonteCarloAugmentedExpectedImprovement(int(1e3)).using(OBJECTIVE),
+                    generate_continuous_optimizer(100),
+                ),
+                id="MonteCarloAugmentedExpectedImprovement",
             ),
             pytest.param(
                 24,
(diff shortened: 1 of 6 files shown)


tests/integration/test_bayesian_optimization.py
```diff
@@ -96,6 +99,22 @@ def GPR_OPTIMIZER_PARAMS() -> Tuple[str, List[ParameterSet]]:
...
+                id="MonteCarloExpectedImprovement",
+            ),
+            pytest.param(
+                20,
+                EfficientGlobalOptimization(
+                    MonteCarloAugmentedExpectedImprovement(int(1e3)).using(OBJECTIVE),
+                    generate_continuous_optimizer(100),
+                ),
+                id="MonteCarloAugmentedExpectedImprovement",
+            ),
```
henrymoss: Perhaps we dont need a full test for the augmented one?

tests/integration/test_bayesian_optimization.py
```diff
@@ -325,7 +344,26 @@ def test_bayesian_optimizer_with_dgp_finds_minima_of_scaled_branin(
...
+            id="MonteCarloExpectedImprovement",
+        ),
+        pytest.param(
+            5,
+            EfficientGlobalOptimization(
+                MonteCarloAugmentedExpectedImprovement(int(1e2)), generate_continuous_optimizer(100)
+            ),
+            id="MonteCarloAugmentedExpectedImprovement",
+        ),
+    ],
```
henrymoss: same comment as above

tests/unit/acquisition/function/test_function.py
```diff
@@ -264,10 +264,13 @@ def test_mc_expected_improvement_builds_expected_improvement_using_best_from_mod
         tf.constant([[-2.0], [-1.0], [0.0], [1.0], [2.0]]),
         tf.constant([[4.1], [0.9], [0.1], [1.1], [3.9]]),
     )
+    # model = QuadraticMeanAndRBFKernel()
```
henrymoss: eh?

tests/unit/acquisition/function/test_function.py
```diff
@@ -264,10 +264,13 @@ def test_mc_expected_improvement_builds_expected_improvement_using_best_from_mod
...
     model = QuadraticMeanAndRBFKernelWithSamplers(dataset)
     acq_fn = MonteCarloExpectedImprovement(int(1e6)).prepare_acquisition_function(model, dataset)
     xs = tf.linspace([[-10.0]], [[10.0]], 100)
     expected = expected_improvement(model, tf.constant([0.0]))(xs)
+    print(expected.shape)
+    print(acq_fn(xs).shape)
```
henrymoss: Quick delete this sloppy code before @uri-granta sees

tests/util/models/gpflow/models.py
```diff
@@ -148,7 +148,7 @@ def covariance_between_points(
     def reparam_sampler(
         self: GaussianProcessWithoutNoise, num_samples: int
     ) -> ReparametrizationSampler[GaussianProcessWithoutNoise]:
-        return GaussianProcessSampler(num_samples, self)
+        return IndependentReparametrizationSampler(num_samples, self)
```
henrymoss, line 152: This is good. I was going to ask where this GaussianProcessSampler came from. Not sure it was ever needed

trieste/acquisition/function/function.py
```diff
@@ -664,14 +666,16 @@ def update_acquisition_function(
...
+        )
+        mean = tf.reduce_mean(samples_at_query_points, axis=-3, keepdims=True)
 
         tf.debugging.assert_shapes(
-            [(mean, ["_", 1])], message="Expected model with output dimension [1]."
+            [(mean, [..., 1])], message="Expected model with output dimension [1]."
         )
 
-        eta = tf.reduce_min(mean, axis=0)
+        eta = tf.squeeze(tf.reduce_min(mean, axis=0))
```
henrymoss, line 678: bit weird that it ends up with this extra dim?

henrymoss (commented):
Looks good. I've only done a quick check so far though.

hstojic (commented):
looks ok to me, but will leave @henrymoss  to approve

tests/integration/test_bayesian_optimization.py
sebastianober: I think we should have at least one just in case someone manages to break it, although you're right probably not both. Should we keep the slow test or the fast test?

tests/unit/acquisition/function/test_function.py
sebastianober: Whoops...

tests/unit/acquisition/function/test_function.py
sebastianober: Done ;)

tests/util/models/gpflow/models.py
sebastianober: It's because the shapes my acquisition functions were expecting were different to the shapes `IndependentReparametrizationSampler` was giving, so I had to write my own utility `GaussianProcessSampler` to give the correct shapes. Probably not the best move in retrospect...

trieste/acquisition/function/function.py
sebastianober: So `mean` has shape `[N, 1, B, L]` where both `B` and `L` end up being 1, which should be consistent with the reparameterization code you already wrote for GPflow models. I can try to get rid of the extra dimensions elsewhere if that would be better?

tests/integration/test_bayesian_optimization.py
hstojic: keep the fast one I think

hstojic (approved):
Henry might need 2 more weeks since he has some  paper deadlines - to me it seems fine so I'll  approve this

PR #556, opened 2022-05-12
Trajectory samplers now work with mean functions

### LINKED ISSUE #543: Decoupled trajectory samplers for GPflow models do not use mean function
**Describe the bug**
The decoupled trajectory samplers for GP models do not make use of the model's mean function. This is fine for e.g. a constant mean function, but in the case of e.g. a linear mean function the acquisition rule would not select the correct point.

**Expected behaviour**
This should be a fairly minor modification - all that needs to be changed is that the mean function needs to be added to the current trajectories. This should also be checked for other trajectory samplers.

henrymoss:
@sebastianober noticed that I had been stupid and forgotten to take into account mean functions in my decoupled sampling implementation.

This is now fixed and I have added extra tests to make sure this cant happen again!

### DIFF
--- a/tests/unit/models/gpflow/test_models.py
+++ b/tests/unit/models/gpflow/test_models.py
@@ -612,14 +612,21 @@ def test_gaussian_process_regression_trajectory_sampler_raises_multi_latent_gp()
 
 @random_seed
 @pytest.mark.parametrize("use_decoupled_sampler", [True, False])
+@pytest.mark.parametrize("use_mean_function", [True, False])
+@pytest.mark.parametrize("noise_var", [1e-5,1e-1])
 def test_gaussian_process_regression_trajectory_sampler_has_correct_samples(
     use_decoupled_sampler: bool,
+    use_mean_function: bool,
+    noise_var: float,
 ) -> None:
     x = tf.constant(np.arange(5).reshape(-1, 1), dtype=gpflow.default_float())
     model = GaussianProcessRegression(
         gpr_model(x, _3x_plus_gaussian_noise(x)), use_decoupled_sampler=use_decoupled_sampler
     )
-    model.model.likelihood.variance.assign(1e-3)
+    model.model.likelihood.variance.assign(noise_var)
+    
+    if use_mean_function:
+        model.model.mean_function = gpflow.mean_functions.Linear()
     model.update_posterior_cache()
 
     num_samples = 100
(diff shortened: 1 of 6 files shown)


henrymoss:
@uri-granta  at the moment the high level FeatureDecompositionTrajectorySampler is defined for any probabilistic model. However, these samplers only really work for very specific models (i.e. FeatureDecompositionInternalDataModel and FeatureDecompositionInducingPointModel). 

Is there a way to specify this requirement for the high level FeatureDecompositionTrajectorySampler rather than at the lower level RandomFourierFeatureTrajectorySampler and DecoupledTrajectorySampler.  At the moment I had to do a slightly weird lazy init for the mean function in the init of FeatureDecompositionTrajectorySampler as general probabilistic models do not have mean functions.

tests/unit/models/gpflow/test_sampler.py
```diff
@@ -292,7 +292,12 @@ def test_rff_trajectory_sampler_returns_trajectory_function_with_correct_shapes(
     sampler = RandomFourierFeatureTrajectorySampler(model, num_features=num_features)
 
     trajectory = sampler.get_trajectory()
-    xs = tf.linspace([-10.0], [10.0], num_evals)  # [N, D]
+    xs = tf.linspace(
```
uri-granta, line 295: (any reason for the reformatting?)

tests/unit/models/gpflow/test_models.py
```diff
@@ -407,6 +407,22 @@ def test_gaussian_process_regression_raises_for_invalid_init() -> None:
...
+
+    m = gpflow.models.GPR((x, y), gpflow.kernels.RBF())
+    model = GaussianProcessRegression(m)
+    assert isinstance(model.get_mean_function(), gpflow.mean_functions.Zero)
+
+    m = gpflow.models.GPR(
+        (x, y), gpflow.kernels.RBF(), mean_function=gpflow.mean_functions.Linear()
+    )
+    model = GaussianProcessRegression(m)
+    assert isinstance(model.get_mean_function(), gpflow.mean_functions.Linear)
```
uri-granta, line 423: (Why not check that it's the actual mean function in the model, rather than just the same type?)

uri-granta (approved):
Looks good! (though I wouldn't trust me to catch any other omissions where the mean function needs to be accounted for but isn't)

uri-granta:
> Is there a way to specify this requirement for the high level FeatureDecompositionTrajectorySampler rather than at the lower level RandomFourierFeatureTrajectorySampler and DecoupledTrajectorySampler.

Not sure I fully understand. If FeatureDecompositionTrajectorySampler only supports the same model types as DecoupledTrajectorySampler then why can't you just define it in the same way:

```python
class FeatureDecompositionTrajectorySampler(
    TrajectorySampler[
        Union[
            FeatureDecompositionInducingPointModel,
            FeatureDecompositionInternalDataModel,
        ]],
    ABC,
):
```

PR #558, opened 2022-05-16
Make trajectory sampler ABC typing more precise

uri-granta:


trieste/models/gpflow/sampler.py
```diff
@@ -196,8 +195,43 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+    A probabilistic model that supports get_kernel, get_observation_noise
+    and get_internal_data methods.
+    """
+
+    pass
+
+
+@runtime_checkable
+class FeatureDecompositionInducingPointModel(
+    SupportsGetKernel, SupportsGetInducingVariables, Protocol
```
hstojic, line 212: while at it, perhaps we could decompose getinducingvariables into inducing points and variational parameters? to disambiguate SGPR and SVGP

PR #564, opened 2022-06-06
Fix k-means behavior when can't calculate stds

henrymoss:
Small bug fix for when empirical stds are 0 in the query points for the K-means inducing point initializer.

The old logic wasn't quite right.

I have updated the tests so that they would have failed before the fix.


uri-granta (approved):
LGTM

PR #567, opened 2022-06-09
Fix deepensemble deepcopy

uri-granta:
Our carefully crafted code for copying (and hence also serialising) DeepEnsemble models was saving only the architecture and not the weights. Fixed, and added various tests that hopefully reassure us that the copying is working.

### DIFF
--- a/tests/unit/models/keras/test_models.py
+++ b/tests/unit/models/keras/test_models.py
@@ -503,3 +503,15 @@ def test_deep_ensemble_deep_copyable() -> None:
     npt.assert_allclose(variance_f_copy_updated, variance_f_copy)
     npt.assert_array_compare(operator.__ne__, mean_f_updated, mean_f)
     npt.assert_array_compare(operator.__ne__, variance_f_updated, variance_f)
+
+    # check that we can also update the copy
+    newer_example_data = _get_example_data([30, 3], [30, 3])
+    model_copy.update(newer_example_data)
+    model_copy.optimize(newer_example_data)
+
+    mean_f_updated_2, variance_f_updated_2 = model.predict(example_data.query_points)
+    mean_f_copy_updated_2, variance_f_copy_updated_2 = model_copy.predict(example_data.query_points)
+    npt.assert_allclose(mean_f_updated_2, mean_f_updated)
+    npt.assert_allclose(variance_f_updated_2, variance_f_updated)
+    npt.assert_array_compare(operator.__ne__, mean_f_copy_updated_2, mean_f_copy_updated)
+    npt.assert_array_compare(operator.__ne__, variance_f_copy_updated_2, variance_f_copy_updated)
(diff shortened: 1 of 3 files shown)


hstojic (commented):
if its only the question of weights, this seems good enough to me

I'm wondering about stateful optimizers like Adam, which is contained in the DeepEnsemble wrapper as well, is its state reset by the deepcopy?

uri-granta:
> I'm wondering about stateful optimizers like Adam, which is contained in the DeepEnsemble wrapper as well, is its state reset by the deepcopy?

It seems to be copied fine (e.g. the weights and iterations). I've added a test.

tests/unit/models/keras/test_models.py
```diff
@@ -479,3 +480,70 @@ def test_deep_ensemble_prepare_data_call(
...
+    example_data = _get_example_data([10, 3], [10, 3])
+    model, _, _ = trieste_deep_ensemble_model(example_data, 2, False, False)
+    new_example_data = _get_example_data([20, 3], [20, 3])
+    model.update(new_example_data)
+    model.optimize(new_example_data)
+
+    model_copy = copy.deepcopy(model)
+    assert model.model.optimizer is not model_copy.model.optimizer
+    npt.assert_allclose(model_copy.model.optimizer.iterations, 1)
+    npt.assert_equal(model.model.optimizer.get_weights(), model_copy.model.optimizer.get_weights())
```
hstojic, line 532: so `weights` contains the optimizer state, wasn't aware of that - can you verify its not empty, i.e. that the state is actually used and then retained with a deepcopy?

trieste/models/keras/models.py
```diff
@@ -369,15 +369,18 @@ def optimize(self, dataset: Dataset) -> None:
...
-            # jsonify all the callback models, pickle the optimizer(!), and revert (ugh!)
+            # serialize all the callback models, pickle the optimizer(!), and revert (ugh!)
             callback: Callback
             saved_models: list[KerasOptimizer] = []
             for callback in self._optimizer.fit_args["callbacks"]:
                 saved_models.append(callback.model)
-                callback.model = callback.model and callback.model.to_json()
+                callback.model = callback.model and (
+                    callback.model.to_json(),
+                    callback.model.get_weights(),
```
hstojic, line 382: it seems like we should do something similar for the optimizer state, no?

tests/unit/models/keras/test_models.py
hstojic: perhaps similar to above check that updating the original doesn't break or change the deepcopy etc

hstojic (approved):
ok, setting weights seems fine, these tests should do the trick and optimizer weights seem to be picked up automatically so probably no need to do it manually (although we are testing only Adam here...)

I'm wondering whether we need to test whether we get the same architecture?
but I guess if we got a different architecture somehow, setting weights to it would complain

PR #569, opened 2022-06-14
Make DeepGPs copyable (ish)

### LINKED ISSUE #570: deepcopy of DeepEnsemble
if the `fit_args` do not contain callbacks an error is raised when a deepcopy is attempted

uri-granta:
Add support for copying DeepGPs by making DeepGaussianProcess accept a closure so we don't have to copy the architecture (something gpflux doesn't support).

Also fix issue with deep ensembles failing to copy when the fit args doesn't include a callbacks key.


hstojic:
not great that we have to use a closure, but its good to have a solution, it seems good otherwise, I don't have cosntructive comments except that it would be good to mention it in the notebook as well

uri-granta:
I've added an explanation to the deepgp notebook and changed it so it no longer uses `track_state=False`. I've also updated the generic exception message which gets printed when track_state fails.

hstojic (approved):
looks good to me!

PR #571, opened 2022-06-16
Test with tf 2.8 and latest gpflux

uri-granta:


hstojic (approved):
looks ok as far as I can see...

PR #573, opened 2022-06-18
Make import work for gpflux>=0.2.7

avullo:
The existing mechanism to guarantee the import for version >= 0.2.7 doesn't work (still generates `AttributeError`). 
Another repo using `trieste==0.11.2` and `gpflux==0.2.7` cannot run tests successfully. 
I've tried several variants but nothing seems to address the issue using the existing mechanism. Here I rely on a direct import, the reassignment is necessary for `mypy`. It works but there's a problem with python==3.9 which is strange since it's not catching and handling the exception.

### DIFF
--- a/trieste/models/gpflow/sampler.py
+++ b/trieste/models/gpflow/sampler.py
@@ -48,9 +48,9 @@ try:
     # code ugliness is due to https://github.com/python/mypy/issues/8823
     RFF: Any = getattr(gpflux.layers.basis_functions, "RandomFourierFeatures")
 except AttributeError:
-    from gpflux.layers.basis_functions.fourier_features import (  # type: ignore
-        RandomFourierFeaturesCosine as RFF,
-    )
+    import gpflux.layers.basis_functions.fourier_features.RandomFourierFeaturesCosine as RFFredef
+
+    RFF = RFFredef
 
 
 class IndependentReparametrizationSampler(ReparametrizationSampler[ProbabilisticModel]):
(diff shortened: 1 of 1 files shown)


hstojic (commented):
@uri-granta' thoughts would be more helpful here I think, @uri-granta do you mind helping out here?

uri-granta:
This is a pain, sorry. The code works or 0.2.3 and 0.3.0 but not 0.2.7 (which we no longer test on, as it's now neither "production" nor "latest"). Trying to figure out how to get it to work while still keeping the type checker happy.

avullo:
> This is a pain, sorry. The code works or 0.2.3 and 0.3.0 but not 0.2.7 (which we no longer test on, as it's now neither "production" nor "latest"). Trying to figure out how to get it to work while still keeping the type checker happy.

Thanks @uri-granta. I'm baffled by 3.9 though.

trieste/models/gpflow/sampler.py
```diff
@@ -48,9 +48,9 @@
     # code ugliness is due to https://github.com/python/mypy/issues/8823
     RFF: Any = getattr(gpflux.layers.basis_functions, "RandomFourierFeatures")
 except AttributeError:
-    RFF = getattr(
-        getattr(gpflux.layers.basis_functions, "fourier_features"), "RandomFourierFeaturesCosine"
-    )
+    import gpflux.layers.basis_functions.fourier_features.RandomFourierFeaturesCosine as RFFredef
```
uri-granta, line 51:
```suggestion
    from gpflux.layers.basis_functions.fourier_features import RandomFourierFeaturesCosine as RFFredef
```

uri-granta: Also you need to update gpflux/sampler.py with a similar change.

Alternatively, I can merge in my fix (https://github.com/secondmind-labs/trieste/pull/574) which is similar and seems to work.

PR #574, opened 2022-06-20
Support gpflux 0.2.7

uri-granta:
The recent changes to support gpflux 0.3.0 and TF 2.8 broke our support for gpflux 0.2.7! This fixes that.

However, we currently only run tests on "production" (gpflux 0.2.3, TF 2.4) and "latest" (which is now gpflux 0.3.0, TF 2.8), so I'm not sure how to test this without adding another test target, which we don't want to do right now. I'll add an issue, though hopefully we'll be able to drop support for 0.2.3 soon anyway.

### DIFF
--- a/trieste/models/gpflow/sampler.py
+++ b/trieste/models/gpflow/sampler.py
@@ -48,6 +48,8 @@ try:
     # code ugliness is due to https://github.com/python/mypy/issues/8823
     RFF: Any = getattr(gpflux.layers.basis_functions, "RandomFourierFeatures")
 except AttributeError:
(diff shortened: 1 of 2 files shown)


hstojic (approved):
seems good to me

hstojic:
by the way, why do we need to support 0.2.3? bcs of older TF versions?

uri-granta:
> by the way, why do we need to support 0.2.3? bcs of older TF versions?

Yes. The current production uses TF 2.4 and gpflux 0.2.3. Hopefully we'll be able to drop support by 1.0.

PR #576, opened 2022-06-23
Detect unreachable code

uri-granta:
Enable mypy checking for unreachable code and fix revealed issues: mostly an indentation issue in test_optimize_continuous_recovery_runs, which revealed an actual bug where continuous optimizer raised the wrong exception when num_recovery_runs was 0 due to an empty tensor dtype issue.


henrymoss (approved):
LGTM

PR #577, opened 2022-06-23
Uri/improve tensorboard logging

### LINKED ISSUE #440: Logging for the parallel acquisiton function optimizer
e.g. which k was it that yielded the best solution, what were the values and points in all k runs, did we enter the recovery runs etc ?

I think it would be useful for monitoring, e.g. to spot whether you might need to decrease/increase the number of runs, or if you are consistently entering recovery runs that something might be off

uri-granta:
Tensorboard improvements:

1. data closure support
    1. support passing in closures to the various logging helper functions, so that the value only gets generated if the summary isn't being filtered out
    2. catch exceptions thrown during logging (most usefully during closure evaluation) and log instead of stopping execution
3. scipy BFGS optimizer logging
    1. log total number of acquisition function evaluations
    3. log whether we needed a recovery run
    4. log the improvement of the best point over the best initial sample

### DIFF
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -54,7 +54,6 @@ from trieste.bayesian_optimizer import (
     BayesianOptimizer,
     OptimizationResult,
     TrainableProbabilisticModelType,
-    stop_at_minimum,
 )
 from trieste.logging import tensorboard_writer
 from trieste.models import TrainableProbabilisticModel, TrajectoryFunctionClass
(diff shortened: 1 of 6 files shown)


trieste/acquisition/optimizer.py
```diff
@@ -314,6 +321,26 @@ def optimize_continuous(
...
+
+                _target_func: AcquisitionFunction = target_func  # make mypy happy
+
+                def improvement() -> tf.Tensor:
+                    best_initial_value = tf.math.reduce_max(_target_func(initial_points))
+                    best_value = tf.math.reduce_max(fun_values)
+                    return best_value - tf.cast(best_initial_value, best_value.dtype)
+
+                logging.scalar("scipy.optimizer/improvement_on_initial_samples", improvement)
+                logging.scalar("scipy.optimizer/af_evaluations", total_nfev)
```
hstojic: how will this work with vectorized acq functions, like parallel cont Thompson sampling?

hstojic (commented):
seems more or less good to me, the question mark for me is how the optimizer info will be logged for vectorized acq functions? could you also have a test for that case?

trieste/acquisition/optimizer.py
uri-granta: I don't think there's any problem with vectorized acq functions, as `batchify_vectorize` calls the underlying parallel continuous optimizer just once per batch (which in turn parallelises the optimization across multiple greenlets, but still batches the acquisition function evaluation).

There is however an issue with greedy acquisition functions, as optimize gets called repeatedly each step, meaning that the logs were being overwritten. I've fixed this by adding appropriately named name scopes.

hstojic: right, but then what is reported as `EGO.optimizer/spo_af_evaluations` and `EGO.optimizer/spo_improvement_on_initial_samples` exactly in this case? I have just run it for PCTS and it shows a single number for a whole batch of acq fnc optimizations

uri-granta: The `spo_af_evaluations` is the number of times the `DecoupledTrajectorySampler` gets evaluated, since that's the thing that's being optimized. Since PCTS is vectorized, num_query_points > 1 results in a single vectorised optimization.

`spo_improvement_on_initial_samples` is a good point though. It currently calculates the difference between the best single returned point objective value and best single initial point objective value. So if there were for example 10 initial batches of 5, and 20 final batches of 5, it would calculate the difference in objective value between the best point inside one of the final batches and the best point inside one of the initial batches.

Not sure how useful that is, or whether there's something better we could return.

hstojic: ok for `spo_af_evaluations` seems fine if it works that way, for batches `spo_improvement_on_initial_samples` would not be good, we do want some meaningful stats how the batch is doing, so a mean across the batch would more suitable, probably also standard deviation - @henrymoss any thoughts here?

henrymoss: Not sure that tells us very much.

uri-granta: How useful do you think `spo_improvement_on_initial_samples` is for num_query_points = 1? We could always just drop it for now.

hstojic: I think its useful for num_query_points = 1

would a histogram for num_query_points >1 be too much work?

uri-granta: Suggest restricting it to V == 1 for now, and looking at adding a histogram at some point in the future.

uri-granta: Added histogram after all. @hstojic can you rereview?

tests/integration/test_bayesian_optimization.py
```diff
@@ -591,7 +590,6 @@ def _test_optimizer_finds_minimum(
                 acquisition_rule,
                 track_state=track_state,
                 track_path=Path(tmpdirname) / "history" if track_state else None,
-                early_stop_callback=stop_at_minimum(minima, minimizers, minimum_rtol=rtol_level),
```
hstojic: I thought this was a good idea, why removing it?

hstojic (approved):
looks good, just one new question mark

tests/integration/test_bayesian_optimization.py
uri-granta: Good spot! Just a typo due to switching branches and pycharm getting confused. Will fix.

PR #578, opened 2022-06-28
Handle empty tagged product spaces

### LINKED ISSUE #566: `TaggedProductSearchSpace` cannot be created from empty `TaggedProductSearchSpace`
**Describe the bug**

When creating a `TaggedProductSearchSpace` from `TaggedProductSearchSpace`s that were created in an inconsistent way, this

https://github.com/secondmind-labs/trieste/blob/f3a9636ce42419b0ad4f766d5eef60f023c7211a/trieste/space.py#L488

fails.

**To reproduce**

```python
>>> from trieste.space import Box, TaggedProductSearchSpace
>>> space_A = Box([-1, -2], [2, 3])
>>> tgpss_A = TaggedProductSearchSpace(spaces=[space_A], tags=["A"])
>>> tgpss_B = TaggedProductSearchSpace(spaces=[], tags=[])
>>> tgpss_C = TaggedProductSearchSpace(spaces=[tgpss_A, tgpss_B], tags=["AA", "BB"])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pyenvs/trieste-dev/trieste-py38/lib/python3.8/site-packages/trieste/space.py", line 488, in __init__
    self._subspace_starting_indices = dict(zip(tags, tf.cumsum(subspace_sizes, exclusive=True)))
  File "/home/pyenvs/trieste-dev/trieste-py38/lib/python3.8/site-packages/tensorflow/python/util/traceback_utils.py", line 153, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/home/pyenvs/trieste-dev/trieste-py38/lib/python3.8/site-packages/tensorflow/python/framework/ops.py", line 7164, in raise_from_not_ok_status
    raise core._status_to_exception(e) from None  # pylint: disable=protected-access
tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute Pack as input #1(zero-based) was expected to be a int32 tensor but is a float tensor [Op:Pack] name: x
```

**Expected behaviour**

This should not raise.

**System information**
 - OS: Ubuntu 20.04
 - Python version: 3.7 / 3.8
 - Trieste version: 0.11.0
 - TensorFlow version: 2.4 and later
 - GPflow version: 2.5.2

**Additional context**

This happens because `tgpss_B = TaggedProductSearchSpace(spaces=[], tags=[])` uses this code in an inconsistent way:

https://github.com/secondmind-labs/trieste/blob/f3a9636ce42419b0ad4f766d5eef60f023c7211a/trieste/space.py#L490

For `spaces=[]` this works out to to `tf.reduce_sum([])` which yields `float32`. In the non-empty case (`tgpss_A` as example) the return value is `int32`. Then, `tf.cumsum(subspace_sizes, exclusive=True)` in `tgpss_C.__init__` fails to combine both types.

This is due to backwards incompatible change in #513 that stopped converting `self._dimension` to `int` but returns `TensorType` directly.

uri-granta:


trieste/space.py
```diff
@@ -501,13 +501,21 @@ def __repr__(self) -> str:
     def lower(self) -> TensorType:
         """The lowest values taken by each space dimension, concatenated across subspaces."""
         lower_for_each_subspace = [self.get_subspace(tag).lower for tag in self.subspace_tags]
-        return tf.concat(lower_for_each_subspace, axis=-1)
+        return (
+            tf.concat(lower_for_each_subspace, axis=-1)
+            if lower_for_each_subspace
+            else tf.constant([], dtype=tf.float64)
```
hstojic: could there be a problem bcs of setting the dtype to some default, can we infer it from some object here? 
(in some cases, e.g. keras models, we would want to instantiate float32 search space, would this make them issues for models where we use float32?)

hstojic (approved):
looks good but I have left a question as well

trieste/space.py
uri-granta: We already default to tf.float64 in Box when we pass in a list of floats (including an empty one), so at least we're consistent. There's probably also a workaround by using a zero-dimension Box (defined with tf.constants) instead of a zero sequence tagged product space. Suggest we don't worry about it unless we hit an actual issue.

hstojic: ok, what about the following small improvement - declaring a constant dtype variable somewhere at the top and using it throughout?

uri-granta: Done

PR #579, opened 2022-06-28
Add early stop callback

### LINKED ISSUE #489: Support `callback` function for `BayesianOptimizer`
In applications, the user may favor a `callback` function (which can take inputs including `models`, `datasets`, `acquisition_state`) for `BayesianOptimizer.optimize()`. This can be useful for several reasons:
1.  get current best (as well as its input) for some intermediate performance check.
2.  easily define some other online inspecting strategies besides the current quantities logged.
2.  optimal stopping: stop the bo process before the `num_steps` if some stopping rule defined in `callback` has been met.

uri-granta:


hstojic (commented):
this is great!

what about  providing a base callback class (say in utils/callbacks.py)and essentially what you have written in the test as one instance, where you simply need to provide the output values at which you should stop?

uri-granta:
> what about providing a base callback class (say in utils/callbacks.py)and essentially what you have written in the test as one instance, where you simply need to provide the output values at which you should stop?

I've added a `stop_at_minimum` helper function and a `EarlyStopCallback` alias type. I don't see the point in defining a class interface as well at the moment; perhaps when we have stateful examples in the codebase.

hstojic (approved):
I agree, this is a single use case so lets keep it simple for now...
this additional stop at minimum function is a great addition!

Issue #580, opened 2022-06-29
Not in use param `sample_size` in `sampler.py`


TsingQAQ:
**Describe the bug**
The doc of `DecoupledTrajectorySampler` (line 273), `RandomFourierFeatureTrajectorySampler` (line 390) and `FeatureDecompositionTrajectorySampler` (line 543) classes within [`sampler.py`](https://github.com/secondmind-labs/trieste/blob/develop/trieste/models/gpflow/sampler.py) seems describe a not required input: `sample_size`. If understood correctly the functionality has been supported within `feature_decomposition_trajectory` , hence the redundant doc maybe can be removed as otherwise cause some confusion to users?

PR #582, opened 2022-07-07
Changing the shapes of trajectory samplers to support multiple outputs

sebastianober:
WIP


hstojic:
could you check #580, if its a simple doc fix to do it in this PR?

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -42,6 +42,16 @@ class GreedyContinuousThompsonSampling(SingleModelGreedyAcquisitionBuilder[HasTr
     :cite:`wilson2020efficiently`.
     """
 
+    def __init__(self, output_dim: int = 0):
```
hstojic: I think it would be useful to have a callable here, say `select_output` instead of `output_dim`, the default one could be a simple function that would choose a dimension, but it would allow for more complex combination of multiple outputs

sebastianober: Ok interesting, so for instance you could take the sum of multiple outputs and optimize that instead?

sebastianober:
TODO:

- DTS
- Tests
- `select_output`

trieste/acquisition/function/continuous_thompson_sampling.py
hstojic: yes, or cover the example that @vpicheny  was giving, with a heteroscedastic model
"Take hetGP, if you want to do TS over the mean, you do TS over the first GP; but if you want to do TS over a quantile, you'd get a quantile trajectory by traj_gp_1 + beta * exp(traj_gp_2)]."

trieste/acquisition/utils.py
```diff
@@ -101,3 +101,15 @@ def split_optimizer(
...
+def select_first_output(x: TensorType) -> TensorType:
+    """
+    A utility function for trajectory sampler-related acquisition functions which selects the first
+    output as the trajectory to be used.
+
+    :param x: Input with shape [..., B, L], where L is the number of outputs of the model.
+    :return: TensorType with shape [..., B], where the first output has been selected to reduce the
+        input.
+    """
+    return x[..., 0]
```
henrymoss: I wonder if we might want this function to be slightly more general, Like if we actually wanna use multiple latents, then we need to be able to access the samples of each one separately? Maybe make a select_nth_output function ?

trieste/acquisition/sampler.py
```diff
@@ -81,14 +96,16 @@ def sample(self, model: ProbabilisticModel, sample_size: int, at: TensorType) ->
         :param sample_size: The desired number of samples.
         :param at: Where to sample the predictive distribution, with shape `[N, D]`, for points
             of dimension `D`.
+        :param select_output: A method that returns the desired output from the model sampler, with
+            shape `[S, N]` where `S` is the number of samples and `N` is the number of locations.
```
henrymoss, line 102: maybe say that the default (and almost always to be used) option is to get the first sample?

henrymoss: and for all the others :param select_output: doc bits

henrymoss (commented):
Looks great.  I have a use case where I am using trajectory samples in a multi-objective loop, i.e. I have a trajectory sample for each of my K models.  Would it be appropriate to use your new trajectory sampler output dim to store this?

trieste/models/gpflux/sampler.py
```diff
@@ -181,7 +179,7 @@ def get_trajectory(self) -> TrajectoryFunction:
         Generate an approximate function draw (trajectory) from the deep GP model.
 
         :return: A trajectory function representing an approximate trajectory from the deep GP,
-            taking an input of shape `[N, B, D]` and returning shape `[N, B]`.
+            taking an input of shape `[N, B, D]` and returning shape `[N, B, L]`.
```
uri-granta, line 182: Remember to update the descriptions of `TrajectoryFunction` and `TrajectorySampler.get_trajectory` in `models.interfaces`. (In particular, having trajectory functions support multiple outputs makes them a bit more different to acquisition functions, right?)

trieste/models/gpflux/models.py
```diff
@@ -197,7 +197,7 @@ def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
         trajectory = self.trajectory_sampler().get_trajectory()
         expanded_query_points = tf.expand_dims(query_points, -2)
         tiled_query_points = tf.tile(expanded_query_points, [1, num_samples, 1])
-        return tf.expand_dims(tf.transpose(trajectory(tiled_query_points), [1, 0]), -1)
+        return tf.transpose(trajectory(tiled_query_points), [1, 0, 2])
```
uri-granta: (shape comments would be nice)

trieste/acquisition/utils.py
uri-granta: Also this could do with some (very simple) unit tests.

trieste/acquisition/function/continuous_thompson_sampling.py
```diff
@@ -149,12 +174,14 @@ def update_acquisition_function(
...
         return self._negated_trajectory
 
 
-def negate_trajectory_function(function: TrajectoryFunction) -> TrajectoryFunction:
+def negate_trajectory_function(
+    function: TrajectoryFunction, select_output: Optional[Callable[[TensorType], TensorType]] = None
+) -> TrajectoryFunction:
     """
     Return the negative of trajectories so that our acquisition optimizers (which are
     all maximizers) can be used to extract the minimizers of trajectories.
```
uri-granta: The logic around negate_trajectory_function is getting increasingly ugly, but I can't immediately think of a better approach.

In the meantime, maybe update the docstring (speciically the next paragraph) as this function now doesn't just negate the call method, but also applies the output selector.

uri-granta (commented):
A few initial comments. Overall looks good!

henrymoss (approved):
LGTM

PR #583, opened 2022-07-08
small refactoring of acquisition functions

hstojic:
dividing previous function.py into some logical sets: improvement and optimism based functions and then rest goes to misc
moved multi-objective file from function into its package in the parent folder, makes more sense that way
also fixed few missing imports in init files


uri-granta (commented):
Looks great, though I'd be happy if you ran the function grouping past someone else too to make sure it's sensible.

(As an aside, it would be nice if git managed to track the history for all the split files, but I don't think this is possible while still squashing the PR.)

hstojic:
ok, will pass it by others

trieste/acquisition/function/multi_objective/dominance.py
```diff
@@ -17,7 +17,7 @@
 
 import tensorflow as tf
 
-from ...types import TensorType
+from ....types import TensorType
```
uri-granta, line 20: I don't understand why you moved `acquisition.multi_objective` inside `acquisition.function`. AFAIUI:

* `acqusition` is everything connected to acquisition functions, including interfaces, rules, samplers, optimizers, etc.
* `acquisition.function` is for _specific_ acquisition function implementations
* `acqusition.multi_objective` is for multi-objective helper functions

uri-granta (commented):
Just one question

trieste/acquisition/function/multi_objective/dominance.py
hstojic: multi objective helper functions in there are really in service of multi BO acq functions, like in hypervolume.py, so it made sense to move them there

PR #584, opened 2022-07-11
Uri/plotting package

uri-granta:
Move plotting utilities to an experimental package, so that people can run the tutorials without installing trieste from source.

- the package is installed with trieste; however the necessary plotting dependencies are only installed by running `pip install trieste[plotting]` (this is explained in an error message if you try to use it otherwise)
- the utilities are type annotated and have docstrings, but are untested beyond their current use in the notebooks


docs/notebooks/expected_improvement.pct.py
```diff
@@ -23,8 +23,7 @@
     SCALED_BRANIN_MINIMUM,
     BRANIN_SEARCH_SPACE,
 )
-from trieste.objectives.utils import mk_observer
```
hstojic, line 26: removing `mk_observer` as well?

docs/notebooks/feasible_sets.pct.py
```diff
@@ -36,8 +36,7 @@
 
 # %%
 from trieste.objectives import branin, BRANIN_SEARCH_SPACE
-from util.plotting_plotly import plot_function_plotly
-from trieste.space import Box
```
hstojic, line 40: removing `Box` as well?

docs/tutorials.rst
```diff
@@ -70,7 +70,7 @@ The following tutorials (or sections thereof) explain how to use and extend spec
 Run the tutorials interactively
 -------------------------------
 
-The above tutorials are built from Jupytext notebooks in the notebooks directory of the repository. These notebooks can also be run interactively. To do so, install the library from sources, along with additional notebook dependencies, with (in the repository root)
+The above tutorials are built from Jupytext notebooks in the notebooks directory of the trieste repository. These notebooks can also be run interactively. To do so, install the library from sources, along with additional notebook dependencies, with (in the repository root)
```
hstojic, line 73: perhaps it would be good to specify how to install library from sources, like exact sequence of commands?

setup.py
```diff
@@ -47,6 +47,6 @@
         "greenlet>=1.1.0",
     ],
     extras_require={
-        "plotting": ["seaborn"],
+        "plotting": ["seaborn", "plotly"],
```
hstojic, line 51: so `matplotlib` is not needed?

docs/tutorials.rst
```diff
@@ -81,3 +81,9 @@ then run
 .. code::
 
    $ jupyter-notebook notebooks
+
+Alternatively, you copy and paste the tutorials into fresh notebooks and avoid installing the library from source. To ensure you have the required plotting dependencies, simply run:
```
hstojic, line 85: but this would install only `plotly`, what about other dependencies in `notebooks/requirements` file?

hstojic (approved):
seems like a good first step, and since it will take some time to develop this properly I agree its best to put it in experimental

left a few minor comments

docs/notebooks/expected_improvement.pct.py
uri-granta: Wasn't being used

docs/notebooks/feasible_sets.pct.py
uri-granta: Ditto

docs/tutorials.rst
uri-granta: The following line does precisely that

setup.py
uri-granta: Seaborn already requires matplotlib

docs/tutorials.rst
uri-granta: The remaining (non-plotting) requirements, `gym`, `ray` and `box2d`, are already mentioned in the notebooks that use them.

PR #585, opened 2022-07-12
Fix versioned documentation

uri-granta:
Fix versioned documentation issues:

- check out pages branch in develop-deploy.yaml
- use full urls in versions.json so that switching works not just at the top level


hstojic (approved):
lgtm

PR #591, opened 2022-07-19
keras backend fix for deep GPs

hstojic:
having to set the keras backend is problematic, here is a fix

raised an issue at gpflux: https://github.com/secondmind-labs/GPflux/issues/76, fix seems to be easy but dont have time at the moment to do it


trieste/models/gpflux/models.py
```diff
@@ -127,7 +125,19 @@ def scheduler(epoch: int, lr: float) -> float:
             }
 
         self._model_gpflux = model
-
+        # inputs and targets need to be redone with a float64 dtype to avoid setting the keras
+        # backend to float64, this is likely to be fixed in GPflux, see issue:
+        # https://github.com/secondmind-labs/GPflux/issues/76
+        self._model_gpflux.inputs = tf.keras.Input(
```
uri-granta, line 131: Is this sufficient, or should you copy the other parameters too (batch_size, sparse, ragged, etc)?

docs/notebooks/deep_gaussian_processes.pct.py
```diff
@@ -40,9 +35,9 @@
 
 search_space = MICHALEWICZ_2_SEARCH_SPACE
 
-fig = plot_function_plotly(function, search_space.lower, search_space.upper)
-fig.update_layout(height=800, width=800)
-fig.show()
+# fig = plot_function_plotly(function, search_space.lower, search_space.upper)
```
uri-granta: Unintentional?

uri-granta (commented):
Looks good. Just one question regarding copying the layers.

docs/notebooks/deep_gaussian_processes.pct.py
hstojic: indeed, good catch :)

trieste/models/gpflux/models.py
hstojic: as far as I can see it should be fine,  this issue would typically surface when calling a method or trying to train the model - here training works, calling all the methods we have works fine in the tests - can you think of a test that would fail?

hstojic: where do you see those parameters? in DeepGP `model.parameters` everything is in float64 as far as I can see

uri-granta: Looks like gpflux doesn't set any of them. I guess if we're planning to fix this in gpflux anyway then this is ok for now.

uri-granta (approved):

PR #592, opened 2022-07-21
Code overview tutorial

uri-granta:
Q: Should I also mention any of the following?

- VectorizedAcquisitionFunctionBuilder
- ReparametrizationSampler
- TrajectorySampler


hstojic:
took a quick look and it nicely fills a gap, explaining important concepts in the toolbox, I'll take a more detailed look at a later point
also, `mk_multi_observer` is useful, I have used that snippet quite a few times already...

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+
+# %% [markdown]
+# ## Key types
+#
+# The following types represent the key concepts in Trieste. For a full listing of all the types in Trieste, see the API reference.
+
+# %% [markdown]
+# ### `Observer`
+#
+# The `Observer` type definition represents the black box objective function. Observers are functions that accept query points and return observations. Observations are either a single objective value that we wish to optimize, or a dictionary of mutiple tagged values that must be combined somehow, for example an objective and an inequality constraint. Objective values can be either single or multi-dimensional (see [multi-objective optimization](multi_objective_ehvi.ipynb)).
```
vpicheny: Observers return a dataset rather than observations, which is slightly unusual so we should state it.

Not sure how to write it, but it might be worth noting that this is very flexible in the sense that the returned dataset can be anything (not necessarily [query points, observations]).

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+
+# %% [markdown]
+# ### `ProbabilisticModel`
+#
+# The `ProbabilisticModel` protocol represents any probabilistic model used to model observations. Like for `Dataset`, observers with multiple observations are modelled by a dictionary o multiple tagged models.
+#
+# At it simplest, a `ProbabilisticModel` is anything that implements a `predict` and `sample` method. However, many algorithms in Trieste depend on models with additional features, which are represented by the various subclasses of `ProbabilisticModel`. The standard Bayesian optimizer uses `TrainableProbabilisticModel` models, which also implement `update` and `optimize` methods. Specific acuqisition functions may require other features, represented by classes like `SupportsPredictJoint` and `SupportsGetObservationNoise`. Since these are defined as protocols, it is possible to define and depend on the intersections of different model types (e.g. only support models that are both `SupportsPredictJoint` and `SupportsGetObservationNoise`).
```
vpicheny: It would be good to explain `update` and `optimize`. Maybe all in brackets?  

(...) which also implement `update` (updating the model structure when new data is added to the training set) and `optimize` (optimizing the model training loss) methods

also 
`SupportsPredictJoint` (ability to predict the joint distribution at several inputs x1, ... xp)
`SupportsGetObservationNoise` (ability to predict the observation noise variance)

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+#
+# Multiple models can also be combined into a single `ModelStack` model that combines their outputs for prediction and sampling. This can be useful when modelling multi-objective observations with independent models. There are also constructors like `TrainableModelStack` and `PredictJointModelStack` that combine specific types of model for use with code that requires that type, delegating the relevant methods where appropriate.
```
vpicheny: Would start with a slightly more general sentence about multiple outputs perhaps? e.g. something along the lines of

There are multiple ways to deal with multiple outputs: one can use a single multi-output model or a dictionary of models.

Maybe also add that `ModelStack` essentially emulates a single multi-output model from independent single-output models? and say it's useful for acquisitions that expect a multioutput model such as our multi-objective ones?

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+
+# %% [markdown]
+# ### `SearchSpace`
+#
+# The `SearchSpace` base class represents the domain over which the objective function is to be optimized. Spaces can currently be either continuous `Box` spaces, discrete `DiscreteSearchSpace` spaces, or a `TaggedProductSearchSpace` product of multipe spaces. All search spaces expose their dimensions, bounds and a sampler.
```
vpicheny, line 34: "over which the objective function is to be optimized"

mmmm I think that actually it's "over which the acquisition function is to be optimized"?? which in the case of trust regions is not the same. Possibly an unnecessary distinction.

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+# %% [markdown]
+# ### `AcquisitionRule`
+#
+# The `AcquisitionRule` base class represents a routine for selecting new query points during a Bayesian optimization loop (via an `acquire` method). It is generic on three types:
+#
+# * **ResultType**: the output of the rule, typically this is just tensors representing the query points. However, it can also be functions that accept some *acquisition state* and return the query points with a new state.
+# * **SearchSpaceType**: the type of the search space; any optimizer that the rule uses must support this.
+# * **ProbabilisticModelType**: the type of the models; any acquisition functions or samplers that the rule uses must support this.
+#
+# Exampes of rules include:
```
vpicheny: typo: examples

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+# 1. `AsynchronousOptimization` is similar to EGO but uses acquisition state to keep track of points whose observations are still pending.
+# 1. `DiscreteThompsonSampling` uses Thompson samplers rather than acquisition functions to select new query points.
+#
+
+# %% [markdown]
+# ### `AcquisitionFunction` and `AcquisitionFunctionBuilder`
+#
+# The `AcquisitionFunction` type definition represents any acquisition function: that is, a function that maps a set of query points to a single value that describes how useful it would be evaluate all these points together.
+#
+# The `AcquisitionFunctionBuilder` base class, meanwhile, represents something that builds and updates acquisition functions. At the start of the Bayesian optimization, the builder's `prepare_acquisition_function` method is called by the acquisition rule to create an acquisition function from the current observations and probabilistic models. For efficiency, most builders also define an `update_acquisition_function` method for updating the function using the updated observations and models. (The ones that don't instead generate a new acquisition function when necessary.)
```
vpicheny: maybe give a little bit more detail about "for efficiency"? e.g. for someone familiar with BO but not tensorflow, this is probably very puzzling (it's still very puzzling for me).

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+#
+# Acquisition functions that support only one probabilistic model are more easily defined using the `SingleModelAcquisitionBuilder` convenience class.
+#
+# Acquisition functions that are suitable for greedily building batches of points can be defined using `GreedyAcquisitionFunctionBuilder` (or `SingleModelGreedyAcquisitionBuilder`) using a `prepare_acquisition_function` method that also accepts the points already chosen to be in the current batch.
+
+# %% [markdown]
+# ### `AcquisitionOptimizer`
+#
+# The `AcquisitionOptimizer` type definition represents an optimizer function that maximizes an acquisition function over a search space. Trieste provides a `generate_continuous_optimizer` function for generating gradient-based optimizers for continuous spaces and `automatic_optimizer_selector` for quickly selecting an appropriate optimizer.
```
vpicheny: worth telling here that we can also deal with discrete and hybrid cases?

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+
+# %% [markdown]
+# ### `BayesianOptimizer` and `AskTellOptimizer`
+#
+# The `BayesianOptimizer` and `AskTellOptimizer` classes are the two Bayesian optimization interfaces provided by Trieste.
+#
+# `BayesianOptimizer` exposes an `optimize` method for running a Bayesian optimization loop with given initial datasets and models, and a given number of steps (and an optional early stop callback).
```
vpicheny, line 73: add that there is some monitoring as well?

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+#
+# `AskTellOptimizer` provides greater control over the loop, by providing separate `ask` and `tell` steps for suggesting query points and updating the models with new observations.
```
vpicheny: link to the notebook?

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+import tensorflow as tf
+from trieste.types import TensorType
+
+# %% [markdown]
+# This section explains how to define new observers, model types and acqusition functions.
+
+# %% [markdown]
+# ### New observers
+#
+# Defining an observer with a single observation is as simple as defining a function that returns that observation:
```
vpicheny: an dataset

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+)
+multiobserver(tf.constant([[0, 1], [1, 1]], dtype=tf.float64))
+
+# %% [markdown]
+# ### New probabilistic model types
+#
+# Defining a new probabilistic model type simply involves writing a class that implements all the relevant methods (at the very least `predict` and `sample`). For clarity, it is best to also explicitly inherit from the supported feature protocols.
+
+# %%
+from trieste.models.interfaces import (
```
vpicheny, line 134: trajectory samplers are advanced methods. Possibly doesn't matter, otherwise use for example a more trivial feature, eg `SupportGetObservationNoise?

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+    TrainableProbabilisticModel,
+    HasReparamSampler,
+    HasTrajectorySampler,
+    ReparametrizationSampler,
+)
+
+
+class GizmoModel(
```
vpicheny, line 142: I'm puzzled, we don't specify a vanilla sample method here.

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+    def reparam_sampler(
+        self, num_samples: int
+    ) -> ReparametrizationSampler[GizmoModel]:
+        ...
+
+    ...
+
+
+# %% [markdown]
+# If the new model type has an additional feature on which you'd like to depend, e.g. in a new acquisition function, then you can define that feature as a protocol. Marking it runtime_checkable will alow you to check for the feature at runtime too.
```
vpicheny: I don't know why I would want or not to check it at runtime. Add a clue?

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+
+# %%
+from trieste.models.interfaces import ProbabilisticModel
+from typing_extensions import Protocol, runtime_checkable
```
vpicheny, line 165: maybe add something specific about `Protocol`?

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+):
+    """A model that supports both gizmo, reparam_sampler and trajectory_sampler."""
+
+    pass
+
+
+# %% [markdown]
+# ### New acquisition function builders
+#
+# To define a new acquisition function builder, you simply need to define a class with a `prepare_acquisition_function` method that returns an `AcquisitionFunction`. If the acquisition function depends on just one model/dataset (as is often the case) then you can define it as a `SingleModelAcquisitionBuilder`; if it depends on more than one (e.g. both an objective and a constraint) then you must define it as a `ModelAcquisitionBuilder` instead. You can also specify (in brackets) the type of probabilistic models that the acquisition function supports: e.g. a `SingleModelAcquisitionBuilder[HasReparamSampler]` only supports models with a reparatemtrization sampler. This allows the type checker to warn you if you try to use an incompatible model type.
```
vpicheny: typo reparatemtrization

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+#
+
+# %% [markdown]
+# ### `AcquisitionFunction` and `AcquisitionFunctionBuilder`
+#
+# The `AcquisitionFunction` type definition represents any acquisition function: that is, a function that maps a set of query points to a single value that describes how useful it would be evaluate all these points together.
+#
+# The `AcquisitionFunctionBuilder` base class, meanwhile, represents something that builds and updates acquisition functions. At the start of the Bayesian optimization, the builder's `prepare_acquisition_function` method is called by the acquisition rule to create an acquisition function from the current observations and probabilistic models. For efficiency, most builders also define an `update_acquisition_function` method for updating the function using the updated observations and models. (The ones that don't instead generate a new acquisition function when necessary.)
+#
+# Acquisition functions that support only one probabilistic model are more easily defined using the `SingleModelAcquisitionBuilder` convenience class.
```
vpicheny: could be good to give a bit more detail about this? why it is convenient, what it changes...

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+
+# %%
+from trieste.acquisition import (
+    AcquisitionFunction,
+    SingleModelAcquisitionBuilder,
+)
+from trieste.data import Dataset
+
+
+class ProbabilityOfValidity(SingleModelAcquisitionBuilder[ProbabilisticModel]):
```
vpicheny, line 203: is it worth having a multi-model example as well? (not sure)

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+    ) -> AcquisitionFunction:
+        def acquisition(at: TensorType) -> TensorType:
+            mean, _ = model.predict_y(tf.squeeze(at, -2))
+            return mean
+
+        return acquisition
+
+
+# %% [markdown]
+# For efficiency, it usually makes sense to compile the generated acquisition function into a TensorFlow graph using `tf.function`. Furthermore, to avoid generating (and compiling) a new acquisition function on each Bayesian optimization loop, you can define an `update_acquisition_function` method that can instead update the previously generated acquisition function using the new models and data. This may involve updating the acquisition function's internal state (which you should store in `tf.Variable`s), though if the function has no internal state then it is suficient to simply return the old function unchanged.
```
vpicheny, line 215: this text should go to the top

docs/notebooks/code_overview.pct.py
```diff
@@ -0,0 +1,238 @@
...
+
+# %%
+class ProbabilityOfValidity2(SingleModelAcquisitionBuilder[ProbabilisticModel]):
```
vpicheny, line 218: maybe more useful to have one example with a state?

vpicheny (commented):
that's a great addition to the doc!

vpicheny:
> Q: Should I also mention any of the following?
> 
>     * VectorizedAcquisitionFunctionBuilder
> 
>     * ReparametrizationSampler
> 
>     * TrajectorySampler

I don't think that it's documented anywhere... which is a problem. But not sure if this is the right place if we want to focus on the main items.

PR #593, opened 2022-07-22
Log AF values in histogram if vectorized

uri-granta:
The code for logging acquisition function values wasn't handling vectorised acquisition functions and was just logging one value as a scalar.

### DIFF
--- a/trieste/acquisition/rule.py
+++ b/trieste/acquisition/rule.py
@@ -267,8 +267,14 @@ class EfficientGlobalOptimization(
         if summary_writer:
             with summary_writer.as_default(step=step_number):
                 batched_points = tf.expand_dims(points, axis=0)
-                value = self._acquisition_function(batched_points)[0][0]
-                logging.scalar("EGO.acquisition_function/maximum_found" + "[0]" * greedy, value)
+                if isinstance(self._builder, VectorizedAcquisitionFunctionBuilder):
+                    values = self._acquisition_function(batched_points)[0]
+                    logging.histogram(
+                        "EGO.acquisition_function/maximum_found" + "[0]" * greedy, values
+                    )
+                else:
+                    value = self._acquisition_function(batched_points)[0][0]
+                    logging.scalar("EGO.acquisition_function/maximum_found" + "[0]" * greedy, value)
 
         if isinstance(self._builder, GreedyAcquisitionFunctionBuilder):
             for i in range(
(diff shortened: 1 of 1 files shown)


trieste/acquisition/rule.py
```diff
@@ -288,8 +295,15 @@ def acquire(
...
                     with summary_writer.as_default(step=step_number):
                         batched_points = tf.expand_dims(chosen_point, axis=0)
-                        value = self._acquisition_function(batched_points)[0][0]
-                        logging.scalar(f"EGO.acquisition_function/maximum_found[{i+1}]", value)
+                        values = self._acquisition_function(batched_points)[0]
+                        if len(values) == 1:
+                            logging.scalar(
+                                f"EGO.acquisition_function/maximum_found[{i + 1}]", values[0]
+                            )
+                        else:  # vectorized acquisition function
```
uri-granta, line 303: (this can't happen at the moment, as we don't have vectorized greedy acquisition functions, but it feels safer to leave it in here for the future)

trieste/acquisition/rule.py
```diff
@@ -267,8 +267,15 @@ def acquire(
...
                 batched_points = tf.expand_dims(points, axis=0)
-                value = self._acquisition_function(batched_points)[0][0]
-                logging.scalar("EGO.acquisition_function/maximum_found" + "[0]" * greedy, value)
+                values = self._acquisition_function(batched_points)[0]
+                if len(values) == 1:
+                    logging.scalar(
+                        "EGO.acquisition_function/maximum_found" + "[0]" * greedy, values[0]
+                    )
+                else:  # vectorized acquisition function
+                    logging.histogram(
```
vpicheny, line 276: do we want to do something more complicated than than, e.g. also track individual values? or is this overkill?

trieste/acquisition/rule.py
```diff
@@ -267,8 +267,15 @@ def acquire(
...
+                        "EGO.acquisition_function/maximum_found" + "[0]" * greedy, values
+                    )
 
         if isinstance(self._builder, GreedyAcquisitionFunctionBuilder):
```
vpicheny, line 280: BTW you could replace by if(greedy)

trieste/acquisition/rule.py
uri-granta: what do you mean? generate individual scalar plots for the different batch elements, like we do for greedy batch?

trieste/acquisition/rule.py
vpicheny, line 303: this is a very corner case where we use both greedy and multipoint, in theory why not but not sure this is actually feasible at the moment? this extra bit of code doesn't hurt though

trieste/acquisition/rule.py
uri-granta: IIRC that makes mypy sad as it can no longer figure out that `update_acquisition_function` just below accepts a `pending_points` argument.

trieste/acquisition/rule.py
vpicheny: yes

trieste/acquisition/rule.py
uri-granta: yeah, this can't happen at the moment, but feels like it's safer to leave this in so if it does become possible we don't forget to update it.

trieste/acquisition/rule.py
vpicheny: I'm not entirely sure it's a good idea considering that the batch isn't following any particular order so we might see lots of random jumps

uri-granta: yeah. with greedy at least there is an explicit ordering.

vpicheny (approved):

PR #595, opened 2022-07-28
Add AskTell models setter (and fix generic type)

### LINKED ISSUE #594: provide a setter for models in AskTell / BayesianOptimizer
models sometime need to be recreated and trained from scratch, but then one would like to update the optimizer state as well

uri-granta:


hstojic (approved):
looks fine to me

Issue #599, opened 2022-07-31
keras model cannot be saved with tensorboard callback


hstojic:
if we add a following standard callback:
```
tf.keras.callbacks.TensorBoard(
                    log_dir=".",
                    histogram_freq=0,
                    update_freq="epoch",
                    write_graph=False,
                    write_images=False,
                    embeddings_freq=0,
                )
```
to `fit_args` state cannot be saved, an error occurs:
```
E                       NotImplementedError: Failed to save the optimization state. Some models do not support deecopying or serialization and cannot be saved. (This is particularly common for deep neural network models, though some of the model wrappers accept a model closure as a workaround.) For these models, the `track_state`` argument of the :meth:`~trieste.bayesian_optimizer.BayesianOptimizer.optimize` method should be set to `False`. This means that only the final model will be available.

../../trieste/bayesian_optimizer.py:668: NotImplementedError
------------------------------------------------------------------------------------------- Captured stderr call -------------------------------------------------------------------------------------------
2022-07-31 22:37:26.456257: I tensorflow/core/profiler/lib/profiler_session.cc:126] Profiler session initializing.
2022-07-31 22:37:26.456310: I tensorflow/core/profiler/lib/profiler_session.cc:141] Profiler session started.
2022-07-31 22:37:26.456357: I tensorflow/core/profiler/lib/profiler_session.cc:159] Profiler session tear down.
2022-07-31 22:37:26.457966: W tensorflow/python/util/util.cc:348] Sets are not currently considered sequences, but this may change in the future, so consider avoiding using them.
2022-07-31 22:37:26.554112: I tensorflow/compiler/mlir/mlir_graph_optimization_pass.cc:176] None of the MLIR Optimization Passes are enabled (registered 2)
2022-07-31 22:37:26.554592: I tensorflow/core/platform/profile_utils/cpu_utils.cc:114] CPU Frequency: 1999965000 Hz
2022-07-31 22:37:27.802835: I tensorflow/core/profiler/lib/profiler_session.cc:126] Profiler session initializing.
2022-07-31 22:37:27.802877: I tensorflow/core/profiler/lib/profiler_session.cc:141] Profiler session started.
2022-07-31 22:37:27.805983: I tensorflow/core/profiler/lib/profiler_session.cc:66] Profiler session collecting data.
2022-07-31 22:37:27.808525: I tensorflow/core/profiler/lib/profiler_session.cc:159] Profiler session tear down.
[...]

PR #600, opened 2022-07-31
keras logger

### LINKED ISSUE #597: tensorboard logging model parameters or training stats for keras models

hstojic:
this closes #597 - logging keras model losses and metrics to the tensorboard, only summary stats available,
epoch loss traces have to be done differently, by providing a callback in  `fit_args`, the only issue is that errors is raised when trying to save the model with tensorboard callback... (see #599


trieste/ask_tell_optimization.py
```diff
@@ -294,6 +294,13 @@ def __init__(
                 model.update(dataset)
                 model.optimize(dataset)
 
+            # summary_writer = logging.get_tensorboard_writer()
+            # if summary_writer:
+            #     with summary_writer.as_default(step=-1):
+            #         for tag, model in self._models.items():
+            #             with tf.name_scope(f"{tag}.model"):
+            #                 model.log()
+
```
hstojic: @uri-granta we forgot to log model at initial fit, this code is not working, what would be the best way to do this?

trieste/bayesian_optimizer.py
```diff
@@ -675,6 +675,13 @@ def optimize(
...
                                 model.optimize(dataset)
 
+                            # summary_writer = logging.get_tensorboard_writer()
+                            # if summary_writer:
+                            #     with summary_writer.as_default(step=-1):
+                            #         for tag, model in models.items():
+                            #             with tf.name_scope(f"{tag}.model"):
+                            #                 model.log()
+
                     with Timer() as query_point_generation_timer:
```
hstojic: @uri-granta same here?

trieste/models/keras/interface.py
```diff
@@ -73,6 +74,23 @@ def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
...
+        if summary_writer:
+            with summary_writer.as_default(step=logging.get_step_number()):
+                logging.scalar(f"num_epochs", len(self.model.history.epoch))
+                for k, v in self.model.history.history.items():
+                    logging.histogram(f"epoch_{k}", v)
+                    logging.scalar(f"{k}_final", v[-1])
+                    logging.scalar(f"{k}_diff", v[0] - v[-1])
+                    logging.scalar(f"{k}_min", tf.reduce_min(v))
+                    logging.scalar(f"{k}_max", tf.reduce_max(v))
+
```
hstojic: @willcowley what do you think of these summaries, would you include anything else?

trieste/ask_tell_optimization.py
uri-granta: If the issue is `step=-1` then the cleanest solution is probably to use `step=0` here and update the BO loop to start counting from 1 (`for step in range(1, num_steps + 1)`, and remembering to update the `fix_initial_model` check). This would arguably be more user friendly anyway (the first step is currently "step #0") and the only thing apart from logging that depends on the number is the tracked state filenames, which will work fine too.

trieste/models/keras/models.py
```diff
@@ -359,15 +367,82 @@ def optimize(self, dataset: Dataset) -> None:
...
+                        logging.histogram(
+                            f"variance_predict_{i}", predict_ensemble_variance[i, ...]
+                        )
+                        logging.scalar(
+                            f"variance_predict_mean_{i}",
+                            tf.reduce_mean(predict_ensemble_variance[i, ...]),
+                        )
+                    # data stats
+                    empirical_variance = tf.math.reduce_variance(dataset.observations)
+                    logging.scalar("variance_empirical", empirical_variance)
```
hstojic: @willcowley what do you think of the summaries above, would you include anything else?

trieste/models/keras/models.py
```diff
@@ -359,15 +367,82 @@ def optimize(self, dataset: Dataset) -> None:
...
+
+        :param dataset: Optional data that can be used to log additional data-based model summaries.
+        """
+        summary_writer = logging.get_tensorboard_writer()
+        if summary_writer:
+            with summary_writer.as_default(step=logging.get_step_number()):
+                logging.scalar("num_epochs", len(self.model.history.epoch))
+                for k, v in self.model.history.history.items():
+                    logging.histogram(f"{k}_epoch", v)
+                    logging.scalar(f"{k}_final", v[-1])
```
uri-granta: (If there's any chance that these can throw an exception (e.g. v is empty) then you can wrap the logged data in a closure, which will catch the exception.)

```suggestion
                    logging.scalar(f"{k}_final", lambda: v[-1])
```

trieste/models/keras/models.py
```diff
@@ -359,15 +367,82 @@ def optimize(self, dataset: Dataset) -> None:
...
+        For custom logs user will need to subclass the model and overwrite this method.
+
+        :param dataset: Optional data that can be used to log additional data-based model summaries.
+        """
+        summary_writer = logging.get_tensorboard_writer()
+        if summary_writer:
+            with summary_writer.as_default(step=logging.get_step_number()):
+                logging.scalar("num_epochs", len(self.model.history.epoch))
+                for k, v in self.model.history.history.items():
+                    logging.histogram(f"{k}_epoch", v)
```
uri-granta: Perhaps make all these collapsible under `k`? Also what sort of names do `k`s have? Should there be a prefix?

```suggestion
                    logging.histogram(f"{k}/epoch", v)
```

trieste/models/keras/models.py
```diff
@@ -359,15 +367,82 @@ def optimize(self, dataset: Dataset) -> None:
...
+                    logging.scalar(f"{k}_final", v[-1])
+                    logging.scalar(f"{k}_diff", v[0] - v[-1])
+                    logging.scalar(f"{k}_min", tf.reduce_min(v))
+                    logging.scalar(f"{k}_max", tf.reduce_max(v))
+                if dataset:
+                    predict = self.predict(dataset.query_points)
+                    # training accuracy
+                    diffs = predict[0] - tf.cast(dataset.observations, predict[0].dtype)
+                    logging.histogram("accuracy_absolute_errors", tf.math.abs(diffs))
```
uri-granta: Again, I'd collate all `accuracy` logs together, and similarly for `variance` logs.

```suggestion
                    logging.histogram("accuracy/absolute_errors", tf.math.abs(diffs))
```

uri-granta (approved):
LGTM, though it'd be good to run a screenshot of the logs past someone with a better sense of how useful they are.

trieste/models/keras/models.py
avullo: How about logging the histogram of the normalised residuals? You'd get a better indication of the behaviour of the model independent on the scale of the differences.
On a similar note, you could log the Z-score of the normalised residuals (should be close to 1).

trieste/models/keras/models.py
hstojic: tried to improve that part further

trieste/models/keras/models.py
hstojic: good idea, I have added 
- z residuals histogram
- variance error histogram
- z residuals std scalar
- root mean variance error scalar

trieste/models/keras/models.py
```diff
@@ -409,39 +409,59 @@ def log(self, dataset: Optional[Dataset] = None) -> None:
...
-                logging.scalar("num_epochs", len(self.model.history.epoch))
+                logging.scalar("epochs/num_epochs", len(self.model.history.epoch))
                 for k, v in self.model.history.history.items():
-                    logging.histogram(f"{k}_epoch", v)
-                    logging.scalar(f"{k}_final", v[-1])
-                    logging.scalar(f"{k}_diff", v[0] - v[-1])
-                    logging.scalar(f"{k}_min", tf.reduce_min(v))
-                    logging.scalar(f"{k}_max", tf.reduce_max(v))
+                    k_sep = k.split("_")
+                    if "loss" in k and "model" in k:
```
uri-granta: This is very arcane! Without knowing the contents of history I can't tell what the names will be (and whether this will ever throw an exception due to k_sep being size 1). Maybe a few comments with examples?

uri-granta (approved):
Still looks good!

trieste/models/keras/models.py
hstojic: I was unhappy with this bit, feels a bit brittle to rely on output names assigned by KerasEnsemble - here I was trying to separate loss as a group and identify if any other metrics is in the output (here we have `mse` by default in `DeepEnsemble`) and if so make it a grouping

the alternative is to go with the unprocessed variant but then loss and whatever other metrics are defined will be mixed in ... perhaps I can get metrics names somehow from the compiled model, that would make it cleaner

hstojic: the default output looks like: 
```
{"loss":[...], "model_i_output_loss":[...], "model_i_output_mse":[...]}
```
where i is the member of ensemble, so we would have e.g. 10 of those

uri-granta: Maybe use a regex based whitelist instead then? Clearer and more resilient.

```python
KEY_SPLITTER = {
    # map history keys to prefix and suffix
    "loss" : ("loss", ""),
    r"(?P<model>model_\d+)_output_loss" : ("loss", r"_\g<model>"),
    r"(?P<model>model_\d+)_(?P<metric>.+)" : (r"\g<metric>", r"_\g<model>"),
}
for pattern, (pre_sub, post_sub) in KEY_SPLITTER.items():
    if re.match(pattern, k):
        pre = re.sub(pattern, pre_sub, k)
        post = re.sub(pattern, post_sub, k)
        break
else:
    # unrecognised history key; ignore
    continue
```

PR #601, opened 2022-08-08
Alex/qhsri

AlexGoodallAI:


tox.ini
```diff
@@ -41,6 +41,7 @@ commands =
     format: black -l 80 --include \.pct\.py$ --check notebooks
     # tests (run with "-- --runslow yes" to run all tests)
     tests: pip install . -r tests/requirements.txt -c tests/constraints.txt
```
uri-granta, line 43: You can put it all on this line:

```suggestion
    tests: pip install .[qhsri] -r tests/requirements.txt -c tests/constraints.txt
```

Though if you follow the suggestion for updating `generate_constraints.sh`, changing this file might not be necessary at all (as the qhsri requirements ill be in constraints.txt).

setup.py
```diff
@@ -50,5 +50,6 @@
     ],
     extras_require={
         "plotting": ["seaborn", "plotly"],
+        "qhsri": ["pymoo", "cvxpy"],
```
uri-granta, line 53: To ensure the tests and notebooks always run with the same version of pymoo and cvxpy (and their dependencies) you should also update `generate_constraints.sh` to do `pip install -e .[qhsri]` on line 31. Rerunning the script will then regenerate the constraints.txt files for the tests and notebooks so that they include these libraries.

trieste/acquisition/rule.py
```diff
@@ -1068,3 +1081,118 @@ def state_func(
...
         return state_func
+
+
+class BatchHypervolumeSharpeRatioIndicator(
+    AcquisitionRule[TensorType, SearchSpaceType, ProbabilisticModelType]
+):
+    def __init__(
+        self, batch_size: int = 5, population_size: int = 500, filter_threshold: float = 0.4
+    ):
+        if pymoo is None:
```
uri-granta, line 1125: (remember to add a docstring explaining the rule and parameters: half the people will try and figure out how to use this from the tutorial, the other half from here)

trieste/acquisition/rule.py
```diff
@@ -1068,3 +1081,118 @@ def state_func(
             return state_, points
 
         return state_func
+
+
+class BatchHypervolumeSharpeRatioIndicator(
+    AcquisitionRule[TensorType, SearchSpaceType, ProbabilisticModelType]
```
uri-granta: For now, you can change this (and the other occurrences) of `ProbabilisticModelType` to `ProbabilisticModel`. The former is a type variable that lets us restrict rules to support only certain types of model. This rule can (currently) support any model (just like ProbabilityOfImprovement). If in future we allow users to specify e.g other builders then we'll need to make this a type variable again, so that they can pass e.g. a builder that only supports certain types of model (so long as they only use the rule with that type of model).

```suggestion
    AcquisitionRule[TensorType, SearchSpaceType, ProbabilisticModel]
```

trieste/acquisition/rule.py
```diff
@@ -1068,3 +1081,118 @@ def state_func(
...
+
+        self._builder: AcquisitionFunctionBuilder[ProbabilisticModelType] = builder
+        self._batch_size: int = batch_size
+        self._population_size: int = population_size
+        self._filter_threshold: float = filter_threshold
+        self._acquisition_function: Optional[AcquisitionFunction] = None
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""BatchHypervolumeSharpeRatioIndicator(
```
uri-granta, line 1142: Why not include the other parameters too?

```suggestion
        return f"""BatchHypervolumeSharpeRatioIndicator(batch_size={batch_size}, population_size={population_size}, filter_threshold={filter_threshold})"""
```

trieste/acquisition/rule.py
```diff
@@ -1068,3 +1081,118 @@ def state_func(
...
+
+        return out_points, out_mean_std
+
+    def acquire(
+        self,
+        search_space: SearchSpaceType,
+        models: Mapping[str, ProbabilisticModelType],
+        datasets: Optional[Mapping[str, Dataset]] = None,
+    ) -> TensorType:
+
```
uri-granta, line 1211: (docstring here can just be copy and pasted from the parent class)

uri-granta (commented):
Just a couple of early comments. Looks nice and clean overall.

AlexGoodallAI:
Still to do:

- [x] Add unit tests for `Pareto.sample` sampling method
- [x] Add unit tests for `BatchHypervolumeSharpeRatioIndicator` rule
- [x] Add integration tests
- [ ]  Refactor `Pareto.sample` to make use of existing `hypervolume_indicator` function. 
- [ ]  Add option to use noisy sampling method in `Pareto.sample`

trieste/acquisition/rule.py
```diff
@@ -368,7 +381,8 @@ def _remove_point(pending_points: TensorType, point_to_remove: TensorType) -> Te
 
         tf.debugging.assert_shapes(
             [(self.pending_points, [None, "D"]), (points_to_remove, [None, "D"])],
-            message=f"""Point to remove shall be 1xD where D is the last dimension of pending points.
+            message=f"""Point to remove shall be 1xD where D is the last dimension
```
uri-granta: Probably a bit cleaner like this (though really I should go through the codebase an clean all these up so they indent properly when raised)

```suggestion
            message=f"""
            Point to remove shall be 1xD where D is the last dimension of pending points.
            Got {tf.shape(self.pending_points)} for pending points
            and {tf.shape(points_to_remove)} for other points.
            """,
```

trieste/acquisition/rule.py
uri-granta: (remember to change this to ProbabilisticModel. You can always change it back to ProbabilisticModelType later if you need to when you make the rule more general)

trieste/acquisition/rule.py
```diff
@@ -1068,3 +1082,173 @@ def state_func(
             return state_, points
 
         return state_func
+
+
+class BatchHypervolumeSharpeRatioIndicator(
+    AcquisitionRule[TensorType, SearchSpaceType, ProbabilisticModelType]
+):
+    """Implements the Batch Hypervolume Sharpe-ratio indicator acquisition
+    rule, designed for large batches"""
```
uri-granta: You should probably include a citation here (see other docstrings for examples, or ask).

trieste/acquisition/rule.py
```diff
@@ -1068,3 +1082,173 @@ def state_func(
...
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""BatchHypervolumeSharpeRatioIndicator(
+        batch_size={self._batch_size}, ga_population_size={self._population_size},
+        ga_n_generations={self._n_generations}, filter_threshold={self._filter_threshold}
+        )
+        """
+
+    class MeanStdTradeoff(PymooProblem):  # type: ignore[misc]
```
uri-granta: In general it's best to avoid inner classes, as they prevent objects from being pickled. You could instead make this a top level class but prefix its name with `_` (so `_MeanStdTradeoff`) to indicate that's it's private.

trieste/acquisition/rule.py
```diff
@@ -1068,3 +1082,173 @@ def state_func(
...
+        # Filter out points below a threshold probability of improvement
+        filtered_points, filtered_mean_std = self._filter_points(nd_points, nd_mean_std)
+
+        # Set up a Pareto set of the filtered points
+        pareto_set = Pareto(filtered_mean_std, already_non_dominated=True)
+
+        # Sample points from set using qHSRI
+        _, batch_ids = pareto_set.sample(self._batch_size)
+
+        batch = filtered_points[batch_ids]
```
uri-granta, line 1235: Is there anything you want to log to tensorboard here?

tox.ini
```diff
@@ -40,14 +40,14 @@ commands =
...
+    tests: pip install .[qhsri] -r tests/requirements.txt -c tests/constraints.txt
     tests: pytest --ignore trieste/experimental {posargs}
     # docs build
-    docs: pip install . -r notebooks/requirements.txt -c notebooks/constraints.txt
+    docs: pip install .[qhsri] -r notebooks/requirements.txt -c notebooks/constraints.txt
     docs: pip install -r docs/requirements.txt -c docs/constraints.txt
     docs: bash -c "cd docs; make html"
     # quickdocs build
-    quickdocs: pip install . -r notebooks/requirements.txt -c notebooks/constraints.txt
+    quickdocs: pip install .[qhsri] -r notebooks/requirements.txt -c notebooks/constraints.txt
```
uri-granta, line 50: You'll also need to update `tests_latest`.

Though I'm now concerned that if we always test with `[qhsri]` then we won't spot a dependency break in the basic trieste package. So maybe you should _only_ test qhsri with `tests_latest` (and `docs` and `quickdocs`), and have the normal `tests` just not use qhsri and instead skip the qhsri tests?

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +81,135 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
         )
         return hypervolume_indicator
 
+    def sample(self, sample_size: int) -> Tuple[TensorType, TensorType]:
```
uri-granta: Since we have `from __future__ import annotations`, you can (and should) do this instead.

```suggestion
    def sample(self, sample_size: int) -> tuple[TensorType, TensorType]:
```

tests/unit/acquisition/test_rule.py
```diff
@@ -817,3 +818,77 @@ def test_asynchronous_rule_add_pending_points() -> None:
...
+def test_qhsri_raises_for_invalid_dataset_keys(
+    datasets: dict[str, Dataset], models: dict[str, ProbabilisticModel]
+) -> None:
+    search_space = Box([-1], [1])
+    rule = BatchHypervolumeSharpeRatioIndicator()
+    with pytest.raises(ValueError):
+        rule.acquire(search_space, models, datasets=datasets)
+
+
+# @pytest.mark.parametrize("num_query_points", [5, 10])
```
uri-granta: WIP or are you stuck here?

Also don't forget to add an integration test (which is the proof of the pudding). Feel free to ask me if you have any questions.

uri-granta (commented):
Looks really good. A few more comments. Feel free to ask me about remaining tests.

tests/unit/acquisition/test_rule.py
AlexGoodallAI: Ah yeah, I couldn't get it to work because I was getting: ` NotImplementedError: Model QuadraticMeanAndRBFKernel() does not support predicting observations, just the latent function`, and I wasn't sure of the best way to proceed, whether to use a different kernel, or add a dummy prediction to that kernel.

tests/unit/acquisition/multi_objective/test_pareto.py
```diff
@@ -104,3 +105,74 @@ def test_get_reference_point_extract_based_on_pareto_front(
     observations: tf.Tensor, expected: tf.Tensor
 ) -> None:
     tf.debugging.assert_equal(get_reference_point(observations), expected)
+
+
+@pytest.mark.qhsri
```
vpicheny, line 110: what does this do? `@pytest.mark.qhsri`

maybe I'm misreading this but it looks like it's a restricted test, while this tests our generic Pareto object?

trieste/acquisition/function/function.py
```diff
@@ -42,6 +42,89 @@
...
+        tf.debugging.Assert(dataset is not None, [])
+        dataset = cast(Dataset, dataset)
+        tf.debugging.assert_positive(len(dataset), message="Dataset must be populated.")
+        mean, _ = model.predict(dataset.query_points)
+        eta = tf.reduce_min(mean, axis=0)
+        function.update(eta)  # type: ignore
+        return function
+
+
+class probability_of_improvement(AcquisitionFunctionClass):
```
vpicheny, line 93: do we need this? it is basically equivalent to probability_of_feasibility
 https://github.com/secondmind-labs/trieste/blob/develop/trieste/acquisition/function/function.py#L390

if I'm not mistaken?

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,184 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
         )
         return hypervolume_indicator
 
+    def sample(self, sample_size: int, allow_repeats: bool = True) -> tuple[TensorType, TensorType]:
```
vpicheny, line 82: maybe use a more explicit name? e.g. `sample_diverse_subset` or something along those lines?

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,184 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+
+        if cp is None:
+            raise ImportError(
+                "Pareto.sample method requires cvxpy, "
+                "this can be installed via `pip install trieste[qhsri]`"
+            )
+
+        front_size, front_dims = self.front.shape
+
+        if front_size < sample_size:
```
vpicheny, line 99: is it still a problem if you allow repeats?

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,184 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+        deltas = [
+            (float(max(self.front[:, i])) - float(min(self.front[:, i]))) * 0.2
+            for i in range(front_dims)
+        ]
+
+        # Define lower bound and reference point
+        lower_bound = [float(min(self.front[:, i])) - deltas[i] for i in range(front_dims)]
+
+        # Use deltas and max values to create reference point
+        reference_point = [float(max(self.front[:, i])) + deltas[i] for i in range(front_dims)]
```
vpicheny, line 256: why is this not called upper_bound?

vpicheny: if you prefer technical terms you can use utopia_point = lower bound and nadir_point = upper bound

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,184 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+                    " in a dimension, this means you either have multiples"
+                    " of a single point, or only one point"
+                )
+            denominator *= reference_point[i] - lower_bound[i]
+
+        # Fill entries of p
+        for i in range(front_size):
+            for j in range(front_size):
+                pij = 1
+                for k in range(front_dims):
```
vpicheny, line 234: I think that it's possible to remove this line with `cp.cummax` but not sure if worth it

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,184 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+
+        return p
+
+    def _get_bounds(self) -> tuple[TensorType, TensorType]:
+
+        front_dims = self.front.shape[1]
+
+        # Calculate the deltas to add to the bounds to get the reference point and lower bound
+        deltas = [
+            (float(max(self.front[:, i])) - float(min(self.front[:, i]))) * 0.2
```
vpicheny, line 248: why the 0.2? should this be a parameter of the method instead?

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,184 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+        # Calculate denominator value for p matrix elements
+        denominator: float = 1
+        for i in range(front_dims):
+            if reference_point[i] - lower_bound[i] == 0:
+                raise ValueError(
+                    "Pareto set has identical upper and lower bounds"
+                    " in a dimension, this means you either have multiples"
+                    " of a single point, or only one point"
+                )
+            denominator *= reference_point[i] - lower_bound[i]
```
vpicheny, line 228: Will crash later if they are equal - by construction they shouldn't, unless your Pareto front is a single point where this would happen. In that case your delta is zero and your bounds are equal. 

Single point Pareto front is not a corner case so we should have something that works in this case.

PR #605, opened 2022-08-15
logging improvements

hstojic:
2 main things:
1. now logging initial model fitting as step 0 
2. allow dataset as an argument to log method to allow for data-based model diagnostics

few smaller things
- updated notebook a bit to reflect the changes
- also removed restriction that step numer is non-negative, its not really necessary and in ask-tell case users might want to log steps differently


trieste/bayesian_optimizer.py
```diff
@@ -669,11 +669,23 @@ def optimize(
...
-                        if step == 0 and fit_initial_model:
+                        if step == 1 and fit_initial_model:
                             for tag, model in models.items():
                                 dataset = datasets[tag]
                                 model.update(dataset)
                                 model.optimize(dataset)
+                            
+                    if step == 1 and fit_initial_model and summary_writer:
+                        logging.set_step_number(0)
+                        with summary_writer.as_default(step=0):
```
hstojic, line 678: @uri-granta for some reason this didn't work without first doing `logging.set_step_number(0)`, although if I understand correctly it should have worked without issues...

uri-granta: The summary writer step argument is separate from the logging step state so you have to specify both. You can use a context manager if you prefer though, so you don't have to worry about resetting it afterwards.

```suggestion
                        with logging.step_number(0):
                            with summary_writer.as_default(step=0):
```

hstojic: thanks for the suggestion, that's definitely a better way to do it

hstojic: ok, that didnt work at the end for logging.scalar line below, logging wallclock model fitting time, so I reverted back

trieste/bayesian_optimizer.py
```diff
@@ -667,14 +667,25 @@ def optimize(
                             "will be available."
                         ) from e
 
-                with Timer() as total_step_wallclock_timer:
```
hstojic, line 670: @uri-granta I have removed this bit as initial model fitting wallclock is now shown as step 0, so also excluded from total step time

trieste/bayesian_optimizer.py
```diff
@@ -707,7 +718,6 @@ def optimize(
                             models,
                             query_points,
                             tagged_output,
-                            initial_model_fitting_timer,
```
hstojic, line 710: not needed anymore as logged separately now

trieste/logging.py
```diff
@@ -83,8 +83,6 @@ def set_step_number(step_number: int) -> None:
     :raise ValueError: if step_number < 0
     """
     global _STEP_NUMBER
-    if step_number < 0:
-        raise ValueError(f"step_number must be non-negative (got {step_number})")
```
hstojic, line 87: unnecessary in my opiniopn, in ask-tell one might log initial model stuff as -1 for example

trieste/models/interfaces.py
```diff
@@ -90,9 +90,11 @@ def predict_y(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
             f"Model {self!r} does not support predicting observations, just the latent function"
         )
 
-    def log(self) -> None:
+    def log(self, dataset: Dataset) -> None:
         """
         Log model-specific information at a given optimization step.
+
+        :param dataset: The data that can be used to log additional data-based model summaries.
```
hstojic: @uri-granta added dataset argument as its more flexible, there are some metrics we would like to log that are based on the training data

hstojic: perhaps we should make it optional though?

docs/notebooks/visualizing_with_tensorboard.pct.py
```diff
@@ -153,17 +153,18 @@ def acquire(self, search_space, models, datasets=None):
...
 # ## Using Tensorboard with Ask-Tell Optimization
 #
-# To use Tensorboard logging with the [Ask-Tell interface](ask_tell_optimization.ipynb), you must also explicitly set the optimization step number before calling ask or tell:
+# To use Tensorboard logging with the [Ask-Tell interface](ask_tell_optimization.ipynb), you must also explicitly set the optimization step number before calling ask or tell methods. If fitting the model on initialization of the Ask-Tell interface (default), then you must also set the optimization step number before initializing the interface to log initial model fitting as a different step.
 
 # %%
 summary_writer = tf.summary.create_file_writer("logs/tensorboard/experiment4")
 trieste.logging.set_tensorboard_writer(summary_writer)
 
+trieste.logging.set_step_number(0)
```
uri-granta, line 162: (step number does deault to 0, so strictly speaking this isn't necessary, but probably best to leave it in)

trieste/logging.py
uri-granta: I don't think that works with tensorboard (can't remember if you get an error or if the data just disappears)

trieste/models/gpflow/interface.py
```diff
@@ -144,9 +144,11 @@ def optimize(self, dataset: Dataset) -> None:
         """
         self.optimizer.optimize(self.model, dataset)
 
-    def log(self) -> None:
+    def log(self, dataset: Dataset) -> None:
         """
         Log model-specific information at a given optimization step.
+
+        :param dataset: The data that can be used to log additional data-based model summaries.
```
uri-granta: Should probably be optional and default to None (like acquisition function builders)

trieste/models/interfaces.py
uri-granta: Yes, make this optional, and default it to None (like acquisition function builders)

trieste/bayesian_optimizer.py
uri-granta: :+1:

trieste/bayesian_optimizer.py
uri-granta: Was the indentation correct? What error did you get? The context manager should be totally equivalent to setting and resetting like you're doing now, but cleaner and more robust.

trieste/bayesian_optimizer.py
```diff
@@ -667,14 +667,25 @@ def optimize(
...
+                            model.update(dataset)
+                            model.optimize(dataset)
+                    if summary_writer:
+                        logging.set_step_number(0)
+                        with summary_writer.as_default(step=0):
                             for tag, model in models.items():
-                                dataset = datasets[tag]
-                                model.update(dataset)
-                                model.optimize(dataset)
+                                with tf.name_scope(f"{tag}.model"):
```
uri-granta: maybe move this to a `_write_summary_initial_fit` function (to go alongside `_write_summary_step` and `_write_summary_init`)?

tests/unit/test_logging.py
```diff
@@ -88,11 +88,6 @@ def test_set_get_step_number(step: int) -> None:
     assert get_step_number() == 0
 
 
-def test_set_step_number_error() -> None:
-    with pytest.raises(ValueError):
-        set_step_number(-1)
```
uri-granta, line 93: (remember to restore this if you bring back the check)

tests/unit/models/gpflow/test_models.py
```diff
@@ -263,14 +263,15 @@ def test_gpflow_wrappers_log(
...
+    data = Dataset(x, fnc_3x_plus_10(x))
 
     if kernel is not None:
         model.model.kernel = kernel
 
     mocked_summary_writer = unittest.mock.MagicMock()
     with tensorboard_writer(mocked_summary_writer):
         with step_number(42):
-            model.log()
+            model.log(data)
```
uri-granta: (If you make data optional you can skip passing it in here)

uri-granta (commented):
Looks good. Various comments.

trieste/bayesian_optimizer.py
hstojic: yes, indentation was correct I think, there was no error, scalar simply wasn't logged at all

trieste/logging.py
hstojic: it works, tried it with ask-tell interface

trieste/models/interfaces.py
hstojic: done!

trieste/bayesian_optimizer.py
hstojic: done

uri-granta (approved):
LGTM!

PR #607, opened 2022-08-22
Make Acquisition Functions pickleable

uri-granta:


hstojic (approved):
there are some pieces I'm not sure I get, but looks goof to me

PR #611, opened 2022-08-25
fixing trajectory sampler for deep ensemble

### LINKED ISSUE #609: bug with `deep_ensemble_trajectory` diversify option
Trajectory sampler at the moment with `diversify` option samples quantiles from randomly chosen networks and uses that as a tproxy for the trajectory - this is then repeated for the whole batch and idea was to use it to be able to get more trajectories than the networks we have in the ensemble.

This was not correct idea unfortunately, as we end up not drawing from distributions that contain epistemic/estimation uncertainty - this part in deep ensembles comes from combining all the networks. This particularly is problematic with noiseless data where variances of individual network vanish quickly and batch becomes not diverse enough very quickly. 

The fix is easy, we should sample trajectories as quantiles from the approximate posterior of the whole ensemble.

hstojic:
this PR closes #609


trieste/models/gpflux/__init__.py
```diff
@@ -21,3 +21,10 @@
 from .builders import build_vanilla_deep_gp
 from .interface import GPfluxPredictor
 from .models import DeepGaussianProcess
+from .sampler import (
```
uri-granta, line 24: Are all of these really worth exporting from trieste.models.gpflux or are some of them implementation details that can remain in trieste.models.gpflux.sampler?

trieste/models/keras/sampler.py
```diff
@@ -121,12 +131,14 @@ def __init__(self, model: DeepEnsembleModel, diversify: bool):
 
         self._initialized = tf.Variable(False, trainable=False)
         self._batch_size = tf.Variable(0, dtype=tf.int32, trainable=False)
-        self._indices = tf.Variable(tf.zeros([0], dtype=tf.int32), shape=[None], trainable=False)
 
         if self._diversify:
-            self._sample_size = tf.Variable(0, dtype=tf.int32, trainable=False)
-            self._seeds = tf.Variable(
-                tf.zeros([0, 0], dtype=tf.float32), shape=[None, None], trainable=False
+            self._indices = tf.Variable(
```
uri-granta: Why is this now called `_indices` if it's still a collection of floats? Maybe call it `_percentiles`?

trieste/models/keras/sampler.py
```diff
@@ -154,20 +164,13 @@ def __call__(self, x: TensorType) -> TensorType:  # [N, B, d] -> [N, B, 1]
...
-                    for i, dist in enumerate(ensemble_distributions)
-                ]
-            )  # [E, B/E, N*B, 1]
-            flattened_samples = tf.reshape(
-                samples, [self._ensemble_size * self._sample_size, *samples.shape[2:]]
-            )  # [E*B/E, N*B, 1]
-            predictions = tf.gather(flattened_samples, self._indices)  # [B, N*B, 1]
+            predicted_means, predicted_vars = self._model.predict(flat_x)  # ([N*B, 1], [N*B, 1])
+            normal = tfp.distributions.Normal(predicted_means, tf.sqrt(predicted_vars))
+            predictions = tf.map_fn(normal.quantile, self._indices)  # [B, N*B, 1]
```
uri-granta: Nice and clean

trieste/models/keras/sampler.py
```diff
@@ -188,15 +191,25 @@ def resample(self) -> None:
...
                     minval=0.000001,
                     maxval=0.999999,
                     dtype=tf.float32,
                 )
-            )  # [E, B/E]
+            )  # [B]
         else:
             self._indices.assign(sample_model_index(self._ensemble_size, self._batch_size))  # [B]
+
+    def get_state(self) -> Dict[str, TensorType]:
```
uri-granta, line 205: Any reason to having this, rather than say three properties `self.initialized`, `self.batch_size`, `self.indices`?

uri-granta (approved):
Looks good! Just a couple of questions.

trieste/models/keras/sampler.py
hstojic: just thinking ahead, so you can get state info from multiple types of samplers with the same method

trieste/models/gpflux/__init__.py
hstojic: I have just noticed that nothing was exported and took a look at gpflow, there everything was exported so I have jusst copied the approach

PR #612, opened 2022-08-25
additional query point diagnostics for batch BO

### LINKED ISSUE #608: additional tensorboard logs for batches
currently we have histograms of query points when it comes to batch BO that are useful for checking the diversity, that all seems to work fine - there are two more that would be quite useful:

1. histogram of pairwise distances between query points - its easy to compute pairwise euclidean distances with Scipy's pdist and then use `logging.histogram`  
2. pairplot with individual scatterplots for each input dimension and model predicted output - this one is a bit more work and we would need to use image summary

hstojic:
closes #608


trieste/ask_tell_optimization.py
```diff
@@ -457,6 +458,7 @@ def ask(self) -> TensorType:
                             logging.scalar(f"query_points/[{i}]", float(query_points[0, i]))
                         else:
                             logging.histogram(f"query_points/[{i}]", query_points[:, i])
+                    logging.histogram("query_points/euclidean_distances", pdist(query_points))
```
uri-granta: Probably fine here, but generally better to use a closure for two reasons: it doesn't waste time doing the computation if you disable the log, and it catches any exception that the computation might raise.

```suggestion
                    logging.histogram("query_points/euclidean_distances", lambda: pdist(query_points))
```

trieste/bayesian_optimizer.py
```diff
@@ -910,15 +911,22 @@ def _write_summary_step(
                     logging.scalar(f"query_points/[{i}]", float(query_points[0, i]))
                 else:
                     logging.histogram(f"query_points/[{i}]", query_points[:, i])
+            logging.histogram("query_points/euclidean_distances", pdist(query_points))
 
         if pd and sns and logging.include_summary("query_points/_pairplot"):
             columns = [f"x{i}" for i in range(tf.shape(query_points)[1])]
-            query_new_df = pd.DataFrame(query_points, columns=columns).applymap(float)
+            qp_preds = query_points
```
uri-granta, line 918: What exactly is this doing? Is it changing the pairplots to show the predictions at the query points rather than the query points themselves? Have you run this past Victor (who originally demanded this plot)?

uri-granta (commented):
One question regarding the pairplot change

trieste/bayesian_optimizer.py
hstojic: That's right, adds predictions at qp's. This is uncontraversial change, qp's are still there and this adds useful information.

trieste/ask_tell_optimization.py
hstojic: done

uri-granta (approved):
LGTM!

hstojic:
here are pics of how the additional plots look like

euclidean distances
Screenshot from 2022-08-26 11-06-49

qp scatter plot with model predictions
Screenshot from 2022-08-26 10-51-14

PR #615, opened 2022-08-31
Deepcopy keras model history callback

uri-granta:
Turns out the history callback is created automatically and stored separately from the other callbacks. Make sure we copy it (handling the fact that, like other callbacks, it may contain a non-pickleable model).

The other callback that's added automatically is ProgbarLogger for logging, but that's not stored so we don't need to copy it.


hstojic (approved):
looks good, though I it seems we are duplicating a lot of code between keras and gpflux for deepcopying (in previous PRs as well), I wonder whether we can have them one place as utility functions that we can call for both models

PR #620, opened 2022-09-05
gpflux log method

### LINKED ISSUE #598: tensorboard logging model parameters or training stats for GPflux models

hstojic:
introducing the log method, largely based on keras and gpflow loggers, which should help with diagnosing the health of the model

this would close #598


trieste/models/gpflux/models.py
```diff
@@ -60,11 +62,10 @@ def __init__(
...
             rather than a model can help when copying or serialising.
-        :param optimizer: The optimizer configuration for training the model. Defaults to
-            :class:`~trieste.models.optimizer.KerasOptimizer` wrapper with
-            :class:`~tf.optimizers.Adam` optimizer. The ``optimizer`` argument to the wrapper is
-            used when compiling the model and ``fit_args`` is a dictionary of arguments to be used
-            in the Keras ``fit`` method. Defaults to 400 epochs, batch size of 1000, and verbose 0.
+        :param optimizer: The optimizer wrapper with necessary specifications for compiling and
+            training the model. Defaults to :class:`~trieste.models.optimizer.KerasOptimizer` with
+            :class:`~tf.optimizers.Adam` optimizer, mean squared error metric and a dictionary of
+            default arguments for Keras `fit` method: 400 epochs, batch size of 1000, and verbose 0.
```
sebastianober:
```suggestion
            default arguments for the Keras `fit` method: 400 epochs, batch size of 1000, and verbose 0.
```

trieste/models/gpflux/models.py
```diff
@@ -354,3 +358,69 @@ def optimize(self, dataset: Dataset) -> None:
         # so that the next time we call `optimize` the starting learning rate would be different.
         # Therefore we make sure the learning rate is set back to its initial value.
         self.optimizer.optimizer.lr.assign(self.original_lr)
+
+    def log(self, dataset: Optional[Dataset] = None) -> None:
+        """
+        Log model training information at a given optimization step to the Tensorboard.
+        We log few summary statistics of losses, layer KL divergences and metrics (as provided in
```
sebastianober:
```suggestion
        We log a few summary statistics of losses, layer KL divergences and metrics (as provided in
```

sebastianober:
Looks generally good to me, but I don't quite see where you've tracked KLs for each layer - is that in here?

sebastianober:
Maybe also include an SNR for the last layer kernel variance and likelihood variance? Also look at what to do with more complicated likelihoods (e.g. heteroskedastic)

hstojic:
@sebastianober I think I have fixed the likelihood now, but had issues with implementing a model with het likelihood to test it out so haven't included a test - SNR is a nice idea but would be a lot of additional code to justify the value, those who want it can compare the two figures

sebastianober (approved):
LGTM!

PR #621, opened 2022-09-06
add support for multifidelity modeling

AlexGoodallAI:


hstojic (commented):
interesting stuff, curious to see how it develops

seems like we could have a specific Dataset object for multifidelity problems and outsource as much data manipulation as possible there

trieste/data.py
```diff
@@ -108,3 +110,65 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+    to be reasonably considered to represent fidelities.
+    The final input column of multi-fidelity data should be a reference to
+    the fidelity of the query point. We cannot have mixed type tensors, but
+    we can check that thhe final column values are suitably close to integers.
+    :param query_points: Data to check final column of.
+    :raise AssertionError: If any value in the final column is far from an integer
+    """
+
+    fidelity_col = query_points[:, -1]
+    npt.assert_allclose(
```
uri-granta: Maybe use `tf.debugging.assert_equal` instead? Expect it plays more nicely with tensorflow compilation.

trieste/data.py
```diff
@@ -108,3 +110,65 @@ def astuple(self) -> tuple[TensorType, TensorType]:
         :return: A 2-tuple of the :attr:`query_points` and :attr:`observations`.
         """
         return self.query_points, self.observations
+
+
```
uri-granta, line 113: I'll try to get back to this comment once I've looked at more of the code, but does it make sense to have a `FidelityDataset` type alias for describing Datasets with a fidelity column? It'll allow you to have a docstring describing the expected format (i.e. that the final column corresponds to an integral fidelity), and will make functions like split_dataset_by_fidelity easier to read. (It might also make it easier to refactor if we ever want to change how we pass around per-fidelity observations.)

trieste/data.py
```diff
@@ -108,3 +110,65 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+        fidelity_col,
+        err_msg="Fidelity column should be float(int), but"
+        "got a float that was not close to an int",
+    )
+
+
+def split_dataset_by_fidelity(dataset: Dataset, num_fidelities: int) -> Sequence[Dataset]:
+    """Split dataset into individual datasets without fidelity information
+    :param dataset: Dataset for which to split fidelities
+    :param num_fidlities: Number of fidelities in the problem (not just dataset)
```
uri-granta, line 162: Assuming you normally want to get _all_ the fidelities, a possibly cleaner approach might be to return a `Mapping[int, Dataset]` dictionary containing the datasets for all the fidelities in the input? This might be more efficient too, as it might be possible to get tensorflow to extract everything in one go (in numpy I'd use `np.array_split`, not sure about tensorflow). You would still be able to access the specific datasets almost as before, using `fidelities.get(fidelity, Dataset())`.

If this turns out to be too complicated then I wouldn't worry about it though.

trieste/data.py
```diff
@@ -108,3 +110,65 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+    :return: Ordered list of datasets with lowest fidelity at index 0 and highest at -1
+    """
+    assert_valid_fidelity_query_points(dataset.query_points)
+    datasets = []
```
uri-granta: List comprehension?

```suggestion
    return [get_dataset_for_fidelity(dataset, fidelity) for fidelity in range(num_fidelities)]
```

trieste/data.py
```diff
@@ -108,3 +110,65 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+    )
+
+
+def split_dataset_by_fidelity(dataset: Dataset, num_fidelities: int) -> Sequence[Dataset]:
+    """Split dataset into individual datasets without fidelity information
+    :param dataset: Dataset for which to split fidelities
+    :param num_fidlities: Number of fidelities in the problem (not just dataset)
+    :return: Ordered list of datasets with lowest fidelity at index 0 and highest at -1
+    """
+    assert_valid_fidelity_query_points(dataset.query_points)
```
uri-granta: (unnecessary since you already do this each time you call get_dataset_for_fidelity)

trieste/data.py
```diff
@@ -108,3 +110,65 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+    return Dataset(query_points=inputs_for_fidelity, observations=observations_for_fidelity)
+
+
+def convert_query_points_for_fidelity(query_points: TensorType, fidelity: int) -> TensorType:
+    """Add fidelity column to query_points without fidelity data
+    :param query_points: query points without fidelity to add fidelity column to
+    :param fidelity: fidelity to populate fidelity column with
+    :return: TensorType of query points with fidelity column added
+    """
+    if not isinstance(fidelity, int):
```
uri-granta: (We don't normally bother with straight type checks that would be easily caught by the type checker.)

trieste/data.py
```diff
@@ -108,3 +110,65 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+        raise TypeError("Fidelity must be an integer")
+    fidelity_col = tf.ones((tf.shape(query_points)[0], 1), dtype=query_points.dtype) * fidelity
```
uri-granta: Was going to suggest using `tf.fill` but looks like that doesn't support dtype, so fine as is.

trieste/models/gpflow/__init__.py
```diff
@@ -31,6 +31,7 @@
     SparseGaussianProcessRegression,
     SparseVariational,
     VariationalGaussianProcess,
+    AR1,
```
uri-granta: do you want to expose build_ar1_models too or is that only internal?

trieste/models/gpflow/builders.py
```diff
@@ -423,3 +424,29 @@ def _get_inducing_points(
...
 
     return inducing_points
+
+
+def build_ar1_models(dataset: Dataset, num_fidelities: int, input_search_space: SearchSpace):
+    """
+    Build the individual GPR models required for constructing an AR1 model
+    with `num_fidelities` fidelities.
+
+    :param dataset: Dataset of points with which to initialise the individual models
```
uri-granta: (mention that the datasets contain fidelity information?)

trieste/models/gpflow/builders.py
```diff
@@ -423,3 +424,29 @@ def _get_inducing_points(
         inducing_points = search_space.sample(num_inducing_points)
 
     return inducing_points
+
+
+def build_ar1_models(dataset: Dataset, num_fidelities: int, input_search_space: SearchSpace):
```
uri-granta:
```suggestion
def build_ar1_models(dataset: Dataset, num_fidelities: int, input_search_space: SearchSpace) -> Sequence[GaussianProcessRegression]:
```

tests/unit/test_data.py
```diff
@@ -196,3 +203,52 @@ def test_dataset_astuple() -> None:
...
+def test_assert_valid_multifidelity_dataset(query_points: Dataset, is_valid: bool) -> None:
+    if is_valid:
+        assert_valid_fidelity_query_points(query_points)
+    else:
+        with pytest.raises(AssertionError):
+            assert_valid_fidelity_query_points(query_points)
+
+
+def test_multifidelity_split_dataset_by_fidelity() -> None:
+    fidelity_0 = Dataset(tf.constant([[0.456, 0.0]]), tf.constant([[0.2]]))
```
uri-granta: Probably worth testing with more than one point in one of the fidelities.

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,242 @@ def covariance_between_points(
...
+    ):
+        """
+        :param fidelity_models: List of
+            :class:`~trieste.models.gpflow.models.GaussianProcessRegression`
+            models, one for each fidelity.
+        """
+        self.num_fidelities = len(fidelity_models)
+
+        self.lowest_fidelity_signal_model = fidelity_models[0]
+        self.fidelity_residual_models: Sequence[Optional[GaussianProcessRegression]] = [
```
uri-granta: Why can't you just use `self.fidelity_residual_models = fidelity_models`? The code already avoids calling predict/sample/etc for `fidelity == 0`. Happy to discuss.

trieste/data.py
```diff
@@ -108,3 +110,65 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+
+
+def get_dataset_for_fidelity(dataset: Dataset, fidelity: int) -> Dataset:
+    """Get a dataset with only the specified fidelity of data in
+    :param dataset: The dataset from which to extract the single fidelity data
+    :param fidelity: The fidelity to extract the data for
+    :return: Dataset with a single fidelity and no fidelity column
+    """
+    assert_valid_fidelity_query_points(dataset.query_points)
+    input_points = dataset.query_points[:, :-1]  # [..., D+1]
```
uri-granta: Wack suggestion: how about renaming `assert_valid_fidelity_query_points` to `check_and_extract_fidelity_query_points` and have it return `input points, fidelity_col`? You could then also use it in predict and sample.

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,242 @@ def covariance_between_points(
...
+        assert_valid_fidelity_query_points(query_points)
+        query_points_wo_fidelity = query_points[:, :-1]
+        query_points_fidelity_col = query_points[:, -1:]
+
+        signal_mean, signal_var = self.lowest_fidelity_signal_model.predict(
+            query_points_wo_fidelity
+        )
+
+        for fidelity in range(self.num_fidelities):
+            if fidelity > 0:
```
uri-granta: Perhaps more natural to have 0 case first? (and ditto for sample)

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,242 @@ def covariance_between_points(
...
+        extracted by splitting the observations in ``dataset``  by fidelity level.
+        Note that we have to code up a custom loss function when optimizing our residual
+        model, so that we can include the correlation parameter as an optimisation variable.
+        :param dataset: The query points and observations for *all* the wrapped models.
+        """
+        assert_valid_fidelity_query_points(dataset.query_points)
+        dataset_per_fidelity = split_dataset_by_fidelity(dataset, self.num_fidelities)
+
+        for fidelity, dataset_for_fidelity in enumerate(dataset_per_fidelity):
+            print(f"Working on fidelity {fidelity}")
```
uri-granta: You should use tf.print if you want to log to console (search for other examples in the codebase).

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,242 @@ def covariance_between_points(
             query_points_2=query_points_2,
             whiten=whiten,
         )
+
+
+class AR1(TrainableProbabilisticModel):
+    """
```
uri-granta:
```suggestion
    r"""
```

uri-granta (commented):
Looks good! Various comments, feel free to chat. Still to do: unit and integration tests for AR1, some missing type annotations.

trieste/data.py
AlexGoodallAI: After discussion decided against this

trieste/data.py
```diff
@@ -108,3 +109,90 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+    we can check that thhe final column values are suitably close to integers.
+
+    :param query_points: Data to check final column of.
+    :raise: ValueError: If there are not enough columns to be multifidelity data
+    :raise InvalidArgumentError: If any value in the final column is far from an integer
+    :return: Query points without fidelity column
+        and the fidelities of each of the query points
+    """
+    # Check we have sufficient columns
+    if query_points.shape[1] < 2:
```
vpicheny: here and below: it might be useful to assume that the querypoints can be higher order tensors? eg
`got {query_points.shape[-1]}`
`input_points = query_points[..., :-1]`
etc.

trieste/data.py
```diff
@@ -108,3 +109,90 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+    :param dataset: The dataset from which to extract the single fidelity data
+    :param fidelity: The fidelity to extract the data for
+    :return: Dataset with a single fidelity and no fidelity column
+    """
+
+    input_points, fidelity_col = check_and_extract_fidelity_query_points(
+        dataset.query_points
+    )  # [..., D], [..., 1]
+    mask = fidelity_col == fidelity  # [..., ]
+    inds = tf.where(mask)[:, 0]  # [..., ]
```
vpicheny: `tf.where(mask)[..., 0]`?

trieste/data.py
```diff
@@ -108,3 +109,90 @@ def astuple(self) -> tuple[TensorType, TensorType]:
...
+    inputs_for_fidelity = tf.gather(input_points, inds, axis=0)  # [..., D]
+    observations_for_fidelity = tf.gather(dataset.observations, inds, axis=0)  # [..., 1]
+    return Dataset(query_points=inputs_for_fidelity, observations=observations_for_fidelity)
+
+
+def convert_query_points_for_fidelity(query_points: TensorType, fidelity: int) -> TensorType:
```
vpicheny: how about `add_fidelity_column`?

trieste/models/gpflow/builders.py
```diff
@@ -423,3 +424,44 @@ def _get_inducing_points(
...
+        if len(fidelity_data) < 2:
+            raise ValueError(
+                f"Not enough data to create model for fidelity {i},"
+                f" need at least 2 datapoints, got {len(fidelity_data)}"
+            )
+
+    gprs = [
+        GaussianProcessRegression(
+            build_gpr(
+                data[fidelity], input_search_space, likelihood_variance=1e-6, kernel_priors=False
```
vpicheny: does it make sense to make `likelihood_variance` and `kernel_priors` optional inputs?

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+                 and var: Th variance at query_points with shape [N, P]
+        """
+        (
+            query_points_wo_fidelity,
+            query_points_fidelity_col,
+        ) = check_and_extract_fidelity_query_points(
+            query_points, max_fidelity=self.num_fidelities - 1
+        )
+
+        signal_mean, signal_var = self.lowest_fidelity_signal_model.predict(
```
vpicheny, line 1430: just to check my understanding, this fetches the fidelity-1 level?

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+            query_points_wo_fidelity
+        )
+
+        for fidelity in range(self.num_fidelities):
```
vpicheny: This predict is quite expensive since it computes the prediction of all the models at all the data points. I appreciate it's difficult, but would it be possible to have a solution based on masks? The current one may become very slow if we have multiple fidelities and lots of query points. If too challenging for now, keep this solution and raise an issue?

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+
+            mask = query_points_fidelity_col >= fidelity
+            signal_mean = tf.where(mask, new_fidelity_signal_mean, signal_mean)
+            signal_var = tf.where(mask, new_fidelity_signal_var, signal_var)
+
+        return signal_mean, signal_var
+
+    def _calculate_residual(self, dataset: Dataset, fidelity: int) -> TensorType:
+        """
+        Calculate the true residuals for a set of datapoints at a given fidelity.
```
vpicheny, line 1469: "true residuals" to me means purely data based. Might be good to add the equation here of what's computed for clarity?

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+            query_points_fidelity_col,
+        ) = check_and_extract_fidelity_query_points(
+            query_points, max_fidelity=self.num_fidelities - 1
+        )
+
+        signal_sample = self.lowest_fidelity_signal_model.sample(
+            query_points_wo_fidelity, num_samples
+        )
+
+        for fidelity in range(self.num_fidelities):
```
vpicheny: same as predict, going through all fidelities and query points is going to be very expensive.

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+        Predict the marginal mean and variance at `query_points` including observation noise
+
+        :param query_points: Query points with shape [N, D+1], where the
+            final column of the final dimension contains the fidelity of the query point
+        :return: mean: The mean at query_points with shape [N, P],
+                 and var: Th variance at query_points with shape [N, P]
+        """
+
+        f_mean, f_var = self.predict(query_points)
+        query_points_fidelity_col = query_points[:, -1:]
```
vpicheny: should work with `query_points[..., -1:]` I think?

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+            * self.lowest_fidelity_signal_model.get_observation_noise()
+        )
+
+        for fidelity in range(1, self.num_fidelities):
+
+            fidelity_observation_noise = (
+                self.rho[fidelity] ** 2
+            ) * observation_noise + self.fidelity_residual_models[fidelity].get_observation_noise()
+
+            mask = query_points_fidelity_col >= fidelity
```
vpicheny, line 1551: why `>=` and not equal?

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+            dataset.query_points, max_fidelity=self.num_fidelities - 1
+        )
+        dataset_per_fidelity = split_dataset_by_fidelity(dataset, self.num_fidelities)
+        for fidelity, dataset_for_fidelity in enumerate(dataset_per_fidelity):
+            if fidelity == 0:
+                self.lowest_fidelity_signal_model.update(dataset_for_fidelity)
+            else:
+                # Make query points but with final column corresponding to
+                # fidelity we wish to predict at
+                self.fidelity_residual_models[fidelity].update(
```
vpicheny, line 1574: For my understanding: I thought that the AR1 model was taking into account the prediction uncertainty of the lower layer. But this is not the case here since we only pass the residual values? Is it because I have it wrong, or because it's hard to do? In the latter case then this model isn't an AR1? which might be OK for now but we'd need to write an issue to fix it.

vpicheny: looks like it's taken into account in the loss, but possibly not in predict. We could use the new heteroscedastic likelihood of gpflow maybe? (I might also have this wrong)

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+                gpf_residual_model = self.fidelity_residual_models[fidelity].model
+
+                fidelity_observations = dataset_for_fidelity.observations
+                fidelity_query_points = dataset_for_fidelity.query_points
+                prev_fidelity_query_points = convert_query_points_for_fidelity(
+                    dataset_for_fidelity.query_points, fidelity - 1
+                )
+                predictions_from_lower_fidelity = self.predict(prev_fidelity_query_points)[0]
+
+                def loss() -> TensorType:  # hardcoded log liklihood calculation for residual model
```
vpicheny, line 1608: I haven't checked but I'm assuming this is mimicking the `gpflow` version as much as possible? there might be subtleties we could miss otherwise that could affect performance

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,259 @@ def covariance_between_points(
...
+                )
+                self.fidelity_residual_models[fidelity].optimizer.optimizer.minimize(
+                    loss, trainable_variables
+                )
+                residuals = self._calculate_residual(dataset_for_fidelity, fidelity)
+                self.fidelity_residual_models[fidelity].update(
+                    Dataset(fidelity_query_points, residuals)
+                )
+
+    def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
```
vpicheny, line 1630: I don't understand what this is computing. It doesn't look like a covariance, just a variance?

trieste/models/gpflow/models.py
AlexGoodallAI: No, this fetches the fidelity==0 level

trieste/models/gpflow/models.py
AlexGoodallAI: Because we need to continually feed predictions from lower fidelities through higher fidelities to get high fidelity predictions

trieste/models/gpflow/models.py
AlexGoodallAI: It's used in the multifidelity acquisition function that I'll be adding after this. Maybe I should leave this out until then? 

It's computing the covariance between a point at a lower fidelity and the highest fidelity. Because of independence between residual and signal models it does end up looking like a variance.

trieste/models/gpflow/models.py
AlexGoodallAI: I've changed it to a method using `tf.tensor_scatter_nd_update`, which avoids calling `predict` and `sample` on uneccesary points, but is not a particularly fast operation itself, so I'm not actually sure which is better. It may depend on `num_fidelities` and the distribution of queries across fidelities (i.e how much wasted computation there would have been with the old method)

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,305 @@ def covariance_between_points(
             query_points_2=query_points_2,
             whiten=whiten,
         )
+
+
+class AR1(TrainableProbabilisticModel):
```
hstojic: since we haven't been using abbreviations in naming our models/wrappers, lets be consistent and use a full name here - this stands for autoregressive, right? so perhaps `AutoRegressive1`, or as a compromise, since it s a multi-fidelity model, perhaps we could have `MultiFidelityAR1`

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,305 @@ def covariance_between_points(
...
+    from :cite:`Kennedy2000`. This is a multi-fidelity model that works with an
+    arbitrary number of fidelities. It relies on there being a linear relationship
+    between fidelities, and may not perform well for more complex relationships.
+    Precisely, it models the relationship between sequential fidelities as
+
+    .. math:: f_{i}(x) = \rho f_{i-1}(x) + \delta(x)
+
+    where :math:`\rho` is a scalar and :math:`\delta` models the residual between the fidelities.
+    All the base models used in this implementation are :class:`~gpflow.models.GPR` models.
+    Note: Currently only supports single output problems.
```
hstojic, line 1385: does it support anything else than GPR?

trieste/models/gpflow/models.py
```diff
@@ -1362,3 +1368,305 @@ def covariance_between_points(
...
+    """
+
+    def __init__(
+        self,
+        fidelity_models: Sequence[GaussianProcessRegression],
+    ):
+        """
+        :param fidelity_models: List of
+            :class:`~trieste.models.gpflow.models.GaussianProcessRegression`
+            models, one for each fidelity.
```
hstojic: the order matters so perhaps write explicitly how it matters?

trieste/data.py
hstojic: Can you clarify the choice here?

It seems to me as well that something like a `FidelityDataset` would facilitate things rather than having all these functions to manipulate data, why not maintaining `Mapping[int, Dataset]` for the fidelity levels? would it make all that compute in `predict` and `sample` more expensive? 
or perhaps an object with additional tensor that records fidelity, adding it to the query points does not seem to me it has any benefits?

hstojic (commented):
I took a (very) quick look and left some comments
do you plan to add a notebook?

trieste/data.py
AlexGoodallAI: Yeah, I think this is the better way to do it, but I was told not to do this when I was initially implementing it. It was mentioned that this made the typing quite tricky, and might require changes throughout the repo, for relatively little actual gain.

 I'm happy to change it, but it's a fairly large change I think.

trieste/models/gpflow/models.py
AlexGoodallAI: I believe it is, yes

trieste/models/gpflow/models.py
AlexGoodallAI: I'm not sure. I've only tried it with GPRs so far, and I think that's what the paper uses. I think because we have to do a reimplementation of the loss function we may be limited to GPRs, but @alansaul probably has more idea.

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,102 @@
...
+    test_xs = tf.linspace(0, 1, 11)[:, None]
+    test_xs_w_fid = add_fidelity_column(test_xs, fidelity=3)
+
+    predictions = model.predict(test_xs_w_fid)[0]
+    gt_obs = observer(test_xs_w_fid).observations
+
+    npt.assert_allclose(predictions, gt_obs, rtol=0.20)
+
+
+def test_multifidelity_autoregressive_gets_expected_rhos() -> None:
```
vpicheny, line 77: We might need a few more tests for the AR model predict: 1) we should test that predicting at LF amounts to using LF predict_f. Then we could create dummy models and check that predict does what is expected, e.g. the HF predict reverts to the HF predict_f predict if rho is zero or if the LF model is flat, or it reverts to the LF prediction if the HF one is flat and zero, etc.

trieste/models/gpflow/models.py
vpicheny: possibly good enough for this PR but indeed it could be really nice to have an SGPR / SVGP implementation?

trieste/models/gpflow/models.py
vpicheny: makes sense - maybe just make the docstring a bit more explicit? e.g. 

Calculate the covariance of the output at `query_point` and a given fidelity with the highest fidelity output at the same `query_point`.

trieste/models/gpflow/models.py
hstojic: Its fine if its restricted to GPR but make it clear in the docs

trieste/data.py
hstojic: with "this" you mean current implementation that you have is better?

@uri-granta I guess you had reasons against the above route due to typing?

AlexGoodallAI: No sorry, I mean the way that you suggested feels much neater

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -100,3 +103,165 @@ def test_multifidelity_autoregressive_gets_expected_rhos() -> None:
...
+        )
+    )
+
+    model.update(dataset)
+    model.optimize(dataset)
+
+    test_locations = tf.Variable(np.linspace(0, 10, 32), dtype=tf.float64)[:, None]
+    lf_test_locations = add_fidelity_column(test_locations, 0)
+
+    lf_prediction = model.predict(lf_test_locations)
```
vpicheny: Here and for the two tests below: a possible source of bugs is when we predict at a batch that contains both low and high fidelities  at the same time. 
We should perhaps make test_locations the concatenation of two locations and two fidelities?

tests/unit/models/gpflow/test_models.py
```diff
@@ -1626,3 +1628,331 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
     actual_covariance = model.covariance_between_points(query_points_1, query_points_2)
 
     np.testing.assert_allclose(expected_covariance, actual_covariance, atol=1e-4)
+
+
+def multifidelity_autoregressive_nd_dataset(n_dims: int = 1) -> Dataset:
```
vpicheny, line 1633: does it make sense to have those tests here and not in `test_multifidelity_models`? (my guess is it does, but just double-checking)

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,574 @@
...
+    model.update(initial_data)
+    model.optimize(initial_data)
+
+    expected_rho = [1.0] + [(fidelity + 1) / fidelity for fidelity in range(1, n_fidelities)]
+    rhos = [float(rho.numpy()) for rho in model.rho]
+
+    npt.assert_allclose(np.array(expected_rho), np.array(rhos), rtol=0.30)
+
+
+def test_multifidelity_autoregressive_predict_low_fidelity_are_consistent() -> None:
```
vpicheny: do we still need this test if we have the next?

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,574 @@
...
+    model.update(dataset)
+    model.optimize(dataset)
+
+    # Add some high fidelity points to check that predict on different fids works
+    test_locations_30 = tf.Variable(np.linspace(0, 10, 60), dtype=tf.float64)[:, None]
+    lf_test_locations = add_fidelity_column(test_locations_30, 0)
+    test_locations_32 = tf.Variable(np.linspace(0, 10, 32), dtype=tf.float64)[:, None]
+    hf_test_locations = add_fidelity_column(test_locations_32, 1)
+
+    concat_test_locations = tf.concat([lf_test_locations, hf_test_locations], axis=0)
```
vpicheny, line 143: Could we add some leading dimensions to the locations? e.g. such that it's `[2, 92, 2]`

Easiest way would be 
`test_locations_30_bis = tf.Variable(np.linspace(0.5, 10.5, 60), dtype=tf.float64)[:, None]`
etc. to get `concat_test_locations_bis`
then you do
`tf.concat([concat_test_locations[None, ...], concat_test_locations_bis[None, ...]], axis=0)`

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,574 @@
...
+        build_multifidelity_autoregressive_models(
+            dataset, num_fidelities=2, input_search_space=search_space
+        )
+    )
+
+    model.update(dataset)
+    model.optimize(dataset)
+
+    kernel = gpflow.kernels.Matern52(variance=1e-18, lengthscales=[0.2, 0.2])
+    gpr = gpflow.models.GPR(dataset.astuple(), kernel, noise_variance=1e-5)
```
vpicheny: This is weird, it should crash, not return a flat model. 

Besides your `dataset` contains a fidelity column?

Maybe instead keep a kernel variance of one and make all your observations in `lf_dataset` equal to zero? Prediction variance will not be zero but it doesn't matter for this test unless I'm mistaken.

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,574 @@
...
+        build_multifidelity_autoregressive_models(initial_data, n_fidelities, input_search_space)
+    )
+
+    model.update(initial_data)
+    model.optimize(initial_data)
+
+    expected_rho = [1.0] + [(fidelity + 1) / fidelity for fidelity in range(1, n_fidelities)]
+    rhos = [float(rho.numpy()) for rho in model.rho]
+
+    npt.assert_allclose(np.array(expected_rho), np.array(rhos), rtol=0.30)
```
vpicheny, line 105: I don't understand this test...

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,574 @@
...
+    search_space = Box([0.0], [10.0])
+
+    model = MultifidelityAutoregressive(
+        build_multifidelity_autoregressive_models(
+            dataset, num_fidelities=2, input_search_space=search_space
+        )
+    )
+
+    model.update(dataset)
+    model.optimize(dataset)
```
vpicheny: for this test and all the others: there is a small risk that when calling `optimize` the rho value becomes (or stay) 0 or 1, which weakens the tests. Could we set it to any random value after optimize instead? 

actually we could even get rid of `optimize`?

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,574 @@
...
+
+    kernel = gpflow.kernels.Matern52(variance=1e-18, lengthscales=[0.2, 0.2])
```
vpicheny: see above comment

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,574 @@
...
+
+    hf_prediction, _ = model.predict(hf_test_locations)
+    hf_prediction_direct = (
+        model.rho[1] * model.lowest_fidelity_signal_model.predict(test_locations)[0]
+    )
+
+    npt.assert_allclose(hf_prediction, hf_prediction_direct)
+
+
+def test_multifidelity_autoregressive_sample_low_fidelity_are_consistent() -> None:
```
vpicheny: do we need this test if we have test_multifidelity_autoregressive_sample_lf_are_consistent_with_multiple_fidelities?

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,574 @@
...
+    concat_samples = model.sample(concatenated_test_locations, 1_000_000)
+    hf_samples = concat_samples[:, 30:]
+    hf_samples_direct = model.fidelity_residual_models[1].sample(test_locations_32, 1_000_000)
+
+    hf_samples_mean = tf.reduce_mean(hf_samples, axis=0)
+    hf_samples_var = tf.math.reduce_variance(hf_samples, axis=0)
+    hf_samples_direct_mean = tf.reduce_mean(hf_samples_direct, axis=0)
+    hf_samples_direct_var = tf.math.reduce_variance(hf_samples_direct, axis=0)
+
+    npt.assert_allclose(hf_samples_mean, hf_samples_direct_mean, atol=1e-3)
```
vpicheny: maybe the way the LF model is "flattened" creates a little bit of noise, if we do it in a different way we might be able to tighten this tolerance? at least for the mean with 1e6 samples it seems a bit high...

tests/unit/models/gpflow/test_models.py
```diff
@@ -1626,3 +1628,331 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
...
+
+    test_locations = tf.Variable(np.linspace(0, 10, 32), dtype=tf.float64)[:, None]
+    lf_test_locations = add_fidelity_column(test_locations, 0)
+    hf_test_locations = add_fidelity_column(test_locations, 1)
+
+    lf_true_means, lf_true_vars = model.predict(lf_test_locations)
+
+    hf_true_means, hf_true_vars = model.predict(hf_test_locations)
+
+    lf_samples = model.sample(lf_test_locations, 100000)
```
vpicheny: maybe best to merge the two `sample` calls into one? to also capture a possible bug if we get it wrong with multiple fidelity calls at the same time

tests/integration/models/multifidelity/test_multifidelity_models.py
```diff
@@ -0,0 +1,504 @@
...
+    npt.assert_allclose(hf_samples_var, hf_samples_direct_var, atol=1e-3)
```
vpicheny: have you tried decreasing the tolerance?

vpicheny (approved):

PR #622, opened 2022-09-07
Expected TF function signature.

avullo:
The assert function requires to have a non-empty list of tensors to print when the condition is not met.

### DIFF
--- a/trieste/models/gpflow/sampler.py
+++ b/trieste/models/gpflow/sampler.py
@@ -315,8 +315,8 @@ class FeatureDecompositionTrajectorySampler(
         :param trajectory: The trajectory function to be resampled.
         :return: The new resampled trajectory function.
         """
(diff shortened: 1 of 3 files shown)


hstojic:
@uri-granta could you check this one? we might need to propagate this to gpflow/gpflux trajectory samplers...

uri-granta (approved):
Looks good, though do you know why this isn't causing test failures? Are we only ever executing this code eagerly?

Also, here is a PR that also fixes the other instances in trieste: https://github.com/secondmind-labs/trieste/pull/623. Maybe check in that instead?

avullo:
> Looks good, though do you know why this isn't causing test failures? Are we only ever executing this code eagerly?

I believe so. I've spotted the problem while trying to export the graph of some of the models' methods.

PR #624, opened 2022-09-08
Fix failing alltest_latest tests

uri-granta:


hstojic (approved):
looks good, worthwhile keeping it

PR #626, opened 2022-09-15
Remove model config support

uri-granta:


hstojic (approved):
it feels great to burn this code :)

PR #627, opened 2022-09-20
Fix FastUpdateModel docstrings

st--:
The docstrings of the FastUpdateModel abstract methods seemed to be in error; I believe this is what is meant, but please double-check.


trieste/models/interfaces.py
```diff
@@ -224,7 +224,7 @@ def conditional_predict_joint(
         :param additional_data: Dataset with query_points with shape [..., N, D] and observations
                  with shape [..., N, L]
         :return: The mean and covariance of the joint marginal distribution at each batch of points
```
st--, line 285: Also, what is the "joint marginal distribution"? is this the joint distribution over all *f*s? So should this be
```suggestion
        :return: The mean and covariance of the joint distribution at each batch of points
```
?

hstojic:
good to hear from you @st-- !
@vpicheny is the main author of this code (though @uri-granta chipped in as well I think), Victor care to check this out?

PR #628, opened 2022-09-22
Wrap single objective test functions in a container class

### LINKED ISSUE #596: inconsistent definitions of single and multi objective problems
multi-objectives are defined as classes while all the relevant components of single objectives are individual functions and constants, we should define an unifying interface for both types and use class type definition for single objectives as well since typically we need to use all components (function, minima, search space...) whenever using a problem, so best to load a single object that has all required components

uri-granta:


trieste/objectives/single_objectives.py
```diff
@@ -93,6 +114,24 @@ def scaled_branin(x: TensorType) -> TensorType:
 """ The search space for the :func:`branin` function. """
 
 
+Branin = SingleObjectiveTestProblem(
+    objective=branin,
+    search_space=BRANIN_SEARCH_SPACE,
+    minimizers=BRANIN_MINIMIZERS,
+    minimum=BRANIN_MINIMUM,
```
hstojic: not sure it makes sense to keep those SEARCH_SPACE, MINIMIZERS, MINIMUM objects around, why not defining them here when creating the class?

trieste/objectives/multi_objectives.py
```diff
@@ -79,6 +80,7 @@ class VLMOP2(MultiObjectiveTestProblem):
     bounds = [[-2.0] * 2, [2.0] * 2]
     dim = 2
```
hstojic: I don't see a good reason why we have bounds here and search space in single objective test problems, lets define `search_space` here as well, instead of this `bounds` property

although it can be derived from search space, `dim` property is probably a useful shortcut, but then lets add it to the single objectives ones as well

`name` property or some nice string representation would be very useful as well - when we run experiments with lots of problems, they would come handy for saving the results

hstojic (commented):
thanks @uri-granta! 
I provided few comments, how we can improve it further

with the suggested changes it would also be good to have an interface for both single and multi problems I think

trieste/objectives/single_objectives.py
uri-granta: Prefer to keep them both to avoid breaking other people's code and to avoid wasting time updating our unit tests. The cost of having them is minimal.

trieste/objectives/multi_objectives.py
uri-granta: Good suggestions. I've:

1. updated SingleObjectiveTestProblem and MultiObjectiveTestProblem so they're both types of ObjectiveTestProblem
2. added a name member variable
3. added bounds and dim properties which are calculated from the search space

trieste/objectives/single_objectives.py
hstojic: huh, we are keeping a lot of objects around and its not a nice self contained code for each test problem, there is value in having a clean code and its time to clean up before 1.0.0

I see that containers are saving some effort now, but to me its a bit unnatural way of defining these problems

re breaking other ppl code, its mostly us that are using it, personally its not a big cost to change a few lines

trieste/objectives/__init__.py
```diff
@@ -58,6 +58,20 @@
     TRID_10_MINIMIZER,
     TRID_10_MINIMUM,
     TRID_10_SEARCH_SPACE,
+    Ackley5,
```
hstojic, line 22: can we please also import here `multi_objective` ones so they are available from `objectives` package?
also interfaces?

hstojic (commented):
hello Uri, changes in the right direction, but I don't like this container type of solution where all relevant objects are spread around independent - in some cases, like dltz, it also resulted in unnecessary complications

trieste/objectives/single_objectives.py
uri-granta: Done (since you clearly care about this way more than I do). Please review all the files very carefully, as with so much copy and pasting there is a good chance of typos.

hstojic: Thank you @uri-granta :)

hstojic (approved):
looks much better now :)

PR #629, opened 2022-09-30
Faster pareto front calculation

uri-granta:


trieste/acquisition/multi_objective/dominance.py
```diff
@@ -15,30 +15,46 @@
...
-    )
-
-    return tf.boolean_mask(observations, dominance == 0), dominance
+    orig_points = observations.shape[0]
+    orig_indices = np.arange(orig_points)
+
+    if orig_points > 0:
+        # Reordering the observations beforehand speeds up the search
+        obs_np = to_numpy(observations)
+        idx = np.argsort(((obs_np - obs_np.mean(axis=0)) / (obs_np.std(axis=0) + 1e-7)).sum(axis=1))
```
hstojic, line 44: will this trigger retracing if this function is used anywhere in tf.function?

hstojic (approved):
looks good, the only concern is that perhaps numpy might trigger retracing?
but perhaps non_dominated is not used anywhere inside tf.function?

trieste/acquisition/multi_objective/dominance.py
uri-granta: Very good spot! In fact the use of not just `to_numpy` but also `np.any` and `np.sum` in the main loop all fail inside a tf.function! I'll have a go at rewriting this in pure tensorflow.

uri-granta: I've added a test, but fixing this will probably be a pain :(

PR #630, opened 2022-10-05
Merge TB logging between BO and AskTell

uri-granta:


trieste/bayesian_optimizer.py
```diff
@@ -788,152 +822,152 @@ def _write_summary_init(
...
+
+    if pd and sns and logging.include_summary("query_points/_pairplot"):
+        columns = [f"x{i}" for i in range(tf.shape(query_points)[1])]
+        qp_preds = query_points
+        for tag in datasets:
+            pred = models[tag].predict(query_points)[0]
+            qp_preds = tf.concat(
+                [qp_preds, tf.cast(pred, query_points.dtype)], 1
+            )
+            if tf.shape(pred)[-1] == 1:
```
vpicheny: here you have two cases: either "pred" is one-dimensional, then you use the tag directly, or it is multidimensional (typically, a ModelStack with tag OBJECTIVE), then you need more column names, e.g. OBJECTIVE1, OBJECTIVE2

vpicheny: might be tidier using `output_dim = tf.shape(tagged_output[tag].observations)[-1]`

trieste/bayesian_optimizer.py
```diff
@@ -788,152 +822,152 @@ def _write_summary_init(
...
-
-        logging.scalar("wallclock/step", total_step_wallclock_timer.time)
-        logging.scalar(
-            "wallclock/query_point_generation",
-            query_point_generation_timer.time,
-        )
-        logging.scalar(
-            "wallclock/model_fitting",
-            model_fitting_timer.time,
+            observation_plot_df["pareto"] = non_dominated(datasets[tag].observations)[1] == 0
```
vpicheny: actually, it only makes sense to look at the Pareto if the tag is OBJECTIVE?

trieste/bayesian_optimizer.py
```diff
@@ -788,152 +822,152 @@ def _write_summary_init(
...
+            observation_plot_df["observation type"] = observation_plot_df.apply(
+                lambda x: x["observations"] + x["pareto"] * " (non-dominated)",
+                axis=1,
+            )
+            pairplot = sns.pairplot(
```
vpicheny, line 907: a bit trickier is that if you have constraints, then your Pareto front is different, see 

https://secondmind-labs.github.io/trieste/0.12.0/notebooks/multi_objective_ehvi.html#Acquisition-function-for-multiple-objectives-and-constraints

uri-granta: Not immediately obvious how we can do this out of the box: all the default logger knows is that there are multiple tagged datasets (one of which is probably, but not necessarily, called OBJECTIVE), not the relation between them. Any ideas?

Suggest creating an issue for this and fixing it in a separate PR.

uri-granta: (We could delegate the plotting to the acquisition function builder, but that will require some more infrastructure, so definitely makes sense to do in a separate PR.)

vpicheny: that's fair - I thought the EHVI etc. were expected specifically an OBJECTIVE tag but you can specify whatever tag you want as `objective_tag`.

trieste/bayesian_optimizer.py
vpicheny: actually I thought the EHVI etc. were expected specifically an OBJECTIVE tag but you can specify whatever tag you want as `objective_tag`, so this may be very tricky.

trieste/bayesian_optimizer.py
vpicheny: I think this makes sense. Showing pairs plot of predictions with markers based on multi-objective considerations only makes sense if it matches your acquisition function builder.

trieste/bayesian_optimizer.py
uri-granta: (I've changed it so we only highlight pareto fronts for OBJECTIVE tag or for any tag in a single-tag dataset — at least for now)

vpicheny (approved):
works well on the reaction engine (although the Pareto front visualisation is not very useful since it ignores constraints but this goes in a separate PR)

PR #638, opened 2022-10-21
Add ProbabilityOfImprovement acqusition function

AlexGoodallAI:


trieste/acquisition/function/function.py
```diff
@@ -42,6 +42,54 @@
 )
 
 
+class ProbabilityOfImprovement(SingleModelAcquisitionBuilder[ProbabilisticModel]):
```
vpicheny, line 45: Note: PoI is generally not a usable acquisition because it massively overexploits. It is only useful to have it because of the portfolio approach. 
There is one way to use make it useful, which is to allow for a parameter epsilon to shift the threshold eta. See for instance:
https://distill.pub/2020/bayesian-optimization/

Not sure if only worth keeping in mind for now, or add epsilon as an extra parameter.

vpicheny (approved):

PR #643, opened 2022-10-21
Add Pareto diverse sample method

AlexGoodallAI:


trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,194 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+        sample_size: int,
+        allow_repeats: bool = True,
+        bounds_delta_scale_factor: float = 0.2,
+        bounds_min_delta: float = 1e-9,
+    ) -> tuple[TensorType, TensorType]:
+        """
+        Sample a set of diverse points from the Pareto set using
+        Hypervolume Sharpe-Ratio Indicator
+
+        :param sample_size: The number of point to sample from the Pareto front
```
vpicheny: pointS

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,194 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+        :param allow_repeats: Whether the sample may contain repeats
+        :param bounds_delta_scale_factor: The factor by which to grow the distance
+            between extrema when calculating lower and upper bounds
+        :param bounds_min_delta: The minimum value of the distance between extrema
+        """
```
vpicheny, line 100: add return to docstring?

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,194 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+            samples, sample_ids = self._choose_batch_no_repeats(x_star, sample_size)
+
+        return samples, sample_ids
+
+    def _choose_batch_with_repeats(
+        self, x_star: TensorType, sample_size: int
+    ) -> tuple[TensorType, TensorType]:
+
+        # Calculate number of times each point is sampled
+        n_times_sampled = x_star * float(sample_size)
```
vpicheny: is it OK to convert with `float()`? naive question (maybe the answer is yes)

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,194 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
...
+
+        # Calculate the deltas to add to the bounds to get the reference point and lower bound
+        deltas = [
+            ((float(max(self.front[:, i])) - float(min(self.front[:, i]))) * delta_scaling_factor)
+            + min_delta
+            for i in range(front_dims)
+        ]
+
+        # Define lower bound and reference point
+        lower_bound = [float(min(self.front[:, i])) - deltas[i] for i in range(front_dims)]
```
vpicheny: same question as above about the usage of `float()`

trieste/acquisition/multi_objective/pareto.py
```diff
@@ -73,6 +79,194 @@ def hypervolume_indicator(self, reference: TensorType) -> TensorType:
         )
         return hypervolume_indicator
 
+    def sample_diverse_subset(
+        self,
+        sample_size: int,
+        allow_repeats: bool = True,
+        bounds_delta_scale_factor: float = 0.2,
```
vpicheny, line 86: check those values are strictly positive?

tests/unit/acquisition/multi_objective/test_pareto.py
```diff
@@ -104,3 +105,74 @@ def test_get_reference_point_extract_based_on_pareto_front(
...
+    pareto_set = Pareto(observations)
+    sample, sample_ids = pareto_set._choose_batch_no_repeats(x_star, sample_size=2)
+    expected_sample = tf.constant([[-1.0, 1.0], [1.0, -1.0]])
+    expected_sample_ids = tf.constant([3, 1])
+    npt.assert_array_equal(expected_sample, sample)
+    npt.assert_array_equal(expected_sample_ids, sample_ids)
+
+
+@pytest.mark.parametrize(
+    "x_star,expected_ids",
```
vpicheny, line 173: do you want to add a test where choosing 3 points from a batch of 3 returns the same front?

vpicheny (approved):
LGTM! left a few minor comments

trieste/acquisition/multi_objective/pareto.py
AlexGoodallAI: Yes, I'm pretty sure that's what python does under the hood if you do `float * int`. It's unnecessary here, not sure why I've got it in, so I'll remove it

trieste/acquisition/multi_objective/pareto.py
AlexGoodallAI: Also probably unnecessary, but it means we're storing a list of floats rather than a list of numpy arrays, which is probably more natural

vpicheny: maybe if this doesn't impact performance we should try to be consistent w.r.t the rest of the multiobj code in trieste?

AlexGoodallAI: Do you mean convert it to tf?

vpicheny: sorry that wasn't clear, maybe it makes sense to do as in `get_reference_point` for instance? (but ignore if that doesn't make sense)

AlexGoodallAI: Ah yeah, it could (and should) be vectorised. I'll do that

PR #646, opened 2022-11-03
Handle isotropic priors for non-isotropic parameters

uri-granta:


tests/unit/models/gpflow/test_utils.py
```diff
@@ -75,6 +75,21 @@ def test_randomize_hyperparameters_randomizes_kernel_parameters_with_priors(dim:
...
+@random_seed
+def test_randomize_hyperparameters_randomizes_kernel_parameters_with_const_priors(dim: int) -> None:
+    kernel = gpflow.kernels.RBF(variance=1.0, lengthscales=[0.2] * dim)
+    kernel.lengthscales.prior = tfp.distributions.LogNormal(
+        loc=tf.math.log(0.2), scale=1.0  # constant loc should be applied to every dimension
+    )
+    randomize_hyperparameters(kernel)
+
+    npt.assert_allclose(1.0, kernel.variance)
+    npt.assert_array_equal(dim, kernel.lengthscales.shape)
```
Andrew878, line 98: Optional: Maybe worth adding another assert to check if the lengthscale values are different?

Andrew878 (approved):
Looks great

tests/unit/models/gpflow/test_utils.py
uri-granta: I did! (see lines 92 and 78)

Andrew878: Sorry, I meant to check that after sampling, the lengthscale for dimension 1 is different to dimension 2

Andrew878: if this 
`np.count_nonzero(kernel.lengthscales == kernel.lengthscales[0]) == 1 `
achieves it though then no problem at all 

I would have thought it would be something like `assert np.unique(kernel.lengthscales) > 1`

uri-granta: You're right. I was just checking that the first lengthscale was unique, not all of them (for no obvious reason). Fixed.

Andrew878 (approved):

PR #649, opened 2022-11-04
Unpin gpflow and dill

uri-granta:
GPflow 2.6 introduced a breaking change. Rather than support both versions, I've increased the minimum supported version. Before this can be merged though, we also need to upgrade to a newer gpflux.

Also using this opportunity to up the minimum supported version of TF from 2.4 to 2.5 (which is the minimum supported by latest gpflux) and moving our tests to test against 2.8 and 2.10 (rather than 2.4 and 2.8).


hstojic (commented):
I didn't realize we were pinning to 2.5.2...

tests/requirements.txt
```diff
@@ -15,9 +15,7 @@
 packaging
 pytest
 PyYAML
-importlib-metadata<5.0
 
 # pin to version of TF used in prod
-gpflux == 0.2.3
-tensorflow ~= 2.4.1
-tensorflow-probability ~= 0.12.0
+tensorflow ~= 2.8.0
```
AlexGoodallAI, line 20: Why is this `~=2.8.0` and not `~=2.5.0`?

uri-granta: Because "product" is now using 2.8, so we want that to be out main test target. I think there's already an issue to increase the number of versions we test against (we could at least add a "minimal" target), but I'd like to postpone doing that until after I've fixed the memory issues we're seeing.

AlexGoodallAI (approved):
LGTM

hstojic (approved):
looks good

PR #651, opened 2022-11-15
Add support for portfolio method for batch BO

AlexGoodallAI:


tox.ini
```diff
@@ -55,8 +55,8 @@ commands =
 basepython = python3.9
 commands =
     # test but with python 3.9 and latest supported library versions
```
uri-granta, line 57:
```suggestion
    # test but with python 3.9 and latest supported library versions
    # unlike tests this includes the optional qhsri support
```

uri-granta (approved):
LGTM! Make sure you've manually run the slow tests too before you merge (either on GitHub or locally) and ask me if you have any problems with fixing the docs.

trieste/acquisition/rule.py
```diff
@@ -1080,3 +1096,187 @@ def state_func(
...
+    AcquisitionRule[TensorType, SearchSpace, ProbabilisticModel]
+):
+    """Implements the Batch Hypervolume Sharpe-ratio indicator acquisition
+    rule, designed for large batches, introduced by Binois et al, 2021.
+    See :cite:`binois2021portfolio` for details.
+    """
+
+    def __init__(
+        self,
+        batch_size: int = 5,
```
vpicheny: for consistency with other rules, batch_size => num_query_points: int = 1

trieste/acquisition/rule.py
```diff
@@ -1080,3 +1096,187 @@ def state_func(
...
+        self._n_generations: int = ga_n_generations
+        self._filter_threshold: float = filter_threshold
+        self._noisy_observations: bool = noisy_observations
+        self._acquisition_function: Optional[AcquisitionFunction] = None
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""BatchHypervolumeSharpeRatioIndicator(
+        batch_size={self._batch_size}, ga_population_size={self._population_size},
+        ga_n_generations={self._n_generations}, filter_threshold={self._filter_threshold}
```
vpicheny: add noisy observations

vpicheny (approved):
LGTM! would have been ideal to have a unit test that shows that the rule returns what we expect when the problem is trivial (say using a dummy model with closed-form mean and variance such that we know the Pareto) but it's probably overkill.

PR #655, opened 2022-11-22
Allow scipy optimizer to be changed from args

khurram-ghani:


trieste/acquisition/optimizer.py
```diff
@@ -523,13 +523,13 @@ def value_and_gradient(
 
             return cast("np.ndarray[Any, Any]", cache_y), cast("np.ndarray[Any, Any]", cache_dy_dx)
 
+        default_optimizer_args = dict(method="l-bfgs-b")
```
uri-granta: Perhaps simpler like this

```suggestion
        optimizer_args = dict(method="l-bfgs-b", **(optimizer_args or {}))
```

and then just use `**optimizer_args` below?

trieste/acquisition/optimizer.py
```diff
@@ -490,10 +490,10 @@ def _objective_value_and_gradient(x: TensorType) -> Tuple[TensorType, TensorType
     return (successes, fun_values, chosen_x, nfev)
 
 
-class ScipyLbfgsBGreenlet(gr.greenlet):  # type: ignore[misc]
+class ScipyOptimizerGreenlet(gr.greenlet):  # type: ignore[misc]
     """
-    Worker greenlet that runs a single Scipy L-BFGS-B. Each greenlet performs all the L-BFGS-B
-    update steps required for an individual optimization. However, the evaluation
+    Worker greenlet that runs a single Scipy L-BFGS-B (by default). Each greenlet performs all the
```
uri-granta, line 495: There should really be some sort of unit test for `optimizer_args`, though it looks like currently there isn't. Do you think you could add something simple to test_optimizer (maybe passing through the args to `generate_continuous_optimizer` and checking that they get used or defaulted as appropriate)? Alternatively if you think this might be tested by a future PR then it might be ok to skip for now.

uri-granta (approved):
LGTM. Just one suggestion.

trieste/acquisition/optimizer.py
khurram-ghani: This was going to be tested in a future PR. However, I'll add something now. It makes sense.

trieste/acquisition/optimizer.py
khurram-ghani: Actually that doesn't work as is. You have to wrap the defaults in a dict, otherwise you get multiple keyword args error when you try to override. So this would work instead:
`        optimizer_args = dict(dict(method="l-bfgs-b"), **(optimizer_args or {}))
`

trieste/acquisition/optimizer.py
hstojic: I agree with Uri, a test for defaults (and successfully changing them) would be important

hstojic (commented):
looks ok so far, but I agree with Uri, tests are needed

trieste/acquisition/optimizer.py
khurram-ghani: Updated that in the latest commits.

trieste/acquisition/optimizer.py
khurram-ghani: I have added some tests.

PR #656, opened 2022-11-22
Add support for explicit input constraints to Box search-space

### LINKED ISSUE #537: cheap-to-evaluate constraint

**Additional context**

khurram-ghani:
The first commit adds support and tests for linear and non-linear constraints to the Box search-space.
The second commit adds 3 new methods and tests for sampling of feasible points.


trieste/space.py
```diff
@@ -127,6 +309,24 @@ def __eq__(self, other: object) -> bool:
...
+        """
+        Checks if points satisfy the explicit constraints of this :class:`SearchSpace`.
+        Note membership of the search space is not checked.
+
+        :param points: The points to check constraints feasibility for, with shape [..., D].
+        :return: A tensor of booleans. Returns `True` for each point if it is feasible in this
+            search space, else `False`.
+        """
+        # By default assume everything is feasible; can be overridden by a subclass.
+        return True
```
uri-granta: Should we perhaps ensure that this gets overridden whenever there are any constraints by defaulting to raising a `NotImplementedError` exception if `self.has_constraints` is `True`?

trieste/space.py
```diff
@@ -246,17 +446,31 @@ class Box(SearchSpace):
     """
 
     @overload
-    def __init__(self, lower: Sequence[float], upper: Sequence[float]):
+    def __init__(
+        self,
+        lower: Sequence[float],
+        upper: Sequence[float],
+        constraints: Optional[Sequence[LinearConstraint | NonlinearConstraint]] = None,
```
uri-granta: Maybe define a `Constraint = LinearConstraint | NonLinearConstraint` type alias?

Alternatively you could define an ABC with the relevant methods (if they're obvious) and have both LinearConstraint and NonLinearConstraint inherit from it.

trieste/space.py
```diff
@@ -429,12 +796,42 @@ def __eq__(self, other: object) -> bool:
         if not isinstance(other, Box):
             return NotImplemented
         return bool(
-            tf.reduce_all(self.lower == other.lower) and tf.reduce_all(self.upper == other.upper)
+            tf.reduce_all(self.lower == other.lower)
+            and tf.reduce_all(self.upper == other.upper)
+            and self._constraints == other._constraints
```
uri-granta, line 809: (maybe add a comment about how we're insisting on the exact same list of constraints?)

trieste/space.py
```diff
@@ -32,6 +34,180 @@
...
+        <tf.Tensor: shape=(3, 2), dtype=float64, numpy=
+        array([[ 0.01058163,  3.28941832],
+               [-0.1646425 ,  3.46464245],
+               [ 1.01058163,  2.28941832]])>
+
+
+        :param points: The points to calculate the residuals for, with shape [..., D].
+        :return: A tensor containing the lower and upper residual values with shape [..., M*2].
+        """
+        non_d_axes = tf.ones_like(tf.shape(points))[:-1]
```
uri-granta: Maybe add some shape/rank check on points?

trieste/space.py
```diff
@@ -32,6 +34,180 @@
...
+
+    def __eq__(self, other: object) -> bool:
+        """
+        :param other: A constraint.
+        :return: Whether the constraint is identical to this one.
+        """
+        if not isinstance(other, NonlinearConstraint):
+            return False
+        return bool(
+            tf.reduce_all(self._orig_fun == other._orig_fun)
```
uri-granta: Are the `tf.reduce_all`s for _orig_fun and keep_feasible necessary?

trieste/space.py
```diff
@@ -290,9 +506,15 @@ def __init__(
 
         self._dimension = tf.shape(self._upper)[-1]
 
+        if constraints is None:
+            self._constraints: Sequence[LinearConstraint | NonlinearConstraint] = []
+        else:
+            self._constraints = constraints
+        self._ctol = ctol
```
uri-granta, line 514: check that ctol > 0?

trieste/space.py
```diff
@@ -395,6 +622,146 @@ def sample_sobol(self, num_samples: int, skip: Optional[int] = None) -> TensorTy
             dim=dim, num_results=num_samples, dtype=self._lower.dtype, skip=skip
         ) + self._lower
 
+    def _sample_feasible_loop(
+        self,
+        num_samples: int,
+        sampler: Callable[[], TensorType],
+        max_tries: int = 100,
+    ) -> TensorType:
+        # Rejection sampling using provided callable. Try ``max_tries`` number of times to find
```
uri-granta: Still worth adding a params docstring?

trieste/space.py
```diff
@@ -395,6 +622,146 @@ def sample_sobol(self, num_samples: int, skip: Optional[int] = None) -> TensorTy
...
+        :param num_samples: The number of feasible points to sample from this search space.
+        :param seed: Random seed for reproducibility.
+        :param max_tries: Maximum attempts to sample the requested number of points.
+        :return: ``num_samples`` i.i.d. random points, sampled uniformly,
+            from this search space with shape '[num_samples, D]' , where D is the search space
+            dimension.
+        :raise SampleTimeoutError: If ``max_tries`` are exhausted before ``num_samples`` are
+            sampled.
+        """
+        # Without constraints use the normal sample method directly.
```
uri-granta: Is there any way to reduce the duplicated code in these methods? Maybe add a `_sample(num_samples, seed, feasible, max_tries)` internal method that gets called from both `sample` and `sample_feasible`?

uri-granta (commented):
Looks good! Some initial comments. Will finish looking at this tomorrow.

trieste/space.py
```diff
@@ -246,17 +446,31 @@ class Box(SearchSpace):
     """
 
     @overload
-    def __init__(self, lower: Sequence[float], upper: Sequence[float]):
+    def __init__(
+        self,
+        lower: Sequence[float],
+        upper: Sequence[float],
+        constraints: Optional[Sequence[LinearConstraint | NonlinearConstraint]] = None,
+        ctol: float = 1e-7,
```
hstojic: default tolerance should depend on dtype I think

hstojic (commented):
just a very quick pass for now, few comments

trieste/space.py
khurram-ghani: Updated in 30a17e7

trieste/space.py
khurram-ghani: Updated in 30a17e7

trieste/space.py
khurram-ghani: Added in 30a17e7

trieste/space.py
khurram-ghani: Added in 30a17e7

trieste/space.py
khurram-ghani: It isn't necessary for `NonlinearConstraint` (updated in 30a17e7). It is necessary for `keep_feasible` in `LinearConstraint` as scipy internally broadcasts it.

trieste/space.py
khurram-ghani: Added in 30a17e7

trieste/space.py
khurram-ghani: I did have it originally but removed it, thinking we don't want internal methods to show up in the user docs. I have added it back again in 30a17e7.

trieste/space.py
khurram-ghani: I have had a go in 30a17e7. Please have a look.

trieste/space.py
khurram-ghani: I have updated the type to optionally include `TensorType`:
https://github.com/secondmind-labs/trieste/blob/f5a33b42342e41e464960972c3deffcccbbbeeda/trieste/space.py#L464

However, I am not sure what I should do with the value `dtype`. I copied the style from `ExpectedConstrainedImprovement` for default `min_feasibility_probability`:
https://github.com/secondmind-labs/trieste/blob/30a17e7fe4099e8ecd4aead386d787d8136d11f2/trieste/acquisition/function/function.py#L483

trieste/space.py
```diff
@@ -32,6 +34,186 @@
...
+            return False
+        return bool(
+            tf.reduce_all(self.A == other.A)
+            and tf.reduce_all(self.lb == other.lb)
+            and tf.reduce_all(self.ub == other.ub)
+            and tf.reduce_all(self.keep_feasible == other.keep_feasible)
+        )
+
+
+# Constraint type alias.
```
uri-granta: If you turn this comment into a docstring it will show up in the API docs (e.g. see the docstring for AcquisitionFunction)

tests/unit/test_space.py
```diff
@@ -390,93 +398,178 @@ def _assert_correct_number_of_unique_constrained_samples(
...
+    box = Box(tf.zeros((3,)), tf.ones((3,)), constraints)
+    samples = box.sample_feasible(num_samples)
     _assert_correct_number_of_unique_constrained_samples(num_samples, box, samples)
 
 
-@pytest.mark.parametrize("num_samples", [0, 1, 10])
-def test_box_sobol_sampling_returns_correct_shape(num_samples: int) -> None:
-    box = Box(tf.zeros((3,)), tf.ones((3,)))
-    sobol_samples = box.sample_sobol(num_samples)
+@pytest.mark.parametrize(
```
uri-granta: Probably more natural to write this as two parametrizes?

```suggestion
@pytest.mark.parametrize("num_samples", [0, 1, 1])
@pytest.mark.parametrize("constraints", [None, _box_sampling_constraints()])
```

uri-granta (approved):
Tests look good! No explicit tests for the two constraint wrappers: though the residuals are tested indirectly in `test_box_constraints_residuals_and_feasibility` and presumably the inheritance from spo.Constraints will be tested when you update the optimizer code?

khurram-ghani:
> No explicit tests for the two constraint wrappers: though the residuals are tested indirectly in `test_box_constraints_residuals_and_feasibility` and presumably the inheritance from spo.Constraints will be tested when you update the optimizer code?

@uri-granta the tests for constraints wrappers are in the respective docstrings for `residual` methods. They show up in the list of pytests (in at least VSCode). Would it better if I move them to `test_space.py`?
I am not sure I understand your question about inheriting from `spo.Constraints`.

trieste/space.py
khurram-ghani: Done in b76379c.

tests/unit/test_space.py
khurram-ghani: Done in b76379c.

uri-granta:
> > No explicit tests for the two constraint wrappers: though the residuals are tested indirectly in `test_box_constraints_residuals_and_feasibility` and presumably the inheritance from spo.Constraints will be tested when you update the optimizer code?
> 
> @uri-granta the tests for constraints wrappers are in the respective docstrings for `residual` methods. They show up in the list of pytests (in at least VSCode). Would it better if I move them to `test_space.py`? I am not sure I understand your question about inheriting from `spo.Constraints`.

While the docstring examples do get run, I think it's probably clearer to have them as explicit tests in test_space. At least that's what we do elsewhere!

My point about inheriting from spo was that as far as I can see you don't currently test e.g. that NonlinearConstraint is inheriting from spo.NonlinearConstraint: you're only testing the residual function. But that this will presumably be tested when you test calling the optimizer with these constraints?

khurram-ghani:
> While the docstring examples do get run, I think it's probably clearer to have them as explicit tests in test_space. At least that's what we do elsewhere!

I have moved them to `test_space.py`.

> My point about inheriting from spo was that as far as I can see you don't currently test e.g. that NonlinearConstraint is inheriting from spo.NonlinearConstraint: you're only testing the residual function. But that this will presumably be tested when you test calling the optimizer with these constraints?

Yes, this will get tested when calling the optimizer with constraints. That is when the inheritance will be important.

PR #660, opened 2022-11-29
Add feasibility builder for explicit constraints

### LINKED ISSUE #537: cheap-to-evaluate constraint

khurram-ghani:
Also includes initial notebook showing penalty-EI method. This notebook will be expanded further in the PR for constrained optimisation.


trieste/acquisition/function/function.py
```diff
@@ -468,6 +468,84 @@ def acquisition(x: TensorType) -> TensorType:
     return acquisition
 
 
+class FastConstraintsFeasibility(SingleModelAcquisitionBuilder[ProbabilisticModel]):
+    """
+    Uses the :func:`fast_constraints_feasibility` function to build a
```
uri-granta: I don't think this needs to mention that it uses `fast_constraints_feasibility` (which after all is an implementation detail)?

```suggestion
    Builder for a probability of feasiblity acquisition function that uses the residuals of 
    direct constraints defined in the search space.
```

trieste/acquisition/function/function.py
```diff
@@ -468,6 +468,84 @@ def acquisition(x: TensorType) -> TensorType:
...
+        smoothing_function: Optional[Callable[[TensorType], TensorType]] = None,
+    ):
+        """
+        :param search_space: The global search space over which the feasibility of the constraints
+            is defined.
+        :param smoothing_function: The smoothing function used for constraints residuals. The
+            default is CDF of the Normal distribution with a small scale.
+        """
+        self._search_space = search_space
+        self._smoothing_function = smoothing_function
```
uri-granta, line 499: Should we check that `search_space.has_constraints` here and raise an exception if it doesn't?

trieste/acquisition/function/function.py
```diff
@@ -468,6 +468,84 @@ def acquisition(x: TensorType) -> TensorType:
...
+    def __init__(
+        self,
+        search_space: SearchSpace,
+        smoothing_function: Optional[Callable[[TensorType], TensorType]] = None,
+    ):
+        """
+        :param search_space: The global search space over which the feasibility of the constraints
+            is defined.
+        :param smoothing_function: The smoothing function used for constraints residuals. The
+            default is CDF of the Normal distribution with a small scale.
```
uri-granta: (explicitly state the default scale so people don't have to dig through the code?)

tests/unit/acquisition/function/test_function.py
```diff
@@ -15,7 +15,7 @@
 from __future__ import annotations
```
uri-granta, line 15: Do we also want to also add an integration test case to `test_bayesian_optimization.py`? (something similar to the notebook setup perhaps)

docs/notebooks/explicit_constraints.pct.py
```diff
@@ -0,0 +1,190 @@
...
+observations = dataset.observations.numpy()
+
+_, ax = plot_function_2d(
+    ScaledBranin.objective,
+    search_space.lower,
+    search_space.upper,
+    grid_density=30,
+    contour=True,
+)
+plot_bo_points(query_points, ax[0, 0], num_initial_points, arg_min_idx)
```
uri-granta: The plot is a bit unclear for me: maybe make the dots a bit bigger and the purple one brighter? (Alternatively, ask @vpicheny for a second opinion!)

trieste/acquisition/function/function.py
```diff
@@ -468,6 +468,84 @@ def acquisition(x: TensorType) -> TensorType:
...
+) -> AcquisitionFunction:
+    """
+    :param search_space: The global search space over which the feasibility of the constraints
+        is defined.
+    :param smoothing_function: The smoothing function used for constraints residuals. The
+        default is CDF of the Normal distribution with a small scale.
+    :return: The function for probability of feasibility of constraints.
+    """
+
+    @tf.function
```
uri-granta, line 553: Looking at the notebook, it looks like this is being retraced every loop, which is not ideal! I think the issue is because ExpectedConstrainedImprovement calls the constraint function with the query points, which change shape each time. We may be able to specify an input_signature here to handle this. If you can't quickly figure it out don't worry though: I'll raise an issue since I expect it's happening for the normal ProbabilityOfFeasibility too.

uri-granta (commented):
Looks good!

trieste/acquisition/function/function.py
khurram-ghani: I copied that style from ProbabilityOfFeasibility, but I can remove it.

trieste/acquisition/function/function.py
khurram-ghani: Yes I'll add that in.

trieste/acquisition/function/function.py
khurram-ghani: Yes, `ProbabilityOfFeasibility` has that problem too. I did spend a little bit of time looking at it, but didn't try changing the signature.

trieste/acquisition/function/function.py
khurram-ghani: Done in 1e41a57.

trieste/acquisition/function/function.py
khurram-ghani: Done in 1e41a57.

trieste/acquisition/function/function.py
khurram-ghani: Done in 1e41a57.

docs/notebooks/explicit_constraints.pct.py
khurram-ghani: That is an existing plotting function. I don't think the point size can be changed from its args but the colour can be, which I now have. Also, I have made the overall figure larger. Hopefully that is good enough.

tests/unit/acquisition/function/test_function.py
khurram-ghani: This will take a bit longer to implement. I'll look into it and maybe add it as part of the next PR where we will have more optimisation methods.

uri-granta (approved):
Happy with this

trieste/acquisition/function/function.py
uri-granta: I've raised an issue, thanks: https://github.com/secondmind-labs/trieste/issues/661

PR #663, opened 2022-12-01
Add qhsri to github workflows

AlexGoodallAI:


AlexGoodallAI:
This seems to have fixed the problem with develop-deploy: https://github.com/secondmind-labs/trieste/actions/runs/3594017393

docs/notebooks/quickrun/qhsri-tutorial.yaml
```diff
@@ -0,0 +1,2 @@
+replace:
+  - { from: "num_steps=8,", to: "num_steps=1," }
```
uri-granta, line 2: (The advantage of making this a regex (`num_steps=\d+`) is that it would still work if we change the number of steps in the normal run. But since we'll get an error anyway if it doesn't match anything, fine to leave as is.)

uri-granta (approved):
LGTM!

PR #664, opened 2022-12-02
Add constraints support to EI and ECI

### LINKED ISSUE #537: cheap-to-evaluate constraint

khurram-ghani:
Switch to trust-constr optimisation method when constraints are present.
Updated explicit_constraints notebook.
New bayesian_optimization integration tests.

`ExpectedImprovement` and `ExpectedConstrainedImprovement` now take an optional search_space argument. It is optional to maintain backwards compatibility.

I also had to change `NonlinearConstraint` class to only support constraint functions that return a single output - that means a single constraint per instance. The user can still provide a list of constraints. Even though the _trieste_ code itself supports multi-output constraints, the `tfp.math.value_and_gradient` call used does not work correctly. It returns the sum of gradients instead of the jacobian required. It will require a more extensive change to support this correctly. For now I have limited the functions to be single output only.

The use of `FastConstraintsFeasibility` with `ExpectedConstrainedImprovement` is slightly modified as shown in the updated `expected_improvement` notebook. The constraints (as part of the search_space) should only be passed to `FastConstraintsFeasibility`. This is so the new functionality of constraints is bypassed in _trieste_ for the penalty method (e.g. we don't want to switch to _trust-constr_ acquisition function optimizer).


trieste/acquisition/function/function.py
```diff
@@ -98,9 +98,17 @@ class ExpectedImprovement(SingleModelAcquisitionBuilder[ProbabilisticModel]):
     of the posterior mean at observed points.
```
vpicheny, line 98: need to update the docstring to explain the `predictive_mean_fn` trick?

khurram-ghani: I have added some text. Hopefully it is clear.

trieste/acquisition/function/function.py
```diff
@@ -96,11 +96,23 @@ class ExpectedImprovement(SingleModelAcquisitionBuilder[ProbabilisticModel]):
...
+
+    In the presence of constraints in the search_space the "best" value is computed only at the
+    feasible query points. If there are no feasible points, this builder builds the
+    `NegativePredictiveMean` instead to help the optimizer find a first feasible point.
     """
 
+    def __init__(self, search_space: Optional[SearchSpace] = None):
+        """
+        :param search_space: The global search space over which the optimisation is defined.
```
uri-granta: Perhaps mention that this is only used for constraints?

trieste/acquisition/function/function.py
```diff
@@ -118,9 +130,25 @@ def prepare_acquisition_function(
         tf.debugging.Assert(dataset is not None, [tf.constant([])])
         dataset = cast(Dataset, dataset)
         tf.debugging.assert_positive(len(dataset), message="Dataset must be populated.")
-        mean, _ = model.predict(dataset.query_points)
+
+        self._predictive_mean_fn = NegativePredictiveMean().prepare_acquisition_function(
```
uri-granta: Maybe only create this when there's a search space with constraints?

trieste/acquisition/function/function.py
```diff
@@ -577,12 +626,14 @@ def __init__(
...
         constraint_builder: AcquisitionFunctionBuilder[ProbabilisticModelType],
         min_feasibility_probability: float | TensorType = 0.5,
+        search_space: Optional[SearchSpace] = None,
     ):
         """
         :param objective_tag: The tag for the objective data and model.
         :param constraint_builder: The builder for the constraint function.
         :param min_feasibility_probability: The minimum probability of feasibility for a
             "best point" to be considered feasible.
+        :param search_space: The global search space over which the optimisation is defined.
```
uri-granta: (again, maybe mention that this is only used for explicit constraints?)

tests/integration/test_constrained_bayesian_optimization.py
```diff
@@ -99,3 +107,90 @@ def observer(query_points: TensorType) -> dict[str, Dataset]:
...
+    arg_min_idx = tf.squeeze(tf.argmin(dataset.observations, axis=0))
+
+    best_y = dataset.observations[arg_min_idx]
+    best_x = dataset.query_points[arg_min_idx]
+
+    relative_minimizer_err = tf.abs(best_x - MINIMIZER)
+    # these accuracies are the current best for the given number of optimization steps, which makes
+    # this is a regression test
+
+    print("best_y:", best_y)
```
uri-granta: (do you want to keep this in or were they just for debugging?)

tests/integration/test_constrained_bayesian_optimization.py
```diff
@@ -99,3 +107,90 @@ def observer(query_points: TensorType) -> dict[str, Dataset]:
...
+
+@random_seed
+@pytest.mark.parametrize(
+    "num_steps, acquisition_function_builder",
+    [
+        pytest.param(12, ExpectedImprovement, id="ExpectedImprovement"),
+        pytest.param(12, ExpectedConstrainedImprovement, id="ExpectedConstrainedImprovement"),
+    ],
+)
+def test_constrained_optimizer_finds_minima_of_custom_problem(
```
uri-granta, line 121: How fast are these tests? If they're very slow then consider marking this with a `@pytest.mark.slow`.

tests/integration/test_constrained_bayesian_optimization.py
```diff
@@ -99,3 +107,90 @@ def observer(query_points: TensorType) -> dict[str, Dataset]:
...
+        c0 = x[..., 0] - 0.2 - tf.sin(x[..., 1])
+        c0 = tf.expand_dims(c0, axis=-1)
+        return c0
+
+    def _nlc_func1(x: TensorType) -> TensorType:
+        c1 = x[..., 0] - tf.cos(x[..., 1])
+        c1 = tf.expand_dims(c1, axis=-1)
+        return c1
+
+    constraints = [
```
uri-granta: Maybe define this as a `SingleObjectiveTestProblem` (either here or even in `trieste.objectives.single_objectives`), calling it something like `ConstrainedScaledBranin`?

uri-granta (approved):
Looks great! Just a few small comments.

trieste/acquisition/function/function.py
khurram-ghani: Done in a874ae1

trieste/acquisition/function/function.py
khurram-ghani: Done in a874ae1

trieste/acquisition/function/function.py
khurram-ghani: Done in a874ae1

tests/integration/test_constrained_bayesian_optimization.py
khurram-ghani: Ooops.. removed that in a874ae1

tests/integration/test_constrained_bayesian_optimization.py
khurram-ghani: One of them takes around 2 mins. I have added the mark in a874ae1.

tests/integration/test_constrained_bayesian_optimization.py
khurram-ghani: Added it to `single_objectives` in a874ae1. I have updated the notebook to also use the new test objective.

khurram-ghani:
The fragile failing test has been skipped and I have raised an issue for it: https://github.com/secondmind-labs/trieste/issues/666

docs/notebooks/explicit_constraints.pct.py
```diff
@@ -168,27 +145,84 @@ def constraint(input_data):
...
+
+
+plot_bo_results()
+
+# %% [markdown]
+# ## Penalty method
+#
+# ### Acquisition function (penalty method)
+#
+# An alternative to using a constrained optimization method is to construct the _expected constrained improvement_ acquisition function similar to the [inequality-constraints notebook](inequality_constraints.ipynb). However, instead of using probability of feasibility with respect to the constraint model, we construct feasibility from the explicit input constraints. Feasibility is calculated by passing all the constraints residuals (to their respective limits) through a smoothing function and taking the product.
```
vpicheny: "through a smoothing function and taking the product."

through a smooth step function and taking the product.

docs/notebooks/explicit_constraints.pct.py
```diff
@@ -168,27 +145,84 @@ def constraint(input_data):
...
+#
+# For this method only the `FastConstraintsFeasibility` class should be passed constraints via the search space. `ExpectedConstrainedImprovement` and `BayesianOptimizer` should be set up in the normal way without constraints.
```
vpicheny: That's a bit misleading. How about:

For this method, the `FastConstraintsFeasibility` class should be passed constraints via the search space. `ExpectedConstrainedImprovement` and `BayesianOptimizer` should be set up as usual.

docs/notebooks/explicit_constraints.pct.py
```diff
@@ -125,26 +105,23 @@ def constraint(input_data):
...
 
 # %% [markdown]
-# ## Acquisition function
+# ## Constrained optimization method
 #
-# We can construct the _expected constrained improvement_ acquisition function similar to the [inequality-constraints notebook](inequality_constraints.ipynb). However, instead of using probability of feasibility with respect to the constraint model, we construct feasibility from the explicit input constraints. Feasibility is calculated by passing all the constraints residuals (to their respective limits) through a smoothing function and taking the product.
+# ### Acquisition function (constrained optimization)
 #
-# Note this method penalises the expected improvement acquisition outside the feasible region. The optimizer uses unconstrained L-BFGS method to find the max of the acquistion function.
+# We can construct the _expected improvement_ acquisition function as usual. In order to handle the constraints, the search space must be passed as a constructor argument.
```
vpicheny: I don't understand "the search space must be passed as a constructor argument."

Also, I think it would make sense to explicitly build the search space in the notebook. This is possibly the main thing a user would have to replicate on their own problem?

docs/notebooks/explicit_constraints.pct.py
```diff
@@ -168,27 +145,84 @@ def constraint(input_data):
...
+
+rule = EfficientGlobalOptimization(eci)
+
+# %% [markdown]
+# ### Run the optimization loop (penalty method)
+#
+# We can now run the optimization loop.
+
+# %%
+# Note: use the search space without constraints for the penalty method.
```
vpicheny, line 245: does this search space not have constraints then? it's a bit confusing since we don't see how it's built.

trieste/acquisition/function/function.py
vpicheny: all good

vpicheny (approved):
Looks great! Just a few minor comments to improve the notebook

docs/notebooks/explicit_constraints.pct.py
khurram-ghani: Done in 72a3b9c.

docs/notebooks/explicit_constraints.pct.py
khurram-ghani: Done in 72a3b9c.

docs/notebooks/explicit_constraints.pct.py
khurram-ghani: I have resurrected the code I had previously - the constraints are explicitly created in the notebook. I also now create both the constrained and unconstrained search spaces. This text is updated; hopefully it makes more sense now.

docs/notebooks/explicit_constraints.pct.py
khurram-ghani: It does not have constraints. The markdown above mentions it. The penalty method is a bit strange to setup. The constrained search space is only passed to `FastConstraintsFeasibility` in this mode, to skip the constraints specific functionality in the other classes.

I now explicitly pass the unconstrained search space. Do you think that is clear enough now?

vpicheny: quite clear now!

docs/notebooks/explicit_constraints.pct.py
vpicheny: perfect

PR #665, opened 2022-12-02
Add NARGP multifidelity model

AlexGoodallAI:


trieste/models/gpflow/builders.py
```diff
@@ -474,3 +497,97 @@ def build_multifidelity_autoregressive_models(
     ]
 
     return gprs
+
+
+def build_multifidelity_nonlinear_autoregressive_models(
```
vpicheny, line 486: can we make the kernel type an optional argument? or does it only work with `Matern32`?

trieste/models/gpflow/builders.py
```diff
@@ -474,3 +497,97 @@ def build_multifidelity_autoregressive_models(
...
+    kernel_base_class: Type[Stationary], n_fidelities: int, n_input_dims: int, ard: bool = False
+) -> Sequence[Stationary]:
+
+    dims = list(range(n_input_dims + 1))
+    if ard:
+        lengthscales = tf.ones([n_input_dims])
+    else:
+        lengthscales = 1.0
+    scale_lengthscale = 1.0
+    kernels = [kernel_base_class(lengthscales=lengthscales)]
```
vpicheny, line 594: This assumes that the input space is always scaled such that a lenghtscale of 1.0 is a initial value. Might be best to have the `search_space` as input and use `_get_kernel`? would give you the priors for free.

Unless there is a good practical reason not to use those?

trieste/models/gpflow/models.py
```diff
@@ -1647,3 +1647,367 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+    ):
+        """
+        :param fidelity_models: List of
+            :class:`~trieste.models.gpflow.models.GaussianProcessRegression`
+            models, one for each fidelity. The model at index 0 should take
+            inputs with the same number of dimensions as `x` and can use any kernel,
+            whilst the later models should take an extra input dimesion, and use the kernel
+            described in :cite:`perdikaris2017nonlinear`.
+        :param num_monte_carlo_samples: The number of Monte Carlo samples to use for the
+            ections of prediction and sampling that require the use of Monte Carlo methods.
```
vpicheny: typo

trieste/models/gpflow/models.py
```diff
@@ -1647,3 +1647,367 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+        sample_var = tf.repeat(sample_var, self.monte_carlo_random_numbers.shape[0], axis=-1)[
+            ..., :, None, :
+        ]  # [..., N, 1, S]
+
+        # Repeat fidelity points for each sample to match shapes for masking
+        query_points_fidelity_col = tf.repeat(
+            query_points_fidelity_col, self.monte_carlo_random_numbers.shape[0], axis=-1
+        )[
+            ..., :, None, :
+        ]  # [N, 1, S]
```
vpicheny: I guess it's [..., N, 1, S] ?

trieste/models/gpflow/models.py
```diff
@@ -1647,3 +1647,367 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+        # Predict for all fidelities but stop updating once we have
+        # reached desired fidelity for each query point
+        for fidelity in range(1, self.num_fidelities):
+
+            # sample_mean [..., N, 1, S]
+            # sample_var [..., N, 1, S]
+            (
+                next_fidelity_sample_mean,
+                next_fidelity_sample_var,
+            ) = self._propogate_samples_through_level(
```
vpicheny: typo: propogate

trieste/models/gpflow/models.py
```diff
@@ -1647,3 +1647,367 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+        ) = check_and_extract_fidelity_query_points(
+            query_points, max_fidelity=self.num_fidelities - 1
+        )  # [..., N, D], [..., N, 1]
+
+        sample_mean, sample_var = self.fidelity_models[0].predict(
+            query_points_wo_fidelity
+        )  # [..., N, 1], [..., N, 1]
+
+        # Create new dimension to store samples for each query point
+        # Repeat the inital sample mean and variance 100 times and add a dimension in the
```
vpicheny: 100 -> S

trieste/models/gpflow/models.py
```diff
@@ -1647,3 +1647,367 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+        # Repeat random numbers for each query point and add middle dimension
+        # for concatentation with query points This also means that it has the same
+        # shape as sample_var and sample_mean, so there's no broadcasting required
+        # Note: at the moment we use the same monte carlo values for every value in the batch dim
+        reshaped_random_numbers = tf.repeat(
+            tf.transpose(self.monte_carlo_random_numbers), N, axis=-2
+        )[
+            ..., None, :
+        ]  # [..., N, 1, S]
+        samples = reshaped_random_numbers * tf.sqrt(sample_var) + sample_mean  # [..., N, 1, S]
```
vpicheny, line 1876: actually I think that if `sample_var` and `sample_mean` are of shape `[..., N, 1, 1]` this will just broadcast? so you don't need the repeats above. But looks like they are useful for the mask?

trieste/models/gpflow/models.py
```diff
@@ -1647,3 +1647,367 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+        """
+
+        (
+            query_points_wo_fidelity,
+            query_points_fidelity_col,
+        ) = check_and_extract_fidelity_query_points(
+            query_points, max_fidelity=self.num_fidelities - 1
+        )  # [..., N, D], [..., N, 1]
+
+        sample_mean, sample_var = self.fidelity_models[0].predict(
```
vpicheny, line 1801: naive question: is this what we want to do? i.e. use a single sample mean and var and duplicate it? or would we ideally want S values of `sample_mean` and `sample_var`?

trieste/models/gpflow/models.py
```diff
@@ -1647,3 +1647,367 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+            self.monte_carlo_random_numbers.shape[0],
+            axis=-1,
+        )  # [..., N, D, S]
+        qp_augmented = tf.concat([qp_repeated, samples], axis=-2)  # [..., N, D+1, S]
+
+        # Flatten sample dimension to n_qp dimension to pass through predictor
+        # Switch dims to make reshape match up correct dimensions for query points
+        # Use Einsum to switch last two dimensions
+        qp_augmented = tf.linalg.matrix_transpose(qp_augmented)  # [..., N, S, D+1]
+        flat_qp_augmented = tf.reshape(qp_augmented, shape=[*B, N * S, D + 1])  # [..., N*S, D+1]
```
vpicheny: we have `flatten_leading_dims` here https://github.com/secondmind-labs/trieste/blob/develop/trieste/utils/misc.py#L240

not sure if it would be useful?

trieste/models/gpflow/models.py
AlexGoodallAI: Yeah, I did everything explicitly and avoided broadcasting to try and keep better track of everything. But yes, I think for mask it's useful to have everything the same shape or it can get confusing (for me at least!)

trieste/models/gpflow/models.py
AlexGoodallAI: Yeah, I think this was something we discussed, and may not have been completely sure on. In the function that is called shortly after we then use the random numbers to turn the repeats into a Cartesian product of [sample, monte carlo random number]. I think that is how we get to the S values of `sample_mean` and `sample_var`

trieste/models/gpflow/models.py
AlexGoodallAI: Yes, that seems to do what I need it to, thanks!

trieste/models/gpflow/builders.py
AlexGoodallAI: We could do. We'd have to extend `_get_kernel` to work for arbitrary kernels if we also want to allow the base kernels be arbitrary for this model. There's also a little bit of trickiness of not fully knowing what the input space is for the models for fidelity 1+ because they take outputs from the lower fidelity as inputs, but maybe there's a straightforward solution or good heuristic for this

trieste/models/gpflow/builders.py
```diff
@@ -474,3 +475,108 @@ def build_multifidelity_autoregressive_models(
...
+    # Split data into fidelities
+    data = split_dataset_by_fidelity(dataset=dataset, num_fidelities=num_fidelities)
+
+    _validate_multifidelity_data_modellable(data, num_fidelities)
+
+    # Input dim requires excluding fidelity row
+    input_dim = dataset.query_points.shape[1] - 1
+
+    # Create kernels
+    kernels = _create_multifidelity_nonlinear_autoregressive_kernels(
```
vpicheny, line 521: looks like you can't change `ard` option to `True`, so need to make it also an optional argument of the builder?

trieste/models/gpflow/builders.py
vpicheny: Right, forgot about the `kernel_base_class` input. It's probably worth modifying `get_kernel`? but not straightforward because we'd need to also have an ard argument.

I agree that the last dimension is going to be tricky... can't think of a good heuristic unless we have access to the data here, which is not the case...

vpicheny: if we leave it as it is, it will break if people use it on search_spaces that are not rescaled to approximately [0, 1], so it's probably worth doing a little bit extra work? on the other hand if changing get_kernel looks too complicated we can leave it as it is, perhaps add some sort of warning in the docstring?

trieste/models/gpflow/models.py
vpicheny: I think I got mislead by the name `sample_mean` - at this point it's actually just a call to `predict` so it's a deterministic value.

trieste/models/gpflow/models.py
vpicheny: if this doesn't incur any extra cost (or memory print) then it's not worth bothering I guess?

trieste/models/gpflow/models.py
AlexGoodallAI: Ah yes, sorry, I think the naming is bad here

vpicheny (approved):

PR #668, opened 2022-12-08
Allow dataset/model tags to be arbitrary hashables

uri-granta:
This handles the case when datasets/models are indexed by something more complicated than a unique string.

Only annoyance is in some of the unit tests: because `Mapping[KT, VT]` is invariant in KT, we have to tell the type checker that the various  dictionaries of the form `{"foo": model}` that the tests use are actually `Mapping[Tag, Model]` not `Mapping[str, Model]`. The simplest way to do this is to predefine the tags, e.g. `FOO: Tag = "foo"`.


hstojic (approved):
seems good to me - what's the underlying motivation for this? I thought there might be an issue related to it?

Issue #64, opened 2020-11-19
Make SearchSpace __contains__ broadcast


joelberkeley:
**Describe the feature you'd like**
To be able to use a tensor of points in `points in search_space` (i.e. broadcast the membership test over all points in `points`)

If the last dimension always corresponds to the search space dimension, `points` can be any rank

uri-granta:
Fixed by https://github.com/secondmind-labs/trieste/pull/677.

PR #677, opened 2023-01-11
Broadcast space containment

uri-granta:
Implements https://github.com/secondmind-labs/trieste/issues/64 (which for some reason I can't figure out how to link to in Development)

### DIFF
--- a/tests/unit/test_space.py
+++ b/tests/unit/test_space.py
@@ -55,7 +55,7 @@ class Integers(SearchSpace):
     def sample(self, num_samples: int, seed: Optional[int] = None) -> tf.Tensor:
         return tf.random.shuffle(tf.range(self.limit), seed=seed)[:num_samples]
 
-    def __contains__(self, point: tf.Tensor) -> bool | TensorType:
+    def _contains(self, point: tf.Tensor) -> bool | TensorType:
         tf.debugging.assert_integer(point)
         return 0 <= point < self.limit
 
@@ -101,6 +101,15 @@ def test_discrete_search_space_points() -> None:
 @pytest.mark.parametrize("point", list(_points_in_2D_search_space()))
 def test_discrete_search_space_contains_all_its_points(point: tf.Tensor) -> None:
     assert point in DiscreteSearchSpace(_points_in_2D_search_space())
+    assert DiscreteSearchSpace(_points_in_2D_search_space()).contains(point)
+
+
+def test_discrete_search_space_contains_all_its_points_at_once() -> None:
+    points = _points_in_2D_search_space()
+    space = DiscreteSearchSpace(points)
+    contains = space.contains(points)
+    assert len(contains) == len(points)
+    assert tf.reduce_all(contains)
 
 
 @pytest.mark.parametrize(
@@ -114,6 +123,31 @@ def test_discrete_search_space_contains_all_its_points(point: tf.Tensor) -> None
 )
 def test_discrete_search_space_does_not_contain_other_points(point: tf.Tensor) -> None:
     assert point not in DiscreteSearchSpace(_points_in_2D_search_space())
+    assert not DiscreteSearchSpace(_points_in_2D_search_space()).contains(point)
+
+
+def test_discrete_search_space_contains_some_points_but_not_others() -> None:
+    points = tf.constant([[-1.0, -0.4], [-1.0, 0.4], [-1.0, 0.5]])
+    space = DiscreteSearchSpace(_points_in_2D_search_space())
+    contains = space.contains(points)
+    assert list(contains) == [False, True, False]
+
+
+@pytest.mark.parametrize(
+    "test_points, contains",
+    [
+        (tf.constant([[0.0, 0.0], [1.0, 1.0]]), tf.constant([True, False])),
+        (tf.constant([[[0.0, 0.0]]]), tf.constant([[True]])),
+    ],
+)
+def test_discrete_search_space_contains_handles_broadcast(
+    test_points: tf.Tensor, contains: tf.Tensor
+) -> None:
+    space = DiscreteSearchSpace(tf.constant([[0.0, 0.0]]))
+    tf.assert_equal(contains, space.contains(test_points))
+    # point in space raises (because python insists on a bool)
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        _ = test_points in space
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


trieste/space.py
```diff
@@ -387,9 +420,9 @@ def dimension(self) -> TensorType:
         """The number of inputs in this search space."""
         return self._dimension
 
-    def __contains__(self, value: TensorType) -> bool | TensorType:
-        tf.debugging.assert_shapes([(value, self.points.shape[1:])])
-        return tf.reduce_any(tf.reduce_all(value == self._points, axis=1))
+    def _contains(self, value: TensorType) -> TensorType:
+        comparison = tf.math.equal(self._points, tf.expand_dims(value, -2))  # [..., N, D]
```
henrymoss, line 425: why is value a different shape to self._points?

trieste/space.py
```diff
@@ -984,7 +1007,7 @@ def fix_subspace(self, tag: str, values: TensorType) -> TaggedProductSearchSpace
 
     def get_subspace_component(self, tag: str, values: TensorType) -> TensorType:
         """
-        Returns the components of ``values`` lying in a particular subspace.
+        Returns the components of ``values`` lying in a particular subspace. Supports broadcasting.
```
henrymoss: maybe not worth writing "support broadcasting". Most other methods do and we dont state it there?

trieste/space.py
```diff
@@ -212,15 +211,49 @@ def sample(self, num_samples: int, seed: Optional[int] = None) -> TensorType:
...
         """
+        tf.debugging.assert_equal(
+            tf.rank(value) == 1,
+            True,
+            message=f"""
+                Rank mismatch: expected 1, got {tf.rank(value)}. To broadcast containment use
+                `space.contains(value)` method rather than `value in space`.
+                """,
+        )
+        return self.contains(value)
```
henrymoss: this confused me? But I guess its for a good reason

trieste/space.py
uri-granta: The idea is that `value` can be a single point, an arbitrary list of points, an arbitrary array of points, etc, while `self._points` is the fixed list of points defining the space. The computation needs to check for every point in `value` whether it is equal to one of the points in `self._points`.

trieste/space.py
uri-granta: The problem is that the Python expression `A in B` automatically converts the result of calling `B.__contains__(A)` to a boolean, which means it can't support broadcasting. That's why I've added a separate `contains` method that we can call to get broadcasting if we want to, while the `in` method will still give a useful error if you pass in multiple points.

trieste/space.py
```diff
@@ -212,15 +211,49 @@ def sample(self, num_samples: int, seed: Optional[int] = None) -> TensorType:
...
+        :param value: A single point to check for membership of this :class:`SearchSpace`.
+        :return: `True` if ``value`` is a member of this search space, else `False`.
         :raise ValueError (or tf.errors.InvalidArgumentError): If ``value`` has a different
             dimensionality from this :class:`SearchSpace`.
         """
+        tf.debugging.assert_equal(
+            tf.rank(value) == 1,
+            True,
+            message=f"""
+                Rank mismatch: expected 1, got {tf.rank(value)}. To broadcast containment use
```
sc336: I believe containment means exclusion, rather than the noun form of contain.  Consider "To get a tensor of boolean membership values from a tensor of points, use..."

tests/unit/test_space.py
```diff
@@ -114,6 +123,31 @@ def test_discrete_search_space_contains_all_its_points(point: tf.Tensor) -> None
 )
 def test_discrete_search_space_does_not_contain_other_points(point: tf.Tensor) -> None:
     assert point not in DiscreteSearchSpace(_points_in_2D_search_space())
+    assert not DiscreteSearchSpace(_points_in_2D_search_space()).contains(point)
+
+
+def test_discrete_search_space_contains_some_points_but_not_others() -> None:
+    points = tf.constant([[-1.0, -0.4], [-1.0, 0.4], [-1.0, 0.5]])
```
sc336, line 132: Might be worth a note specifying that this test will break if the return of _points_in_2D_search_space changes?

tests/unit/test_space.py
```diff
@@ -358,6 +394,7 @@ def test_box_bounds_attributes() -> None:
 )
 def test_box_contains_point(point: tf.Tensor) -> None:
     assert point in Box(tf.constant([-1.0, 0.0, -2.0]), tf.constant([2.0, 1.0, -0.5]))
+    assert Box(tf.constant([-1.0, 0.0, -2.0]), tf.constant([2.0, 1.0, -0.5])).contains(point)
```
sc336: Would be slightly neater and easier to read if 
```python3
box = Box(tf.constant([-1.0, 0.0, -2.0]), tf.constant([2.0, 1.0, -0.5]))
assert point in box
assert box.contains(point)
```
(and similarly in other places which duplicate `point in` with `.contains(point)`)

tests/unit/test_space.py
```diff
@@ -370,11 +407,31 @@ def test_box_contains_point(point: tf.Tensor) -> None:
...
+    assert not Box(tf.constant([-1.0, 0.0, -2.0]), tf.constant([2.0, 1.0, -0.5])).contains(point)
+
+
+@pytest.mark.parametrize(
+    "points, contains",
+    [
+        (tf.constant([[-1.0, 0.0, -2.0], [-1.1, 0.0, -2.0]]), tf.constant([True, False])),
+        (tf.constant([[[0.5, 0.5, -1.5]]]), tf.constant([[True]])),
+    ],
+)
```
sc336, line 423: Is there a reason to pass the points and contains values in using `parametrize` but not to pass the box in?

tests/unit/test_space.py
```diff
@@ -370,11 +407,31 @@ def test_box_contains_point(point: tf.Tensor) -> None:
...
+def test_box_contains_broadcasts(points: tf.Tensor, contains: tf.Tensor) -> None:
+    box = Box(tf.constant([-1.0, 0.0, -2.0]), tf.constant([2.0, 1.0, -0.5]))
+    npt.assert_array_equal(contains, box.contains(points))
+    # point in space raises (because python insists on a bool)
+    with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+        _ = points in box
```
sc336, line 429: Is this just to test that using `in` on a tensor of points raises an error?

sc336 (approved):
Some commenting and clarity points but LGTM generally.

tests/unit/test_space.py
uri-granta: Yes

PR #678, opened 2023-01-12
fix for multi objective integration test

hstojic:
we were computing ref_point on observations instead of the true pareto front

### DIFF
--- a/tests/integration/test_multi_objective_bayesian_optimization.py
+++ b/tests/integration/test_multi_objective_bayesian_optimization.py
@@ -172,10 +172,15 @@ def test_multi_objective_optimizer_finds_pareto_front_of_the_VLMOP2_function(
             )
 
     # A small log hypervolume difference corresponds to a succesful optimization.
-    ref_point = get_reference_point(dataset.observations)
-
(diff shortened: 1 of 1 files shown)


vpicheny (commented):
actually in `get_reference_point` we extract the nadir:

```
    front = Pareto(observations).front
    f = tf.math.reduce_max(front, axis=-2) - tf.math.reduce_min(front, axis=-2)
    return tf.math.reduce_max(front, axis=-2) + 2 * f / tf.cast(tf.shape(front)[-2], f.dtype)
```
so I think this fix is not needed?

vpicheny (approved):
my bad, now I understand the fix.

hstojic:
> actually in `get_reference_point` we extract the nadir:
> 
> ```
>     front = Pareto(observations).front
>     f = tf.math.reduce_max(front, axis=-2) - tf.math.reduce_min(front, axis=-2)
>     return tf.math.reduce_max(front, axis=-2) + 2 * f / tf.cast(tf.shape(front)[-2], f.dtype)
> ```
> 
> so I think this fix is not needed?

That nadir is based on acquired data, not on the true front - for evaluating the performance we should have a fixed one based on the true pareto front, right?

PR #680, opened 2023-01-16
gpflow models logging improvement

hstojic:
I have added model performance on training data, it was missing so far and I think these are important diagnostic metrics whether model training went well

I have also extracted these summaries as functions common to all the models


trieste/models/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    Logging utility for writing TensorBoard summary of likelihood parameters. Provides useful
+    diagnostics for models with a GPflow likelihood. Only trainable parameters are logged.
+
+    :param likelihood: The likelihood to use for computing the metrics.
+    :param prefix: The prefix to add to "likelihood" category of model summaries.
+    """
+    likelihood_components = _merge_leaf_components(leaf_components(likelihood))
+    for k, v in likelihood_components.items():
+        if v.trainable:
+            logging.scalar(f"{prefix}likelihood.{k}", v)
```
avullo, line 104: Could it better belong to `trieste.models.gpflow.utils` with possibly a similar function but for a `GPflow` model which would then invoke these two?

tests/unit/models/test_utils.py
```diff
@@ -0,0 +1,206 @@
...
+                distribution_class=tfp.distributions.Normal,
+                scale_transform=tfp.bijectors.Exp(),
+            ),
+            [],
+            [],
+            id="HeteroskedasticTFPConditional",
+        ),
+    ],
+)
+def test_write_summary_likelihood_parameters(
```
avullo, line 138: The body is the same as before. I think you can group this test and the one above into a single one just by passing the `write_summary_*_parameters` as an extra parameter.

trieste/models/utils.py
hstojic: these are used by gpflux models as well so its more appropriate that these functions are here

tests/unit/models/test_utils.py
```diff
@@ -0,0 +1,206 @@
...
+        assert mocked_summary_scalar.call_args_list[i][0][0] == n
+        assert tf.reduce_all(mocked_summary_scalar.call_args_list[i][0][1].numpy() == v)
+
+
+@unittest.mock.patch("trieste.logging.tf.summary.histogram")
+@unittest.mock.patch("trieste.logging.tf.summary.scalar")
+@pytest.mark.parametrize("prefix", ["", "dummy_"])
+def test_write_summary_data_based_metrics(
+    mocked_summary_scalar: unittest.mock.MagicMock,
+    mocked_summary_histogram: unittest.mock.MagicMock,
```
hstojic, line 157: @uri-granta something weird happens here with scalar and histogram mock - the order matters and it only works correctly this way, not sure how are they exactly mapped to the variables below - any suggestion how to make this more robust?

trieste/models/utils.py
```diff
@@ -0,0 +1,104 @@
...
+from gpflow.utilities.traversal import _merge_leaf_components, leaf_components
+
+from .. import logging
+from ..data import Dataset
+from .interfaces import TrainableProbabilisticModel
+
+
+def write_summary_data_based_metrics(
+    dataset: Dataset,
+    model: TrainableProbabilisticModel,
```
uri-granta: You don't need a trainable model here (and in fact the GPflowPredictor ABC isn't one itself).

```suggestion
    model: ProbabilisticModel,
```

trieste/models/gpflow/interface.py
```diff
@@ -146,22 +150,22 @@ def optimize(self, dataset: Dataset) -> None:
...
-                kernel = self.get_kernel()
-                components = _merge_leaf_components(leaf_components(kernel))
-                for k, v in components.items():
-                    if v.trainable:
-                        if tf.rank(v) == 0:
-                            logging.scalar(f"kernel.{k}", v)
-                        elif tf.rank(v) == 1:
-                            for i, vi in enumerate(v):
-                                logging.scalar(f"kernel.{k}[{i}]", vi)
+                # kernel parameters
```
uri-granta: (don't think the comments are adding anything here other than noise)

trieste/models/keras/models.py
```diff
@@ -430,27 +431,8 @@ def log(self, dataset: Optional[Dataset] = None) -> None:
...
-                    # training variance
-                    variance_error = predict[1] - diffs ** 2
-                    logging.histogram("variance/predict_variance", predict[1])
-                    logging.histogram("variance/variance_error", variance_error)
-                    logging.scalar("variance/predict_variance_mean", tf.reduce_mean(predict[1]))
-                    logging.scalar(
-                        "variance/root_mean_variance_error",
-                        tf.math.sqrt(tf.reduce_mean(variance_error ** 2)),
+                    write_summary_data_based_metrics(
+                        dataset=dataset, model=self, prefix="training_"
```
uri-granta, line 435: What prefix looks better: `"training"` or `"training_"`? You should probably choose one and stick with it for consistency.

trieste/models/gpflux/models.py
```diff
@@ -378,51 +382,14 @@ def log(self, dataset: Optional[Dataset] = None) -> None:
...
                 for idx, layer in enumerate(self.model_gpflux.f_layers):
-                    kernel_components = _merge_leaf_components(leaf_components(layer.kernel))
-                    for k, v in kernel_components.items():
-                        if v.trainable:
-                            if tf.rank(v) == 0:
-                                logging.scalar(f"layer[{idx}]/kernel.{k}", v)
-                            elif tf.rank(v) == 1:
-                                for i, vi in enumerate(v):
-                                    logging.scalar(f"layer[{idx}]/kernel.{k}[{i}]", vi)
+                    write_summary_kernel_parameters(layer.kernel, f"layer[{idx}]/")
```
uri-granta: (suggestion)

```suggestion
                    write_summary_kernel_parameters(layer.kernel, prefix=f"layer[{idx}]/")
```

tests/unit/models/test_utils.py
uri-granta: Yes, the [parameter order matters](https://docs.python.org/3/library/unittest.mock.html#nesting-patch-decorators):

> Note that the decorators are applied from the bottom upwards. This is the standard way that Python applies decorators. The order of the created mocks passed into your test function matches this order.

uri-granta (approved):
Good work, thanks!

trieste/models/utils.py
hstojic: done

trieste/models/gpflow/interface.py
hstojic: ok, removed

trieste/models/keras/models.py
hstojic: done

trieste/models/gpflux/models.py
hstojic: done

hstojic:
thanks @uri-granta, all comments addressed, I'll merge it in once tests are passed

PR #681, opened 2023-01-18
For testing with GPflux introducing Orthogonal DGPs. 

avullo:


hstojic:
we needed this to exist but no need to keep it open, closing it now

PR #682, opened 2023-01-23
deep ensemble faster sampler

hstojic:
it turns out using reparametrisation trick for `diversify=True` option is much faster :)
bonus: now we can do multi output case as well

### DIFF
--- a/tests/unit/models/keras/test_sampler.py
+++ b/tests/unit/models/keras/test_sampler.py
@@ -50,14 +50,6 @@ def _batch_size_fixture(request: Any) -> int:
     return request.param
 
 
-def test_ensemble_trajectory_sampler_raises_for_multi_output_model() -> None:
-    example_data = empty_dataset([2], [2])
-    model, _, _ = trieste_deep_ensemble_model(example_data, _ENSEMBLE_SIZE)
-
-    with pytest.raises(NotImplementedError):
-        DeepEnsembleTrajectorySampler(model, diversify=True)
-
-
 @pytest.mark.parametrize("diversify, num_outputs", [(True, 1), (False, 1), (False, 2)])
 def test_ensemble_trajectory_sampler_returns_trajectory_function_with_correctly_shaped_output(
     num_evals: int,
@@ -102,7 +94,7 @@ def test_ensemble_trajectory_sampler_returns_deterministic_trajectory(
     npt.assert_allclose(eval_1, eval_2)
 
 
-@pytest.mark.skip(reason="Seems fragile. Unrelated changes causing it to fail. Issue being raised.")
+# @pytest.mark.skip(reason="Seems fragile. Unrelated changes causing it to fail. Issue being raised.")
 @pytest.mark.parametrize("seed", [42, None])
 def test_ensemble_trajectory_sampler_is_not_too_deterministic(
     seed: Optional[int], diversify: bool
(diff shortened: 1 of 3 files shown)


trieste/models/keras/sampler.py
```diff
@@ -173,24 +168,27 @@ def __call__(self, x: TensorType) -> TensorType:  # [N, B, d] -> [N, B, 1]
 
         if self._diversify:
             predicted_means, predicted_vars = self._model.predict(flat_x)  # ([N*B, 1], [N*B, 1])
-            normal = tfp.distributions.Normal(predicted_means, tf.sqrt(predicted_vars))
-            predictions = tf.map_fn(normal.quantile, self._quantiles)  # [B, N*B, 1]
+            predicted_vars = predicted_vars + tf.cast(DEFAULTS.JITTER, predicted_vars.dtype)
+            predictions = predicted_means + tf.sqrt(predicted_vars) * tf.tile(
```
vpicheny, line 172: does this work? `self._eps` is `[B, L]` so here with tiling you get `[N*B, L]`? but how do you know that the right epsilon is given to the right trajectory?

Wouldn't it be easier and safer to do without flattening? e.g. with `predicted_means` being `[..., B, L]` and eps `[B, L]`, you just need to add dummy dimensions for broadcasting.

hstojic: I verified it manually that it is correct way of tiling it...
we cannot really do it without flatenning, we would need B forward passes with the networks which would be much slower

hstojic: we actually have a test that makes sure eps are applied correctly, check `test_ensemble_trajectory_sampler_samples_are_distinct_within_batch` - we have same `x` for each batch and we should get different trajectory for each batch

vpicheny: I understand. Yes it would be really important to test this. Possibly you for a test you can assign `[0, 1]` to `eps` and check that one trajectory contains only zeros and the other doesn't contain any?

vpicheny: small point: `# [B, N*B, 1]` should be `# [B, N*B, L]` now, correct?

hstojic: I think the existing test mentioned above should be enough, but fair enough, added another one you have suggested `test_ensemble_trajectory_sampler_eps_broadcasted_correctly` - works in eager mode but not with tf.function, probably its hard to set _eps externally in that mode - @uri-granta can you perhaps help with this test?

uri-granta: Started taking a look but for me it also fails in eager mode. How are you running it? (I added `tf.config.run_functions_eagerly(True)` straight after importing tensorflow)

uri-granta: Regardless, you'll want to `.assign(...)` to `._eps` rather than `= ...` as it's a variable. And just about makes the test pass.

tests/unit/models/keras/test_sampler.py
```diff
@@ -223,6 +218,34 @@ def test_ensemble_trajectory_sampler_samples_are_distinct_within_batch(diversify
...
+    test_data = tf.expand_dims(test_data, -2)  # [N, 1, d]
+    test_data = tf.tile(test_data, [1, 2, 1])  # [N, 2, D]
+
+    model, _, _ = trieste_deep_ensemble_model(example_data, _ENSEMBLE_SIZE)
+
+    trajectory_sampler = DeepEnsembleTrajectorySampler(model, diversify=True)
+    trajectory = trajectory_sampler.get_trajectory()
+
+    _ = trajectory(test_data)  # first call needed to initialize the state
+    trajectory._eps = tf.constant([[0], [1]], dtype=tf.float64)  # type: ignore
```
uri-granta:
```suggestion
    trajectory._eps.assign(tf.constant([[0], [1]], dtype=tf.float64))
```

uri-granta (approved):
Looks good from an engineering and documentation perspective!

tests/unit/models/keras/test_sampler.py
hstojic: thanks a lot, quite obvious now that you pointed it out :)

PR #683, opened 2023-01-25
BO-specific Inducing Point Allocators

henrymoss:
I have added two new inducing point allocators: one from my new paper and one powerful baseline from the existing literature. 

I also fiddled with the existing inducing point allocators so that they do not need to be given search_space when initialized.  This makes it much easier to initialize our SVGP models.


trieste/models/gpflow/inducing_point_selectors.py
```diff
@@ -112,13 +112,22 @@ def _recalculate_inducing_points(
         raise NotImplementedError
 
 
-class UniformInducingPointSelector(InducingPointSelector[ProbabilisticModel]):
+class UniformInducingPointSelector(InducingPointSelector[GPflowPredictor]):
```
uri-granta, line 115: (why, out of interest?)

trieste/models/gpflow/inducing_point_selectors.py
```diff
@@ -163,22 +172,28 @@ def _recalculate_inducing_points(
...
 
         if N < M:  # if fewer data than inducing points then sample remaining uniformly
-            uniform_sampler = UniformInducingPointSelector(self._search_space)
+            data_as_discrete_search_space = DiscreteSearchSpace(dataset.query_points)
+            convex_hull_of_data = Box(
+                lower=data_as_discrete_search_space.lower,
+                upper=data_as_discrete_search_space.upper,
+            )
+
+            uniform_sampler = UniformInducingPointSelector(convex_hull_of_data)
```
uri-granta, line 181: Nice. Should we have a utility function to convert a discrete (or even arbitrary) searchspace into a convex hull box?

trieste/models/gpflow/inducing_point_selectors.py
```diff
@@ -215,10 +230,226 @@ def _recalculate_inducing_points(
...
                 M - len(centroids), model, dataset
             )
             centroids = tf.concat([centroids, extra_centroids], axis=0)  # [M, d]
 
         return centroids  # [M, d]
+
+
+class QualityFunction(ABC):
+    """
+    An :const:`QualityFunction` uses a  `model` to measure the quality of each of
```
uri-granta:
```suggestion
    A :const:`QualityFunction` uses a  `model` to measure the quality of each of
```

trieste/models/gpflow/inducing_point_selectors.py
```diff
@@ -215,10 +230,226 @@ def _recalculate_inducing_points(
...
+    def __call__(self, model: GPflowPredictor, dataset: Dataset) -> TensorType:
+        """
+        Evaluate the quality of the data-points according to the model.
+        :param model: The sparse model.
+        :param dataset: The data from the observer. Must be populated.
+        :return: The quality scores.
+        """
+
+
+class DPPInducingPointSelector(InducingPointSelector[GPflowPredictor]):
```
uri-granta, line 263: Do we want some `__repr__`s for these classes?

trieste/models/gpflow/inducing_point_selectors.py
```diff
@@ -215,10 +230,226 @@ def _recalculate_inducing_points(
...
+    over ``dataset`` following the algorithm of :cite:`chen2018fast`. Note that we are using the
+    quality-diversity decomposition of a DPP, specifying both a similarity ``kernel``
+    and ``quality_scores``.
+
+    :param M: Desired set size.
+    :param kernel: The underlying kernel of the DPP.
+    :param quality_scores: The quality score of each item in ``dataset``.
+    :return: The MAP estimate of the DPP.
+    :raise tf.errors.InvalidArgumentError: If ``dataset`` is empty or if the shape of
+    ``quality_scores`` does not match that of ``dataset.observations``.
```
uri-granta: Probably the reason the docs build is failing

```suggestion
        ``quality_scores`` does not match that of ``dataset.observations``.
```

uri-granta (approved):
Looks good from an engineering perspective! Just a couple of questions.

trieste/models/gpflow/inducing_point_selectors.py
henrymoss: Its what we want, as the only models with inducing points are GP ones really, and we rely on accessing the kernel anyway

trieste/models/gpflow/inducing_point_selectors.py
henrymoss: Sure, will leave for another PR though

trieste/models/gpflow/inducing_point_selectors.py
henrymoss: We didnt have any for the existing inducing point allocators. We can add this in another PR

PR #685, opened 2023-01-30
Workaround for alltests crash

uri-granta:
Alltests is currently OOM crashing for reasons unknown, though it's possible there are just too many tests (since pytest holds on to some monitoring data for each test). We should really:

1. try to fix it if possible
2. split test run across multiple workers to mitigate it (and speed things up!)

This ticket mitigates it another way, which is to only run the slow tests after merging (since the other tests have already been run beforehand).


tests/integration/test_bayesian_optimization.py
```diff
@@ -349,7 +349,7 @@ def test_bayesian_optimizer_with_sgpr_finds_minima_of_simple_quadratic() -> None
             25,
             EfficientGlobalOptimization(
                 ParallelContinuousThompsonSampling(),
-                num_query_points=3,
```
sc336, line 352: How does this change (and the change in the parameters of `_test_optimizer_finds_minimum` above) relate to only running the slow tests?

sc336 (approved):

tests/integration/test_bayesian_optimization.py
uri-granta: These are tests that were previously not being run due to the OOM crash. Removing the crash showed that the tests were fragile: each was passing in one testing environment but failing in the other. The changes are aiming to make them a bit less fragile and pass in both (which they now do: https://github.com/secondmind-labs/trieste/actions/runs/4045012501).

uri-granta: (PS You can see that both the changed tests were decorated with `@pytest.mark.slow`)

PR #687, opened 2023-01-31
PF2ES preview notebook

TsingQAQ:
This is a preview notebook for [PF2ES](https://arxiv.org/abs/2204.05411) (update Arxiv Ver comes later) acquisition function for multi-objective optimization using information theoretic-based search strategy. This standalone notebook has every additional functionality supporting the acquisition function. 

Several noticeable differences between this notebook with the original implementation of PF2ES are:
- This notebook depends on upgrade `Pymoo` ver 0.6.0.1 (instead of 0.5.0), this is observed as having minor speed gain for the sequential acquisition function sample.
- There is no `maxiter` specified for multi-start l-bfgs-b acq optimier in this notebook, hence for the batch PF2ES (which is more difficult for the optimizer to converge due to the MC approximation), the acq optimization process can take a longer time here.
- the constraint < 0  is considered feasible for the constraint problem here.


TsingQAQ:
according to discussions, this notebook will be better suited as a branch of Trieste fork, so turn to close the PR for `develop`.

PR #688, opened 2023-01-31
Test against Python 3.10

uri-granta:
Migrate *_latest tests from Python 3.9 to Python 3.10. Keep the 3.7 tests as before.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -28,7 +28,7 @@ jobs:
         path: gh-pages-history
     - uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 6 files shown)


sc336 (approved):
LGTM.  Feel free to leave a comment about moving from 3.7 to "3.7" throughout if you like, but I doubt it's important.

uri-granta:
> LGTM. Feel free to leave a comment about moving from 3.7 to "3.7" throughout if you like, but I doubt it's important.

(That's because versions aren't numbers, which isn't a problem for 3.9 but is for 3.10, which as a number gets converted to 3.1)

PR #689, opened 2023-02-03
random scalarization - part 1

hstojic:
this PR implements Random scalarization acquisition builder and function, interface for scalarization functions and one example of a scalarization function

part 2 will follow with a notebook
part 3 will follow with some functions for adaptive ideal points (and possible 1-2 more scalarization functions)

still at a draft stage to collect opinions on the design
(tests are missing and docs are not ironed out)

main questions about the design:
- I went with a dict of models approach to combine trajectories instead of a model stack, but its a bit inconsistent with the rest of multi-objective code, I kinda like the dict approach, but don't have a strong preference and we could use a model stack I guess (we would need a new type of model stack for trajectories I think, similar to `HasReparamSamplerModelStack`)
- I have separated the random scalarization from scalarization functions themselves, as we could have many forms of scalarizations - since they can be quite complex, they are objects. As a result `random_scalarization` is more of a container, most of the work is done by `Scalarizer`. Not sure what's the best approach here, perhaps we should integrate the scalarizer in the `random_scalarization` , but then it might be more difficult to use a different scalarization function. I'm also not sure how this separation affects retracing.

@vpicheny @uri-granta @henrymoss any thoughts on these?

### DIFF
--- a/tests/integration/test_multi_objective_bayesian_optimization.py
+++ b/tests/integration/test_multi_objective_bayesian_optimization.py
@@ -21,8 +21,9 @@ from trieste.acquisition import (
     HIPPO,
     BatchMonteCarloExpectedHypervolumeImprovement,
     ExpectedHypervolumeImprovement,
+    RandomScalarization,
 )
-from trieste.acquisition.multi_objective.pareto import Pareto, get_reference_point
+from trieste.acquisition.multi_objective import Pareto, get_reference_point, Chebyshev
 from trieste.acquisition.optimizer import generate_continuous_optimizer
 from trieste.acquisition.rule import (
     AcquisitionRule,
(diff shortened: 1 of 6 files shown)


trieste/acquisition/multi_objective/scalarization.py
```diff
@@ -0,0 +1,212 @@
...
+import tensorflow as tf
+
+from trieste.data import Dataset
+from trieste.models.interfaces import (
+    HasTrajectorySampler,
+)
+from trieste.types import Tag, TensorType
+
+
+IdealSpecCallable = Callable[..., TensorType]
```
uri-granta, line 31: Why not make this precise?

```suggestion
IdealSpecCallable = Callable[[Mapping[Tag, HasTrajectorySampler], datasets: Mapping[Tag, Dataset]], TensorType]
```

trieste/acquisition/multi_objective/scalarization.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def _get_ideal(
+        self, models: Mapping[Tag, HasTrajectorySampler], datasets: Mapping[Tag, Dataset]
+    ) -> TensorType:
+        dtype = self._infer_dtype(datasets)
+        if callable(self._ideal_spec):
+            ideal = tf.cast(self._ideal_spec(models, datasets), dtype=dtype)
+        else:
+            ideal = tf.cast(self._ideal_spec, dtype=dtype)
+        tf.debugging.assert_shapes([(ideal, (self._num_objectives, 1))])
```
uri-granta, line 83: Maybe add a message?

trieste/acquisition/multi_objective/scalarization.py
```diff
@@ -0,0 +1,212 @@
...
+            self._ideal_spec = tf.convert_to_tensor(ideal_spec)
+
+    @staticmethod
+    def _infer_dtype(datasets: Mapping[Tag, Dataset]) -> tf.DType:
+        return next(iter(datasets.values())).query_points.dtype
+
+    def _get_ideal(
+        self, models: Mapping[Tag, HasTrajectorySampler], datasets: Mapping[Tag, Dataset]
+    ) -> TensorType:
+        dtype = self._infer_dtype(datasets)
```
uri-granta, line 78: This is a bit clunky: e.g. what happens if datasets is empty, or has inconsistent dtypes, or if the dtype changes between calls? Can't we just insist that the ideal_spec returns the correct dtype? 

If that's a problem, maybe:

1. make _infer_dtype more resilient by giving an error if datasets is empty or inconsistent?
2. save the dtype when you call prepare and check it hasn't changed in update?
3. add a dtype parameter to _get_ideal like _sample_weights (or alternatively move the casting outside)?

trieste/acquisition/multi_objective/scalarization.py
```diff
@@ -0,0 +1,212 @@
...
+        :return: Scalarized objectives, [N, B].
+        """
+
+    @abstractmethod
+    def prepare(
+        self, models: Mapping[Tag, HasTrajectorySampler], datasets: Mapping[Tag, Dataset]
+    ) -> None:
+        """
+        Generate all the internal variables on initialization. For example, weights in a linear
+        weighted sum scalarization could be sampled.
```
uri-granta, line 101: Docstring should mention parameters

trieste/acquisition/multi_objective/scalarization.py
```diff
@@ -0,0 +1,212 @@
...
+        tf.debugging.assert_non_negative(alpha, f"alpha should be nonnegative but received {alpha}")
+        self._alpha = alpha
+
+    def __repr__(self) -> str:
+        """"""
+        if callable(self._ideal_spec):
+            return (
+                f"Chebyshev({self._batch_size!r},"
+                f"{self._num_objectives!r},"
+                f"{self._ideal_spec.__name__},"
```
uri-granta, line 160: You could easily remove the code duplication if you want:

```suggestion
                + (f"{self._ideal_spec.__name__}," if callable(self._ideal_spec) else f"{self._ideal_spec!r},") +
```

trieste/acquisition/multi_objective/scalarization.py
```diff
@@ -0,0 +1,212 @@
...
+        :return: Sampled weights [K, B].
+        """
+        weights = tf.random.normal([self._num_objectives, self._batch_size], dtype=dtype)
+        return tf.abs(weights / tf.sqrt(tf.reduce_sum(weights ** 2, axis=0, keepdims=True)))
+
+    def prepare(
+        self, models: Mapping[Tag, HasTrajectorySampler], datasets: Mapping[Tag, Dataset]
+    ) -> None:
+        dtype = self._infer_dtype(datasets)
+        self._weights = tf.Variable(  # pylint: disable=attribute-defined-outside-init
```
uri-granta, line 200: I'm worried that this would create tensorflow compilation issues, if the variables don't exist straight after initialisation. I think elsewhere we've initialised similar variables to dummy values, and (if need be) tracked initialisation with a `self._initialized = tf.Variable(False)` variable.

trieste/acquisition/function/multi_objective.py
```diff
@@ -755,3 +756,123 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            if not isinstance(model, HasTrajectorySampler):
+                raise ValueError(
+                    f"We only support models with a trajectory_sampler "
+                    f"method; received {model.__repr__()}"
+                    f"for output {tag}"
+                )
+
+        self._scalarizer.prepare(models, datasets)
+
+        self._trajectory_sampler = {  # pylint: disable=attribute-defined-outside-init
```
uri-granta, line 799: I think this is less likely to cause issues than the `attribute-defined-outside-init` Variables below, but would still be nice to initialise these to empty dicts and update them in place here. Similarly, you should declare `self._negated_trajectory` as an `Optional[TrajectorySampler]` and initialise it to `None`.

tests/integration/test_multi_objective_bayesian_optimization.py
```diff
@@ -55,81 +56,96 @@
...
-            10,
-            EfficientGlobalOptimization(
-                BatchMonteCarloExpectedHypervolumeImprovement(sample_size=250).using(OBJECTIVE),
-                num_query_points=4,
-                optimizer=generate_continuous_optimizer(num_initial_samples=500),
-            ),
-            -3.2095,
-            id="BatchMonteCarloExpectedHypervolumeImprovement/4",
-        ),
+        # pytest.param(
```
uri-granta, line 59: You probably don't want to check this in :-)

uri-granta (commented):
Looks good. Some initial comments. Will obviously need more tests before it can be checked in.

PR #691, opened 2023-02-03
Accessing shape directly inside tf.function might produce None values in some contexts.

avullo:


hstojic (approved):
wasn't aware of that, we should be careful with that... looks good!

PR #692, opened 2023-02-03
`Cbranincurrin` synthetic test function for cmoo

TsingQAQ:
This is a (draft) PR for adding a commonly utilized synthetic test function for the multi-objective constraint function, which can be used in both notebooks and tests.

- since haven't aware of any constraint objective function class but only one [for the input](https://github.com/secondmind-labs/trieste/blob/3f688e3892f7d338db3cb4172d13bf7ec9e7003e/trieste/objectives/single_objectives.py#L169), might worth discussion whether having a  `class ConstrainedMultiObjectiveTestProblem(MultiObjectiveTestProblem):` as in this PR is the ideal approach. Alternatively, one can return a concatenation of objective function value and constraint in one function. 
- follow the [notebook](https://secondmind-labs.github.io/trieste/1.0.0/notebooks/inequality_constraints.html#:~:text=We%20get%20the%20constrained%20objective%20by%20masking%20out%20regions%20where%20the%20constraint%20function%20is%20above%20the%20threshold), for constraint < 0 is feasible


tests/unit/objectives/test_multi_objectives.py
```diff
@@ -137,13 +178,18 @@ def test_gen_pareto_front_is_equal_to_math_defined(
...
+    actual_x: TensorType,
 ) -> None:
     with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
         obj_inst.objective(actual_x)
+    if isinstance(obj_inst, ConstrainedMultiObjectiveTestProblem):
+        with pytest.raises(TF_DEBUGGING_ERROR_TYPES):
+            obj_inst.constraint(actual_x)
 
 
 @pytest.mark.parametrize(
```
hstojic, line 203: please add a test below as well for shape and dtype

trieste/objectives/multi_objectives.py
```diff
@@ -236,3 +237,76 @@ def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
         search_space=Box([0.0], [1.0]) ** d,
         gen_pareto_optimal_points=gen_pareto_optimal_points,
     )
+
+
+@dataclass(frozen=True)
+class ConstrainedMultiObjectiveTestProblem(MultiObjectiveTestProblem):
```
hstojic: could you please move it up just below the `MultiObjectiveTestProblem` interface

trieste/objectives/multi_objectives.py
```diff
@@ -236,3 +237,76 @@ def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
...
+    """
+    Convenience container class for synthetic constrained multi-objective test functions, containing
+    a generator for the pareto optimal points, which can be used as a reference of performance
+    measure of certain multi-objective optimization algorithms.
```
hstojic:
```suggestion
    additionally a constraint function.
```

trieste/objectives/multi_objectives.py
```diff
@@ -236,3 +237,76 @@ def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
...
+
+def branin_currin(x: TensorType) -> TensorType:
+    """
+    The branincurrin synthetic function.
+
+    :param x: The points at which to evaluate the function, with shape [..., d].
+    :raise ValueError (or InvalidArgumentError): If ``x`` has an invalid shape.
+    """
+    tf.debugging.assert_shapes([(x, (..., 2))])
+    return tf.concat([branin(x), currin(x)], axis=-1)
```
hstojic, line 322: mixing objectives and constraints seems like a wrong design, I think it makes sense to keep them separate, since constraints typically have different treatment (e.g. computing prob of feasibility)

what I would suggest to return `currin` as constraint and add another method to the interface for evaluating the constraint (i.e. `evaluate_slack_true`), we could call it `evaluate_constraint`? 

@uri-granta what do you think?

trieste/objectives/multi_objectives.py
```diff
@@ -236,3 +237,76 @@ def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
...
+    See :cite:`belakaria2019max` and :cite:`daulton2020differentiable`
+    (the latter for adding the constraint) for details.
+
+    :return: The problem specification.
+    """
+
+    def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
+        return tf.zeros(shape=0)
+
+    def evaluate_slack_true(x: TensorType) -> TensorType:
```
hstojic: I'm not familiar with the constrained brannin currin function, would it work with any other threshold than 0?

hstojic: but in any case, in general threshold will typically be flexible and user could modify it - so in the new method in the interface lets have an additional argument `threshold: float`, you can set it to 0 as a default in your problem and potentially add to the docs that it shouldn't be used in any other way if it only works with 0

hstojic (commented):
thanks for the proposition!
left some comments - in short I think we need to modify the interface a bit...

trieste/objectives/multi_objectives.py
TsingQAQ: Sorry that i have failed to make this clear: for the `branincurrin` problem the objective function is a concatenation of `[branin, currin]`, so both are objectives. For the constraint we indeed want to have a separate function, i.e.,  `evaluate_constraint` as suggested.

trieste/objectives/multi_objectives.py
TsingQAQ: >would it work with any other threshold than 0?  

the constraint should be well defined on the different thresholds. The default formulation used here is just a copy of the reference paper's equation since its property is better understood (than other thresholds).

trieste/objectives/multi_objectives.py
hstojic: ah, ok, clearer now, thanks!

trieste/objectives/multi_objectives.py
hstojic: ok, then lets simply set it as a default and say something about it in docs

trieste/objectives/multi_objectives.py
```diff
@@ -236,3 +248,66 @@ def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
...
+
+    :return: The problem specification.
+    """
+
+    def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
+        return tf.zeros(shape=0, dtype=tf.float64)
+
+    def evaluate_constraint(x: TensorType, threshold: Optional[float] = 0.0) -> TensorType:
+        """
+        The constraint of branincurrin problem, < threshold is feasible
```
hstojic:
```suggestion
        The constraint of branincurrin problem, < ``threshold`` is feasible.
```

trieste/objectives/multi_objectives.py
```diff
@@ -236,3 +248,66 @@ def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
...
+    """
+    The ConstrainedBraninCurrin problem, typically evaluated over :math:`[0, 1]^2`.
+    See :cite:`belakaria2019max` and :cite:`daulton2020differentiable`
+    (the latter for adding the constraint) for details.
+
+    :return: The problem specification.
+    """
+
+    def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
+        return tf.zeros(shape=0, dtype=tf.float64)
```
hstojic: so there is no known pareto front for this problem I take it?
can you please add a note to the docs of the problem above stating that and that this function will return an empty tensor please

trieste/objectives/multi_objectives.py
```diff
@@ -236,3 +248,66 @@ def gen_pareto_optimal_points(n: int, seed: int | None = None) -> TensorType:
...
+
+    def evaluate_constraint(x: TensorType, threshold: Optional[float] = 0.0) -> TensorType:
```
hstojic, line 291: I think we should have some tests for this, for example:
1. for some selected input and threshold values we know what the result should be
2. the output is boolean

hstojic (commented):
almost there, some further minor comments

trieste/objectives/multi_objectives.py
TsingQAQ: not sure if get point 1 properly but we have a test for different `threshold values` as [here](https://github.com/TsingQAQ/trieste/blob/1ae93d133b019ab4e3a54d3b4fe8c6896fa7a313/tests/unit/objectives/test_multi_objectives.py#L133-L142), would it be possible to elaborate these points especially point 2 please?

trieste/objectives/multi_objectives.py
hstojic: @uri-granta do you think this is ok or we should return `NotImplementedError`? for some problems pareto front will not be available

it's a bit burdensome, but what can be done is to compute offline approximate ground truth with say genetic algos with a large number of function evaluations, store a fixed number of points in the repo as a text file and then load that when the function is called

trieste/objectives/multi_objectives.py
```diff
@@ -38,7 +39,21 @@ def __call__(self, n: int, seed: int | None = None) -> TensorType:
         :param n: The number of pareto optimal points to be generated.
         :param seed: An integer used to create a random seed for distributions that
          used to generate pareto optimal points.
-        :return: The Pareto optimal points
+        :return: The Pareto optimal points.
+        """
+
+
+class Constraint(Protocol):
```
hstojic: @uri-granta we are introducing new interfaces here, it would be good if you could check them out?
they seem fine to me, but you'll know better

the only thing I can think of is that constraint functions could be in their own file as they could be used in other types of problems, not just multi objectives

trieste/objectives/multi_objectives.py
hstojic: indeed, you have covered 1. already, I have missed that, apologies

regarding 2. its quite simple, just add a test that the constraint output has correct dtype, e.g. you can modify `test_objective_and_constraint_has_correct_shape_and_dtype` by adding a check on constraint dtype if constraint is present

hstojic (commented):
thanks @TsingQAQ, added a comment on your question on my suggestion about typing
and I think we need @uri-granta input on interfaces

trieste/objectives/multi_objectives.py
uri-granta: Definitely clearer to  raise a `NotImplementedError` with a message explaining that there is no known pareto front, rather than returning an empty "front".

trieste/objectives/multi_objectives.py
uri-granta: Fine to leave it here for now I think, as we can easily move it "outwards" later if we want to.

Another question is the name: is this purporting to be a general interface for constraints everywhere, or is it specific to specifying test problems?

tests/unit/objectives/test_multi_objectives.py
```diff
@@ -137,33 +185,40 @@ def test_gen_pareto_front_is_equal_to_math_defined(
...
+        (DTLZ2(10, 5), 10, 5, 0),
+        (ConstrainedBraninCurrin(), 2, 2, 1),
     ],
 )
 @pytest.mark.parametrize("num_obs", [1, 5, 10])
 @pytest.mark.parametrize("dtype", [tf.float32, tf.float64])
-def test_objective_has_correct_shape_and_dtype(
-    problem: MultiObjectiveTestProblem,
+def test_objective_and_constraint_has_correct_shape_and_dtype(
+    problem: Union[MultiObjectiveTestProblem, ConstrainedMultiObjectiveTestProblem],
```
uri-granta: Is this change necessary given ConstrainedMultiObjectiveTestProblem is a subclass of MultiObjectiveTestProblem?

uri-granta (commented):
Looks fine to me!

trieste/objectives/multi_objectives.py
hstojic: right, I agree, lets use `ConstraintTestProblem` to be more consistent

tests/unit/objectives/test_multi_objectives.py
hstojic: I agree, probably not

trieste/objectives/multi_objectives.py
hstojic: @TsingQAQ can you please raise an error here, but perhaps in the docs you can indicate to user a way to generate approximate ground truth

TsingQAQ: Sure! for readability have created a new exception class [`NoAnalyticalParetoPointsError`](https://github.com/TsingQAQ/trieste/blob/e0ac5d45f649aef832badd6c7855fb829b92fd7c/trieste/objectives/multi_objectives.py#L32), please check if this is idea.

PR #693, opened 2023-02-10
swap out incorrect tf.repeats for tf.broadcast_tos

AlexGoodallAI:


hstojic:
@AlexGoodallAI are you happy with this PR, ready for the review?

AlexGoodallAI:
@hstojic Yeah, it's ready for review

hstojic (approved):
finally got around checking this, thanks for the patience :)

to me it seems `tf.repeat` was having the same effect as `tf.broadcast_to`, so not sure this is a bug really, but ok broadcast does seem safer as you explicitly provide the shape needed, rather than just point to the axis

regarding testing, I agree, it would amount to testing `broadcast_to`, what we could do perhaps is sprinkle asserts in few places to check that elements of the required dimension are the same, though it might be an expensive check - what do you think, worth it?

hstojic:
(you'll need to resolve conflicts before merging)

AlexGoodallAI:
Yes, I think you're correct, it doesn't actually make a difference here. I think I was mostly alarmed by `tf.repeat` not working in the way I thought it did. 
I don't have a strong opinion on adding asserts. I don't think asserting the shapes would have caught the bug I thought I had anyway, so I'm not sure it would contribute much to this specific concern, but could find unexpected broadcasts.

hstojic:
> Yes, I think you're correct, it doesn't actually make a difference here. I think I was mostly alarmed by `tf.repeat` not working in the way I thought it did. I don't have a strong opinion on adding asserts. I don't think asserting the shapes would have caught the bug I thought I had anyway, so I'm not sure it would contribute much to this specific concern, but could find unexpected broadcasts.

I was not thinking of asserting shapes but rather the content, because we know that certain dimensions should have exactly the same values, right?  I think these kind of checks can be rather expensive...

uri-granta:
Q: does this have a user-visible impact that should be mentioned in the release notes? (i.e. does it actually fix any broken behaviour?)

PR #695, opened 2023-02-14
Investigate broken docs build

uri-granta:
Turns out the latest setuptools is less tolerant than previous versions, exposing a bug in gym. Pin setuptools in tox for now.


hstojic (approved):
looks good

PR #699, opened 2023-02-16
Jumbo mumbo PR

uri-granta:


trieste/objectives/multifidelity_objectives.py
```diff
@@ -0,0 +1,95 @@
+from dataclasses import dataclass
```
AlexGoodallAI, line 17: Does this need a copyright statement?

trieste/objectives/multifidelity_objectives.py
```diff
@@ -0,0 +1,95 @@
...
+    5: tf.constant([-7.933019704], tf.float64),
+}
+
+
+def _linear_multifidelity_search_space_builder(
+    n_fidelities: int, input_search_space: SearchSpace
+) -> TaggedProductSearchSpace:
+
+    fidelity_search_space = DiscreteSearchSpace(
+        np.array([np.arange(n_fidelities, dtype=float)]).reshape(-1, 1)
```
AlexGoodallAI:
```suggestion
        np.arange(n_fidelities, dtype=float).reshape(-1, 1)
```

trieste/acquisition/function/entropy.py
```diff
@@ -617,3 +621,231 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        """
+        :param model: The model.
+        :param dataset: The data from the observer.
+        """
+        query_points = self._search_space.sample(num_samples=self._grid_size)
+        tf.debugging.assert_same_float_dtype([dataset.query_points, query_points])
+        query_points = tf.concat([dataset.query_points, query_points], 0)
+        query_points_on_top_fidelity = tf.concat(
+            [query_points[:, :-1], tf.ones_like(query_points[:, -1:])], -1
+        )
```
AlexGoodallAI: Couple of things here, I think we should use the function designed for this, and I think we have assumed 2 fidelities, 1 is not guaranteed to be the top fidelity.

```suggestion
        query_points_on_top_fidelity = add_fidelity_column(query_points[:, :-1], model.num_fidelities -1)
```

trieste/acquisition/function/entropy.py
```diff
@@ -617,3 +621,231 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        tf.debugging.assert_shapes(
+            [(x, [..., 1, None])],
+            message="This acquisition function only supports batch sizes of one.",
+        )
+
+        top_fidelity_idx = self._model.num_fidelities - 1
+        x_on_top_fidelity = tf.concat(
+            [tf.squeeze(x, -2)[:, :-1], top_fidelity_idx * tf.ones_like(tf.squeeze(x, -2)[:, -1:])],
+            -1,
+        )
```
AlexGoodallAI:
```suggestion
        x_on_top_fidelity = add_fidelity_column(tf.squeeze(x, -2)[:, :-1],  self._model.num_fidelities - 1)
```

trieste/acquisition/function/entropy.py
```diff
@@ -617,3 +621,231 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+        fmean, fvar = self._model.predict(x_on_top_fidelity)
+        fsd = tf.clip_by_value(
+            tf.math.sqrt(fvar), CLAMP_LB, fmean.dtype.max
+        )  # clip below to improve numerical stability
+        ymean, yvar = self._model.predict_y(tf.squeeze(x, -2))
+        cov = self._model.covariance_with_top_fidelity(tf.squeeze(x, -2))
```
AlexGoodallAI: Is it worth just doing this squeeze once and storing the output, rather than 3 times?

AlexGoodallAI (approved):
Looks good to me. I had a few minor comments and a question where I didn't really understand what was going on, but nothing major. I think there was also a plan to add a demo of this to the multifidelity tutorial, but this should probably be done in a new PR.

trieste/acquisition/function/entropy.py
```diff
@@ -617,3 +621,231 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+    @overload
+    def __init__(
+        self: "MUMBO[MUMBOModelType]",
+        search_space: SearchSpace,
+        num_samples: int = 5,
+        grid_size: int = 1000,
+        min_value_sampler: Optional[ThompsonSampler[MUMBOModelType]] = None,
+    ):
+        ...
```
AlexGoodallAI, line 664: I'm not sure I understand what this overloading does?

uri-granta: MUMBO is generic in the type of models it supports. This allows you to use it with a min value sampler that only supports certain types of model (second overload), as long as you only use MUMBO with those types of model. However, if you don't specify a min value sampler (first overload) then MUMBO supports any type of model that has covariance with top fidelity.

trieste/acquisition/function/entropy.py
uri-granta: Good catch!

uri-granta: So interestingly changing this to use `fidelity=model.num_fidelities-1` rather than `fidelity=1` causes `test_mumbo_builder_builds_min_value_samples` to fail on `assert max(min_value_samples) < min(fmean)`. Any idea what's going on?

AlexGoodallAI: I'm not sure, sorry

tests/unit/acquisition/function/test_entropy.py
```diff
@@ -519,3 +525,203 @@ def test_batch_gibbon_is_sum_of_individual_gibbons_and_repulsion_term(
...
+    builder.prepare_acquisition_function(model, dataset=dataset)
+    mocked_mves.assert_called_once()
+
+    # check that the Gumbel samples look sensible
+    min_value_samples = mocked_mves.call_args[0][1]
+    query_points = builder._search_space.sample(num_samples=builder._grid_size)
+    query_points = tf.concat([dataset.query_points, query_points], 0)
+    query_points = add_fidelity_column(query_points[:, :-1], model.num_fidelities - 1)
+    fmean, _ = model.predict(query_points)
+    assert max(min_value_samples) < min(fmean)
```
vpicheny, line 633: I don't see a theoretical reason why this should always be true... it is a bit of a weird thing to test but I guess it does the job of checking that `min_value_samples` returns small values?

A better test would have been to do `model.sample(query_points)` and check if the empirical distribution of the minimum of the samples resembles `min_value_samples`.

tests/unit/acquisition/function/test_entropy.py
```diff
@@ -519,3 +525,203 @@ def test_batch_gibbon_is_sum_of_individual_gibbons_and_repulsion_term(
...
+    )
+    builder.prepare_acquisition_function(model, dataset=dataset)
+    mocked_mves.assert_called_once()
+
+    # check that the Gumbel samples look sensible
+    min_value_samples = mocked_mves.call_args[0][1]
+    query_points = builder._search_space.sample(num_samples=builder._grid_size)
+    query_points = tf.concat([dataset.query_points, query_points], 0)
+    fmean, _ = model.predict(query_points)
+    assert max(min_value_samples) < min(fmean) + 1e-4
```
vpicheny, line 703: see above comment on asserting this vs something else

tests/unit/acquisition/function/test_entropy.py
```diff
@@ -519,3 +525,203 @@ def test_batch_gibbon_is_sum_of_individual_gibbons_and_repulsion_term(
...
+    search_space = Box([0.0, 0.0], [1.0, 1.0])
+    x_range = tf.linspace(0.0, 1.0, 5)
+    x_range = tf.cast(x_range, dtype=tf.float64)
+    xs = tf.reshape(tf.stack(tf.meshgrid(x_range, x_range, indexing="ij"), axis=-1), (-1, 2))
+    ys = quadratic(xs)
+    dataset = Dataset(xs, ys)
+    model = MultiFidelityQuadraticMeanAndRBFKernelWithSamplers(
+        dataset=dataset, noise_variance=tf.constant(1e-10, dtype=tf.float64)
+    )
+    model.kernel = (
```
vpicheny, line 688: why is this not part of `MultiFidelityQuadraticMeanAndRBFKernelWithSamplers`?

vpicheny (approved):
Looks good to me from what I can tell - would be nice to reinforce this by extending the multifi notebook (would act as an additional integration test) in a following PR?

PR #700, opened 2023-02-18
Fix typo in builders.py

eltociear:
propogate -> propagate


hstojic (approved):
Thank you @eltociear for the PR, welcome to Trieste :)

PR #701, opened 2023-02-21
Unpin typing extensions — and hence black and mypy

uri-granta:


trieste/models/interfaces.py
```diff
@@ -86,6 +86,7 @@ def predict_y(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
             ``query_points``. For a predictive distribution with event shape E, the mean and
             variance will both have shape [...] + E.
         """
+        pass  # (required so that mypy doesn't think this method is abstract)
```
hstojic, line 89: aaah, I was getting these errors - I was wondering whats going on with mypy, nice that this resolves it!

hstojic (approved):
looks good

PR #706, opened 2023-03-03
Add redirects for tutorials and autoapi

uri-granta:


hstojic (approved):
that's great, thanks a lot!

PR #708, opened 2023-03-09
Use QMC in reparam sampler

uri-granta:
Update IndependentReparametrizationSampler to support QMC.

### DIFF
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -767,30 +767,12 @@ def test_rff_and_decoupled_trajectory_give_similar_results(noise_var: float) -> 
     )  # variance across samples should (very) roughly agree for different samplers
 
 
-@pytest.mark.parametrize(
-    ("batch_shape", "n_sample_dim"),
-    [
-        ([5], 1),
-        ([5], 2),
-        ([5, 5], 2),
-        ([1, 2, 3, 4], 5),
-    ],
-)
-def test_qmc_normal_samples__various_shapes(batch_shape: list[int], n_sample_dim: int) -> None:
-    samples = qmc_normal_samples(tf.TensorShape(batch_shape), n_sample_dim)
-    assert samples.shape == batch_shape + [
-        n_sample_dim,
-    ]
-
-
 @pytest.mark.parametrize("n_sample_dim", [2, 5])
 @pytest.mark.parametrize("skip", [0, 10_000])
 def test_qmc_samples_return_standard_normal_samples(n_sample_dim: int, skip: int) -> None:
     n_samples = 10_000
 
-    qmc_samples = qmc_normal_samples(
-        batch_shape=tf.TensorShape([n_samples]), n_sample_dim=n_sample_dim, skip=skip
-    )
+    qmc_samples = qmc_normal_samples(num_samples=n_samples, n_sample_dim=n_sample_dim, skip=skip)
 
     # should be multivariate normal with zero correlation
     for i in range(n_sample_dim):
@@ -801,10 +783,10 @@ def test_qmc_samples_return_standard_normal_samples(n_sample_dim: int, skip: int
 
 
 def test_qmc_samples_skip() -> None:
-    samples_1a = qmc_normal_samples(tf.TensorShape([25]), 100)
-    samples_1b = qmc_normal_samples(tf.TensorShape([25]), 100)
+    samples_1a = qmc_normal_samples(25, 100)
+    samples_1b = qmc_normal_samples(25, 100)
     npt.assert_allclose(samples_1a, samples_1b)
-    samples_2a = qmc_normal_samples(tf.TensorShape([25]), 100, skip=100)
-    samples_2b = qmc_normal_samples(tf.TensorShape([25]), 100, skip=100)
+    samples_2a = qmc_normal_samples(25, 100, skip=100)
+    samples_2b = qmc_normal_samples(25, 100, skip=100)
     npt.assert_allclose(samples_2a, samples_2b)
     npt.assert_raises(AssertionError, npt.assert_allclose, samples_1a, samples_2a)
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -39,7 +39,7 @@ from trieste.models.gpflow import (
     RandomFourierFeatureTrajectorySampler,
     feature_decomposition_trajectory,
 )
-from trieste.models.gpflow.sampler import idealised_normal_samples
+from trieste.models.gpflow.sampler import qmc_normal_samples
 from trieste.models.interfaces import ReparametrizationSampler, SupportsPredictJoint
 from trieste.objectives import Branin
 
@@ -734,29 +734,23 @@ def test_rff_and_decoupled_trajectory_give_similar_results(noise_var: float) -> 
         ([1, 2, 3, 4], 5),
     ],
 )
-def test_idealised_normal_samples__various_shapes(
-    batch_shape: list[int], n_sample_dim: int
-) -> None:
-    samples = idealised_normal_samples(tf.TensorShape(batch_shape), n_sample_dim)
+def test_qmc_normal_samples__various_shapes(batch_shape: list[int], n_sample_dim: int) -> None:
+    samples = qmc_normal_samples(tf.TensorShape(batch_shape), n_sample_dim)
     assert samples.shape == batch_shape + [
         n_sample_dim,
     ]
 
 
-def test_kl_divergence_normal_samples() -> None:
+def test_qmc_samples_return_standard_normal_samples() -> None:
     n_samples = 10_000
 
-    idealised_samples = idealised_normal_samples(
-        batch_shape=tf.TensorShape([n_samples]), n_sample_dim=2
-    )
+    qmc_samples = qmc_normal_samples(batch_shape=tf.TensorShape([n_samples]), n_sample_dim=2)
 
     multivariate_gaussian_dist = MultivariateNormalDiag(scale_diag=[1.0, 1.0])
     random_samples = multivariate_gaussian_dist.sample(sample_shape=n_samples)
 
     bins = [np.linspace(-4, 4, 9), np.linspace(-4, 4, 9)]
-    idealised_sample_counts = np.histogram2d(
-        idealised_samples[:, 0], idealised_samples[:, 1], bins=bins
-    )[0]
+    qmc_sample_counts = np.histogram2d(qmc_samples[:, 0], qmc_samples[:, 1], bins=bins)[0]
     random_sample_counts = np.histogram2d(random_samples[:, 0], random_samples[:, 1], bins=bins)[0]
 
     def compute_kl_divergence(
@@ -764,7 +758,5 @@ def test_kl_divergence_normal_samples() -> None:
     ) -> "np.ndarray[Any, Any]":
         return np.sum(np.where(np.logical_and(a != 0, b != 0), a * np.log(a / b), 0))
 
-    kl_div = compute_kl_divergence(
-        idealised_sample_counts / n_samples, random_sample_counts / n_samples
-    )
+    kl_div = compute_kl_divergence(qmc_sample_counts / n_samples, random_sample_counts / n_samples)
     assert kl_div < 0.005
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -14,7 +14,7 @@
 from __future__ import annotations
 
 import math
-from typing import List, Type
+from typing import Any, List, Type
 
 import gpflow
 import numpy as np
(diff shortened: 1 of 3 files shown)


trieste/models/gpflow/sampler.py
```diff
@@ -91,11 +113,23 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
+        if self._eps is None:
+            self._eps = tf.Variable(sample_eps())
+
+        tf.cond(
+            self._initialized,
+            lambda: self._eps,
+            lambda: self._eps.assign(sample_eps()),
+        )
 
         return mean + tf.sqrt(var) * tf.cast(self._eps[:, None, :], var.dtype)  # [..., S, 1, L]
```
vpicheny, line 136: should this be `self._eps[..., None, :]`?

tests/unit/models/gpflow/test_sampler.py
```diff
@@ -719,3 +723,48 @@ def test_rff_and_decoupled_trajectory_give_similar_results(noise_var: float) ->
...
+def test_idealised_normal_samples__various_shapes(
+    batch_shape: list[int], n_sample_dim: int
+) -> None:
+    samples = idealised_normal_samples(tf.TensorShape(batch_shape), n_sample_dim)
+    assert samples.shape == batch_shape + [
+        n_sample_dim,
+    ]
+
+
+def test_kl_divergence_normal_samples() -> None:
```
vpicheny: change name to `test_idealised_samples_return_standard_normal_samples` ? (took me a little while to get what this was testing)

tests/unit/models/gpflow/test_sampler.py
```diff
@@ -719,3 +723,48 @@ def test_rff_and_decoupled_trajectory_give_similar_results(noise_var: float) ->
...
+    multivariate_gaussian_dist = MultivariateNormalDiag(scale_diag=[1.0, 1.0])
+    random_samples = multivariate_gaussian_dist.sample(sample_shape=n_samples)
+
+    bins = [np.linspace(-4, 4, 9), np.linspace(-4, 4, 9)]
+    idealised_sample_counts = np.histogram2d(
+        idealised_samples[:, 0], idealised_samples[:, 1], bins=bins
+    )[0]
+    random_sample_counts = np.histogram2d(random_samples[:, 0], random_samples[:, 1], bins=bins)[0]
+
+    def compute_kl_divergence(
```
vpicheny: this is a bit of a weird test, if you want to see if a sample comes from a particular distribution there are established tests for that, i.e. 
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html 
(see examples)

trieste/models/gpflow/sampler.py
```diff
@@ -44,6 +44,30 @@
 )
 
 
+def idealised_normal_samples(
+    batch_shape: tf.TensorShape, n_sample_dim: int, skip: int = 0
```
vpicheny: we should probably test that skip is working?

trieste/models/gpflow/sampler.py
```diff
@@ -91,11 +113,23 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
 
-        if not self._initialized:
-            self._eps.assign(
-                tf.random.normal([self._sample_size, tf.shape(mean)[-1]], dtype=tf.float64)
-            )  # [S, L]
+        def sample_eps() -> tf.Tensor:
             self._initialized.assign(True)
+            skip = IndependentReparametrizationSampler.skip
+            IndependentReparametrizationSampler.skip.assign(skip + self._sample_size)
+            normal_samples = idealised_normal_samples(
```
vpicheny: Should we have the possibility to revert to regular sampling rather than QMC? as in keep it as default but if asked explicitely revert to basic MC.

Might be OK it ignore if it's complicated.

trieste/models/interfaces.py
```diff
@@ -541,10 +541,12 @@ def __repr__(self) -> str:
...
         """
-        :param at: Input points that define the sampler of shape `[N, D]`.
-        :param jitter: The size of the jitter to use when stabilizing the Cholesky
-            decomposition of the covariance matrix.
-        :return: Samples of shape `[sample_size, D]`.
+        :param at: Where to sample the predictive distribution, with shape `[..., 1, D]`, for points
+            of dimension `D`.
+        :param jitter: The size of the jitter to use when stabilising the Cholesky decomposition of
+            the covariance matrix.
+        :return: The samples, of shape `[..., S, 1, L]`, where `S` is the `sample_size` and `L` is
```
vpicheny: I'm not sure about this. What if you use a `BatchReparametrizationSampler`? then you would get `[..., S, B, L]`, right?

vpicheny:
> Q: should we also update BatchReparametrizationSampler to used idealised_normal_samples? (A quick attempt resulted in test_batch_reparametrization_sampler_samples_approximate_mean_and_covariance no longer passing but I haven't looked in detail yet.)

Yes I think it would make sense to change it as well. I don't see why it would fail.

trieste/models/gpflow/sampler.py
```diff
@@ -44,6 +44,30 @@
...
 
 
+def idealised_normal_samples(
+    batch_shape: tf.TensorShape, n_sample_dim: int, skip: int = 0
+) -> tf.Tensor:
+    """
+    Generates a batch of shape `batch_shape` sobol samples, skipping the first `skip`, where each
+    sample has dimension `n_sample_dim`.
+    """
+    sobol_samples = tf.math.sobol_sample(
```
vpicheny, line 52: I'm actually having second thoughts about all of this... we're using a space-filling sequence, but what "space" are we filling?

If you have n_sample_dim = `[S, 1]` then you're basically filling a 1d space? which I guess is OK

If you have n_sample_dim = `[S, B]` then your Sobol sequence should be defined over B dimensions.

But if you have n_sample_dim = `[N, S, B, L]` ??

trieste/models/gpflow/sampler.py
```diff
@@ -44,6 +44,30 @@
...
+        dtype=tf.float64,
+        skip=skip,
+    )
+
+    dist = tfp.distributions.Normal(
+        loc=tf.constant(0.0, dtype=tf.float64),
+        scale=tf.constant(1.0, dtype=tf.float64),
+    )
+    samples_shape = batch_shape + (n_sample_dim,)
+    normal_samples = tf.reshape(dist.quantile(sobol_samples), samples_shape)
```
hstojic: if we needed uniform random numbers then Sobol does the trick, but does this with quantile transform really works? the goal is to get normally distributed numbers that are more regular, analgous to the usual usage of Sobol sequence, right?

hstojic: would this work for `n_sample_dim>1`?

trieste/models/interfaces.py
uri-granta: Yes, that's a typo. I meant to say B (the key thing being to note that there's an explicit correlated batch dimension).

tests/unit/models/gpflow/test_sampler.py
uri-granta: This test was copied verbatim! Happy to change it to use kstest though I mean need help getting it to work right.

trieste/models/gpflow/sampler.py
uri-granta: I don't think so? `self._eps` is still shape `[S, L]` without any leading dimensions.

trieste/models/gpflow/sampler.py
uri-granta: @vpicheny @ChrisMorter you able to comment on this?

trieste/models/gpflow/sampler.py
uri-granta: Pity I noticed this comment last :-) Happy to discuss offline, but options are:

1. Make this behaviour the default
2. Make this behaviour optional so people can use it when appropriate
3. Leave this out of trieste

trieste/models/gpflow/sampler.py
hstojic: gave it some further thought and tested it, it does work all right, for `n_sample_dim>1` as well

uri-granta: Thanks. I might try to update the test to test with sample_dim = 2 then.

vpicheny: I think testing with larger dimensions is important. I would actually test  with dim>2 and leading dimensions

trieste/models/gpflow/sampler.py
vpicheny: you're right, sorry

uri-granta:
> > Q: should we also update BatchReparametrizationSampler to used idealised_normal_samples? (A quick attempt resulted in test_batch_reparametrization_sampler_samples_approximate_mean_and_covariance no longer passing but I haven't looked in detail yet.)
> 
> Yes I think it would make sense to change it as well. I don't see why it would fail.

Can you take a look at what I've added? The sample mean is coming out different from the model mean, suggesting I've made a mistake somewhere.

trieste/models/gpflow/sampler.py
```diff
@@ -170,13 +212,26 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
 
-        if not self._initialized:
-            self._eps.assign(
-                tf.random.normal(
+        def sample_eps() -> tf.Tensor:
+            self._initialized.assign(True)
+            if self._qmc:
+                skip = IndependentReparametrizationSampler.skip
+                IndependentReparametrizationSampler.skip.assign(skip + self._sample_size)
+                normal_samples = qmc_normal_samples(
```
vpicheny, line 217: I think this is where the issue is. Fundamentally you want the Sobol sequence to be of dimension `B = batch_size`. That is, each sample will be a multi-output Gaussian of dimension `batch_size` and you want to be space-filling in this space. 

Then you want `S = self._sample_size` samples = number of iterates from the Sobol sequence.

Finally there is the number of latents, `L = mean.shape[-1]`. 

Currently this code calls: `tf.math.sobol_sample(dim=[L], num_results=B*S)`.

I think what you want is `tf.math.sobol_sample(dim=[B, L], num_results=S)`.

However sobol doesn't work with more than 1 dimension. So what we should do is:

`tf.math.sobol_sample(dim=[B], num_results=S*L)`

then reshape to obtain `[L, B, S]` shape.

trieste/models/gpflow/sampler.py
```diff
@@ -91,11 +116,25 @@ def sample(self, at: TensorType, *, jitter: float = DEFAULTS.JITTER) -> TensorTy
...
-            self._eps.assign(
-                tf.random.normal([self._sample_size, tf.shape(mean)[-1]], dtype=tf.float64)
-            )  # [S, L]
+        def sample_eps() -> tf.Tensor:
             self._initialized.assign(True)
+            if self._qmc:
+                skip = IndependentReparametrizationSampler.skip
+                IndependentReparametrizationSampler.skip.assign(skip + self._sample_size)
+                normal_samples = qmc_normal_samples(
+                    tf.TensorShape(self._sample_size), mean.shape[-1], skip
```
vpicheny: Now I understand what this is doing, it creates a space-filling design in the space the dimension of the number of latents. In most cases this will just be a 1d sequence (single latent), in other cases it might be mildly beneficial to have this space-filling structure in the latent space, i.e. with 2 latents you would maximise the diversity of the coupled values of `f1` and `f2`.

trieste/models/gpflow/sampler.py
```diff
@@ -44,6 +44,28 @@
...
 
 
+def qmc_normal_samples(batch_shape: tf.TensorShape, n_sample_dim: int, skip: int = 0) -> tf.Tensor:
+    """
+    Generates a batch of shape `batch_shape` sobol samples, skipping the first `skip`, where each
+    sample has dimension `n_sample_dim`.
+    """
+    sobol_samples = tf.math.sobol_sample(
+        dim=n_sample_dim,
+        num_results=tf.reduce_prod(batch_shape),
```
vpicheny: Do we have any reason now to allow for a batch_shape that is not just a single value? in the batch case you do the product outside anyway and in the independent case batch_shape is a single integer.

uri-granta: I guess not! I'll change it from `batch_shape` to `num_samples: int`, which should simplify it a bit.

vpicheny (approved):

trieste/models/gpflow/sampler.py
hstojic: @uri-granta I agree with @vpicheny above on `tf.math.sobol_sample(dim=[B], num_results=S*L)`, but what is implemented below is not correct, we have at the moment `tf.math.sobol_sample(dim=[B*L], num_results=S)`, right?

uri-granta: You're totally right, oops! Here's a fix PR: https://github.com/secondmind-labs/trieste/pull/713

PR #709, opened 2023-03-10
Fix SVGP update equations

### LINKED ISSUE #707: SVGP update equations contain an error
When we change the inducing points before re-training SVGPs and Deep GPs, we need to update q_mu and q_sqrt such that the posterior distribution is unchanged at the new inducing points. 

The current equations only work when the mean function is zero. We need to add a correction term at those places;

(new_q_mu =new_q_mu - model mean prediction at Z)
https://github.com/secondmind-labs/trieste/blob/develop/trieste/models/gpflow/models.py#L1014

(f_mu = f_mu - model mean prediction at Z)
https://github.com/secondmind-labs/trieste/blob/a4b25f8cde453c2a19399e66c3c440d51cfbac84/trieste/models/gpflow/utils.py#L330

DeepGPs currently do not have this problem because we don't have a scheme to update inducing points.

It is unclear whereas we have a test for whiten_points, but we should definitely have one with a non-zero mean function.

vpicheny:
This PR solves [this issue](https://github.com/secondmind-labs/trieste/issues/707):

When we change the inducing points before re-training SVGPs and Deep GPs, we need to update q_mu and q_sqrt such that the posterior distribution is unchanged at the new inducing points. The previous equations only work when the mean function is zero. 

This PR fixes the equations and adds a test to check that equations are correct, by verifying that the posterior distributions at the new inducing points remains unchanged after the model is updated.

closes #707

### DIFF
--- a/tests/unit/models/gpflow/test_models.py
+++ b/tests/unit/models/gpflow/test_models.py
@@ -1655,7 +1655,7 @@ def test_sparse_variational_inducing_updates_preserves_posterior(
 
     old_mu, old_sqrt = model.model.predict_f(xnew, full_cov=True)  # predict old posterior
 
-    model.update(Dataset(x, y1))
+    model.update(Dataset(x, y1))  # this changes inducing points to xnew
 
     npt.assert_raises(
         AssertionError,
--- a/tests/unit/models/gpflow/test_models.py
+++ b/tests/unit/models/gpflow/test_models.py
@@ -1651,8 +1651,6 @@ def test_sparse_variational_inducing_updates_preserves_posterior(
         inducing_point_selector=inducing_point_selector,
     )
 
-    model.optimize(Dataset(x, y1))
-
     np.testing.assert_array_equal(model.model.inducing_variable.Z, x[:num_inducing_points])
 
     old_mu, old_sqrt = model.predict_joint(xnew)  # predict old posterior
--- a/tests/unit/models/gpflow/test_models.py
+++ b/tests/unit/models/gpflow/test_models.py
@@ -1655,16 +1655,16 @@ def test_sparse_variational_inducing_updates_preserves_posterior(
 
     np.testing.assert_array_equal(model.model.inducing_variable.Z, x[:num_inducing_points])
 
-    old_mu, old_sqrt = model.model.predict_f(xnew, full_cov=True)  # predict old posterior
+    old_mu, old_sqrt = model.predict_joint(xnew)  # predict old posterior
 
     model.update(Dataset(x, y1))  # this changes inducing points to xnew
 
     np.testing.assert_array_equal(model.model.inducing_variable.Z, xnew)
 
-    new_mu, new_sqrt = model.model.predict_f(xnew, full_cov=True)  # predict new posterior
+    new_mu, new_sqrt = model.predict_joint(xnew)  # predict new posterior
 
     np.testing.assert_allclose(old_mu, new_mu, atol=1e-4)
-    np.testing.assert_allclose(old_sqrt, new_sqrt, atol=1e-4)
+    np.testing.assert_allclose(old_sqrt, new_sqrt, atol=1.1e-4, rtol=1e-4)
 
 
 def multifidelity_autoregressive_nd_dataset(n_dims: int = 1) -> Dataset:
--- a/tests/unit/models/gpflow/test_models.py
+++ b/tests/unit/models/gpflow/test_models.py
@@ -48,6 +48,7 @@ from tests.util.models.gpflow.models import (
     mock_data,
     sgpr_model,
     svgp_model,
+    svgp_model_with_mean,
     two_output_svgp_model,
     vgp_matern_model,
     vgp_model,
@@ -55,7 +56,7 @@ from tests.util.models.gpflow.models import (
 from tests.util.models.models import fnc_2sin_x_over_3, fnc_3x_plus_10
 from trieste.data import Dataset, add_fidelity_column
 from trieste.logging import step_number, tensorboard_writer
-from trieste.models import TrainableProbabilisticModel
+from trieste.models import ProbabilisticModelType, TrainableProbabilisticModel
 from trieste.models.gpflow import (
     GaussianProcessRegression,
     MultifidelityAutoregressive,
@@ -1599,8 +1600,7 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
     y1 = fnc_3x_plus_10(x)
     y2 = y1 * 0.5
 
-    svgp = two_output_svgp_model(x, mo_type, whiten)
-
+    svgp = svgp_model(x, mo_type, whiten)
     model = SparseVariational(svgp, BatchOptimizer(tf.optimizers.Adam(), max_iter=3, batch_size=10))
     model.model.whiten = whiten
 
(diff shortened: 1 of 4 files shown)


trieste/models/gpflow/utils.py
```diff
@@ -327,7 +327,10 @@ def _whiten_points(
     :para inducing_points: The new inducing point locations.
     :return: The updated q_mu and q_sqrt with shapes [N, L] and [L, N, N], respectively.
     """
+
+    old_mean = model.model.mean_function(inducing_points)
```
uri-granta: Any reason to use a temp variable here but not in the other fix?

tests/unit/models/gpflow/test_models.py
```diff
@@ -1618,6 +1619,57 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
...
+        svgp,
+        BatchOptimizer(tf.optimizers.Adam(), max_iter=3, batch_size=10),
+        inducing_point_selector=inducing_point_selector,
+    )
+
+    model.optimize(Dataset(x, y1))
+
+    old_mu, old_sqrt = model.model.predict_f(xnew, full_cov=True)  # predict old posterior
+
+    model.update(Dataset(x, y1))
```
uri-granta: (so this is changing the inducing points from x[:4] to xnew?)

uri-granta (approved):
LGTM!

tests/unit/models/gpflow/test_models.py
vpicheny: indeed, should I add a comment to make it explicit? below I also make sure that the points have actually changed.

trieste/models/gpflow/utils.py
vpicheny: no, that's a good point, will remove

tests/unit/models/gpflow/test_models.py
alansaul: Can you put an assertion *before* the update as well, testing the opposite (that they *were* equal to `x` and not `xnew`). I.e. checking that the inducing points of the model have indeed changed, yet the posterior predictions at those points are the same.

tests/unit/models/gpflow/test_models.py
```diff
@@ -1618,6 +1619,57 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
...
+    model = SparseVariational(
+        svgp,
+        BatchOptimizer(tf.optimizers.Adam(), max_iter=3, batch_size=10),
+        inducing_point_selector=inducing_point_selector,
+    )
+
+    model.optimize(Dataset(x, y1))
+
+    old_mu, old_sqrt = model.model.predict_f(xnew, full_cov=True)  # predict old posterior
+
```
alansaul, line 1664: Can you put an assertion before the update as well, testing the opposite (that they were equal to `x` and not `xnew`). I.e. checking that the inducing points of the model have indeed changed, yet the posterior predictions at those points are the same; I think this would make the test more explicit (and I guess the comment becomes a bit redundant, but you can keep it in).

tests/unit/models/gpflow/test_models.py
```diff
@@ -1618,6 +1619,57 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
...
+    svgp = svgp_model_with_mean(x, y1, whiten, num_inducing_points)
+    inducing_point_selector = DummyInducingPointSelector(xnew)
+    model = SparseVariational(
+        svgp,
+        BatchOptimizer(tf.optimizers.Adam(), max_iter=3, batch_size=10),
+        inducing_point_selector=inducing_point_selector,
+    )
+
+    model.optimize(Dataset(x, y1))
```
alansaul: I wonder if it might be sensible to disable training of the `Linear` mean function, as we don't care about the quality of the fit, but we do need some mean function contribution to check that this PR is fixing an issue with the mean function wrongfully not being removed..

alansaul (approved):
LGTM, just make sure that the test fails before the fixes were made. A couple of changes to make it perhaps a bit more robust and clear.

tests/unit/models/gpflow/test_models.py
vpicheny: assert added

tests/unit/models/gpflow/test_models.py
vpicheny: I manually checked that the Linear contribution doesn't go to zero and the tests would fail without the fix

vpicheny: OK changed my mind so this is deterministic now

tests/unit/models/gpflow/test_models.py
```diff
@@ -1618,6 +1619,52 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
...
+    old_mu, old_sqrt = model.predict_joint(xnew)  # predict old posterior
+
+    model.update(Dataset(x, y1))  # this changes inducing points to xnew
+
+    np.testing.assert_array_equal(model.model.inducing_variable.Z, xnew)
+
+    new_mu, new_sqrt = model.predict_joint(xnew)  # predict new posterior
+
+    np.testing.assert_allclose(old_mu, new_mu, atol=1e-4)
+    np.testing.assert_allclose(old_sqrt, new_sqrt, atol=1.1e-4, rtol=1e-4)
```
vpicheny: I had to make the tolerance higher to make it pass. Not sure if I should be concerned? There is a additional jitter term by removing it doesn't change anything. There is also some cropping in predict_joint but removing it doesn't change anything either.... @alansaul ?

trieste/models/gpflow/models.py
```diff
@@ -973,6 +972,7 @@ def update(self, dataset: Dataset) -> None:
                 )
             ):  # only bother updating if points actually change
                 self._update_inducing_variables(new_inducing_points)
+                self.update_posterior_cache()
```
vpicheny, line 976: @uri-granta this is not a minor change, previously update didn't change the posterior cache correctly. Didn't matter as long as we called optimize just after update. But unsure if there are other consequences...

tests/unit/models/gpflow/test_models.py
```diff
@@ -1618,6 +1619,60 @@ def test_sparse_variational_pairwise_covariance_for_non_whitened(
...
+@pytest.mark.parametrize("whiten", [False, True])
+def test_sparse_variational_inducing_updates_preserves_posterior(
+    whiten: bool,
+) -> None:
+    from gpflow.config import Config, as_context
+
+    from trieste.utils import DEFAULTS
+
+    default_jitter = 0.0
+    DEFAULTS.JITTER = default_jitter
```
uri-granta: The correct way to do this is to temporarily patch DEFAULTS (that way DEFAULTS.JITTER will be reverted to its original value after the test!)

```
    with unittest.mock.patch.object(DEFAULTS, "JITTER", default_jitter):
        ....
```

vpicheny: @uri-granta this line works fine on my machine but fails otherwise... any clue?

trieste/models/gpflow/models.py
uri-granta: Shouldn't this be run even if `self._inducing_point_selector` is None (like in `SparseGaussianProcessRegression`)?

Regarding testing, is the issue that `test_gpflow_models_cached_predictions_correct` and the other posterior cache tests don't use a `SparseVariational` with an `inducing_point_selector`? If so, it might be useful to add a simple test that checks both this and `SparseGaussianProcessRegression`.

tests/unit/models/gpflow/test_models.py
uri-granta: That's actually encouraging! (we shouldn't be able to override the DEFAULTS accidentally) See my other comment for how to temporarily change them.

trieste/models/gpflow/models.py
vpicheny: I am not sure that it is necessary if you don't change your inducing points, in that case the only change is the data size which has no impact on the posterior.

We could play it safe and run it anyway, but I'm a bit unsure how much overhead this adds.

PR #711, opened 2023-03-15
SeparateIndependent kernel RFF

sebastianober:
Draft PR for SeparateIndependent kernel RFF.

TODO:
 - clean code
 - write tests
 - better shapes


sebastianober:
Note: this will still rely on the updates to GPflux to come through, but tests are passing locally

trieste/models/gpflux/sampler.py
```diff
@@ -237,20 +237,18 @@ def __init__(
                 f"{type(layer)}"
             )
 
+        if isinstance(
+            layer.inducing_variable, gpflow.inducing_variables.SeparateIndependentInducingVariables
```
hstojic, line 241: why it doesn't work in this case?

trieste/models/gpflux/sampler.py
```diff
@@ -337,7 +343,12 @@ def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [B,
 
         q_mu = self._layer.q_mu  # [M, P]
         q_sqrt = self._layer.q_sqrt  # [P, M, M]
-        Kmm = self._kernel.K(inducing_points, inducing_points)  # [M, M]
+        if self._feature_functions.is_multioutput:
+            Kmm = self._kernel.K(
+                inducing_points, inducing_points, full_output_cov=False
+            )  # [P, M, M]
+        else:
+            Kmm = self._kernel.K(inducing_points, inducing_points)
```
hstojic: [M, M]?

trieste/models/gpflux/sampler.py
```diff
@@ -337,7 +343,12 @@ def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [B,
...
         Kmm += tf.eye(tf.shape(inducing_points)[0], dtype=Kmm.dtype) * DEFAULTS.JITTER
```
hstojic, line 352: would this operation on Kmm and those below work fine for both shapes of Kmm?

trieste/models/gpflux/sampler.py
hstojic: perhaps better to make it [1,M,M] and then shape check and all other operations with Kmm below would be safer?

trieste/utils/misc.py
```diff
@@ -249,12 +249,11 @@ def flatten_leading_dims(x: TensorType) -> Tuple[TensorType, Callable[[TensorTyp
     x_flat_shape = tf.concat([[-1], input_shape], axis=0)
 
     def unflatten(y: TensorType) -> TensorType:
-        tf.debugging.assert_rank(y, 2, message="unflatten is expecting a rank two tensor.")
```
hstojic, line 252: hm, are we sure we can remove this, I would assume it was here for a reason?

hstojic: If you think this is safe to do, we would need to expand the tests of (un)flatten

tests/util/models/gpflux/models.py
```diff
@@ -89,6 +90,44 @@ def simple_two_layer_dgp_model(x: TensorType) -> DeepGP:
     return DeepGP([gp_layer_1, gp_layer_2], gpflow.likelihoods.Gaussian(0.01))
 
 
+def separate_independent_kernel_two_layer_dgp_model(x: TensorType) -> DeepGP:
```
hstojic, line 91: perhaps we should have another builder function in `gplux/builders.py`?

hstojic (commented):
looks good, left a few comments

trieste/models/gpflux/sampler.py
sebastianober: Yes, because it will always broadcast to the last two dimensions

tests/util/models/gpflux/models.py
```diff
@@ -89,6 +90,44 @@ def simple_two_layer_dgp_model(x: TensorType) -> DeepGP:
...
+    if isinstance(x, tf.Tensor):
```
vpicheny: you can use https://github.com/secondmind-labs/trieste/blob/develop/trieste/utils/misc.py#L55

hstojic:
@sebastianober I have merged in latest develop with gpflux 4.0 and now all tests are passing - did you want to do anything else here?
did you manage to look into separate independent inducing points?

hstojic (approved):
good to go

PR #713, opened 2023-03-19
Fix Batch QMC sampling

uri-granta:


hstojic (approved):
now looks good :)
does the addition to the test catches the issue?

uri-granta:
> now looks good :) does the addition to the test catches the issue?

Yes: it checks that we call the sampler with the right dimensions.

trieste/models/gpflow/sampler.py
```diff
@@ -215,12 +215,12 @@ def sample_eps() -> tf.Tensor:
...
                 normal_samples = qmc_normal_samples(
-                    self._sample_size, batch_size * mean.shape[-1], skip
-                )  # [S, B*L]
-                normal_samples = tf.transpose(normal_samples)  # [B*L, S]
+                    self._sample_size * mean.shape[-1], batch_size, skip
+                )  # [S*L, B]
                 normal_samples = tf.reshape(
-                    normal_samples, (mean.shape[-1], batch_size, self._sample_size)
-                )  # [L, B, S]
+                    normal_samples, (mean.shape[-1], self._sample_size, batch_size)
```
vpicheny, line 221: naive question, we need the 'B' dimension to be left untouched, are you confident that reshape only messes with the leading dimensions?

uri-granta: In this case, yes: the reshape will simply group the S*L samples of dimension B into L groups each containing S of those samples.

vpicheny (approved):

PR #715, opened 2023-03-22
Support disabling skip in QMC for XLA

uri-granta:
Using the skip parameter to reset QMC samplers isn't supported with XLA, so add an option to disable this (at the expense of not being able to reset samplers).


hstojic (approved):
looks good

PR #718, opened 2023-03-27
Make IndependentReparametrizationSampler more XLA friendly

### LINKED ISSUE #722: Don't decorate acquisition function call methods with tf.function
Trieste's current approach to avoiding retracing is to decorate the `__call__` method of acquisition function classes with `tf.function`. However, this has two negative side effects:

1. It results in early compilation, restricting use of lazy initialisation of variables
2. It keeps compiled graphs in a global cache that accumulates over time (e.g. during the test runs)

To fix this we should move the use of tf.functions further out. Note that this will probably be a significant amount of work.

Following this we can also convert some variables from dynamically shaped to lazy initialised (e.g. `self._eps`).

uri-granta:
Make IndependentReparametrizationSampler support XLA (except when using qmc with qmc_skip set to True).


sam-willis (approved):
LGTM

PR #721, opened 2023-03-29
notebook file sizes fix

### LINKED ISSUE #717: Feasible sets tutorial file too big
**Describe the bug**
The feasible sets notebooks generated for the docs are 28MB each, which is adding up quickly over time. We should make these smaller, and probably backport the change to reduce the repo size.

**To reproduce**
1. See the feasible_sets.html file in the gh-pages-history branch

hstojic:
this PR will close #717 

the culprit are 3D plotly figures, they bloat the html files substantially, matplotlib is fine - I have reduced grid sizes there, that helps but in general we should minimise usage of 3d plots until we start storing files elsewhere
 
I have also :
- made figure sizing consistent and nicer to look at
- added missing quickrun


trieste/experimental/plotting/plotting_plotly.py
```diff
@@ -202,14 +202,16 @@ def plot_model_predictions_plotly(
         fig = add_surface_plotly(xx, yy, lcb, fig, alpha=alpha - 0.35, figrow=1, figcol=k + 1)
         fig = add_surface_plotly(xx, yy, ucb, fig, alpha=alpha - 0.35, figrow=1, figcol=k + 1)
 
+    fig.update_layout(height=600, width=600)
```
uri-granta, line 205: Is your purpose here to ensure all the plotly plots are consistently sized?

Maybe still add height and width as optional parameters to the function so there are no magic numbers?

docs/notebooks/quickrun/explicit_constraints.yaml
```diff
@@ -0,0 +1,2 @@
+replace:
+  - { from: "num_steps = \\d+", to: "num_steps = 2" }
```
uri-granta, line 2: :+1:

uri-granta (approved):
Looks good, thanks! I built the docs locally and they looked fine :-)

trieste/experimental/plotting/plotting_plotly.py
hstojic: correct... this can be easily overriden by using the same command on the object produced by the function, so I think there is no need for that

PR #723, opened 2023-04-05
Allow qmc_normal_samples to be called with tensor parameter values

ChrisMorter:
Allow qmc_normal_samples to be called with tensor parameter values, this is useful when trying to call it in the context of a tf function, where it's more natural to use tensors than ints.

The only thing preventing this being possible already was the shape assertion. I'd argue that that is something which we can, and should, be able to verify using unit tests rather than runtime checks. I've added some tests to verify this, and have removed the assertion. I've also added some tests to check handling of invalid values.

### DIFF
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 
 import math
 import unittest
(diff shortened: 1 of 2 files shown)


uri-granta (approved):
Looks good, thanks!

PR #726, opened 2023-04-26
fix: make metric names unique, so that the same metric name isn't used

hstojic:


trieste/acquisition/optimizer.py
```diff
@@ -350,7 +350,7 @@ def improvements() -> tf.Tensor:
                 if V == 1:
                     logging.scalar("spo_improvement_on_initial_samples", improvements)
                 else:
-                    logging.histogram("spo_improvement_on_initial_samples", improvements)
+                    logging.histogram("spo_improvement_on_initial_samples_across_subspaces", improvements)
```
```suggestion
                    logging.histogram("spo_improvements_on_initial_samples", improvements)
```

trieste/acquisition/rule.py
```diff
@@ -330,7 +330,7 @@ def acquire(
                             )
                         else:  # vectorized acquisition function
                             logging.histogram(
-                                f"EGO.acquisition_function/maximum_found[{i+1}]", values
+                                f"EGO.acquisition_function/maximums_found[{i+1}]", values
```
hstojic, line 333: a bit confused here - I would have thought that `[{i+1}]` bit is part of the name as well, and here the naming would be different than what we have above?

trieste/bayesian_optimizer.py
```diff
@@ -941,7 +941,7 @@ def write_summary_query_points(
     if tf.rank(query_points) == 2:
         for i in tf.range(tf.shape(query_points)[1]):
             if len(query_points) == 1:
-                logging.scalar(f"query_points/[{i}]", float(query_points[0, i]))
+                logging.scalar(f"query_point/[{i}]", float(query_points[0, i]))
```
hstojic, line 944: I think you have the same in the ask-tell optimizer as well

hstojic (commented):
looks ok to me - @uri-granta what do you think?

trieste/acquisition/rule.py
uri-granta: I'm confused by your confusion :-) Both the scalar and historgram were previously called `maximum_found[{i+1}]`.

trieste/bayesian_optimizer.py
alansaul: I think it re-uses this code so it's already covered :+1:

uri-granta (approved):
LGTM!

PR #727, opened 2023-05-04
Handle broadcasting in DeepEnsemble models

uri-granta:
The `DeepEnsemble.predict` method should handle leading batch dimensions, as described in the `ProbabilisticModel` interface.


tests/unit/utils/test_misc.py
```diff
@@ -132,3 +132,29 @@ def test_unflatten_returns_correct_shape() -> None:
...
+    [
+        (1, [120]),
+        (2, [24, 5]),
+        (3, [6, 4, 5]),
+        (4, [2, 3, 4, 5]),
+    ],
+)
+def test_flatten_leading_dims_output_dims(output_dims: int, expected_shape: list[int]) -> None:
+    x_old = tf.random.uniform([2, 3, 4, 5])  # [2, 3, 4, 5]
+    flat_x_old, unflatten = flatten_leading_dims(x_old, output_dims=output_dims)  # [6, 4, 5]
```
khurram-ghani: If you were restarting the tests for the PR, you could remove the shape comment as it doesn't apply to all test cases.

trieste/models/keras/models.py
```diff
@@ -249,7 +250,10 @@ def predict(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
         :return: The predicted mean and variance of the observations at the specified
             ``query_points``.
         """
-        ensemble_distributions = self.ensemble_distributions(query_points)
+        # handle leading batch dimensions
+        input_dims = len(self.model.input_shape[0])
```
khurram-ghani: Just for my understanding, you look at `input_shape[0]` (instead of the whole shape) because that is the input shape of the first network in the ensemble (and we assume all networks have the same inputs shape); is that correct?

khurram-ghani (approved):
LGTM. Just a simple question and a very minor comment change if there is opportunity.

trieste/models/keras/models.py
uri-granta: That's right, though I agree this is not clear. AFAICT there is nothing that actually forces all the networks in the KerasEnsembe to have the same input shapes (though ones generated with `trieste_keras_ensemble_model` will do) but for `predict` to work we assume that we can call the constructed ensemble `tf.keras.Model` with the same query points for every network (@hstojic: is this right?)

uri-granta: (though now that the docs gym issue is fixed, there's an error in the deep ensemble notebook that suggest that I might be not quite right; will investigate...)

uri-granta: Turns out that keras's `Functional` class has code to "allow (None,) and (None, 1) Tensors to be passed interchangeably", which the notebook was making use of. I've changed my fix so it doesn't break that (slightly odd) usage.

trieste/models/keras/models.py
```diff
@@ -249,7 +250,11 @@ def predict(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
         :return: The predicted mean and variance of the observations at the specified
             ``query_points``.
         """
-        ensemble_distributions = self.ensemble_distributions(query_points)
+        # handle leading batch dimensions, while still allowing `Functional` to
+        # "allow (None,) and (None, 1) Tensors to be passed interchangeably"
+        input_dims = min(len(query_points.shape), len(self.model.input_shape[0]))
```
khurram-ghani, line 255: Very minor point, perhaps `tf.rank(query_points)` would be a bit more concise.

uri-granta: Actually that confused AutoGraph for some reason in one of the integration tests! I'm somewhat pressed for time at the moment so might leave as is for now and try and figure out why later.

khurram-ghani: I have a hunch why that may be ... using `tf.rank` would create a tensor and the autograph would try to integrate that into the graph. In my experience adding `.numpy()` at the end, or some other way of making it static usually works. Not worth the hassle in this case though ;-)

PR #728, opened 2023-05-04
Remove redundant pandas dependency from plotting

khurram-ghani:
Removed pandas dependency from plotting code, which wasn't needed. Hence able to also remove the pandas version constraints for notebooks.


trieste/experimental/plotting/plotting_plotly.py
```diff
@@ -88,10 +87,10 @@ def add_surface_plotly(
     :return: updated plotly figure
     """
 
-    d = pd.DataFrame(f.reshape([xx.shape[0], yy.shape[1]]), index=xx, columns=yy)
+    z = f.reshape([xx.shape[0], yy.shape[1]])
```
uri-granta, line 90: Does this impact the generated plots? (i.e. do we still label them correctly?)

khurram-ghani: The generated plots look exactly the same with the same (default) labels. The x and y axes are directly provided to `go.Surface` below, so the index/columns here are redundant.

uri-granta (approved):
LGTM! Testing a workaround for the docs issue now so hopefully you'll be able to merge this soon.

PR #731, opened 2023-05-11
Fix docs buikld issue

uri-granta:
Temporary workaround for the gym bug exposed by the latest version of wheel. In future we'll want to unpin gym and remove the pins to wheel and setuptools somehow (see https://github.com/secondmind-labs/trieste/issues/733).


khurram-ghani (approved):
LGTM

uri-granta:
I've actually got a better fix (upgrading to the latest gym which doesn't have this issue) so closing this PR.

PR #732, opened 2023-05-12
Fix docs build by unpinning gym version

uri-granta:
Unpin gym (fixing the small number of API breaks) which fixes the recent docs build issue.


khurram-ghani (approved):
LGTM

PR #734, opened 2023-05-16
Use latest sphinx for building documentation

uri-granta:
Unpin sphinx used for building documentation.

:warning: before merging this ensure that the documentation _looks_ correct, both for the API and the tutorials (i.e. don't just rely on the build being successful).


khurram-ghani (approved):
Looks good. I generated the docs locally and the only thing odd I noticed is that the table of contents is missing (on the right) for some of the API pages. GPflow has exactly the same behaviour. Must be a new sphinx feature.

PR #736, opened 2023-05-23
Allow SeparateIndependent kernels in gpflow decoupled sampler

### LINKED ISSUE #724: Allow SeparateIndependentKernels type of kernels in gpflow samplers
this change was made for gpflux models in #711  but not for gpflow models

khurram-ghani:
Addresses #724.

This PR adds `SeparateIndependent` kernels support to `DecoupledTrajectorySampler` for gpflow models. As part of updating the tests, I have removed some of the duplication in the sampler tests.

Similar support for `RandomFourierFeatureTrajectorySampler` can be added in a follow-on PR. Also, perhaps there is potential for combining this decoupled trajectory sampler with the one for gpflux `DeepGaussianProcessDecoupledTrajectorySampler` and simplifying. This could be covered in another PR.


trieste/models/gpflow/sampler.py
```diff
@@ -363,7 +364,8 @@ def get_trajectory(self) -> TrajectoryFunction:
         and evaluating the feature functions.
 
         :return: A trajectory function representing an approximate trajectory from the Gaussian
-            process, taking an input of shape `[N, D]` and returning shape `[N, 1]`
+            process, taking an input of shape `[N, B, D]` and returning shape `[N, B, P]`
+            where `P` is the number of outputs of the model.
```
hstojic: TrajectorySampler class spcifies L rather than P

trieste/models/gpflow/sampler.py
```diff
@@ -412,11 +414,11 @@ def resample_trajectory(self, trajectory: TrajectoryFunction) -> TrajectoryFunct
         return trajectory  # return trajectory with resampled weights
 
     @abstractmethod
-    def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [L, B]
+    def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [B, L, P]
```
hstojic: hm, L is already used here for num of features but we should follow the interface and use it for outputs, and smth else for features, say F?

trieste/models/gpflow/sampler.py
```diff
@@ -483,11 +485,11 @@ def __init__(
         )
         super().__init__(self._model, feature_functions)
 
-    def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [L, B]
+    def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [B, L, 1]
```
hstojic: line 482 above should not be necessary, its done by the parent class via super call

trieste/models/gpflow/sampler.py
```diff
@@ -635,75 +636,88 @@ def __init__(
 
         super().__init__(self._model, feature_functions)
 
-    def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:
+    def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [B, L + M, P]
```
hstojic: line 634 above should not be necessary

trieste/models/gpflow/sampler.py
```diff
@@ -749,18 +763,7 @@ def __init__(
...
-        super().__init__(self._kernel, self._n_components, dtype=tf.float64)
-
-        if isinstance(model, SupportsGetInducingVariables):
-            dummy_X = model.get_inducing_variables()[0][0:1, :]
-        else:
-            dummy_X = model.get_internal_data().query_points[0:1, :]
-
-        self.__call__(dummy_X)  # dummy call to force init of weights
-        self.b: TensorType = tf.Variable(self.b)
-        self.W: TensorType = tf.Variable(self.W)  # allow updateable weights
```
hstojic, line 763: this is now done in GPflux so no need for it?

trieste/models/gpflow/sampler.py
```diff
@@ -795,24 +798,32 @@ def __init__(
         :param n_components: The desired number of features.
         """
 
+        super().__init__(model, n_components)
+
         if isinstance(model, SupportsGetInducingVariables):
-            inducing_points = model.get_inducing_variables()[0]  # [M, D]
+            self._inducing_points = model.get_inducing_variables()[0]  # [M, D]
```
hstojic, line 804: why assigning to self?

trieste/models/gpflow/sampler.py
```diff
@@ -795,24 +798,32 @@ def __init__(
...
+                kernel.K(self._inducing_points, x, full_output_cov=False)
+            )
+        else:
+            kernel = model.get_kernel()
+            self._canonical_feature_functions = lambda x: tf.linalg.matrix_transpose(
+                kernel.K(self._inducing_points, x)
+            )
 
-    def __call__(self, x: TensorType) -> TensorType:  # [N,D] -> [N, L + M]
+    def call(self, x: TensorType) -> TensorType:  # [N, D] -> [N, L + M] or [P, N, L + M]
```
hstojic: don't get this change, a bit of education please :)

hstojic (commented):
ok, did a pass and changes look fine to me, left only some minor comments
I didn't have time to do a checkout to verify it in a more detailed manner - I can do that after @vpicheny pass

trieste/models/gpflow/sampler.py
khurram-ghani: Just to make explicit that we keep a handle to it, so this change is not essential. Before it was indirect/hidden via self._canonical_feature_functions lambda. I also found it useful for debugging as I could more easily look at its state.

trieste/models/gpflow/sampler.py
khurram-ghani: The converting of weights to `tf.Variable` hasn't been needed since [PR 48](https://github.com/secondmind-labs/GPflux/pull/48) in GPflux from Aug 2021. That PR created the weights using Keras layer `add_weight` calls, which creates variables.

The deletion of dummy call explained with the change to `__call__`.

trieste/models/gpflow/sampler.py
khurram-ghani: This is a keras layer and the [recommended](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#args_1) methods to override are `build` (for initialisation that depends on input shape) and `call`, instead of directly overriding `__call__`. This means the dummy call in  `ResampleableRandomFourierFeatureFunctions.__init__` can also be removed (as the keras layer will automatically build the weights if required before the call).

trieste/models/gpflow/sampler.py
khurram-ghani: Done

trieste/models/gpflow/sampler.py
khurram-ghani: Done

trieste/models/gpflow/sampler.py
khurram-ghani: Done

trieste/models/gpflow/sampler.py
khurram-ghani: Done

trieste/models/gpflow/sampler.py
```diff
@@ -630,80 +628,92 @@ def __init__(
...
         that takes in a batch size `B` and returns `B` samples for the weights of each of
-        the `L`  RFF features and `N` cannonical features.
+        the `F`  RFF features and `M` canonical features for `L` outputs.
         """
 
+        kernel = self._model.get_kernel()
+        if self._feature_functions.is_multioutput:
+
+            def kernel_K(iv: TensorType) -> TensorType:
+                # mypy is confused and doesn't see that kernel can be of two types
```
uri-granta: Did you ever find a related mypy issue? If not it might be worth spending a little bit of time (but not too long) creating a minimal example and raising one?

trieste/models/gpflow/sampler.py
```diff
@@ -630,80 +628,92 @@ def __init__(
...
         Prepare the sampler function that provides samples of the feature weights
-        for both the RFF and cannonical feature functions, i.e. we return a function
+        for both the RFF and canonical feature functions, i.e. we return a function
         that takes in a batch size `B` and returns `B` samples for the weights of each of
-        the `L`  RFF features and `N` cannonical features.
+        the `F`  RFF features and `M` canonical features for `L` outputs.
         """
 
+        kernel = self._model.get_kernel()
+        if self._feature_functions.is_multioutput:
```
uri-granta: Wouldn't it be simpler to check directly whether the kernel is of type `MultioutputKernel`?

trieste/models/gpflow/sampler.py
```diff
@@ -795,24 +794,32 @@ def __init__(
...
         else:
-            inducing_points = model.get_internal_data().query_points  # [M, D]
+            self._inducing_points = model.get_internal_data().query_points  # [M, D]
 
-        self._cannonical_feature_functions = lambda x: tf.linalg.matrix_transpose(
-            model.get_kernel().K(inducing_points, x)
-        )
-
-        super().__init__(model, n_components)
+        if self.is_multioutput:
```
uri-granta: (same comment as before re checking the kernel type)

trieste/models/gpflow/sampler.py
```diff
@@ -795,24 +794,32 @@ def __init__(
...
+                kernel.K(self._inducing_points, x, full_output_cov=False)
+            )
+        else:
+            kernel = model.get_kernel()
+            self._canonical_feature_functions = lambda x: tf.linalg.matrix_transpose(
+                kernel.K(self._inducing_points, x)
+            )
 
-    def __call__(self, x: TensorType) -> TensorType:  # [N,D] -> [N, L + M]
+    def call(self, x: TensorType) -> TensorType:  # [N, D] -> [N, F + M] or [L, N, F + M]
```
uri-granta, line 813: What was the impact of this previously overriding and calling `__cal__l` rather than `call`? Were we skipping `FourierFeaturesBase.call` altogether?

tests/unit/models/gpflow/test_models.py
```diff
@@ -1502,34 +1502,44 @@ def test_sparse_variational_optimize(batcher: DatasetTransformer, compile: bool)
...
+    kernel_type: str,
 ) -> None:
     x = tf.constant(np.arange(5).reshape(-1, 1), dtype=gpflow.default_float())
     y = _3x_plus_gaussian_noise(x)
     if use_mean_function:
         mean = gpflow.mean_functions.Linear()
     else:
         mean = gpflow.mean_functions.Zero()
+
+    num_latent_gps = 1
```
uri-granta: Maybe put this in a helper function (or even a kernel-generating fixture)?

uri-granta (commented):
Looks good from an engineering perspective! I'll leave the others to confirm the maths.

trieste/models/gpflow/sampler.py
khurram-ghani: I prefer that solution as well and did implement it first! However, mypy caused more issues with that. I could solve it but needed more mypy workarounds/ignores. So I reverted to the solution used by `DeepGaussianProcessDecoupledLayer` as it needed fewer hacks. I can resurrect it since you prefer it as well :-)

trieste/models/gpflow/sampler.py
khurram-ghani: I didn't look for a mypy issue. I'll search for one.

trieste/models/gpflow/sampler.py
khurram-ghani: (As above).

trieste/models/gpflow/sampler.py
khurram-ghani: Not a significant impact for the current usage. `FourierFeaturesBase.call` was still being called because of the `super().__call__(x)` line. Keras `Layer.__call__` internally calls `build` (if needed) and `call`. One disadvantage of the old code would have been the dummy call in `__init__` to create the variables for the weights, instead of simply using `build`; but it turns out we don't need to create the variables at all.

tests/unit/models/gpflow/test_models.py
```diff
@@ -1502,34 +1502,44 @@ def test_sparse_variational_optimize(batcher: DatasetTransformer, compile: bool)
...
+    kernel = gpflow.kernels.Matern32()
+    iv = x
+    if kernel_type != "single":
+        y = tf.tile(y, [1, 2])
+        num_latent_gps = 2
+        iv = gpflow.inducing_variables.SharedIndependentInducingVariables(
+            gpflow.inducing_variables.InducingPoints(x)
+        )
+        if kernel_type == "separate":
+            kernel = gpflow.kernels.SeparateIndependent([kernel, kernel])
```
vpicheny: Might be best to have `[kernel, kernel2]` instead? initialised with different (not 1.) lengthscales and variances so that this model could actually be used for non-trivial tests?

The `model.optimize` below should change those values but it's best to be on the safe side?

tests/unit/models/gpflow/test_models.py
```diff
@@ -1547,19 +1557,19 @@ def test_sparse_variational_trajectory_sampler_has_correct_samples(
...
+    sample_mean = tf.reduce_mean(samples, axis=1)  # [N, P]
+    sample_variance = tf.math.reduce_variance(samples, axis=1)  # [N, P]
 
     true_mean, true_variance = model.predict(x_predict)
 
     # test predictions approx correct away from data
-    npt.assert_allclose(sample_mean[3:] + 1.0, true_mean[3:] + 1.0, rtol=0.1)
-    npt.assert_allclose(sample_variance[3:], true_variance[3:], rtol=0.25)
+    npt.assert_allclose(sample_mean[3:] + 1.0, true_mean[3:] + 1.0, rtol=0.12)
+    npt.assert_allclose(sample_variance[3:], true_variance[3:], rtol=0.3, atol=5e-3)
```
vpicheny: Those tolerances are suspiciously high (in particular for the mean)... might be good to bump `num_samples` to 1000 or more if we can afford it? `rtol=0.3` away from the data worries me.

It is true that the trajectory sampler is an approximation of the exact sampler so we can't expect great accuracy but this still seems a bit high (which is also true for the previous version of this test). Also bump the number of features if possible?

tests/util/models/gpflow/models.py
```diff
@@ -381,6 +381,16 @@ def svgp_model(x: tf.Tensor, y: tf.Tensor, num_latent_gps: int = 1) -> SVGP:
     )
 
 
+def quadratic_mean_rbf_kernel_model(dataset: Dataset) -> QuadraticMeanAndRBFKernelWithSamplers:
+    model = QuadraticMeanAndRBFKernelWithSamplers(
+        noise_variance=tf.constant(1.0, dtype=tf.float64), dataset=dataset
+    )
+    model.kernel = (
+        gpflow.kernels.RBF()
```
vpicheny, line 390: maybe use a non=trivial value for the RBF parameters? e.g. variance=0.9

the default variance = 1 makes every test a bit weaker IMO

tests/unit/models/gpflow/test_models.py
```diff
@@ -1547,19 +1557,19 @@ def test_sparse_variational_trajectory_sampler_has_correct_samples(
     x_predict = tf.constant([[1.0], [2.0], [3.0], [1.5], [2.5], [3.5]], gpflow.default_float())
     x_predict_parallel = tf.expand_dims(x_predict, -2)  # [N, 1, D]
     x_predict_parallel = tf.tile(x_predict_parallel, [1, num_samples, 1])  # [N, B, D]
-    samples = trajectory(x_predict_parallel)  # [N, B, 1]
-    sample_mean = tf.reduce_mean(samples, axis=1)  # [N, 1]
-    sample_variance = tf.math.reduce_variance(samples, axis=1)  # [N, 1]
+    samples = trajectory(x_predict_parallel)  # [N, B, P]
```
vpicheny: not sure if important but we use L, not P, in the model code...

trieste/models/gpflow/sampler.py
```diff
@@ -412,11 +414,11 @@ def resample_trajectory(self, trajectory: TrajectoryFunction) -> TrajectoryFunct
         return trajectory  # return trajectory with resampled weights
 
     @abstractmethod
-    def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [L, B]
+    def _prepare_weight_sampler(self) -> Callable[[int], TensorType]:  # [B] -> [B, F, L]
```
vpicheny, line 432: was that wrong before?

vpicheny: or did we use L for the number of features? (which is a bad idea)

trieste/models/gpflow/sampler.py
```diff
@@ -630,80 +628,92 @@ def __init__(
...
         if isinstance(self._model, FeatureDecompositionInducingPointModel):
             (  # extract variational parameters
                 inducing_points,
                 q_mu,
                 q_sqrt,
                 whiten,
-            ) = self._model.get_inducing_variables()  # [M, d], [M, 1], [1, M, 1]
-            q_sqrt = q_sqrt[0, :, :]  # [M, M]
-            Kmm = self._model.get_kernel().K(inducing_points, inducing_points)  # [M, M]
+            ) = self._model.get_inducing_variables()  # [M, D], [M, L], [L, M, M], []
```
vpicheny, line 665: extra []?

trieste/models/gpflow/sampler.py
vpicheny: Do we use these lines elsewhere? if so worth a helper function? (if not, ignore)

trieste/models/gpflow/sampler.py
```diff
@@ -795,24 +794,32 @@ def __init__(
...
+    def call(self, x: TensorType) -> TensorType:  # [N, D] -> [N, F + M] or [L, N, F + M]
         """
-        combine prior basis functions with cannonical basis functions
+        combine prior basis functions with canonical basis functions
         """
-        fourier_feature_eval = super().__call__(x)  # [N, L]
-        cannonical_feature_eval = self._cannonical_feature_functions(x)  # [N, M]
-        return tf.concat([fourier_feature_eval, cannonical_feature_eval], axis=-1)  # [N, L + M]
+        fourier_feature_eval = super().call(x)  # [N, F] or [L, N, F]
+        cannonical_feature_eval = self._canonical_feature_functions(x)  # [N, M] or [L, N, M]
```
vpicheny: canonical? left one "nn" uncorrected  :-)

trieste/models/gpflow/sampler.py
```diff
@@ -874,19 +881,26 @@ def __call__(self, x: TensorType) -> TensorType:  # [N, B, d] -> [N, B, 1]
...
 
-        flat_x, unflatten = flatten_leading_dims(x)  # [N*B, d]
-        flattened_feature_evaluations = self._feature_functions(flat_x)  # [N*B, m]
-        feature_evaluations = unflatten(flattened_feature_evaluations)  # [N, B, m]
+        flat_x, unflatten = flatten_leading_dims(x)  # [N*B, D]
+        flattened_feature_evaluations = self._feature_functions(
+            flat_x
+        )  # [N*B, F + M] or [L, N*B, F + M]
+        # ensure tensor is always rank 3
+        rank3_shape = tf.concat([[1], tf.shape(flattened_feature_evaluations)], axis=0)[-3:]
```
vpicheny, line 889: we have a flatten / unflatten helper function in utils, perhaps could be used here? (not sure)

vpicheny (commented):
Looks great, I couldn't fully track what was being tested with those changes, just to recap it's really important to check that:
- trajectory samplers return samples from the posterior distribution for a variety of models
- trajectory samplers update when needed 
- and I guess the shapes of weights and samples are consistent for all models.

I am pretty sure that the current tests cover all of this (including the integration tests with Thompson Sampling I imagine?) but it wouldn't hurt to double-check

tests/util/models/gpflow/models.py
khurram-ghani: It was 1.0 originally in all the tests, but I'll change it.

trieste/models/gpflow/sampler.py
khurram-ghani: The comment was wrong before. The dimensions were swapped and it was using `L` for number of features. It should have been `[B, F]`, or similar.

trieste/models/gpflow/sampler.py
khurram-ghani: Missed that one :-)

trieste/models/gpflow/sampler.py
khurram-ghani: Those shapes are for the 4-tuple that `get_inducing_variables()` returns. So `[]` is for `whiten` as it is a scalar.

trieste/models/gpflow/sampler.py
khurram-ghani: Flatten/unflatten is already used for `x`. I tried to also use it here, but can't see a way where it helps with reshaping/rank. There may be a way if I also change the subsequent lines, e.g. transpose/permutation, but that doesn't seem like a big win.

trieste/models/gpflow/sampler.py
khurram-ghani: I have implemented that, in a helper function.

tests/unit/models/gpflow/test_models.py
khurram-ghani: I have now replaced this with the (improved) helper function to create the SVGP model directly.

tests/unit/models/gpflow/test_models.py
khurram-ghani: Done, as part of the improved (and renamed) helper function `svgp_model_by_type`.

tests/unit/models/gpflow/test_models.py
khurram-ghani: By increasing the number of samples to 3000 (without too much test time penalty), I have been able to reduce the `rtol` to `0.04` and `0.1` for mean and variance respectively (so a third compared to before). Is this sufficient?

Increasing the number of features increased the memory usage on the machine (as expected I guess), but didn't seem to help with the tolerances that much. Increasing it too large resulted in the process crashing due to memory issues. So I left it at the default of 1000.

tests/util/models/gpflow/models.py
khurram-ghani: Changed to 0.9.

tests/unit/models/gpflow/test_models.py
khurram-ghani: Done

trieste/models/gpflow/sampler.py
khurram-ghani: There was one other place where we had similar (but not exactly the same) lines in the file. So I have created a helper function.

trieste/models/gpflow/sampler.py
khurram-ghani: Done

tests/unit/models/gpflow/test_models.py
vpicheny: Sounds good! `rtol=0.1` is quite high still but that's OK if the values we're looking at are close to zero, which I think is the case at least for some of them (if all of them are close to zero then this test isn't useful because we'd be testing something trivial and we'd need to change it, for instance by increasing the lengthscales).

vpicheny (approved):
great work!

trieste/models/gpflow/sampler.py
khurram-ghani: I have fixed this by using `TypeGuard`. The issue was that kernel type is dynamically determined so mypy can't figure that out.

khurram-ghani:
> Looks great, I couldn't fully track what was being tested with those changes, just to recap it's really important to check that:
> 
> * trajectory samplers return samples from the posterior distribution for a variety of models
> * trajectory samplers update when needed
> * and I guess the shapes of weights and samples are consistent for all models.
> 

@vpicheny We have tests for these, covering these models:
- single-op simple custom model (`quadratic_mean_rbf_kernel_model`)
- single-op svgp
- two-op shared-independent kernel svgp
- two-op separate-independent kernel svgp

> I am pretty sure that the current tests cover all of this (including the integration tests with Thompson Sampling I imagine?) but it wouldn't hurt to double-check

I found [this](https://github.com/secondmind-labs/trieste/blob/khurram/gpflow_separate_kernel_sampling/tests/integration/test_bayesian_optimization.py#L285-L301) integration test with Thompson sampling that covers this. It tests single output only though. Do we need a multioutput test here? Do we have existing toy problems for multioutputs?

vpicheny:
> I found [this](https://github.com/secondmind-labs/trieste/blob/khurram/gpflow_separate_kernel_sampling/tests/integration/test_bayesian_optimization.py#L285-L301) integration test with Thompson sampling that covers this. It tests single output only though. Do we need a multioutput test here? Do we have existing toy problems for multioutputs?

I don't think that we can get a multi-output integration test easily. The current integration test will check that we haven't introduced a bug for the single-gp case, which is already quite nice I think! All in all it looks like the test coverage is good enough as it is, thanks a lot for checking.

PR #737, opened 2023-06-01
Reduce parallelism in ray notebook to fix OOM

uri-granta:
Testing: manually kicked off develop checks build to check that it no longer fails.


hstojic (approved):
looks good

PR #739, opened 2023-06-02
TURBO

henrymoss:
Hi Y'all
I've implemented a different trust region strategy (TURBO).

pls review


trieste/acquisition/rule.py
```diff
@@ -1097,6 +1104,225 @@ def state_func(
...
+        :param failure tolerance: Number of consecutive failures before changing region size.
+        :param num_samples: Size of sample used for Thompson sampling.
+        """
+
+        if not num_query_points > 0:
+            raise ValueError(f"Num query points must be greater than 0, got {num_query_points}")
+
+        if num_query_points > 1:
+            raise NotImplementedError(
+                f"TURBO does not yet support batches greater than 1, got {num_query_points}"
```
vpicheny: I think it should? you could get a batch within a single trust region, right?

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1104,225 @@ def state_func(
...
+            raise ValueError(f"L_min must be postive, got {L_max}")
+
+        self._num_query_points = num_query_points
+        self._L_min = L_min
+        self._L_init = L_init
+        self._L_max = L_max
+        self._success_tolerance = success_tolerance
+        self._failure_tolerance = failure_tolerance
+        self._num_samples = num_samples
+        self._thompson_sampler = ThompsonSamplerFromTrajectory(sample_min_value=False)
```
vpicheny: there is no particular reason to get tied to TS, right? we could make a slightly more general TurBO that also works with EI, similar to TREGO?

vpicheny: with the above base rule as default so it reverts to the original TurBO

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1104,225 @@ def state_func(
...
+            failure_tolerance = math.ceil(search_space.dimension / num_query_points)
+        search_space_max_width = tf.reduce_max(search_space.upper - search_space.lower)
+        if L_min is None:
+            L_min = (0.5**7) * search_space_max_width
+        if L_init is None:
+            L_init = 0.8 * search_space_max_width
+        if L_max is None:
+            L_max = 1.6 * search_space_max_width
+        if num_samples is None:
+            num_samples = tf.clip_by_value(100 * search_space.dimension, 5_000, 1_000_000)
```
vpicheny: in the paper I think num_sample is min(100d, 5000) since it's exact TS

trieste/acquisition/rule.py
vpicheny: BTW this is not exactly in line with the original TurBO, it should be the `ExactThompsonSampler`, right?

docs/refs.bib
```diff
@@ -521,3 +521,11 @@ @inproceedings{wang2013bayesian
   year={2013}
 }
 
+@article{eriksson2019scalable,
```
uri-granta: (there already is an eriksson2019scalable on line 342!)

trieste/acquisition/rule.py
```diff
@@ -1044,7 +1050,8 @@ def acquire(
         :param models: The model for each tag.
         :param datasets: The known observer query points and observations. Uses the data for key
             `OBJECTIVE` to calculate the new trust region.
-        :return: A function that constructs the next acquisition state and the recommended query
+        :return: A function that constructs the next ac
```
uri-granta: oops?

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1104,225 @@ def state_func(
...
+        failure_counter: int
+        """ Number of consecutive failures (reset if we see a success). """
+
+        success_counter: int
+        """ Number of consecutive successes (reset if we see a failure).  """
+
+        y_min: TensorType
+        """ The minimum observed value. """
+
+        def __deepcopy__(self, memo: dict[int, object]) -> TURBO.State:
```
uri-granta, line 1132: I suspect this isn't necessary (though I'm aware you just copied it from the other trust region implementation). As far as I can see box_copy isn't actually a copy (due to box's override on `__deepcopy__`) and the returned `TURBO.State` is just what the default deepcopy would have done anyway?

If you're feeling virtuous you could remove the deepcopy and see whether the tests still all pass. Otherwise feel free to leave as is.

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1104,225 @@ def state_func(
...
+        self._L_init = L_init
+        self._L_max = L_max
+        self._success_tolerance = success_tolerance
+        self._failure_tolerance = failure_tolerance
+        self._num_samples = num_samples
+        self._thompson_sampler = ThompsonSamplerFromTrajectory(sample_min_value=False)
+
+    def __repr__(self) -> str:
+        """"""
+        return f"TURBO({self._num_query_points!r})"
```
uri-granta: do we want to include any of the other parameters too?

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1104,225 @@ def state_func(
...
+                if L < self._L_min:  # if gets too small then start again
+                    L, failure_counter, success_counter = self._L_init, 0, 0
+
+            # build region with volume according to length L but stretched according to lengthscales
+            xmin = dataset.query_points[tf.argmin(dataset.observations)[0], :]  # centre of region
+            lengthscales = (
+                model.get_kernel().lengthscales
+            )  # stretch region according to model lengthscales
+            tr_width = (
+                lengthscales * L / tf.reduce_prod(lengthscales) ** (1.0 / global_lower.shape[-1])
```
uri-granta, line 1314: Not necessary but a bit easier to read?

```suggestion
                lengthscales * L / (tf.reduce_prod(lengthscales) ** (1.0 / global_lower.shape[-1]))
```

tests/unit/acquisition/test_utils.py
```diff
@@ -64,3 +71,25 @@ def test_select_nth_output() -> None:
...
+def test_random_mixing_raises_for_invalid_input(x: TensorType, other_x: TensorType) -> None:
+    with pytest.raises(ValueError):
+        randomly_mix_x_with_other_x(x, other_x, 0.5)
+
+
+@random_seed
+def test_random_mixing_raises_works() -> None:
+    x = tf.random.normal([10_000, 6])
+    x_other = tf.random.normal([1, 6])
+    x_mixed = randomly_mix_x_with_other_x(x, x_other, 0.5)
```
uri-granta: Could you please also parametrize by:

1. prob? (i.e. don't just test 0.5)
2. tf.function compilation? (like in test_randomize_hyperparameters_randomizes_kernel_parameters_with_priors and various other tests)

trieste/acquisition/utils.py
```diff
@@ -114,3 +115,23 @@ def select_nth_output(x: TensorType, output_dim: int = 0) -> TensorType:
...
     return x[..., output_dim]
+
+
+def randomly_mix_x_with_other_x(x: TensorType, other_x: TensorType, prob: float) -> TensorType:
+    """
+    A utility function that takes a tensor of x and returns a new tensor with some
+    of its entries replaced by that of other_x according to a given probability.
+
+    :param x: Input with shape [N, d].
+    :param other_x: The other point to be mixed into x with shape [1, d].
```
uri-granta: (Missing docstring for `prob`)

tests/unit/acquisition/test_rule.py
```diff
@@ -737,6 +739,314 @@ def test_trust_region_state_deepcopy() -> None:
...
+        dataset, noise_variance=tf.constant(1e-5, dtype=tf.float64)
+    )
+    model.kernel = gpflow.kernels.RBF(
+        lengthscales=tf.constant([4.0, 1.0], dtype=tf.float64), variance=1e-5
+    )  # need a gpflow kernel for TURBO
+    state, query_point = tr.acquire_single(search_space, model, dataset=dataset)(None)
+
+    assert state is not None
+    npt.assert_array_almost_equal(
+        query_point, [[0.0, 0.0]], 2
```
uri-granta:
```suggestion
        query_point, [[0.0, 0.0]], decimal=2
```

tests/unit/acquisition/test_rule.py
```diff
@@ -737,6 +739,314 @@ def test_trust_region_state_deepcopy() -> None:
...
+        npt.assert_array_almost_equal(current_state.L, tf.cast(0.4, dtype=tf.float64))
+        npt.assert_array_almost_equal(current_state.acquisition_space.lower, lower_bound)
+        npt.assert_array_almost_equal(
+            current_state.acquisition_space.upper, tf.constant([0.4, 0.1], dtype=tf.float64)
+        )
+        assert current_state.success_counter == 0
+        assert current_state.failure_counter == 0
+
+
+def test_turbo_restarts_tr_when_to_small() -> None:
```
uri-granta:
```suggestion
def test_turbo_restarts_tr_when_too_small() -> None:
```

uri-granta (approved):
LGTM!

trieste/acquisition/rule.py
henrymoss: yep fixed

trieste/acquisition/rule.py
henrymoss: Ive allowed this to be custom

trieste/acquisition/rule.py
henrymoss: yep fixed

trieste/bayesian_optimizer.py
```diff
@@ -327,6 +328,7 @@ def optimize(
         track_state: bool = True,
         track_path: Optional[Path | str] = None,
         fit_initial_model: bool = True,
+        fit_global_model: bool = True,
```
vpicheny: I don't like the name `fit_global_model`, it's too TurBO-specific...

We could either change it to `refit_models` or `fit_model_at_each_iteration` or something better?

Alternatively to avoid inflating the number of arguments we could use a single `skip_model_training` with possible values `'none'`, `'all'`, `'init'`? 

Also, we would need to do the same for `ask_tell` perhaps?

vpicheny (commented):
If this doesn't trigger retracing, I think this is the right way to code TurBO with local models!

The current code isn't modular (we can' reuse the same elements or logic easily elsewhere without duplicating code) but maybe that's OK for this PR? we can refactor later...

henrymoss:
@uri-granta ive done a lot more fiddling. Please can you fix the types for me.

uri-granta:
> @uri-granta ive done a lot more fiddling. Please can you fix the types for me.

What's the deal with get_local_dataset (added in one commit, removed in another, but still mentioned)?

henrymoss:
> > @uri-granta ive done a lot more fiddling. Please can you fix the types for me.
> 
> What's the deal with get_local_dataset (added in one commit, removed in another, but still mentioned)?

@uri-granta sorry that was my mistake. should be back now

uri-granta:
> > > @uri-granta ive done a lot more fiddling. Please can you fix the types for me.
> > 
> > 
> > What's the deal with get_local_dataset (added in one commit, removed in another, but still mentioned)?
> 
> @uri-granta sorry that was my mistake. should be back now

Some more issues:

`rule.py:1181` How are you going to initialize `failure_tolerance` if the rule is not a `DiscreteThompsonSampling`?
`rule.py:1256` You don't need to redeclare `self._local_models`'s type here (mypy should be able to figre out it's not None).
`test_rule.py:844` Should assert that `_local_models` is NOT None after acquire.
`test_rule.py:844+` Should assert the precise type of the model: at the moment all mypy knows is that it's a `TrainableSupportsGetKernel`, which doesn't necessarily have a `.kernel` (though it does have `get_kernel()`) or `._dataset`.

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1105,225 @@ def state_func(
...
+            acquisition_space = Box(
+                tf.reduce_max([global_lower, xmin - tr_width / 2.0], axis=0),
+                tf.reduce_min([global_upper, xmin + tr_width / 2.0], axis=0),
+            )
+
+            # fit the local model using just data from the trust region
+            local_dataset = get_local_dataset(acquisition_space, dataset)
+            local_model.update(local_dataset)
+            local_model.optimize(local_dataset)
+
```
vpicheny, line 1325: here you're possibly missing

`self._local_models[OBJECTIVE] = local_model`

vpicheny: @uri-granta experimenting with TURBO I found that  if I use this local model to plot some quantities during the BO run e.g.

`local_model = ask_tell._acquisition_rule._local_models['OBJECTIVE']
local.model.predict(...)`

the predictions do not get updated unless I add the line above. Does it make any sense to you?

trieste/bayesian_optimizer.py
henrymoss: nice

trieste/acquisition/rule.py
henrymoss: I think it might be something to do with the deepcopy of the models @uri-granta ?

if i use deepcopy, then we get above and the local model just stays as the global one
if i use copy, then we get the local model correctly updating but seems to also overwrite the global one

uri-granta: Not sure what's going on yet but `self._local_models` should definitely initialised with `deepcopy`, as otherwise it's just shallow copying the dictionary of models, leaving references to the global ones.

(As an aside, wouldn't it be simpler if we use a single `local_model` everywhere rather than a mapping, since that's all it supports anyway?)

I'll see if I can reproduce the issue...

uri-granta: I'm not managing to reproduce the issue (the model seems to be updating fine for me). Can you push a temporary notebook or test so I can see how you're setting things up?

trieste/ask_tell_optimization.py
```diff
@@ -218,6 +220,14 @@ def __init__(
         else:
             self._acquisition_rule = acquisition_rule
 
+        if (fit_model) and isinstance(acquisition_rule, TURBO):
+            warnings.warn(
+                """
+                Are you sure you want to keep fitting the global model even though you
+                are using TURBO which uses local models? This is a waste of computation.
```
hstojic, line 227: perhaps a bit more polite message :)

trieste/acquisition/utils.py
```diff
@@ -114,3 +115,23 @@ def select_nth_output(x: TensorType, output_dim: int = 0) -> TensorType:
         reduce the input.
     """
     return x[..., output_dim]
+
+
+def get_local_dataset(local_space: SearchSpaceType, dataset: Dataset) -> Dataset:
```
hstojic, line 120: this seems like it would be better to move it to data package where we already have some data manipulation utils?

tests/unit/acquisition/test_utils.py
```diff
@@ -64,3 +70,30 @@ def test_select_nth_output() -> None:
...
+    search_space_1 = Box([0, 0, 0], [1, 1, 1])
+    search_space_2 = Box([5, 5, 5], [10, 10, 10])
+    points_1 = search_space_1.sample(10)
+    points_2 = search_space_2.sample(20)
+    dataset_1 = Dataset(points_1, points_1[:, 0:1])
+    dataset_2 = Dataset(points_2, points_2[:, 0:1])
+    combined = dataset_1 + dataset_2
+
+    assert tf.shape(get_local_dataset(search_space_1, combined).query_points)[0] == 10
+    assert tf.shape(get_local_dataset(search_space_2, combined).query_points)[0] == 20
```
hstojic, line 99: it would be good to check also that the points are from the specified ranges

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1105,235 @@ def state_func(
...
+            raise ValueError(f"Num trust regions must be greater than 0, got {num_trust_regions}")
+
+        if num_trust_regions > 1:
+            raise NotImplementedError(
+                f"TURBO does not yet support multiple trust regions, but got {num_trust_regions}"
+            )
+
+        # implement heuristic defaults for TURBO if not specified by user
+        if rule is None:  # default to Thompson sampling with batches of size 1
+            rule = DiscreteThompsonSampling(tf.minimum(100 * search_space.dimension, 5_000), 1)
```
hstojic, line 1176: sample size seems on the small side to me, perhaps you wanted max here?

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1105,235 @@ def state_func(
         return state_func
 
 
+class TURBO(
+    AcquisitionRule[
+        types.State[Optional["TURBO.State"], TensorType], Box, TrainableSupportsGetKernel
+    ]
+):
+    """Implements the TURBO algorithm as detailed in :cite:`eriksson2019scalable,`."""
```
hstojic: it would be good to expand the description a bit, what is the main use case, advantages, limits - we shouldn't expect users to go and read a paper if they wish to use it for their application

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1105,235 @@ def state_func(
...
+        L_max: Optional[float] = None,
+        success_tolerance: int = 3,
+        failure_tolerance: Optional[int] = None,
+        local_models: Optional[Mapping[Tag, TrainableSupportsGetKernel]] = None,
+    ):
+        """
+        Note that the optional parameters are set by a heuristic if not given by the user.
+
+        :param search_space: The search space.
+        :param num_trust_regions: Number of trust regions controlled by TURBO
```
hstojic, line 1154: currently limited to 1, right?

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1105,235 @@ def state_func(
...
+        :param rule: rule used to select points from within the trust region, using the local model.
+        :param L_min: Minimum allowed length of the trust region.
+        :param L_init: Initial length of the trust region.
+        :param L_max: Maximum allowed length of the trust region.
+        :param success_tolerance: Number of consecutive successes before changing region size.
+        :param failure tolerance: Number of consecutive failures before changing region size.
+        :param local_models: Optional model to act as the local model. This will be refit using
```
hstojic, line 1161: is there a case where you would use both global and local?, why not then continue using the way we usually supply the model and here as a default just copy it?

trieste/acquisition/rule.py
```diff
@@ -1097,6 +1105,235 @@ def state_func(
...
+        self._L_init = L_init
+        self._L_max = L_max
+        self._success_tolerance = success_tolerance
+        self._failure_tolerance = failure_tolerance
+        self._rule = rule
+        self._local_models = local_models
+
+    def __repr__(self) -> str:
+        """"""
+        return f"TURBO({self._num_trust_regions!r}, {self._rule})"
```
hstojic, line 1227: what about other parameters?

hstojic (commented):
I did a quick pass and left some comments

not entirely sure whats the best way to handle global-local model distinction, I don't particularly like the new string flag for ask-tell/BOoptimizer - it seems like only some acq rule will rely on local models, perhaps we could have an `AcquisitionRule` subclass that would indicate its the one with local models and model update would not be triggered for these rules?

trieste/acquisition/rule.py
uri-granta: So an important difference between copy and deecopy is that the tell method of AskTellOptimizer calls update and optimize on the original model (unlike BayesianOptimizer it doesn't implement fit_model="never"), so the model gets updated twice for copy: once locally with the local dataset, and once globally with the new points combined with the original dataset.Yet it's this version that seems to behave better and not track the original data points, so I'm still confused!

trieste/acquisition/rule.py
henrymoss: we have plans for a mixed globa/local setup

uri-granta:
@henrymoss Are you happy to merge this or is there anything else you still want to change? The format, typing and docs issues are all fixed.

PR #740, opened 2023-06-07
fixing an issue with resetting learning rate in gpflux and keras

hstojic:
in keras and gpflux models we can use TF keras optimizers, instead of a callback learning rate schedule can be set as an instance of `tf.keras.optimizers.schedules.LearningRateSchedule` in which case our way of resetting the learning rate is not necessary and would throw an error

### DIFF
--- a/trieste/models/gpflux/models.py
+++ b/trieste/models/gpflux/models.py
@@ -111,7 +111,10 @@ class DeepGaussianProcess(
                 f"received {type(self.optimizer.optimizer)} instead."
             )
 
-        self.original_lr = self.optimizer.optimizer.lr.numpy()
+        if not isinstance(
+            self.optimizer.optimizer.lr, tf.keras.optimizers.schedules.LearningRateSchedule
+        ):
+            self.original_lr = self.optimizer.optimizer.lr.numpy()
 
         epochs = 400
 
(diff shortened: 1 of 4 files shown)


trieste/models/gpflux/models.py
```diff
@@ -111,7 +111,10 @@ def __init__(
                 f"received {type(self.optimizer.optimizer)} instead."
             )
 
-        self.original_lr = self.optimizer.optimizer.lr.numpy()
+        if not isinstance(
+            self.optimizer.optimizer.lr, tf.keras.optimizers.schedules.LearningRateSchedule
+        ):
+            self.original_lr = self.optimizer.optimizer.lr.numpy()
```
uri-granta, line 117: Maybe add a one-line explanation of the potentially uninitialised instance variable? Also do we want a test with LearningRateSchedule and tf.function to make sure that compilation doesn't break with uninitialised instance variables (I don't think it does but it's probably good practice to check all our code paths under compilation).

uri-granta (approved):
Looks good!

khurram-ghani:
Is it worth replacing the original callback/`original_lr` code with a default/custom schedule derived from `LearningRateSchedule`, say something like [this](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/schedules/LearningRateSchedule#example_2)? Then we wouldn't need to detect the difference between two scheduling mechanisms.

trieste/models/gpflux/models.py
hstojic: perhaps we can still initialise it to `None` in case of `LearningRateSchedule`?

hstojic: otherwise, how can I test that compilation doesn't break?

hstojic:
> Is it worth replacing the original callback/`original_lr` code with a default/custom schedule derived from `LearningRateSchedule`, say something like [this](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/schedules/LearningRateSchedule#example_2)? Then we wouldn't need to detect the difference between two scheduling mechanisms.

we could easily do that but the user could still use the callback mode, we would then have to rely on them carefully reading the docs

khurram-ghani:
> we could easily do that but the user could still use the callback mode, we would then have to rely on them carefully reading the docs

If I understand correctly, you mean they would have to carefully read the keras docs to use their external callback properly, right? That is standard keras usage though. Seems to me it would be better to use keras where we can, but I guess we can't expect our users to be always familiar with keras.

trieste/models/gpflux/models.py
uri-granta: You might be able to explicitly call tf.function on the optimize method in the test before executing it. But don't worry if you don't: I'm just starting to take a look at the way we handle compilation and as part of that I'll try to get a sense of what testing we're missing.

khurram-ghani (approved):
Looks good!

hstojic:
> > we could easily do that but the user could still use the callback mode, we would then have to rely on them carefully reading the docs
> 
> If I understand correctly, you mean they would have to carefully read the keras docs to use their external callback properly, right? That is standard keras usage though. Seems to me it would be better to use keras where we can, but I guess we can't expect our users to be always familiar with keras.

No, I mean they would need to read our docs carefully, to use only one type of scheduler and not through callbacks as it would result in wrong learning rates

PR #741, opened 2023-06-08
Correctly deepcopy posterior cache and VGP parameters

uri-granta:


trieste/models/gpflow/models.py
```diff
@@ -1228,13 +1228,19 @@ def _ensure_variable_model_data(self) -> None:
...
+            old_q_sqrt = model.q_sqrt
             model.__init__(  # type: ignore[misc]
                 variable_data,
                 model.kernel,
                 model.likelihood,
                 model.mean_function,
                 model.num_latent_gps,
             )
+            model.q_mu.assign(old_q_mu)
+            model.q_sqrt.assign(old_q_sqrt)
```
khurram-ghani, line 1231: The VGP trieste model follows a different pattern to the others, i.e. it recreates the gpflow model; which seems messy. Also, update to `model.num_data` is missing. Would it be better to follow, e.g., what SGPR does?

```suggestion
        if not all(is_variable(x) and x.shape[0] is None for x in self.model.data):
            self.model.data = (
                tf.Variable(
                    self._model.data[0],
                    trainable=False,
                    shape=[None, *self._model.data[0].shape[1:]],
                ),
                tf.Variable(
                    self._model.data[1],
                    trainable=False,
                    shape=[None, *self._model.data[1].shape[1:]],
                ),
            )

        if not is_variable(self.model.num_data):
            self.model.num_data = tf.Variable(self.model.num_data, trainable=False)
```

trieste/models/gpflow/interface.py
```diff
@@ -71,9 +71,15 @@ def create_posterior_cache(self) -> None:
...
 
+    @abstractmethod
+    def _ensure_variable_model_data(self) -> None:
+        """Ensure GPflow data, which is normally stored in Tensors, is instead stored in
+        dynamically shaped Variables."""
+
     def __setstate__(self, state: dict[str, Any]) -> None:
         # when unpickling we may need to regenerate the posterior cache
         self.__dict__.update(state)
+        self._ensure_variable_model_data()
```
khurram-ghani, line 82: ~~Question: how do we ensure that after the deepcopy, the posterior cache will have dynamic shape, or is that not required?~~

khurram-ghani: Never mind.. I didn't process the following 2 lines correctly. It is clear now how that works :-)

trieste/models/gpflow/models.py
uri-granta: The problem is that VGP uses more complex internal state, including a couple of GPflow Parameters for q_mu and q_sqrt that also need to updated to be dynamically shaped. We could copy more of the VGP initialiser into this method, but that actually makes the code more complex not less. (Regarding num_data, the initializer converts that to a dynamically shaped Parameter too.)

uri-granta: And as far as I know there is no simpler way of changing the shape of gpflow Parameters than recreating them and assigning the old value.

khurram-ghani (approved):
LGTM!

PR #744, opened 2023-06-09
additional test problems

hstojic:
I have modified rosenbrock so its easier to create problems of different dimensionality and introduced the levy problem

### DIFF
--- a/trieste/objectives/single_objectives.py
+++ b/trieste/objectives/single_objectives.py
@@ -335,7 +335,7 @@ def levy(x: TensorType, d: int) -> TensorType:
     The Levy test function over :math:`[0, 1]^d`. This function has many local
     minima and a single global minimum. See https://www.sfu.ca/~ssurjano/levy.html for details.
     Note that we rescale the original problem, which is typically defined
-    over `[-10, 10]^d`, to be defined over a unit hypercube :math:`[0, 1]^d.
+    over `[-10, 10]^d`, to be defined over a unit hypercube :math:`[0, 1]^d`.
 
     :param x: The points at which to evaluate the function, with shape [..., d].
     :param d: The dimension of the function.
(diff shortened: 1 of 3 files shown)


trieste/objectives/single_objectives.py
```diff
@@ -330,31 +330,92 @@ def shekel_4(x: TensorType) -> TensorType:
...
+    :param x: The points at which to evaluate the function, with shape [..., d].
+    :param d: The dimension of the function.
     :return: The function values at ``x``, with shape [..., 1].
     :raise ValueError (or InvalidArgumentError): If ``x`` has an invalid shape.
     """
-    tf.debugging.assert_shapes([(x, (..., 4))])
+    tf.debugging.assert_greater_equal(d, 1)
+    tf.debugging.assert_shapes([(x, (..., d))])
+
+    w: TensorType = 1 + ((x * 20.0 - 10) - 1) / 4
```
uri-granta, line 348: Stupid question: why does levy and rosenstock need to know the dimension? Can't we just a dimension-agnostic function to define dimension-specific test problems (which after all also include a search space, so evaluating the function with incorrect inputs shouldn't be an issue)?

uri-granta (approved):

trieste/objectives/single_objectives.py
```diff
@@ -330,31 +330,92 @@ def shekel_4(x: TensorType) -> TensorType:
...
+    term1 = tf.pow(tf.sin(pi * w[..., 0:1]), 2)
+    term3 = (w[..., -1:] - 1) ** 2 * (1 + tf.pow(tf.sin(2 * pi * w[..., -1:]), 2))
+    wi = w[..., 0:-1]
+    wi_sum = tf.reduce_sum(
+        (wi - 1) ** 2 * (1 + 10 * tf.pow(tf.sin(pi * wi + 1), 2)), axis=-1, keepdims=True
+    )
+    return term1 + wi_sum + term3
```
vpicheny, line 355: It would make sense to have some basic rescaling I think? It doeesn't need to be finely calibrated, just to avoid having values beyond `[10, 10]`.

It can be done here or in `levy_8` if you want to keep this function intact?

From the website looks like just dividing by 100 would do the job.

trieste/objectives/single_objectives.py
hstojic: good point, they don't really... but lets leave it in, we can then outsource those asserts and have them only in that function

trieste/objectives/single_objectives.py
hstojic: good point, I'll do the rescaling

PR #745, opened 2023-06-12
Convert feature-function call inputs to be a tensor to fix pickling

khurram-ghani:
Fixes pickling issue introduced by #736, causing two integration tests to fail on develop.

It is important that the feature-function is called with a tensor, instead of a parameter (which inducing points can be). This is to ensure pickling works correctly. First time a Keras layer (i.e. feature-functions) is built, the shape of the input is used to set the input-spec. If the input is a parameter, the input-spec will not be for an ordinary tensor and pickling will fail.

Note: the call [here](https://github.com/secondmind-labs/trieste/blob/47accd9c7539b2619a0f5c2bbae0c76c3f401f5a/trieste/models/gpflow/sampler.py#L890) was not wrapped in `tf.convert_to_tensor`, since `flat_x` should already be a tensor.

### DIFF
--- a/trieste/models/gpflow/sampler.py
+++ b/trieste/models/gpflow/sampler.py
@@ -527,7 +527,7 @@ class RandomFourierFeatureTrajectorySampler(
         and observation noise variance :math:`\sigma^2`.
         """
         dataset = self._model.get_internal_data()
-        phi = self._feature_functions(dataset.query_points)  # [n, m]
+        phi = self._feature_functions(tf.convert_to_tensor(dataset.query_points))  # [n, m]
         D = tf.matmul(phi, phi, transpose_a=True)  # [m, m]
         s = self._model.get_observation_noise() * tf.eye(self._num_features, dtype=phi.dtype)
         L = tf.linalg.cholesky(D + s)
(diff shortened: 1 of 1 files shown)


trieste/models/gpflow/sampler.py
```diff
@@ -527,7 +527,7 @@ def _prepare_theta_posterior_in_design_space(self) -> tfp.distributions.Multivar
         and observation noise variance :math:`\sigma^2`.
         """
         dataset = self._model.get_internal_data()
-        phi = self._feature_functions(dataset.query_points)  # [n, m]
+        phi = self._feature_functions(tf.convert_to_tensor(dataset.query_points))  # [n, m]
```
hstojic, line 530: `Dataset` object should always have tensors, no?

hstojic, line 530: `Dataset` object should always have tensors, no?

hstojic: right, its `TensorType`, and although in practice we typically use tensors, it could be a numpy array

hstojic (approved):
looks ok to me

trieste/models/gpflow/sampler.py
khurram-ghani: Yes, I wasn't 100% sure what it could be, so I went for the safe option :-)

PR #746, opened 2023-06-13
Test with latest TF and numpy

uri-granta:


tests/unit/acquisition/test_rule.py
```diff
@@ -348,7 +348,7 @@ def test_joint_batch_acquisition_rule_acquire(
         _, query_point = points_or_stateful(None)
     else:
         query_point = points_or_stateful
-    npt.assert_allclose(query_point, [[0.0, 0.0]] * num_query_points, atol=1e-3)
+    npt.assert_allclose(tf.constant(query_point), [[0.0, 0.0]] * num_query_points, atol=1e-3)
```
hstojic, line 351: is this bcs `query_point` can be a numpy array?

tests/unit/models/gpflow/test_models.py
```diff
@@ -323,7 +328,7 @@ def test_gpflow_models_cached_predictions_correct(
     model, _ = gpflow_interface_factory(x, y)
 
     if after_model_optimize:
-        model._optimizer = BatchOptimizer(tf.optimizers.SGD(), max_iter=1)
+        model._optimizer = BatchOptimizer(gpflow.optimizers.Scipy(), max_iter=1)
```
hstojic, line 331: `BatchOptimizer` is meant to be used for stochastic optimizers in TF, why this change? was this for the VGP model that we have discussed?

hstojic (approved):
looks good to me, had only few minor comments

tests/unit/models/gpflow/test_models.py
uri-granta: Yes, it's due to https://github.com/secondmind-labs/trieste/issues/754. We can chat about how important we think working around this is in the next planning meeting.

tests/unit/acquisition/test_rule.py
uri-granta: I think it's because numpy typing is now better than tensorflow, so it's not sure that it's ok to pass query points into an npt function. (The stateful acquisition functions may also be confusing it.)

tests/util/models/keras/models.py
```diff
@@ -76,3 +78,11 @@ def trieste_deep_ensemble_model(
...
 
     return model, keras_ensemble, optimizer_wrapper
+
+
+def keras_optimizer_weights(optimizer: tf.keras.optimizers.Optimizer) -> Optional[TensorType]:
+    # optimizer weight API was changed in TF 2.11: https://github.com/keras-team/keras/issues/16983
+    if Version(tf.__version__) < Version("2.11"):
+        return optimizer.get_weights()
+    else:
+        return optimizer.variables[0]
```
khurram-ghani, line 88: Why return only the first variable for TF >= 2.11, is it because the sequence is hierarchical?

khurram-ghani (approved):

tests/util/models/keras/models.py
uri-granta: The new version now includes the weights among the variables. Previously when the weights hadn't been set yet `get_weights` would return None; now with `variables` we get a list with a single variable showing the unset weight. I expect this function is fairly fragile for other use cases, but is currently only being used in tests where it seems to work well enough.

PR #750, opened 2023-06-22
Support saving models as tf.saved_model

### LINKED ISSUE #725: Support saving models via tf.saved_models
It's currently difficult to export trieste model wrappers using tf.saved_models, in part because they don't inherit from tf.Module. We should investigate how to do this as cleanly as possible.

uri-granta:
Support exporting trieste models using tf.saved_model. Note that we don't inherit from tf.Module for two reasons:

1. it would limit type checking on models (since mypy treats tf.Module as Any)
2. it wouldn't be enough by at itself anyway as some of our wrappers include other components that are themselves not tf.Modules

Instead we add a `get_module_with_variables` method for all models, which returns a tf.Module containing references to all the variables contained within the model (using recursive traversal). This can then be extended with any tf,function-decorated methods we wish to export.

### DIFF
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -280,6 +280,38 @@ ax[0, 0].set_xlabel(r"$x_1$")
 ax[0, 0].set_xlabel(r"$x_2$")
 
 # %% [markdown]
+# ## Save the results
+#
+# Trieste provides two ways to save and restore optimization results. The first uses pickling to save the results (including the datasets and models), allowing them to be easily reloaded. **Note however that is not portable and not secure**. You should only try to load optimization results that you generated yourself on the same system (or a system with the same version libraries).
+
+# %%
+# save the results to a given path
+result.save("results_path")
+
+# load the results
+saved_result = trieste.bayesian_optimizer.OptimizationResult.from_path(
+    "results_path"
+)
+saved_result.try_get_final_model().model
+
+# %% [markdown]
+# The second approach is to save the model using the tensorflow SavedModel format. This requires explicitly exporting the methods to be saved and results in a portable model than can be safely loaded and evaluated, but which can no longer be used in subsequent BO steps.
+
+# %%
+# save the model to a given path, exporting just the predict method
+module = result.try_get_final_model().get_module_with_variables()
+module.predict = tf.function(
+    model.predict,
+    input_signature=[tf.TensorSpec(shape=[None, 2], dtype=tf.float64)],
+)
+tf.saved_model.save(module, "model_path")
+
+# load the model
+saved_model = tf.saved_model.load("model_path")
+saved_model.predict(initial_query_points)
+
+
+# %% [markdown]
 # ## LICENSE
 #
 # [Apache License 2.0](https://github.com/secondmind-labs/trieste/blob/develop/LICENSE)
--- a/docs/notebooks/expected_improvement.pct.py
+++ b/docs/notebooks/expected_improvement.pct.py
@@ -310,6 +310,11 @@ tf.saved_model.save(module, "model_path")
 saved_model = tf.saved_model.load("model_path")
 saved_model.predict(initial_query_points)
 
+# compare prediction results
+query_points = search_space.sample_sobol(1)
+print("Original model prediction: ", model.predict(query_points))
+print("Saved model prediction: ", saved_model.predict(query_points))
+
 
 # %% [markdown]
 # ## LICENSE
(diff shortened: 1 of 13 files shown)


khurram-ghani (commented):
LGTM, but will let @avullo review/approve it.

docs/notebooks/expected_improvement.pct.py
```diff
@@ -279,6 +279,38 @@ def build_model(data):
...
+module = result.try_get_final_model().get_module_with_variables()
+module.predict = tf.function(
+    model.predict,
+    input_signature=[tf.TensorSpec(shape=[None, 2], dtype=tf.float64)],
+)
+tf.saved_model.save(module, "model_path")
+
+# load the model
+saved_model = tf.saved_model.load("model_path")
+saved_model.predict(initial_query_points)
```
avullo, line 311: Nit: would it be worth showing we're getting the same predictions with the original and the saved models?

tests/unit/models/gpflow/test_utils.py
```diff
@@ -78,6 +81,51 @@ def test_gaussian_process_deep_copyable(gpflow_interface_factory: ModelFactoryTy
     npt.assert_array_compare(operator.__ne__, variance_f_copy_updated, variance_f_copy_updated2)
 
 
+def test_gaussian_process_tf_saved_model(gpflow_interface_factory: ModelFactoryType) -> None:
+    x = tf.constant(np.arange(5).reshape(-1, 1), dtype=gpflow.default_float())
+    model, _ = gpflow_interface_factory(x, fnc_2sin_x_over_3(x))
+
+    with tempfile.TemporaryDirectory() as path:
+        # create a trajectory sampler (used for smple method)
```
avullo:
```suggestion
        # create a trajectory sampler (used for sample method)
```

tests/unit/models/gpflux/test_models.py
```diff
@@ -380,6 +381,25 @@ def test_deepgp_deep_copyable() -> None:
     npt.assert_array_compare(operator.__ne__, variance_f_copy_updated_2, variance_f_copy_updated)
 
 
+@pytest.mark.skip(reason="Saving DeepGPs not working yet")
```
avullo: What's precisely not working atm?

tests/unit/models/gpflux/test_models.py
```diff
@@ -380,6 +381,25 @@ def test_deepgp_deep_copyable() -> None:
...
+def test_deepgp_tf_saved_model() -> None:
+    x = tf.constant(np.arange(5).reshape(-1, 1), dtype=gpflow.default_float())
+    model = DeepGaussianProcess(partial(single_layer_dgp_model, x))
+    with tempfile.TemporaryDirectory() as path:
+        module = model.get_module_with_variables()
+        module.predict = tf.function(
+            model.predict, input_signature=[tf.TensorSpec(shape=[None, 1], dtype=tf.float64)]
+        )
+        tf.saved_model.save(model, str(path))
```
avullo: Is there a reason as to why the `sample` method is not exported in this case?

tests/unit/models/gpflux/test_models.py
uri-granta: Was going to go back to this oops. Turns out it was just a typo in the test, now fixed.

tests/unit/models/gpflux/test_models.py
uri-granta: The sample test was originally written as a regression test for https://github.com/secondmind-labs/trieste/pull/753 but no reason not to include it in all the saved_model tests. Will fix.

tests/unit/models/gpflow/test_utils.py
```diff
@@ -78,6 +81,50 @@ def test_gaussian_process_deep_copyable(gpflow_interface_factory: ModelFactoryTy
     npt.assert_array_compare(operator.__ne__, variance_f_copy_updated, variance_f_copy_updated2)
 
 
+def test_gaussian_process_tf_saved_model(gpflow_interface_factory: ModelFactoryType) -> None:
```
avullo, line 84: Why is this in `test_utils` and not `test_models` for consistency with the gplux and keras cases?
Moreover, there's a lot of machinery with these tests that can be shared across models, perhaps collapse these into parametrised tests at the `test.unit.models.interfaces` level?

avullo (approved):
LGTM

tests/unit/models/gpflow/test_utils.py
uri-granta: Agree, though I think the (significant) refactoring should be in a separate PR. I put this in `test_utils` because that's where the deepcopy tests were, but both should probably go in `test_models`. And the interface tests are currently model agnostic (using toy models when necessary) but there is definitely scope for sharing lots more code.

PR #752, opened 2023-06-26
Make ensure_variable_model_data optional

uri-granta:
#741 added a `GaussianProcessRegression` abstract method `_ensure_variable_model_data`. This is an unnecessary breaking change, as some subclasses may not need to implement this (and any that do should know about it by failing garden path tests). This PR makes it optional instead.


khurram-ghani (approved):
LGTM!

PR #753, opened 2023-06-26
Always build features so trajectories can be saved before being used

khurram-ghani:
Always build feature-functions using a dummy shape. This is important for `tf.saved_model.save` to work, in cases where the user hasn't called the trajectory yet. By calling `build` in `__init__` we ensure all the variables always exist.


hstojic (approved):
I presume the test for this will be in @uri-granta 's PR so no need to introduce a test here?

khurram-ghani:
> I presume the test for this will be in @uri-granta 's PR so no need to introduce a test here?

Yes, that is correct.

PR #755, opened 2023-06-29
Add continue_optimization method to BO

### LINKED ISSUE #124: Add method to recover optimization from failed step
As a user, I want to be able to **easily** recover the optimization process from a failed step. Based on the relevant notebook, the following API may be a good approach
```python
class BayesianOptimizer:
    ...
    def recover(
        self, total_steps: int, history: List[Record[S]], acquisition_rule[S, SP]
    ) -> OptimizationResult[S]:
       result, new_history = self.optimize(
            total_steps - len(history),
            history[-1].datasets,
            history[-1].models,
            acquisition_rule,
            history[-1].acquisition_state
       )
       return OptimizationResult(result, history + new_history)
```
though we may wish to be able to substitute in a new model/state if desired

uri-granta:
Add a `continue_optimization` method to `BayesianOptimizer` to make it simpler to continue a previous optimization, either one which failed or one which passed but that you wish to continue for longer.

### DIFF
--- a/docs/notebooks/recovering_from_errors.pct.py
+++ b/docs/notebooks/recovering_from_errors.pct.py
@@ -61,18 +61,14 @@ acquisition_rule = trieste.acquisition.rule.TrustRegion()
 #
 # In this tutorial we'll try to complete fifteen optimization loops, which, with the broken observer, may take more than one attempt. The optimizer returns an `OptimizationResult`, which is simply a container for both:
 #
-#   * the final result, which uses a `Result` type (not to be confused with `OptimizationResult`) to safely encapsulate the final data, models and acquisition state if the process completed successfully, or an error if one occurred
-#   * the history of the successful optimization steps.
-#
-# We can access these with the `astuple` method.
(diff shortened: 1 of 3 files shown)


docs/notebooks/recovering_from_errors.pct.py
```diff
@@ -117,7 +105,7 @@ def __call__(self, x):
 from trieste.experimental.plotting import plot_bo_points, plot_function_2d
 
 if result.is_ok:
```
hstojic, line 107: interestingly I was never using this, but that aside, we should perhaps be consistent and use it in all notebooks as a pattern?

trieste/bayesian_optimizer.py
```diff
@@ -577,6 +599,8 @@ def optimize(
         :param early_stop_callback: An optional callback that is evaluated with the current
             datasets, models and optimization state before every optimization step. If this
             returns `True` then the optimization loop is terminated early.
+        :param start_step: The step number to start with. This number is removed from the
+            number of optimization steps run and is useful for restarting previous computations.
```
hstojic: perhaps explictly mention it is removed from `num_steps`?

hstojic (approved):
looks good, I think its a good addition

Issue #464, opened 2022-01-05
argmin in ExactThompsonSampler does not break ties


hstojic:
**Describe the bug**
`tf.math.argmin` used in `ExactThompsonSampler` does not break ties, it just takes the first element if the min is repeated. Its probably a fringe case since we are usually working with floats, but we should deal with this to be certain. Most likely argmin and argmax functions are used elsewhere in the code as well.

**To reproduce**
```python
import tensorflow as tf
a = [11, 10, 10, 15]
tf.math.argmin(input = a)
```

**Expected behaviour**
If ties are encountered they should be broken randomly.

hstojic:
some pointers (numpy): https://stackoverflow.com/questions/42071597/numpy-argmax-random-tie-breaking

a sketch:
```python
import tensorflow as tf
a = [1.0001, 1.0001, 26.9, 2.8, 166.32, 62.3]
a == tf.reduce_min(a)
```
This will produce a boolean tensor and then we probably can use
`tf.random.categorical` or `tf.random.shuffle`, but needs to be checked for multidimensional tensors and what if we get different numbers of ties in each dimension etc.

uri-granta:
Due to the likely impact on both performance and code complexity I don't think we should worry about this edge case. Suggest we close.

hstojic:
I was not thinking that performance would be hit much - hypothetically, how would you solve it, any thoughts?

uri-granta:
The simplest (and probably most efficient) approach is to call `tf.random.shuffle` before calling `tf.math.argmin`. Still not sure this is worthwhile though.

hstojic:
its a minor issue, but would be good to be correct here

uri-granta:
So I had a go at implementing this (see https://github.com/secondmind-labs/trieste/pull/757) and while it does work, it also has a **very significant** performance impact. The problem is that to ensure all valid argmins are possible we have to perform a complicated shuffle on the entire array (or really a combination of mulitple shuffles as it needs to handle different ties in different dimensions). There may be a more efficient way to handle this but I can't think of one. (We could add some noise instead of shuffling but that might result in the wrong answer.)

Suggest closing as wontfix, especially since we have yet to see that this is ever an issue in practice.

hstojic:
ok, agreed, its likely to be a very small issue and worth the additional cost - lets close it then but perhaps add a quick note in the docstring of the affected thompson samplers (probably all of them) that if there are ties they won't be broken randomly

PR #757, opened 2023-07-02
Use tie breakers in Thompson sampling

uri-granta:
Investigating adding tie breakers to argmin as suggested in https://github.com/secondmind-labs/trieste/issues/464.

While this does work, the solution here has a **significant** performance impact as it requires a complicated transposition and shuffle of the entire array before calling tf.math.argmin. Suggest abandoning as not worthwhile.


hstojic:
@uri-granta we can close this one, right?

PR #759, opened 2023-07-03
Tidy up Reducer class

### LINKED ISSUE #136: Add `map` equivalent to the acquisition reducer
As a user, I want to be able to apply arbitrary functions to the result of an acquisition function with minimal boilerplate, so that ...

One solution:
```python
def map(builder: AcquisitionFunctionBuilder, f: Callable[[...], ]) -> AcquisitionFunctionBuilder:
    class _Acq(AcquisitionFunctionBuilder):
        def prepare(...):
            return f(builder.prepare(...))

    return _Acq()
```

uri-granta:
* Make Reducer generic in model types
* Simplify repr method
* Add Map Reducer subclass

### DIFF
--- a/trieste/acquisition/combination.py
+++ b/trieste/acquisition/combination.py
@@ -44,8 +44,9 @@ class Reducer(AcquisitionFunctionBuilder[ProbabilisticModelType]):
 
         self._acquisitions = builders
 
-    def _repr_builders(self) -> str:
-        return ", ".join(map(repr, self._acquisitions))
+    def __repr__(self) -> str:
+        """"""
+        return "{}({})".format(self.__class__.__name__, ", ".join(map(repr, self._acquisitions)))
 
     def prepare_acquisition_function(
         self,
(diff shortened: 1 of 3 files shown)


trieste/acquisition/combination.py
```diff
@@ -130,19 +127,42 @@ def _reduce(self, inputs: Sequence[TensorType]) -> TensorType:
...
         """
         :param inputs: The outputs of each acquisition function.
         :return: The element-wise product of the ``inputs``.
         """
         return tf.reduce_prod(inputs, axis=0)
+
+
+class Map(Reducer[ProbabilisticModelType]):
+    """
+    :class:`Reducer` that accepts just one acquisition function builder and applies a
```
hstojic, line 146: can you add more details here, some example perhaps - a user would not easily see what is this for based on this description...

hstojic (approved):
approved with small request to add further description in the docstring

PR #760, opened 2023-07-03
Document pickling acquisition functions

### LINKED ISSUE #616: add to ask-tell interface notebook a note on pickling acquisition functions
some acquisition functions are stateful, its a useful use case to save it and load it when loading the record to be able to recreate the rule exactly - this allows e.g. checking exactly the acq function behaviour that was previously run, very useful for debugging if smth went wrong during an expensive ask-tell step

uri-granta:


docs/notebooks/ask_tell_optimization.pct.py
```diff
@@ -168,4 +168,23 @@ def plot_ask_tell_regret(ask_tell_result):
 
 
 # %% [markdown]
-# A word of warning. This serialization technique is not guaranteed to work smoothly with every Tensorflow-based model, so apply to your own problems with caution.
+# In some more complicated scenarios we may also wish to serialise the acquisition function, rather than creating a new one from the models and data, as it may contain stochastic internal data. This is not an issue here (where we used the default `EfficientGlobalOptimization` rule and `ExpectedImprovement` function) but we can demonstrate it neverthless:
```
hstojic: perhaps mention an acq function where there is an internal state?

hstojic (approved):

docs/notebooks/ask_tell_optimization.pct.py
uri-granta: Can you remind me what the archetypal use case for this was?

hstojic: any that is using trajectory samplers I think, bcs they have internal state acq func has stochastic internal state... not sure about any others, perhaps one of the Henry's acq functions has internal states?

PR #762, opened 2023-07-04
Fix slowtests

uri-granta:
**Related issue(s)/PRs:** https://github.com/secondmind-labs/trieste/pull/739

## Summary

Fix failures introduced by TURBO PR:

* Always test at least one BO step even if the initial points happen to have found the minimum!
* Increment max number of steps for a few failing tests (these are failing just due to rng changes and the early_stop_callback will prevent unnecessary steps anyway)
* Remove unnecessary breaking API change: keep `fit_initial_model` and make `fit_model` a simpler bool.

## PR checklist

### DIFF
--- a/trieste/bayesian_optimizer.py
+++ b/trieste/bayesian_optimizer.py
@@ -1071,6 +1071,7 @@ def stop_at_minimum(
     minimizers_atol: float = 0,
     minimizers_rtol: float = 0.05,
     objective_tag: Tag = OBJECTIVE,
+    minimum_step_number: Optional[int] = None,
 ) -> EarlyStopCallback[TrainableProbabilisticModel, object]:
     """
     Generate an early stop function that terminates a BO loop when it gets close enough to the
@@ -1083,6 +1084,7 @@ def stop_at_minimum(
     :param minimizers_atol: Absolute tolerance for minimizer point.
     :param minimizers_rtol: Relative tolerance for minimizer point.
     :param objective_tag: The tag for the objective data.
+    :param minimum_step_number: Minimum step number to stop at.
     :return: An early stop function that terminates if we get close enough to both the minimum
         and any of the minimizer points.
     """
@@ -1092,6 +1094,8 @@ def stop_at_minimum(
         _models: Mapping[Tag, TrainableProbabilisticModel],
         _acquisition_state: object,
     ) -> bool:
+        if minimum_step_number is not None and logging.get_step_number() < minimum_step_number:
+            return False
         dataset = datasets[objective_tag]
         arg_min_idx = tf.squeeze(tf.argmin(dataset.observations, axis=0))
         if minimum is not None:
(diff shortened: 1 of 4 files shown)


trieste/bayesian_optimizer.py
```diff
@@ -634,18 +645,12 @@ def optimize(
...
             raise ValueError("dicts of datasets and models must be populated.")
 
-        if fit_model not in ["all", "all_but_init", "never"]:
-            raise ValueError(
-                f"Expecting fit_model to be 'all', 'all_but_init' or 'never', "
-                f"but received {fit_model!r}"
-            )
-
-        if (fit_model in ["all", "all_but_init"]) and isinstance(acquisition_rule, TURBO):
+        if fit_model and isinstance(acquisition_rule, TURBO):
```
hstojic, line 648: why not having a subclass of `AcquisitionRule` like `AcquisitionRuleWithLocalModels`  that we check against, rather than specific acq rules?

trieste/bayesian_optimizer.py
```diff
@@ -1092,6 +1099,8 @@ def early_stop_callback(
         _models: Mapping[Tag, TrainableProbabilisticModel],
         _acquisition_state: object,
     ) -> bool:
+        if minimum_step_number is not None and logging.get_step_number() < minimum_step_number:
```
hstojic, line 1102: is perhaps getting the step variable this way problematic? recall the discussion with Alan about this being a global variable and when we have parallel processes it could give incorrect values?

we should probably have step variable as an attribute of the `BayesianOptimizer` object and pass it to the function? or can it be inferred from the state?

trieste/bayesian_optimizer.py
```diff
@@ -545,7 +555,8 @@ def optimize(
         *,
         track_state: bool = True,
         track_path: Optional[Path | str] = None,
-        fit_model: str = "all",
+        fit_model: bool = True,
+        fit_initial_model: bool = True,
```
khurram-ghani, line 559: Was the agreement to add multiple `fit*` arguments instead of a single one?

hstojic (commented):
looks ok but left some comments to consider first

khurram-ghani (approved):
LGTM!

trieste/bayesian_optimizer.py
uri-granta: Probably makes sense if we end up with another example of this. Best not to preemptively overengineer.

trieste/bayesian_optimizer.py
uri-granta: We can address this if and when we change how step_number is exposed, though I suspect we will find a way to make it accessible in the separate processes without breaking the current API. Certainly we don't want to change the EarlyStopCallback function type without a strong reason.

trieste/bayesian_optimizer.py
uri-granta: The TuRBO PR replaced the existing `fit_initial_model: bool` argument with a `fit_model: str` argument that covers all three use cases. This works, but is a breaking change. Furthermore, having two boolean arguments is actually simpler and less error prone.

PR #764, opened 2023-07-06
Fix versions.json links for latest sphinx

uri-granta:
**Related issue(s)/PRs:** #734

## Summary

Unpinning the version of sphinx theme used to generate the docs seems to have broken the links between versions (using the dropdown menu on the topright). Previously the links were absolute (always going to e.g. https://secondmind-labs.github.io/trieste/1.2.0/index.html), but now they're relative, appending the appropriate document path. This is a good thing, but we need to fix and deploy versions.json to get this to work.

## PR checklist

### DIFF
--- a/tests/unit/test_setup.py
+++ b/tests/unit/test_setup.py
@@ -73,7 +73,7 @@ def test_versions_is_valid(versions: list[dict[str, Any]]) -> None:
         assert isinstance(v, dict)
         assert set(v.keys()) == {"version", "url"}
         assert all(isinstance(value, str) for value in v.values())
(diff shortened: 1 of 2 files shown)


hstojic (approved):
looks okish to me (but not very confidently)

khurram-ghani (approved):
LGTM. I don't fully understand how it was broken, but trust that the links weren't working before and now they are fixed!

Was the issue that the new version of sphinx was appending paths at the end of `../index.html`?

uri-granta:
> Was the issue that the new version of sphinx was appending paths at the end of ../index.html?

Yes. It used to just use the links as they were, but at some point it started appending the relative path to them.

PR #768, opened 2023-07-18
Fix README code example

uri-granta:
**Related issue(s)/PRs:** 

## Summary

The code examples in README.md were failing due to imports.

I also investigated using [byexample](https://byexamples.github.io/byexample/languages/python) to automatically run code extracts in README.md but this is a pain due to how it expects both input and output. So instead suggest we rely on semantic versioning to not break these simple examples.

## PR checklist


hstojic (approved):
lgtm

PR #770, opened 2023-07-18
Start adding some check_shape decorations

### LINKED ISSUE #130: Document and check at runtime all function shapes
As a user, I want to know what shape tensors are expected where, and what shape tensor I should expect from functions and values.

As a contributor, I want arg and return value shapes documented so that I can more easily add features that conform to the design, and reason about how to fit functionality together.

This probably means adding shapes to all values and functions with tensors in their signatures, and shape checks for function arguments (shape checks for function return values are unnecessary - tests are more suitable for that). This is already done over much of the code base, but not all

uri-granta:
**Related issue(s)/PRs:** #130

## Summary

Start adding `check_shape` decorations. This has lots of pros over our existing ad hoc use of `tf.debugging.assert_shapes`:

1. better checking  (e.g. can check batch dimensions, input broadcasting, return types, object attributes and more)
2. more consistent checking (e.g. can inherit checks from abstract methods)
3. clearer code documentation (potentially including docstring rewriting, but see below)
4. clearer error messages, e.g.:

```
E             ShapeMismatchError: 
E             Tensor shape mismatch.
E               Function: expected_improvement.__call__
E                 Declared: /home/uri.granta/code/trieste/trieste/acquisition/function/function.py:215
E                 Argument: x
E                   Note:     This acquisition function only supports batch sizes of one
E                   Expected: [N..., 1, D]
E                   Actual:   [1, 2, 1]
```

However, note that:

* sphinx-autoapi (which we use for our docs, probably because it's simple and quick) doesn't support docstring rewriting, which means that the argument shapes won't automatically appear in the online docs (though they will appear in interactive docstrings)
* by default shape checking is disabled for anything wrapped in a tf.function (though we explicitly enable this in the unit tests) so performance impact should be limited; see below for further discussion (including impact on test runtime)

## Performance impact

Looking at integration test runtimes, the additional shape checks added here don't (currently) seem to have a significant perf impact by themselves. However:

* calling `set_enable_check_shapes(ShapeCheckingState.ENABLED)` for all the tests does slow things down somewhat, though this is primarily due to additional the gpflow shape checks which are now enabled rather than the new checks added in this PR. For example, for `test_bayesian_optimizer_with_gpr_finds_minima_of_scaled_branin`
  * `EfficientGlobalOptimization` goes from 20s to 40s
  * `AugmentedExpectedImprovement` goes from 30s to 50s (even though we didn't add any more checks for that function)
  * `Fantasizer` goes from 60s to 80s.
* conversely, explicitly disabling shape checking sometimes speeds things up over before, as it removes all gpflow checks; this doesn't affect  `EfficientGlobalOptimization` or `AugmentedExpectedImprovement`, but for `Fantasizer` it speeds it up from 60s to 30s! we should definitely document and advertise this (though it's more likely to affect research than production, as the latter is more likely to be compiling everything).

Overall, the test runtime for the unit tests doesn't seem to be impacted, but the (non-slow) integration test runtime increases from around 31 minutes to 38 minutes. If this becomes an issue we can easily split the integration tests into two parallel jobs.

## PR checklist

### DIFF
--- a/trieste/models/interfaces.py
+++ b/trieste/models/interfaces.py
@@ -19,6 +19,7 @@ from typing import Any, Callable, Generic, Optional, TypeVar
 
 import gpflow
 import tensorflow as tf
+from check_shapes import check_shapes
 from typing_extensions import Protocol, runtime_checkable
 
 from ..data import Dataset
@@ -46,6 +47,11 @@ class ProbabilisticModel(Protocol):
     """
 
     @abstractmethod
+    @check_shapes(
+        "query_points: [batch..., N, D]",
+        "return[0]: [batch..., N, E]",
+        "return[1]: [batch..., N, E]",
+    )
     def predict(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
         """
         Return the mean and variance of the independent marginal distributions at each point in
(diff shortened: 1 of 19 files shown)


trieste/models/interfaces.py
```diff
@@ -75,6 +85,10 @@ def sample(self, query_points: TensorType, num_samples: int) -> TensorType:
         """
         raise NotImplementedError
 
+    @check_shapes(
+        "query_points: [broadcast batch..., D]",
+        "return: [batch..., E...]",
+    )
```
khurram-ghani: Should this be returning two tensors, for mean and variance?
```suggestion
    @check_shapes(
        "query_points: [broadcast batch..., D]",
        "return[0]: [batch..., E...]",
        "return[1]: [batch..., E...]",
    )
```

khurram-ghani (commented):
Looks good. Just one comment.

trieste/models/interfaces.py
uri-granta: Good catch. I'll investigate why the predict_y unit tests didn't catch this!

uri-granta: Looks like the previous code was treating the tuple as a single tensor with an additional leading dimension of 2 as far as shape checking is concerned. Which kinda of makes sense: e.g. you can use this to specify that a function returns a list of certain size.

I've fixed it so that the two checks are specified separately like you suggested, though it's also possible to specify it more succinctly (but less clearly) as  `"return: [2, batch..., E...]"`.

khurram-ghani (approved):
LGTM.

Issue #617, opened 2022-08-31
model optimization history is not saved for gpflow models?


hstojic:
needs to be checked - e.g. keras and deep GPs attach a history object to the model after model fitting

uri-granta:
Since there is no mention of optimization history in the [gpflow documentation](https://gpflow.github.io/GPflow/2.5.2/search.html?q=history#), suggest closing this and raising a gpflow issue instead.

uri-granta:
Actually this might be most easily fixable in trieste (using step_callbacks) so reopening.

PR #772, opened 2023-07-24
Try using step_callback to get gpflow training loss history

uri-granta:
**Related issue(s)/PRs:** #617 

## Summary

Investigate using the the the step callback mechaism to track training loss history for the gpflow scipy optimizer.

Having said that, it may be better to fix this in gpflow directly, though that would involve figuring out how (or whether) to handle interaction with other callbacks, as well as how to extend OptimizeResult.

## PR checklist

### DIFF
--- a/trieste/models/gpflow/optimizer.py
+++ b/trieste/models/gpflow/optimizer.py
@@ -18,12 +18,14 @@ This module registers the GPflow specific loss functions.
 
 from __future__ import annotations
 
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Optional, Sequence
 
 import tensorflow as tf
 from gpflow.models import ExternalDataTrainingLossMixin, InternalDataTrainingLossMixin
+from gpflow.optimizers.scipy import StepCallback
 from tensorflow.python.data.ops.iterator_ops import OwnedIterator as DatasetOwnedIterator
 
+from ...types import TensorType
 from ..optimizer import LossClosure, TrainingData, create_loss_function
 
 
@@ -92,3 +94,30 @@ def _create_loss_function_external(
         builder.closure_builder = closure_builder
 
     return builder.closure_builder((X, Y))
+
+
+def loss_history_callback(
+    model: InternalDataTrainingLossMixin | ExternalDataTrainingLossMixin,
+    data: TrainingData,
+    history: list[TensorType],
+    compile: bool = True,
+) -> StepCallback:
+    """
+    A step callback to use with the :class:`~gpflow.optimizers.Scipy` optimizer that tracks
+    the training loss history. To use this with a `GPflowPredictor model, assign the result
+    to `model.optimizer.minimize_args["step_callback"]`.
+
+    :param model: The model to track the loss function for.
+    :param dataset: The data with which to track the loss function.
+    :param history: A list to append loss values to.
+    :param compile: Whether to compile with :func:`tf.function`.
+    :return: A `StepCallback` to use with the :class:`~gpflow.optimizers.Scipy` optimizer.
+    """
+    training_loss = create_loss_function(model, data, compile=compile)
+
+    def step_callback(
+        step: int, variables: Sequence[tf.Variable], values: Sequence[tf.Tensor]
+    ) -> None:
+        history.append(training_loss())
+
+    return step_callback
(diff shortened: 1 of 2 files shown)


trieste/models/gpflow/optimizer.py
```diff
@@ -92,3 +94,30 @@ def compiled_closure() -> tf.Tensor:
         builder.closure_builder = closure_builder
 
     return builder.closure_builder((X, Y))
+
+
+def loss_history_callback(
```
hstojic, line 99: I was thinking of getting the whole `OptimizeResult` object, rather than just loss: https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html

when we call `minimize` in gpflow.scipy, it should return it, right?
then `fun` should be actual loss value, but one gets all the other stuff that is useful for evaluating the optimisation: success, status, nfev, nit ...

hstojic (commented):
I think there should be a way of getting the output of `minimize` call which should return all that we need (see my comment) - and then there is no need to incur the extra cost of computing the loss again

trieste/models/gpflow/optimizer.py
uri-granta: We can certainly get the OptimizeResult, but I thought that didn't include any loss history (unlike the keras optimizer) just the number of evaluations? Isn't that what you wanted?

uri-granta: As an aside, how were you expecting to get the result? TrainableProbabilisticModel.optimize currently doesn't return anything (though Optimizer.optimize already does). If you're using BO or AskTell then there is no obvious way to return all the results when optimizing all the models, but I guess we could place them somewhere in the model wrappers (a bit like how the keras history is already available inside DeepEnsemble.model.history).

hstojic: ok, you want to say we would get here a loss with every iteration of the optimiser? rather than a single end result with `fun` in `OptimizeResult`?

hstojic: > we could place them somewhere in the model wrappers (a bit like how the keras history is already available inside DeepEnsemble.model.history)

yes, this was the solution I was thinking of - but not sure if we can attach a history object to gpflow model like we have with keras model object, if not and we store it in the wrapper, perhaps we should have a unified way for all models, store it in the wrapper object

uri-granta: > ok, you want to say we would get here a loss with every iteration of the optimiser? rather than a single end result with fun in OptimizeResult?

Yes, that's what this does. Essentially equivalent to `history.history["loss"]` on the keras result.

>  perhaps we should have a unified way for all models, store it in the wrapper object

We could potentially add some property to `TrainableProbabilisticModel` but note that the `keras.callbacks.History` is the result of calling `fit` on a keras model, not of calling `minimize` on the optimizer (which returns nothing) so the two aren't directly comparable. Will have a think.

uri-granta: We could also just store the most recent result of calling `Optimizer.optimize` in the optimizer itself, which would let us easily access the OptimizeResult object. Or even all the previous results (enabled via a `save_results: bool` option in the Optimizer).

hstojic: > > ok, you want to say we would get here a loss with every iteration of the optimiser? rather than a single end result with fun in OptimizeResult?

> Yes, that's what this does. Essentially equivalent to history.history["loss"] on the keras result.

I see, that would be useful, but since it can expensive perhaps leaving off by default? (depends how expensive it is...)

hstojic: > We could also just store the most recent result of calling Optimizer.optimize in the optimizer itself, which would let us easily access the OptimizeResult object. Or even all the previous results (enabled via a save_results: bool option in the Optimizer).

I think that is the best solution, all the history from each active learning step

for keras we can simply copy it from the model attribute, just so that we have a unified way of accessing the optimization history?

uri-granta: Remember that for keas the history is generated by the model calling the fit method, not the optimizer's minimize method.

hstojic: we could still attach the resulting history to the optimizer attribute, we would just need to do it from the model wrapper, in the optimize method, no?

uri-granta:
This will be covered by https://github.com/secondmind-labs/trieste/pull/774 and https://github.com/GPflow/GPflow/pull/2080

PR #773, opened 2023-07-26
Multi trust regions

khurram-ghani:
**Related branch:** `victor/trustregions`

## Summary
This PR adds support for multi-trust-regions acquisition rule; based-on and refactored from branch `victor/trustregions`.

The main classes are:
  * `CollectionSearchSpace`
    -  A new abstract class representing search spaces made up of a collection of sub-spaces.
  * `TaggedMultiSearchSpace`
    - A concrete implementation of `CollectionSearchSpace` for collections where all sub-spaces have the same dimensionality. This is the class used for current trust-region rules.
  * `TaggedProductSearchSpace`
    - The existing product search space class now inherits from `CollectionSearchSpace`.
  * `UpdatableTrustRegion`
    - A new abstract class representing search spaces that can be updated or reinitialised.
  * `BatchTrustRegion`
    - A new abstract class for trust region acquisition rules.
  * `SingleObjectiveTrustRegionBox`
    - A concrete implementation of `UpdatableTrustRegion` for the algorithm implemented in `victor/trustregions`.
  * `BatchTrustRegionBox`
    - A concrete implementation of `BatchTrustRegion` for the algorithm implemented in `victor/trustregions`.

The intention is that whenever a new trust region algorithm is implemented, in most cases, it would be sufficient to simply sub-class `UpdatableTrustRegion` and `BatchTrustRegion`.

Question: since `SingleObjectiveTrustRegionBox`/`BatchTrustRegionBox` is a specific trust-regions implementation, should it be moved to either a separate file or perhaps to a notebook? Is it generally useful?
Answer: yes these are useful and should be kept with the rules for now.

An example notebook will be added in a follow-on PR.

## PR checklist


trieste/space.py
```diff
@@ -920,17 +917,217 @@ def __init__(self, spaces: Sequence[SearchSpace], tags: Optional[Sequence[str]]
...
+
+        :param num_samples: The number of points to sample from each subspace.
+        :param seed: Optional tf.random seed.
+        :return: ``num_samples`` i.i.d. random points, sampled uniformly,
+            from each search subspace with shape '[num_samples, V, D]' , where V is the number of
+            subspaces and D is the search space dimension.
+        """
+        return tf.stack(self.subspace_sample(num_samples, seed), axis=1)
+
+    def _contains(self, value: TensorType) -> TensorType:
```
hstojic: why is the method private?

docs/notebooks/multi_trust_region.pct.py
```diff
@@ -0,0 +1,200 @@
...
+ff = obj_fun(Xplot).numpy()
+
+for step in range(num_steps):
+    print(f"step number {step}")
+    trieste.logging.set_step_number(step)
+
+    new_points = ask_tell.ask()
+    new_data = observer(new_points)
+    # monitor models after each tell
+    if summary_writer:
```
hstojic: notebooks would have didactic purpose hence we can eliminate some of the unnecessary components, like TB logging here

hstojic (commented):
did a pass, few minor comments for now - but still thinking about the code in `rule.py` file, trying to make up my mind about updatable class - are there efficiency gains with that?

trieste/space.py
khurram-ghani: That is not called directly by users, instead is called internally by `SearchSpace.contains`. That is the standard pattern used by all the search spaces, e.g. `DiscreteSearchSpace`, `Box`.

docs/notebooks/multi_trust_region.pct.py
khurram-ghani: Thanks. This is the original notebook. I haven't touched it yet (other than the obvious fixes to make it run). So this is not ready for review yet.

khurram-ghani:
> trying to make up my mind about updatable class - are there efficiency gains with that?

@hstojic
There aren't computation efficiency gains. This is about good design patterns (encapsulation); what is easy for developers to understand and maintain. In my opinion, having a separate update-able class makes it clear what it does and keeps most of the code self contained. Only when something applies to multiple regions (e.g. `maybe_reinitialise_subspaces`), it exists in the rule iteself.

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+
+class MultiTrustRegion(
+    AcquisitionRule[
+        types.State[Optional["MultiTrustRegion.State"], TensorType],
+        SearchSpace,
+        ProbabilisticModelType,
+    ],
+    Generic[ProbabilisticModelType, UpdateableSearchSpaceType],
+):
+    """Abstract class for multi trust region acquisition rules."""
```
hstojic: a bit more explanation would be useful, e.g. its a batch algorithm where each element in the batch is optimised with its own TR, with optimisation done in parallel

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+        """Reinitialise the box."""
+        _, dataset = self.get_single_model_and_dataset(models, datasets)
+
+        self.location = tf.squeeze(self.global_search_space.sample(1), axis=0)
+
+        global_lower = self.global_search_space.lower
+        global_upper = self.global_search_space.upper
+
+        if self._initialised:
+            scale = 0.1
```
hstojic: this was an error in the original code, this should be the same as the case below 0.5

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+        models: Optional[Mapping[Tag, ProbabilisticModelType]] = None,
+        datasets: Optional[Mapping[Tag, Dataset]] = None,
+    ) -> Tuple[Optional[ProbabilisticModelType], Optional[Dataset]]:
+        """Get a single model and dataset from the given arguments."""
+        if models is None:
+            model = None
+        elif OBJECTIVE in models.keys():
+            model = models[OBJECTIVE]
+        else:
+            raise ValueError(f"""if models is provided, it must contain the key {OBJECTIVE}""")
```
hstojic: this is too restrictive, we wouldn't be able to apply it then to active learning problems

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+
+        if datasets is None:
+            dataset = None
+        elif OBJECTIVE in datasets.keys():
+            dataset = datasets[OBJECTIVE]
+        else:
+            raise ValueError(f"""if datasets is provided, it must contain the key {OBJECTIVE}""")
+
```
hstojic: same as above

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+        global_lower = self.global_search_space.lower
+        global_upper = self.global_search_space.upper
+
+        if self._initialised:
+            scale = 0.1
+        else:
+            scale = 0.5
+            self._initialised = True
+
+        self._eps = scale * (global_upper - global_lower) / (5.0 ** (1.0 / global_lower.shape[-1]))
```
hstojic: this should be a method, as its something we would likely want to vary / experiment with

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+    @abstractmethod
+    def get_reinitialise_subspaces_mask(
+        self,
+        subspaces: Sequence[UpdateableSearchSpaceType],
+        models: Mapping[Tag, ProbabilisticModelType],
+        datasets: Optional[Mapping[Tag, Dataset]] = None,
+    ) -> TensorType:
+        """Get mask for subspaces that need to be reinitialised."""
+        ...
+
```
hstojic, line 1285: we should add some basic logging, but in a separate PR I guess

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+        tr_volume = tf.reduce_prod(self.upper - self.lower)
+        step_is_success = y_min < self.y_min - self._kappa * tr_volume
+        self._eps = self._eps / self._beta if step_is_success else self._eps * self._beta
+
+        self._lower = tf.reduce_max(
+            [self.global_search_space.lower, self.location - self._eps], axis=0
+        )
+        self._upper = tf.reduce_min(
+            [self.global_search_space.upper, self.location + self._eps], axis=0
+        )
```
hstojic: these might be private functions, being used in several spots...

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+
+        in_tr = self.contains(dataset.query_points)
+        in_tr_obs = tf.where(
+            tf.expand_dims(in_tr, axis=-1),
+            dataset.observations,
+            tf.constant(np.inf, dtype=dataset.observations.dtype),
+        )
+        ix = tf.argmin(in_tr_obs)
+        x_min = tf.gather(dataset.query_points, ix)
+        y_min = tf.gather(in_tr_obs, ix)
```
hstojic, line 1391: I know this was Victor's implementation, but its actually incorrect, it was a hack - say that the batch size is 50, here we check what was the best observation collected in all 50 TRs, while we actually want to get the observation corresponding to the point optimised with a particular TR in the previous iteration

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+        ),
+        axis=1,
+    )
+
+    # Return the mask of unique points; can get the actual points with:
+    #   tf.boolean_mask(points, mask)
+    return mask
+
+
+class MultiTrustRegionBox(MultiTrustRegion[ProbabilisticModelType, TrustRegionBox]):
```
hstojic: I find this to be a bit of an overhead - say I want to implement a different type of reinit, e.g. I subclass the `TrustRegionBox` and change things, but then I also likely need to subclass this one with a bit of repetition in the init,  overwriting `create_subspace`...

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,380 @@ def state_func(
...
+
+    def create_subspace(
+        self,
+        search_space: SearchSpace,
+    ) -> UpdateableSearchSpaceType:
+        """Create a subspace from the given search space.
+        This is the default implementation. Can be overridden by subclasses."""
+        return self._subspace_type()
+
+    def maybe_reinitialise_subspaces(
```
hstojic: so If I have something to compute that would be used in reinitializing each TR (e.g. location candidates), I would have to change this method?

I think this would be a common pattern, but then there would be some extra objects to pass to `reinitialize` method below, rather than just models and datasets - so perhaps allowing kwargs there?

hstojic (commented):
did another round, focusing on the changes in `rule`

trieste/acquisition/rule.py
khurram-ghani: Thanks. I still need to check and update the docstrings.

trieste/acquisition/rule.py
khurram-ghani: This is a helper method for the simple default case, to make developers' life easier. If a different model/dataset is required, than the expectation is to use the full datasets/models dictionaries directly instead of calling this.

trieste/acquisition/rule.py
khurram-ghani: Even-though it would start with observations from all 50 TRs, it will only consider the subset of points that are in the current TR. I am assuming this is what you mean by a hack. The downside being that if TRs overlap, the optimised point in the current TR may have been queried by a different region. So are we saying it will be better if we maintained separate datasets for each TR?

trieste/acquisition/rule.py
khurram-ghani: I'll add the method. However, I have a related question.

Are the concrete `TrustRegionBox` and `MultiTrustRegionBox` classes generally useful and worth keeping directly in Trieste (instead of e.g. a notebook)? My current intention was that the users would inherit from the abstract classes to implement their algorithms; and the above classes are just one implementation of an example algorithm.

trieste/acquisition/rule.py
khurram-ghani: Yes that is correct. I'll add kwargs to both `UpdateableSearchSpace.reinitialize` and `UpdateableSearchSpace.update`.

trieste/acquisition/rule.py
khurram-ghani: For significant changes to an algorithm (e.g. new parameters), yes it would require subclassing both.

To support simpler changes, e.g. to only the internals of `TrustRegionBox.reinitialize`, I can make a small tweak to `MultiTrustRegionBox` to use a new generic type instead of `TrustRegionBox`. This is exactly what the base `MultiTrustRegion` class does with `UpdateableSearchSpaceType`. It would avoid the need to override `create_subspace` and save subclassing in some cases.

trieste/space.py
```diff
@@ -875,20 +875,17 @@ def has_constraints(self) -> bool:
         return len(self._constraints) > 0
 
 
-class TaggedProductSearchSpace(SearchSpace):
+class CollectionSearchSpace(SearchSpace):
     r"""
-    Product :class:`SearchSpace` consisting of a product of
-    multiple :class:`SearchSpace`. This class provides functionality for
-    accessing either the resulting combined search space or each individual space.
+    A :class:`SearchSpace` consisting of a collection of multiple :class:`SearchSpace` objects,
```
```suggestion
    An abstract :class:`SearchSpace` consisting of a collection of multiple :class:`SearchSpace` objects,
```

trieste/space.py
```diff
@@ -920,17 +917,242 @@ def __init__(self, spaces: Sequence[SearchSpace], tags: Optional[Sequence[str]]
             )
 
         self._spaces = dict(zip(tags, spaces))
+        self._tags = tuple(tags)  # avoid accidental modification by users
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""CollectionSearchSpace(spaces =
```
uri-granta: Maybe use a default implementation to avoid having to override?

```suggestion
        return f"""{self.__class__.__name__}(spaces =
```

trieste/space.py
```diff
@@ -920,17 +917,242 @@ def __init__(self, spaces: Sequence[SearchSpace], tags: Optional[Sequence[str]]
...
+            return NotImplemented
+        return self._tags == other._tags and self._spaces == other._spaces
+
+
+class TaggedMultiSearchSpace(CollectionSearchSpace):
+    r"""
+    A :class:`SearchSpace` consisting of a collection of multiple :class:`SearchSpace` objects,
+    each with a unique tag. This class provides functionality for accessing each individual space.
+
+    Note that the individual spaces are not combined in any way, however all subspaces should have
```
uri-granta: Add longer explanation of how TaggedMultiSearchSpace can be used? At the moment it's a bit cryptic!

Also maybe have TaggedProductSearchSpace come before this in the file as it's conceptually simpler and used more widely?

trieste/space.py
```diff
@@ -920,17 +917,242 @@ def __init__(self, spaces: Sequence[SearchSpace], tags: Optional[Sequence[str]]
...
+
+    def __repr__(self) -> str:
+        """"""
+        return f"""TaggedMultiSearchSpace(spaces =
+                {[self.get_subspace(tag) for tag in self.subspace_tags]},
+                tags = {self.subspace_tags})
+                """
+
+    @property
+    def lower(self) -> TensorType:
```
uri-granta: Not sure how easily/accurately we can shape annotate this (though it might be possible, e.g. see the shapes of custom types section at https://gpflow.github.io/check_shapes/main/index.html) but we should at least docstring the behaviour, as it is counterintuitive.

trieste/acquisition/utils.py
```diff
@@ -135,3 +136,36 @@ def get_local_dataset(local_space: SearchSpaceType, dataset: Dataset) -> Dataset
...
+    # Create a boolean mask for each point.
+    mask = tf.reduce_all(
+        tf.logical_or(
+            distances_masked > tolerance, tf.eye(distances_masked.shape[0], dtype=tf.bool)
+        ),
+        axis=1,
+    )
+
+    # Return the mask of unique points; can get the actual points with:
+    #   tf.boolean_mask(points, mask)
```
uri-granta: Maybe include this comment int he docstring instead?

trieste/acquisition/optimizer.py
```diff
@@ -246,8 +249,26 @@ def optimize_continuous(
...
+                message=(
+                    f"""
+                    The batch shape of initial samples {tf.shape(candidates)[1]} must be equal to
+                    the vectorization of the target function {V}.
+                    """
+                ),
+            )
+            tiled_candidates = candidates  # [num_initial_samples, V, D]
+        else:
+            tiled_candidates = tf.tile(
```
uri-granta, line 279: maybe assert that the rank is 2 here? (partly for robustness in case we decide to break the shape sample api again, but mainly for code readability)

uri-granta (commented):
(a few minor initial comments while I was reminding myself of the space and optimizer work; will look at the rule next)

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,361 @@ def state_func(
         return state_func
 
 
+class UpdateableSearchSpace(SearchSpace):
```
uri-granta: Maybe rename this to `MultiTrustRegionSearchSpace` for now? In future, if we generalise this to other uses we can always rename it, but it's certainly not a totally generic updateable search space.

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,361 @@ def state_func(
...
+    """A search space that can be updated."""
+
+    @abstractmethod
+    def reinitialize(
+        self,
+        models: Optional[Mapping[Tag, ProbabilisticModelType]] = None,
+        datasets: Optional[Mapping[Tag, Dataset]] = None,
+        **kwargs: Any,
```
uri-granta: What's the purpose of kwargs?

More importantly, what's the purpose of the whole abstract class? As far as I can see we currently only use it for `TrustRegionBox`, which only updates the bounds (and initialises the eps) using the data, not the models. Do we gain anything by not just restricting `MultiTrustRegion` to `TrustRegionBox` and having init and update methods that just take a single dataset? We could always generalise it later.

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,361 @@ def state_func(
...
+    def update(
+        self,
+        models: Optional[Mapping[Tag, ProbabilisticModelType]] = None,
+        datasets: Optional[Mapping[Tag, Dataset]] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Update the search space using the given models and datasets."""
+        ...
+
+    def get_single_model_and_dataset(
```
uri-granta: Should this be a standalone function? Perhaps even a simple generic one?

```python
def get_tag(mapping: Optional[Mapping[Tag, T]], tag: Tag = OBJECTIVE) -> Optional[T]:
    if mapping is None:
        return None
    elif tag in mapping.keys():
        return mapping[tag]
    else:
        raise ValueError(f"tag {tag} not found in mapping")
```

which you can then use as `get_tag(models), get_tag(datasets)` (or elsehow as necessary).

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,361 @@ def state_func(
         return state_func
 
 
+class UpdateableSearchSpace(SearchSpace):
+    """A search space that can be updated."""
+
+    @abstractmethod
+    def reinitialize(
```
uri-granta: Can we call this `initialize` instead?

uri-granta (commented):
(Some initial rule comments in case you want to respond/chat. Will carry on looking.)

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1115,361 @@ def state_func(
...
+        acquisition_space: TaggedMultiSearchSpace
+        """ The search space. """
+
+        def __deepcopy__(self, memo: dict[int, object]) -> MultiTrustRegion.State:
+            acquisition_space_copy = copy.deepcopy(self.acquisition_space, memo)
+            return MultiTrustRegion.State(acquisition_space_copy)
+
+    def __init__(
+        self: "MultiTrustRegion[ProbabilisticModelType, UpdateableSearchSpaceType]",
+        subspace_type: Type[UpdateableSearchSpaceType],
```
uri-granta: (sorry if this comment comes twice; i thought i submitted it but it seems to have disappeared!)

This seems like a code smell! Is a runtime subspace_type argument only required for the default implementation of create_subspace? If so, why does the ABC even need a default implementation?

trieste/space.py
khurram-ghani: I have swapped the order of the classes and updated the docstring for `TaggedMultiSearchSpace`. Is that clearer? I can add a code example if that helps.

trieste/space.py
khurram-ghani: I have added just the docstrings for now.

trieste/acquisition/rule.py
khurram-ghani: We should discuss. This class just has `initialize` and `update` abstract methods, which seem pretty general.
Note: initially I wanted to make this a mixin, but I had issues with some bits of code expecting a search space.

trieste/acquisition/rule.py
khurram-ghani: `kwargs` were requested by @hstojic (see earlier comments), for use cases where you need other objects to decide when, e.g., to initialize.

The `TrustRegionBox` is just one example implementation and we need some generalisation already. @hstojic has various algorithms in the other repo that would also use these base classes; and some of them do use models. 

Datasets and models are the objects that are always passed to rules, so it made sense to at least keep these around.

khurram-ghani: There is also the potential follow-on work of using these base classes for TREGO and TURBO, where we could use this mechanism for local models.

trieste/acquisition/rule.py
khurram-ghani: Good idea. I'll add that in. Thanks.

trieste/acquisition/rule.py
khurram-ghani: It is also used in the `MultiTrustRegionBox` class (recent change), not just the default class. The default implementation is just for simple cases where the update-able search space doesn't take any arguments, and maybe save some typing. However, I agree that is unlikely to be very useful.

The sole reason `subspace_type` exists is for use as a factory, to create the correct update-able search space; potentially allowing the users to only subclass the update-able search space and re-use an existing rule.

However, I am open to suggestions for a better mechanism for this.

trieste/acquisition/rule.py
uri-granta: Do you foresee the same concrete rule supporting different types of UpdateableSearchSpace? This is also relevant to my question below about subspace_type. At the moment something feels very off about the abstractions here. Maybe we can chat tomorrow morning?

trieste/acquisition/rule.py
uri-granta: Except the assumptions about the type initialisers are all implicit. Passing in a subtype of TrustRegionBox with a different initialiser signature to MultiTrustRegionBox would type check but then fail to create a subspace.

Before I can suggest an alternative though I need to understand whether and how MultiTrustRegionBox might be used with a subtype of TrustRegionBox.

uri-granta: Another way of asking the same question: why can't we just

1. remove subspace_type from the initialisers
2. keep create_subpace abstract (no default implementation)
3. make MultiTrustRegionBox.create_subspace create and return a TrustRegionBox

trieste/acquisition/rule.py
uri-granta: Regarding kwargs: any rule that plans to use them will have to know that it's using a compatiable subtype of UpdateableSearchSpace. In which case we could instead add an additional update method to that type. E.g.

```python
subspace = self.create_subspace(search_space)
subspace.set_foo(bar) # defined for UpdateableSearchSpaceWithFoo subtypes
subspace.initialize(models, datasets)
```

Which will catch setting the extra parameter when it's not supported, something the current approach doesn't do.

trieste/acquisition/rule.py
```diff
@@ -1218,13 +1198,13 @@ def state_func(
                     MultiTrustRegion acquisition rule {self._tags}"""
 
             subspaces = []
-            for tag in self._tags:
+            for index, tag in enumerate(self._tags):
```
uri-granta: (simpler with zip?)

```suggestion
            for tag, init_subspace in zip(self._tags, self._init_subspaces):
```

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1103,275 @@ def state_func(
...
+                subspace.initialize(models, datasets)
+
+    @abstractmethod
+    def get_initialize_subspaces_mask(
+        self,
+        subspaces: Sequence[UpdateableSearchSpaceType],
+        models: Mapping[Tag, ProbabilisticModelType],
+        datasets: Optional[Mapping[Tag, Dataset]] = None,
+    ) -> TensorType:
+        """Get mask for subspaces that need to be initialized."""
```
uri-granta: (This could probably use a more in depth docstring, since it's an abstract method that people may want to override)

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1103,275 @@ def state_func(
...
+        self,
+        subspaces: Sequence[UpdateableSearchSpaceType],
+        models: Mapping[Tag, ProbabilisticModelType],
+        datasets: Optional[Mapping[Tag, Dataset]] = None,
+    ) -> None:
+        """Initialize subspaces if necessary.
+        Get a mask of subspaces that need to be initialized using an abstract method.
+        Initialize individual subpaces by calling the method of the UpdateableSearchSpaceType class.
+        """
+        mask = self.get_initialize_subspaces_mask(subspaces, models, datasets)
```
uri-granta, line 1254: (maybe add some checking on the mask shape? alternatively you may be able to add shape checking to get_initialize_subspaces_mask)

uri-granta (commented):
I'm happy with the architecture now, thanks! Tell me when you want me to review the little details (docstrings, shape checks, tests, etc).

trieste/acquisition/rule.py
khurram-ghani: `create_subspace` has now been removed. The users are required to create the initial sequence of subspaces externally and pass them to the rule. This has the benefit that the users can decide which subspaces to create.

trieste/acquisition/rule.py
khurram-ghani: I added and then reverted the kwargs change. Keeping the kwargs would strongly tie rules and subspaces together, potentially reducing re-usability (see other comments in the PR). Instead, users can add additional optional arguments to `initialize` and `update` methods in the derived search-space classes.

trieste/space.py
khurram-ghani: Added `check_shapes`.

tests/unit/acquisition/test_optimizer.py
```diff
@@ -203,6 +204,15 @@ def test_optimize_continuous_raises_with_invalid_vectorized_batch_size(batch_siz
         generate_continuous_optimizer()(search_space, (acq_fn, batch_size))
 
 
+def test_optimize_continuous_raises_with_mismatch_multi_search_space() -> None:
```
hstojic, line 207: can we add a simple test where we use several TRs/box search spaces that do not overlap at all and check that optim result for each element in the batch is indeed within the correspoinding box?

perhaps also check the ordering, so we have some additional confidence that points in the optim result are ordered according to TR ordering

trieste/space.py
```diff
@@ -976,10 +957,102 @@ def get_subspace(self, tag: str) -> SearchSpace:
...
+        }
+
+        self._subspace_starting_indices = dict(
+            zip(self._tags, tf.cumsum(subspace_sizes, exclusive=True))
+        )
+
+        self._dimension = tf.cast(tf.reduce_sum(subspace_sizes), dtype=tf.int32)
+
+    @property
+    @check_shapes("return: [D]")
```
uri-granta, line 1034: How much work would it be to specify this in the SearchSpace API and use inherit_shapes wherever appropriate (i.e. everywhere apart from TaggedMultiSearchSpace)? Ditto for sample and dimension, the last of which should presumably be trivial as it's the same everywhere.

trieste/space.py
```diff
@@ -1060,11 +1131,136 @@ def product(self, other: TaggedProductSearchSpace) -> TaggedProductSearchSpace:
...
+        is a member if it is a member of any of the subspaces.
+
+        :param value: A point or points to check for membership of this :class:`SearchSpace`.
+        :return: A boolean array showing membership for each point in value.
+        """
+        return tf.reduce_any(
+            [self.get_subspace(tag)._contains(value) for tag in self.subspace_tags], axis=0
+        )
+
+    def product(self, other: TaggedMultiSearchSpace) -> TaggedMultiSearchSpace:
```
uri-granta, line 1245: Are there unit tests for contains, product and discretize? (and if it turns out that any of the methods don't make sense then it might be better to raise NotImplementedError rather than have untested code)

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1104,324 @@ def state_func(
...
+                acquisition_space = TaggedMultiSearchSpace(subspaces, self._tags)
+            else:
+                acquisition_space = state.acquisition_space
+
+            state_ = MultiTrustRegion.State(acquisition_space)
+            points = self._rule.acquire(acquisition_space, models, datasets=datasets)
+
+            return state_, points
+
+        return state_func
```
uri-granta, line 1239: (this method reads nicely!)

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1104,324 @@ def state_func(
...
+        self,
+        global_search_space: SearchSpace,
+        beta: float = 0.7,
+        kappa: float = 1e-4,
+        min_eps: float = 1e-2,
+    ):
+        """
+        Calculates the bounds of the box from the location/centre and global bounds.
+
+        :param global_search_space: The global search space this search space lives in.
```
uri-granta, line 1300: Other parameters?

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1104,324 @@ def state_func(
...
+
+        super().__init__(global_search_space.lower, global_search_space.upper)
+
+    @property
+    def global_search_space(self) -> SearchSpace:
+        """The global search space this search space lives in."""
+        return self._global_search_space
+
+    @property
+    def location(self) -> TensorType:
```
uri-granta: We generally don't use properties where plain attributes will do (like here and eps) as it's not considered Pythonic. mypy will still be able to typecheck assignments and usages based on the initialisation (and it's always possible to add properties later if we need to modify setting or getting).

uri-granta (commented):
A few minor comments, but generally happy from an engineering persepctive.

trieste/acquisition/utils.py
```diff
@@ -135,3 +137,59 @@ def get_local_dataset(local_space: SearchSpaceType, dataset: Dataset) -> Dataset
...
+        mask = tf.tensor_scatter_nd_update(mask, [[idx]], [is_unique_point])
+
+    return mask
+
+
+T = TypeVar("T")
+""" An unbound type variable. """
+
+
+def get_value_for_tag(mapping: Optional[Mapping[Tag, T]], tag: Tag = OBJECTIVE) -> Optional[T]:
```
hstojic: this is likely to be useful more widely, perhaps move to trieste/util/misc.py?

trieste/space.py
```diff
@@ -976,10 +957,102 @@ def get_subspace(self, tag: str) -> SearchSpace:
...
+        """The lowest values taken by each space dimension, concatenated across subspaces."""
+        lower_for_each_subspace = self.subspace_lower
+        return (
+            tf.concat(lower_for_each_subspace, axis=-1)
+            if lower_for_each_subspace
+            else tf.constant([], dtype=DEFAULT_DTYPE)
+        )
+
+    @property
+    @check_shapes("return: [D]")
```
hstojic, line 1045: a bit of education for me if I may - how is this checked, i.e. against what? it picks up D from elsewhere in the class?

trieste/space.py
```diff
@@ -976,10 +957,102 @@ def get_subspace(self, tag: str) -> SearchSpace:
...
+        """The highest values taken by each space dimension, concatenated across subspaces."""
+        upper_for_each_subspace = self.subspace_upper
+        return (
+            tf.concat(upper_for_each_subspace, axis=-1)
+            if upper_for_each_subspace
+            else tf.constant([], dtype=DEFAULT_DTYPE)
+        )
+
+    @property
+    @check_shapes("return: []")
```
hstojic, line 1056: why empty `[]` here?

trieste/space.py
```diff
@@ -1060,11 +1131,136 @@ def product(self, other: TaggedProductSearchSpace) -> TaggedProductSearchSpace:
...
-    def __eq__(self, other: object) -> bool:
+
+class TaggedMultiSearchSpace(CollectionSearchSpace):
+    r"""
+    A :class:`SearchSpace` made up of a collection of multiple :class:`SearchSpace` subspaces,
+    each with a unique tag. All subspaces must have the same dimensionality.
+
+    Each subspace is treated as an independent space and not combined in any way. This class
+    provides functionality for accessing all the subspaces at once by using the usual search space
+    methods, as well as for accessing individual subspaces.
```
hstojic, line 1147: I would explicitly mention here TR as an example of how this class can be useful

trieste/space.py
```diff
@@ -976,10 +957,102 @@ def get_subspace(self, tag: str) -> SearchSpace:
...
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        return self._tags == other._tags and self._spaces == other._spaces
+
+
+class TaggedProductSearchSpace(CollectionSearchSpace):
+    r"""
+    Product :class:`SearchSpace` consisting of a product of
+    multiple :class:`SearchSpace`. This class provides functionality for
+    accessing either the resulting combined search space or each individual space.
```
hstojic, line 995: lets add here an explicit example of how this class can be useful (e.g. mixed search spaces)

trieste/space.py
```diff
@@ -1060,11 +1131,136 @@ def product(self, other: TaggedProductSearchSpace) -> TaggedProductSearchSpace:
...
+
+    Each subspace is treated as an independent space and not combined in any way. This class
+    provides functionality for accessing all the subspaces at once by using the usual search space
+    methods, as well as for accessing individual subspaces.
+
+    When accessing all subspaces at once from this class (e.g. `lower()`, `upper()`, `sample()`),
+    the returned tensors have an extra dimension corresponding to the subspaces.
+    """
+
+    def __init__(self, spaces: Sequence[SearchSpace], tags: Optional[Sequence[str]] = None):
```
hstojic, line 1157: shouldn't tags be Hashable's here as well?
@uri-granta ?

trieste/space.py
```diff
@@ -1060,11 +1131,136 @@ def product(self, other: TaggedProductSearchSpace) -> TaggedProductSearchSpace:
...
+
+        :return: The lower bounds of shape [V, D], where V is the number of subspaces and D is
+            the dimensionality of each subspace.
         """
-        if not isinstance(other, TaggedProductSearchSpace):
-            return NotImplemented
-        return self._tags == other._tags and self._spaces == other._spaces
+
+        lower = self.subspace_lower
+        return tf.stack(lower, axis=0) if lower else tf.constant([], dtype=DEFAULT_DTYPE)
```
hstojic: can this ever be `[]` if when class is initialised it has to have at least 1 space?

tests/unit/test_space.py
```diff
@@ -662,36 +664,359 @@ def test_box_deepcopy() -> None:
     npt.assert_allclose(box.upper, box_copy.upper)
 
 
-def test_product_space_raises_for_non_unqique_subspace_names() -> None:
+@pytest.mark.parametrize("search_space_type", [TaggedMultiSearchSpace, TaggedProductSearchSpace])
```
hstojic: perhaps make a fixture out of it somewhere on the top, it is repeated a lot of times below...

trieste/acquisition/rule.py
hstojic: exactly, if TRs overlap the y_min  could have been obtained by a different TR, or it could be from an old data - the issue I think is that then `success` is not correctly computed if the y_min here is not based on the optimised point

hstojic: @vpicheny any thoughts here? should the optimised point be a default or its something to test empirically?

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1104,324 @@ def state_func(
...
+    ) -> None:
+        """
+        Update this box, including centre/location, using the given dataset. If the size of the
+        box is less than the minimum size, re-initialize the box.
+        """
+        dataset = get_value_for_tag(datasets)
+
+        if tf.reduce_any(self.eps < self._min_eps):
+            self.initialize(models, datasets)
+            return
```
hstojic, line 1363: are we missing the duplicate rule for reinitialisation (i.e. same location as for another TR)?

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1104,324 @@ def state_func(
...
+
+        :param subspaces: The sequence of subspaces.
+        :param models: The model for each tag.
+        :param datasets: The dataset for each tag.
+        :return: A boolean mask of length V, where V is the number of subspaces.
+        """
+        ...
+
+
+class TrustRegionBox(Box, UpdateableSearchSpace):
```
hstojic: the name is a bit off I think - its technically correct, but it would be better to reflect the intended use-case I think, say `SingleObjectiveTrustRegionBox`?

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1104,324 @@ def state_func(
...
+            tf.constant(np.inf, dtype=dataset.observations.dtype),
+        )
+        ix = tf.argmin(in_tr_obs)
+        x_min = tf.gather(dataset.query_points, ix)
+        y_min = tf.gather(in_tr_obs, ix)
+
+        return tf.squeeze(x_min, axis=0), tf.squeeze(y_min)
+
+
+class MultiTrustRegionBox(MultiTrustRegion[ProbabilisticModelType, TrustRegionBox]):
```
hstojic: re the name, perhaps `batch` instead of `multi`?

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1104,324 @@ def state_func(
...
+    """Implements the :class:`MultiTrustRegion` *trust region* acquisition algorithm for a box."""
```
hstojic: add that this is intended for batching and single objetive optimization?

trieste/acquisition/utils.py
```diff
@@ -135,3 +137,59 @@ def get_local_dataset(local_space: SearchSpaceType, dataset: Dataset) -> Dataset
...
+    :param points: A tensor of points, with the first dimension being the number of points.
+    :param tolerance: The tolerance within which points are considered equal.
+    :return: A boolean mask for the unique points.
+    """
+
+    tolerance = tf.constant(tolerance, dtype=points.dtype)
+    n_points = tf.shape(points)[0]
+    mask = tf.zeros(shape=(n_points,), dtype=tf.bool)
+
+    for idx in tf.range(n_points):
```
hstojic, line 162: seems like there should be a more efficient way to do this, but perhaps its fast enough for our use case with batch sizes of up to few thousands?
any clue about scaling with dimensionality and number of points?

hstojic (commented):
looks good, I have left a bunch of comments, but no major point I think - I'm mostly curious how it would look like when starting to use the new division between the updatable search space and rule class itself, in the sense of subclassing them to create new variants

trieste/space.py
khurram-ghani: It should have been simple. So I gave it a go, but have discovered that `@inherit_check_shapes` doesn't seem to work with `@property` methods, whereas `@check_shapes` does. This requires further investigation, so I think we should leave this to a separate PR.

trieste/space.py
khurram-ghani: Yes there are unit tests for all of those. E.g.
- test_collection_space_contains_point
- test_multi_space___mul___
- test_collection_space_discretize_returns_search_space_with_only_points_contained_within_box
- ... plus others

trieste/acquisition/rule.py
khurram-ghani: Yes, will add them. Thanks.

trieste/acquisition/rule.py
khurram-ghani: That makes sense for `location` and I'll remove it. However, we already have a use case for overriding `eps` (in #778), so I'll keep that one.

khurram-ghani: Actually, as you have hinted, the derived class in #778 can add the `eps` properties there, so no need to keep them here :-)

trieste/space.py
khurram-ghani: `D` is an arbitrary string and in this case it is not checked against anything else; so the name is purely for documentation. This is checking that the method returns a vector of any length; so only checking that the rank is correct.

trieste/space.py
khurram-ghani: This checks that the method returns a scalar.

trieste/space.py
khurram-ghani: Good spot; it can't be []. I'll remove it.

trieste/acquisition/utils.py
khurram-ghani: Yes I agree that it is not ideal to use a loop. I had a [previous implementation](https://github.com/secondmind-labs/trieste/blob/309ef09dafd170d195b273a2a3fb2a46626947fb/trieste/acquisition/utils.py#L142-L179) that was more efficient without any loops, i.e. with parallel computation. However, it turned out to be faulty with further [testing](https://github.com/secondmind-labs/trieste/blob/492455722fa5099a0a6068bf51e9e215d71a092b/tests/unit/acquisition/test_utils.py#L118-L122). The fail occurred when multiple points were close to each other and the previous implementation would drop too many of them. Basically we should not drop a subsequent point that is close to a point already being dropped. I **think** the required operation is inherently sequential, but probably requires further investigation.

I ran some quick tests to check scaling. With 10K 2D points it took 40s, which isn't great. However the same test with `tf.function` compilation took 0.6s. Another test with 1K 20D points took 4s and 0.1s respectively. So we can at least compile it for cases where we need large number of points.

trieste/acquisition/rule.py
khurram-ghani: My proposal is to not check for duplication at initialisation or re-initialisation time inside the search space itself. Duplication would be checked on the next call to `maybe_initialize_subspaces` in the rule.  BTW we also don't check that recursively in `maybe_initialize_subspaces` either. We only check for duplicates once in that method, and don't check again after some regions may have been re-initialised.

tests/unit/acquisition/test_optimizer.py
khurram-ghani: There is a [test](https://github.com/secondmind-labs/trieste/blob/044f07cc52101bf5f5bb83acfe8c08af71ab00cc/tests/unit/acquisition/test_optimizer.py#L216-L238) for this now.

Here is a plot of the optimisation result, hopefully self-explanatory:
newplot (3)

trieste/acquisition/rule.py
khurram-ghani: We plan to integrate TURBO with these classes in a subsequent PR. TURBO would require us to put in place a mechanism to handle local models and datasets. My suggestion is to defer the topic of local datasets to that PR.

trieste/space.py
uri-granta: No. Tagged models and datasets use Hashables due to a specific product use case. However, tagged product spaces (and now multi search spaces) don't require this (and are unlikely to in future either), so keeping this as str is best for now.

trieste/acquisition/rule.py
khurram-ghani: That is a bit long. How about `SingleObjectiveTRBox` instead? I have changed to that for now.

trieste/space.py
khurram-ghani: I have added some text.

trieste/space.py
khurram-ghani: Done.

trieste/acquisition/rule.py
khurram-ghani: Done.

trieste/acquisition/rule.py
khurram-ghani: I have added some text.

trieste/acquisition/utils.py
uri-granta: My suggestion is to check in as is and raise an enhancement issue to improve the algorithm. (There may be something simple we can do using grid rounding but not sure whether this is worth spending time on now.)

trieste/acquisition/rule.py
uri-granta: I actually prefer the longer version as it makes it easier both to understand and to search for. Plus we already have other similarly named classes (e.g. `SingleObjectiveMultifidelityTestProblem`) and autocomplete means you won't have to type it all out :-)

uri-granta (approved):

trieste/acquisition/utils.py
khurram-ghani: Raised an enhancement issue.

trieste/acquisition/rule.py
vpicheny: I agree this is a hack that needs fixing, as we have done it in SD, which is to use the latest query point only that corresponds to this TR. 

I would suggest to write an issue and merge?

trieste/acquisition/rule.py
vpicheny: I agree, this is for a TR taken in isolation, duplicates are checked elsewhere.

trieste/acquisition/rule.py
```diff
@@ -1103,6 +1105,312 @@ def state_func(
         return state_func
 
 
+class UpdateableSearchSpace(SearchSpace):
```
vpicheny: I am a bit uncomfortable with this class, but not sure how to phrase my concern. The `SearchSpace` concept should be agnostic to any model or dataset.

Here we have much more than just a search space, we also encode the trust region logic to initialise it and then update it.

But we could have part of this logic elsewhere. For instance, initialise could depend only on a base search space (the global search space), and update only depends on the base search space and a Boolean (success or not). This way, the UpdatableSeachSpace is agnostic to the model, data, etc.

Then the success logic is deferred to the rule.

That being said, in my project with ellipsoidal TRs, I do need the model as well to update the trust region shape so maybe there's no way around it.

trieste/acquisition/rule.py
khurram-ghani: Issue created.

trieste/acquisition/rule.py
khurram-ghani: One option is to not call it a "search-space", since at least for now it is for trust regions only. My assumption was that in the future we may use the concept for other things. So, for the question "which generally useful states we need for a region/search-space to be able to update itself?", datasets and models seemed like the natural answer.

vpicheny: That makes sense. Perhaps yes the simplest way to avoid confusion is to call it something like `UpdatableTrustRegion`.

khurram-ghani: Rename done :-)

PR #774, opened 2023-08-01
Expose optimize results for trainable models

uri-granta:
**Related issue(s)/PRs:** #617

## Summary

Allow trainable models to return an optimization result and provide a backwards compatible method to make it accessible in the models. This currently includes OptimizeResult for gpflow models using the SciPy optimizer, keras.callback.History for keras models using the keras fit method, and lists of results for model stacks.

Note that exposing this in the model makes more sense than exposing it in the optimizer wrappers, as these are essentially an implementation detail and behave very differently between e.g. gpflow and keras.

## PR checklist

### DIFF
--- a/trieste/models/interfaces.py
+++ b/trieste/models/interfaces.py
@@ -140,12 +140,12 @@ class TrainableProbabilisticModel(ProbabilisticModel, Protocol):
 
     def optimize_and_save_result(self, dataset: Dataset) -> None:
         """
-        Optimize the model objective and save the optimization result in last_result.
+        Optimize the model objective and save the optimization result in last_optimization_result.
         """
         setattr(self, "_last_optimization_result", self.optimize(dataset))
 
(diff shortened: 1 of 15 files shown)


trieste/models/interfaces.py
```diff
@@ -143,15 +143,29 @@ def update(self, dataset: Dataset) -> None:
...
 
         :param dataset: The data with which to train the model.
+        :return: Any (optimizer-specific) optimization result.
         """
         raise NotImplementedError
 
+    def optimize_and_save_result(self, dataset: Dataset) -> None:
+        """
+        Optimize the model objective and save the optimization result in last_optimization_result.
+        """
```
hstojic, line 162: missing an argument in docstring here

hstojic (approved):
this is a neat solution, kudos! looks great!

PR #777, opened 2023-08-09
Notebook for trust regions

khurram-ghani:
**Related issue(s)/PRs:** #773 

## Summary
Added example notebook for three trust region algorithms.

## PR checklist


docs/notebooks/multi_trust_region.pct.py
```diff
@@ -0,0 +1,164 @@
+# ---
```
uri-granta: (haven't looked at yet, but you'll need to update tutorials.rst so there's a link to it; probably worth manually running the quality checks job on it too, or building locally, to catch any sphinx errors)

docs/notebooks/multi_trust_region.pct.py
```diff
@@ -0,0 +1,164 @@
...
+base_rule = trieste.acquisition.rule.EfficientGlobalOptimization(  # type: ignore[var-annotated]
+    builder=trieste.acquisition.ParallelContinuousThompsonSampling(),
+    num_query_points=num_query_points,
+)
+acq_rule = trieste.acquisition.rule.MultiTrustRegionBox(
+    init_subspaces, base_rule
+)
+
+# %% [markdown]
+# ## Run the optimization loop
```
hstojic: I think I would retain the boxes to illustrate what is going on with trust regions, its a nice visualisation and its useful to have that piece of code when developing new variants - its also a cool visualisation, we should have some notebooks with those as well :)

@vpicheny what do you think?

hstojic (commented):
its ok, but I would get back the visualisation of the TR boxes, see my comment below

docs/notebooks/multi_trust_region.pct.py
hstojic: we can add some new code to support that in experimental plotting package...

docs/notebooks/multi_trust_region.pct.py
khurram-ghani: I have added a link to that file. I had already run quality checks, but now locally also ran sphinx on it.

docs/notebooks/multi_trust_region.pct.py
vpicheny: I think it would make sense,  with add_patch(matplotlib.patches.Rectangle(...)) this is very simple and compact.

A notebook without visualisation of the TRs would not add as much value IMO

khurram-ghani: I have added the visualisation at the end. It is not particularly compact though, since it is plotting the contours/points (re-using existing functions) and creating the gif. Would it be useful to move any of it to the plotting package?

vpicheny (commented):
I am wondering if it would make sense to have a first section showing TREGO. Or ideally, TREGO, then batch-TR with Thompson sampling, and finally TurBO?

So that this notebook would give an overview of the TR features we have in trieste and would also act as an integration test?

khurram-ghani:
> I am wondering if it would make sense to have a first section showing TREGO. Or ideally, TREGO, then batch-TR with Thompson sampling, and finally TurBO?

Yes, makes sense. I'll add that in the notebook using the existing versions of TREGO and TurBO. We can then migrate to the batch implementations later in the respective PRs.

docs/notebooks/batch_trust_region.pct.py
```diff
@@ -0,0 +1,244 @@
...
+from matplotlib.colors import rgb2hex
+from matplotlib.patches import Rectangle
+from matplotlib.pyplot import cm
+
+colors = [
+    rgb2hex(color) for color in cm.rainbow(np.linspace(0, 1, num_query_points))
+]
+frames = []
+
+for step, hist in enumerate(result.history + [result.final_result.unwrap()]):
```
hstojic: thats' cool, reusing history :)

khurram-ghani:
@vpicheny @hstojic I have updated the notebook to cover all 3 trust region algorithms.

docs/notebooks/trust_region.pct.py
```diff
@@ -0,0 +1,283 @@
...
+    frames: List[np.ndarray], duration=5000
+) -> IPython.display.HTML:
+    gif_file = io.BytesIO()
+    imageio.mimsave(gif_file, frames, format="gif", loop=0, duration=duration)
+    gif = IPython.display.HTML(
+        '<img src="data:image/gif;base64,{0}"/>'.format(
+            base64.b64encode(gif_file.getvalue()).decode()
+        )
+    )
+    return gif
```
hstojic: this is fairly general code that is likely to be useful again, we should move this to experimental/plotting package?

docs/notebooks/trust_region.pct.py
```diff
@@ -0,0 +1,283 @@
...
+# ## Trust region `TurBO` acquisition rule
+#
+# Finally, we show how to run Bayesian optimization with the `TurBO` algorithm. This is a
+# trust region algorithm that uses local models and datasets to approximate the objective function
+# within one trust region.
+#
+# ### Create `TurBO` rule and run optimization loop
+#
+# This rule requires the specification of an aquisition base-rule for performing
+# optimization within the trust region; for our example we use `DiscreteThompsonSampling`.
```
hstojic: add a note that it is important to switch off global model fitting?

hstojic (approved):
looks good to me, few minor comments

docs/notebooks/trust_region.pct.py
khurram-ghani: Moving the second function to `plotting.py` will add dependencies to `imageio` and `IPython` to the main trieste requirements. Currently those dependencies are only present for notebooks. We probably don't want to do that, do we?

docs/notebooks/trust_region.pct.py
khurram-ghani: Done

docs/notebooks/trust_region.pct.py
hstojic: we have those plotting extras at https://github.com/secondmind-labs/trieste/blob/93f44b321e7b9112016288b48325c8863feb173e/setup.py#L52, is this for `experimental/plotting`? then we could add `imageio` to this list? 

Ipython dependency makes less sense but the function could return the `gif_file` instead and final conversion to `gif` could be done within notebook

docs/notebooks/trust_region.pct.py
```diff
@@ -0,0 +1,286 @@
...
+    gpflow_model = build_gpr(
+        initial_data, search_space, likelihood_variance=1e-7
+    )
+    return GaussianProcessRegression(gpflow_model)
+
+
+# %% [markdown]
+# ## Trust region `TREGO` acquisition rule
+#
+# First we show how to run Bayesian optimization with the `TREGO` algorithm. This is a trust region
```
vpicheny, line 66: Add here link to publication?

docs/notebooks/trust_region.pct.py
```diff
@@ -0,0 +1,286 @@
...
+# algorithm that alternates between regular EGO steps and local steps within one trust region.
+#
+# ### Create `TREGO` rule and run optimization loop
+#
+# We can run the Bayesian optimization loop by defining a `BayesianOptimizer` and calling its
+# `optimize` method with the trust region rule. Once the optimization loop is complete, the
+# optimizer will return one new query point for every step in the loop; that's 5 points in total.
+
+# %%
+acq_rule = trieste.acquisition.rule.TrustRegion()
```
vpicheny: Possibly it's important here to explain that TREGO is "meta" rule: it applies a base rule, either inside a Trust Region, or over the whole space. I guess the default is to use EfficientGlobalOptimization if nothing is specified but it would be good to make it explicit here? Would also show users what they need to change if they want to use a custom rule?

docs/notebooks/trust_region.pct.py
```diff
@@ -0,0 +1,286 @@
...
+    )
+
+    plot_bo_points(query_points, ax[0, 0], num_initial_data_points, arg_min_idx)
+
+
+plot_final_result(dataset)
+
+# %% [markdown]
+# We can also visualize the progress of the optimization by plotting the trust regions at each step.
+# The trust regions are shown as translucent boxes, with the current optimum point in each region
```
vpicheny: For TREGO it is either the TR or the full search space. Change to

We can also visualize the progress of the optimization by plotting the acquisition space at each step.
This space is either the full search space or the trust region, depending on the step, and is shown as a translucent box, ...
?

docs/notebooks/trust_region.pct.py
```diff
@@ -0,0 +1,286 @@
...
+#
+# This rule needs an initial number `num_query_points` of sub-spaces (or trust regions) to be
+# provided and performs optimization in parallel across all these sub-spaces. Each region
+# contributes one query point, resulting in each acquisition step collecting `num_query_points`
+# points overall. As the optimization process continues, the bounds of these sub-spaces are
+# dynamically updated.
+#
+# In addition, this rule requires the specification of a batch aquisition base-rule for performing
+# optimization; for our example we use `EfficientGlobalOptimization` coupled with
+# `ParallelContinuousThompsonSampling`.
```
vpicheny: suggestion:
... coupled with the `ParallelContinuousThompsonSampling` acquisition function.

docs/notebooks/trust_region.pct.py
```diff
@@ -0,0 +1,286 @@
...
+#
+# Note: the number of sub-spaces/regions must match the number of batch query points.
+
+# %%
+num_query_points = 5
+
+init_subspaces = [
```
vpicheny, line 198: Not related to the notebook in itself, but have you considered doing this by default at initialisation stage within `BatchTrustRegionBox` if `init_subspaces` isn't provided?

khurram-ghani: Yes, sort of. A previous version of the code created the regions within `BatchTrustRegionBox`. The issue was that we needed some sort of factory mechanism to allow the user to be able to create different types of regions (`SingleObjectiveTrustRegionBox` vs `TREGOBox` etc.), and that code was quite messy. Either we needed factory functions or the user provides type information. Creating the regions outside removed that complexity. Plus with the added benefit of allowing mix and match regions -- i.e. the framework supports mixing, e.g., a TREGO region with batch regions in the same rule. Though I don't know if that is an interesting use case.

Having said that, we could have a default behaviour where if `init_subspaces` is empty, `BatchTrustRegionBox` creates `N` regions of a fixed hard-coded type, e.g. `SingleObjectiveTrustRegionBox`, or perhaps `TREGOBox`. With `N` set equal to `num_query_points` inside the `base_rule`. Would that be useful?

vpicheny: I think it would be useful to have this initialisation, yes. More in line with the design choices we've made for trieste so far, which is to provide default for many methods so the entry cost for the user is reduced?

vpicheny: For the notebook, possibly it's good to leave it as it is so that the user understands that a choice is being made. 

But it would be good perhaps to add a sentence explaining (high level) what `SingleObjectiveTrustRegionBox` encodes? (ie initialisation and update)

docs/notebooks/trust_region.pct.py
```diff
@@ -0,0 +1,281 @@
...
+
+# %%
+plot_history(result)
+
+# %% [markdown]
+# ## Trust region `TurBO` acquisition rule
+#
+# Finally, we show how to run Bayesian optimization with the `TurBO` algorithm. This is a
+# trust region algorithm that uses local models and datasets to approximate the objective function
+# within one trust region.
```
vpicheny, line 243: not sure what the right wording is, but perhaps "within their respective trust regions"? otherwise it could sound like a single TR algorithm? ignore if you think this is sufficiently clear as is

docs/notebooks/trust_region.pct.py
vpicheny: I would actually explain just a bit more, ie that trieste maintains a global model but it is unused.

vpicheny (approved):

docs/notebooks/trust_region.pct.py
khurram-ghani: I did have slightly different wording before that implied multiple regions, but explicitly changed it to one region. The current TurBO implementation only supports one region. When that is replaced with new classes in a future PR, I can update the wording.

docs/notebooks/trust_region.pct.py
khurram-ghani: I have added some explanation for `SingleObjectiveTrustRegionBox` in the notebook.

It will be simpler to add the default initialisation in #778, as that PR already has changes to `init_subspaces` behaviour to allow a single value to be specified (in addition to a list). I'll expand that to include default initialisation.

khurram-ghani: I have added some explanation for `SingleObjectiveTrustRegionBox` in the notebook.

It will be simpler to add the default initialisation in #778, as that PR already has changes to `init_subspaces` behaviour to allow a single value to be specified (in addition to a list). I'll expand that to include default initialisation.

docs/notebooks/trust_region.pct.py
khurram-ghani: Yes EGO is the default. I have made that explicit now, with an explanation.

PR #778, opened 2023-08-10
TREGO using batch trust regions

khurram-ghani:
**Related issue(s)/PRs:** #773 

## Summary
This PR adds a TREGO implementation based on the new batch-trust-region classes and replaces the original TREGO `TrustRegion` implementation.

Existing notebook, unit tests and integration tests have been updated to use the new classes.

We have agreed that we don't need backwards compatibility -- see discussion below.

Previously, a `TREGO` rule was created like this:
``` python
rule = TrustRegion()
```

This will need to be replaced with the following:
``` python
rule = BatchTrustRegionBox(TREGOBox(search_space))
```

## PR checklist


hstojic (commented):
this is quite neat! surprisingly small amount of changes :)

tests/integration/test_bayesian_optimization.py
```diff
@@ -194,6 +195,23 @@ def GPR_OPTIMIZER_PARAMS() -> Tuple[str, List[ParameterSet]]:
                 ),
                 id="TrustRegion/MinValueEntropySearch",
             ),
+            pytest.param(
+                20,
+                MultiTrustRegionBox([TREGOBox(ScaledBranin.search_space)]),
```
hstojic: can we add it with batch size > 1 instead?

trieste/acquisition/rule.py
```diff
@@ -1422,6 +1422,45 @@ def get_initialize_subspaces_mask(
         return tf.logical_not(get_unique_points_mask(centres, tolerance=1e-6))
 
 
+class TREGOBox(TrustRegionBox):
```
hstojic: @vpicheny this is great no? we can now easily have TREGO as a subclass of `TrustRegionBox`

hstojic (commented):
I think this is great, but if we eliminate TREGO, this would break backward compatibility, right?
I think we should still do it...
@vpicheny ? @uri-granta ?

tests/integration/test_bayesian_optimization.py
khurram-ghani: It works with `batch_size > 1`, but have to of course use a vectorised acquisition function -- I have added a new test with `ParallelContinuousThompsonSampling`. My initial thoughts were what it would not be very useful, since the regions would not maintain diversity. TREGO simply uses the global dataset minimum to centre each region, so I suspected all regions would overlap.

However the mechanism to re-initialise the regions (when they are too close) works reasonably well here. The first region behaves as normal, but the other regions end up being in the opposite mode. So we have a mixture of regions in local and global modes.

vpicheny: I think it's good to allow TREGO to work with batches. Indeed, not necessarily useful, but also we should allow the framework to work in the most general way possible if it doesn't involve cumbersome code.

trieste/acquisition/rule.py
vpicheny: Yes this is good!

vpicheny (approved):
We don't need to be back-compatible so it's fine to use TREGOBox

trieste/acquisition/rule.py
```diff
@@ -1411,6 +1266,89 @@ def get_initialize_subspaces_mask(
...
+    def get_local_min(self, dataset: Optional[Dataset]) -> Tuple[TensorType, TensorType]:
+        if dataset is None:
+            raise ValueError("""dataset must be provided""")
+
+        # Always return the global minimum.
+        ix = tf.argmin(dataset.observations)
+        x_min = tf.gather(dataset.query_points, ix)
+        y_min = tf.gather(dataset.observations, ix)
+
+        return tf.squeeze(x_min, axis=0), tf.squeeze(y_min)
```
khurram-ghani, line 1395: @vpicheny @hstojic this is needed to match the behaviour of the previous implementation. It always used the global minimum regardless of the trust-region size or local/global mode. I have assumed we want to keep that behaviour here.

tests/integration/test_bayesian_optimization.py
```diff
@@ -182,17 +183,33 @@ def GPR_OPTIMIZER_PARAMS() -> Tuple[str, List[ParameterSet]]:
                 ),
                 id="MultipleOptimismNegativeLowerConfidenceBound",
             ),
-            pytest.param(20, TrustRegion(), id="TrustRegion"),
+            pytest.param(
+                20,
+                BatchTrustRegionBox(TREGOBox(ScaledBranin.search_space)),
+                id="TREGO",
+            ),
```
khurram-ghani, line 190: @vpicheny @hstojic double checking; are we okay with the instantiation of the rule being a bit more complex now? So, instead of `TrustRegion()`, we now need `BatchTrustRegionBox(TREGOBox(ScaledBranin.search_space))`.

vpicheny: yes I think this is fine

trieste/acquisition/rule.py
vpicheny: yes this makes sense, in particular because this is an implementation of an existing algorithm so we have to make sure it matches the publication.

vpicheny (approved):

PR #783, opened 2023-09-12
Allow TRs to be created with empty/single region

khurram-ghani:

## Summary
This PR updates batch trust-region classes to allow `init_subspaces` to be empty. The request came from [this](https://github.com/secondmind-labs/trieste/pull/777#discussion_r1321545492) discussion. Default subspaces are created if `init_subspaces` is None, with the number of spaces equal to the `batch_size`/`num_query_points` in the base rule.

Unfortunately, there is no mechanism or concept in the base `AcquistionRule` for batch size. So this PR uses the dynamic type of the base rule, with current support for `EfficientGlobalOptimization`.

This PR also allows users to specify a single init-subspace more easily. That is, `init_subspaces` can now be `None`, a single value or a sequence.

## PR checklist

### DIFF
--- a/trieste/acquisition/rule.py
+++ b/trieste/acquisition/rule.py
@@ -1435,6 +1435,16 @@ class BatchTrustRegionBox(BatchTrustRegion[ProbabilisticModelType, SingleObjecti
             )
             self._tags = tuple([str(index) for index in range(len(self._init_subspaces))])
 
+        # Ensure passed in global search space is always the same as the search space passed to
+        # the subspaces.
+        for subspace in self._init_subspaces:
+            assert subspace.global_search_space == search_space, (
+                "The global search space of the subspaces should be the same as the "
+                "search space passed to the BatchTrustRegionBox acquisition rule. "
+                "If you want to change the global search space, you should recreate the rule. "
+                "Note: all subspaces should be initialized with the same global search space."
+            )
+
         return super().acquire(search_space, models, datasets)
 
     @inherit_check_shapes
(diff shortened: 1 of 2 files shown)


trieste/acquisition/rule.py
```diff
@@ -1399,6 +1415,28 @@ class BatchTrustRegionBox(BatchTrustRegion[ProbabilisticModelType, SingleObjecti
     This is intended to be used for single-objective optimization with batching.
     """
 
+    def acquire(
+        self,
+        search_space: SearchSpace,
+        models: Mapping[Tag, ProbabilisticModelType],
+        datasets: Optional[Mapping[Tag, Dataset]] = None,
+    ) -> types.State[BatchTrustRegion.State | None, TensorType]:
+        if self._init_subspaces is None:
```
vpicheny, line 1424: naive question: why is initialisation done here and not within `__init__`?

khurram-ghani: It is because the subspaces we have (e.g. `SingleObjectiveTrustRegionBox`) need the global search space at initialisation time; which is not available in the rule `__init__`. I could have added it there as a required argument, but since that is already provided to `acquire`, I deferred that creation till then.

Note this does raise an important point. The subspaces take the reference to the global search space once (e.g. in first call to `acquire`). Theoretically the users can change the global search space to subsequent calls to `acquire`, but that would be ignored. I don't know how other rules deal with that or if that is even valid.

vpicheny: That makes sense!

Regarding the second point, if you change the global search space, then your trust region may become invalid. In that case it would make sense for the user to manually re-initialise it.

I guess we could either assert that the spaces match at every `acquire` call, or just ignore this corner case?

vpicheny (approved):

trieste/acquisition/rule.py
khurram-ghani: For the case where users create the subspaces outside the class, the global search space passed to `acquire` is completely ignored -- it is only used for this default initialisation case. So the full strict check check would be to assert the search space in every call to `acquire` is equal to the one stored inside every subspace. I hope that check is not too restrictive. I'll go ahead and add that.

Issue #782, opened 2023-08-25
Batch trust regions should use latest query point


khurram-ghani:
**Describe the bug**
The new batch trust region algorithms use the full dataset to determine the best point in each region. This is a hack, since the best point may have been queried by a different region in the previous iteration. Each region should only use its own query point from the previous iteration.

**To reproduce**
Previous discussion on this issue is in [this PR](https://github.com/secondmind-labs/trieste/pull/773#discussion_r1280707594).

khurram-ghani:
This was fixed in #788.

PR #788, opened 2023-10-10
Local models and datasets

khurram-ghani:
**Related issue(s)/PRs:** #782 

## Summary
This PR adds support for local models and datasets. The following scenarios are supported:
- global model and global dataset -- as previously
- local models and local datasets -- one-to-one mapping
- global model and local datasets -- many-to-one mapping

The initial dataset can be a global dataset sampled from the whole search space. This data will be replicated to each of the regions on the first iteration and subsequently each region will have an associated local dataset. For batch-TR algorithm, the dataset for each region are filtered after each iteration to only contain the points in the region (but `TREGO` doesn't do this).

Note: this replication of initial data can potentially cause an issue when a global model is being used, as the points may be repeated. This will only be an issue if regions overlap and both contain that initial data-point (as filtering would otherwise remove duplicates). The main way to avoid this issue is to provide local initial datasets, instead of a global initial dataset.

~~The `trust_region` notebook contains a new temporary `TEST` section, just to show how local models can be used in the notebook. It is worth noting in the `gif` that the query-points are filtered to be only inside the boxes for each iteration. This section is only for demonstration and will be removed before merging this PR. A follow-on PR will update the `TURBO` section to use local models and demonstrate this functionality.~~

The `BatchTrustRegion` rule acquisition returns rank 3 points, instead of rank 2 as for other rules (and previous trust-region rules). This means the users should use the new batched observer with this rule. That is already taken care of in `BayesianOptimizer`. However, with `AskTellOptimizer` the users should use the batched observer as follows:

``` python
from trieste.objectives.utils import mk_batch_observer

observer = ...
batch_observer = mk_batch_observer(observer)

new_points = ask_tell.ask()
new_data = batch_observer(new_points)
ask_tell.tell(new_data)
```

## PR checklist

### DIFF
--- a/docs/notebooks/trust_region.pct.py
+++ b/docs/notebooks/trust_region.pct.py
@@ -246,44 +246,6 @@ plot_final_result(dataset)
 plot_history(result)
 
 # %% [markdown]
-# ## TEST
-
-# %%
-num_query_points = 5
-
-init_subspaces = [
-    trieste.acquisition.rule.SingleObjectiveTrustRegionBox(search_space)
-    for _ in range(num_query_points)
-]
-base_rule = trieste.acquisition.rule.EfficientGlobalOptimization(
-    builder=trieste.acquisition.ParallelContinuousThompsonSampling(),
-    num_query_points=1,
-)
-batch_acq_rule = trieste.acquisition.rule.BatchTrustRegionBox(
-    init_subspaces, base_rule
-)
-
-bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
-
-num_steps = 5
-result = bo.optimize(
-    num_steps,
-    {trieste.observer.OBJECTIVE: initial_data},
-    trieste.acquisition.utils.copy_to_local_models(
-        build_model(), num_query_points
-    ),
-    batch_acq_rule,
-    track_state=True,
-)
-dataset = result.try_get_final_dataset()
-
-# %%
-plot_final_result(dataset)
-
-# %%
-plot_history(result)
-
-# %% [markdown]
 # ## Trust region `TurBO` acquisition rule
 #
 # Finally, we show how to run Bayesian optimization with the `TurBO` algorithm. This is a
--- a/docs/notebooks/trust_region.pct.py
+++ b/docs/notebooks/trust_region.pct.py
@@ -242,7 +242,9 @@ dataset = result.try_get_final_dataset()
 # %% [markdown]
 # ### Visualizing batch trust region results
 #
-# We visualize the results as before.
+# We visualize the results as before. However, please note that the initial query points (crosses) are
+# not highlighted in these plots. On each iteration, the batch trust region rule filters out points
+# that are not in the regions anymore; so there isn't an easy way to track the initial points.
 
 # %%
 plot_final_result(dataset, num_init_points=0)
--- a/docs/notebooks/trust_region.pct.py
+++ b/docs/notebooks/trust_region.pct.py
@@ -106,7 +106,9 @@ dataset = result.try_get_final_dataset()
 from trieste.experimental.plotting import plot_bo_points, plot_function_2d
 
 
-def plot_final_result(_dataset: trieste.data.Dataset) -> None:
+def plot_final_result(
+    _dataset: trieste.data.Dataset, num_init_points=num_initial_data_points
+) -> None:
     arg_min_idx = tf.squeeze(tf.argmin(_dataset.observations, axis=0))
     query_points = _dataset.query_points.numpy()
     _, ax = plot_function_2d(
@@ -117,7 +119,7 @@ def plot_final_result(_dataset: trieste.data.Dataset) -> None:
         contour=True,
     )
 
-    plot_bo_points(query_points, ax[0, 0], num_initial_data_points, arg_min_idx)
+    plot_bo_points(query_points, ax[0, 0], num_init_points, arg_min_idx)
 
 
 plot_final_result(dataset)
@@ -144,7 +146,10 @@ from trieste.experimental.plotting import (
 )
 
 
-def plot_history(result: trieste.bayesian_optimizer.OptimizationResult) -> None:
+def plot_history(
+    result: trieste.bayesian_optimizer.OptimizationResult,
+    num_init_points=num_initial_data_points,
+) -> None:
     frames = []
     for step, hist in enumerate(
         result.history + [result.final_result.unwrap()]
@@ -154,7 +159,7 @@ def plot_history(result: trieste.bayesian_optimizer.OptimizationResult) -> None:
             search_space.lower,
             search_space.upper,
             hist,
-            num_init=num_initial_data_points,
+            num_init=num_init_points,
         )
 
         if fig is not None:
@@ -203,15 +208,15 @@ plot_history(result)
 # contribute `Q` query points to the overall batch.
 
 # %%
-num_query_points = 6
+num_query_points = 5
 
 init_subspaces = [
-    trieste.acquisition.rule.SingleObjectiveTrustRegionBox(search_space, i)
-    for i in range(num_query_points)
+    trieste.acquisition.rule.SingleObjectiveTrustRegionBox(search_space)
+    for _ in range(num_query_points)
 ]
 base_rule = trieste.acquisition.rule.EfficientGlobalOptimization(  # type: ignore[var-annotated]
     builder=trieste.acquisition.ParallelContinuousThompsonSampling(),
-    # num_query_points=num_query_points,
+    num_query_points=num_query_points,
 )
 batch_acq_rule = trieste.acquisition.rule.BatchTrustRegionBox(
     init_subspaces, base_rule
@@ -230,12 +235,7 @@ bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
 num_steps = 5
 result = bo.optimize(
-    # num_steps, initial_data, build_model(), batch_acq_rule, track_state=True
-    num_steps,
-    {trieste.observer.OBJECTIVE: initial_data},
-    trieste.acquisition.utils.copy_to_local_models(build_model(), 2),
-    batch_acq_rule,
-    track_state=True,
+    num_steps, initial_data, build_model(), batch_acq_rule, track_state=True
 )
 dataset = result.try_get_final_dataset()
 
@@ -245,10 +245,48 @@ dataset = result.try_get_final_dataset()
 # We visualize the results as before.
 
 # %%
-plot_final_result(dataset)
+plot_final_result(dataset, num_init_points=0)
 
 # %%
-plot_history(result)
+plot_history(result, num_init_points=0)
+
+# %% [markdown]
+# ## TEST
+
+# %%
+num_query_points = 5
+
+init_subspaces = [
+    trieste.acquisition.rule.SingleObjectiveTrustRegionBox(search_space)
+    for _ in range(num_query_points)
+]
+base_rule = trieste.acquisition.rule.EfficientGlobalOptimization(
+    builder=trieste.acquisition.ParallelContinuousThompsonSampling(),
+    num_query_points=1,
+)
+batch_acq_rule = trieste.acquisition.rule.BatchTrustRegionBox(
+    init_subspaces, base_rule
+)
+
+bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
+
+num_steps = 5
+result = bo.optimize(
+    num_steps,
+    {trieste.observer.OBJECTIVE: initial_data},
+    trieste.acquisition.utils.copy_to_local_models(
+        build_model(), num_query_points
+    ),
+    batch_acq_rule,
+    track_state=True,
+)
+dataset = result.try_get_final_dataset()
+
+# %%
+plot_final_result(dataset, num_init_points=0)
+
+# %%
+plot_history(result, num_init_points=0)
 
 # %% [markdown]
 # ## Trust region `TurBO` acquisition rule
--- a/docs/notebooks/trust_region.pct.py
+++ b/docs/notebooks/trust_region.pct.py
@@ -211,7 +211,7 @@ init_subspaces = [
 ]
 base_rule = trieste.acquisition.rule.EfficientGlobalOptimization(  # type: ignore[var-annotated]
     builder=trieste.acquisition.ParallelContinuousThompsonSampling(),
-    #num_query_points=num_query_points,
+    # num_query_points=num_query_points,
 )
 batch_acq_rule = trieste.acquisition.rule.BatchTrustRegionBox(
     init_subspaces, base_rule
@@ -230,9 +230,9 @@ bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
 num_steps = 5
 result = bo.optimize(
-    #num_steps, initial_data, build_model(), batch_acq_rule, track_state=True
+    # num_steps, initial_data, build_model(), batch_acq_rule, track_state=True
     num_steps,
-    initial_data,
+    {trieste.observer.OBJECTIVE: initial_data},
     trieste.acquisition.utils.copy_to_local_models(build_model(), 2),
     batch_acq_rule,
     track_state=True,
--- a/docs/notebooks/trust_region.pct.py
+++ b/docs/notebooks/trust_region.pct.py
@@ -106,9 +106,7 @@ dataset = result.try_get_final_dataset()
 from trieste.experimental.plotting import plot_bo_points, plot_function_2d
 
 
-def plot_final_result(
-    _dataset: trieste.data.Dataset, num_init_points=num_initial_data_points
-) -> None:
+def plot_final_result(_dataset: trieste.data.Dataset) -> None:
     arg_min_idx = tf.squeeze(tf.argmin(_dataset.observations, axis=0))
     query_points = _dataset.query_points.numpy()
     _, ax = plot_function_2d(
@@ -119,7 +117,7 @@ def plot_final_result(
         contour=True,
     )
 
-    plot_bo_points(query_points, ax[0, 0], num_init_points, arg_min_idx)
+    plot_bo_points(query_points, ax[0, 0], num_initial_data_points, arg_min_idx)
 
 
 plot_final_result(dataset)
@@ -146,10 +144,7 @@ from trieste.experimental.plotting import (
 )
 
 
-def plot_history(
-    result: trieste.bayesian_optimizer.OptimizationResult,
-    num_init_points=num_initial_data_points,
-) -> None:
+def plot_history(result: trieste.bayesian_optimizer.OptimizationResult) -> None:
     frames = []
     for step, hist in enumerate(
         result.history + [result.final_result.unwrap()]
@@ -159,7 +154,7 @@ def plot_history(
             search_space.lower,
             search_space.upper,
             hist,
-            num_init=num_init_points,
+            num_init=num_initial_data_points,
         )
 
         if fig is not None:
@@ -242,15 +237,13 @@ dataset = result.try_get_final_dataset()
 # %% [markdown]
 # ### Visualizing batch trust region results
 #
-# We visualize the results as before. However, please note that the initial query points (crosses) are
-# not highlighted in these plots. On each iteration, the batch trust region rule filters out points
-# that are not in the regions anymore; so there isn't an easy way to track the initial points.
+# We visualize the results as before.
 
 # %%
-plot_final_result(dataset, num_init_points=0)
+plot_final_result(dataset)
 
 # %%
-plot_history(result, num_init_points=0)
+plot_history(result)
 
 # %% [markdown]
 # ## TEST
@@ -285,10 +278,10 @@ result = bo.optimize(
 dataset = result.try_get_final_dataset()
 
 # %%
-plot_final_result(dataset, num_init_points=0)
+plot_final_result(dataset)
 
 # %%
-plot_history(result, num_init_points=0)
+plot_history(result)
 
 # %% [markdown]
 # ## Trust region `TurBO` acquisition rule
--- a/docs/notebooks/trust_region.pct.py
+++ b/docs/notebooks/trust_region.pct.py
@@ -199,15 +199,15 @@ plot_history(result)
 # Note: the number of sub-spaces/regions must match the number of batch query points.
 
 # %%
-num_query_points = 5
+num_query_points = 6
 
 init_subspaces = [
-    trieste.acquisition.rule.SingleObjectiveTrustRegionBox(search_space)
-    for _ in range(num_query_points)
+    trieste.acquisition.rule.SingleObjectiveTrustRegionBox(search_space, i)
+    for i in range(num_query_points)
 ]
 base_rule = trieste.acquisition.rule.EfficientGlobalOptimization(  # type: ignore[var-annotated]
     builder=trieste.acquisition.ParallelContinuousThompsonSampling(),
-    num_query_points=num_query_points,
+    #num_query_points=num_query_points,
 )
 batch_acq_rule = trieste.acquisition.rule.BatchTrustRegionBox(
     init_subspaces, base_rule
@@ -226,7 +226,12 @@ bo = trieste.bayesian_optimizer.BayesianOptimizer(observer, search_space)
 
 num_steps = 5
 result = bo.optimize(
-    num_steps, initial_data, build_model(), batch_acq_rule, track_state=True
+    #num_steps, initial_data, build_model(), batch_acq_rule, track_state=True
+    num_steps,
+    initial_data,
+    trieste.acquisition.utils.copy_to_local_models(build_model(), 2),
+    batch_acq_rule,
+    track_state=True,
 )
 dataset = result.try_get_final_dataset()
 
--- a/tests/integration/test_ask_tell_optimization.py
+++ b/tests/integration/test_ask_tell_optimization.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 import copy
 import pickle
 import tempfile
-from typing import Callable
+from typing import Callable, Tuple
 
 import numpy.testing as npt
 import pytest
@@ -49,30 +49,27 @@ from trieste.types import State, TensorType
 # We use a copy of these for a quicker test against a simple quadratic function
 # (copying is necessary as some of the acquisition rules are stateful).
 OPTIMIZER_PARAMS = (
-    "num_steps, reload_state, acquisition_rule_fn, num_models",
+    "num_steps, reload_state, acquisition_rule_fn",
     [
         pytest.param(
-            20, False, lambda: EfficientGlobalOptimization(), 1, id="EfficientGlobalOptimization"
+            20, False, lambda: EfficientGlobalOptimization(), id="EfficientGlobalOptimization"
         ),
         pytest.param(
             20,
             True,
             lambda: EfficientGlobalOptimization(),
-            1,
             id="EfficientGlobalOptimization/reload_state",
         ),
         pytest.param(
             15,
             False,
             lambda: BatchTrustRegionBox(TREGOBox(ScaledBranin.search_space)),
-            1,
             id="TREGO",
         ),
         pytest.param(
             16,
             True,
             lambda: BatchTrustRegionBox(TREGOBox(ScaledBranin.search_space)),
-            1,
             id="TREGO/reload_state",
         ),
         pytest.param(
@@ -85,20 +82,21 @@ OPTIMIZER_PARAMS = (
                     num_query_points=3,
                 ),
             ),
-            1,
             id="BatchTrustRegionBox",
         ),
         pytest.param(
             10,
             False,
-            lambda: BatchTrustRegionBox(
-                [SingleObjectiveTrustRegionBox(ScaledBranin.search_space) for _ in range(3)],
-                EfficientGlobalOptimization(
-                    ParallelContinuousThompsonSampling(),
-                    num_query_points=2,
+            (
+                lambda: BatchTrustRegionBox(
+                    [SingleObjectiveTrustRegionBox(ScaledBranin.search_space) for _ in range(3)],
+                    EfficientGlobalOptimization(
+                        ParallelContinuousThompsonSampling(),
+                        num_query_points=2,
+                    ),
                 ),
+                3,
             ),
-            3,
             id="BatchTrustRegionBox/LocalModels",
         ),
         pytest.param(
@@ -110,7 +108,6 @@ OPTIMIZER_PARAMS = (
                 ).using(OBJECTIVE),
                 num_query_points=3,
             ),
-            1,
             id="LocalPenalization",
         ),
         pytest.param(
@@ -121,10 +118,22 @@ OPTIMIZER_PARAMS = (
                     ScaledBranin.search_space,
                 ).using(OBJECTIVE),
             ),
-            1,
             id="LocalPenalization/AsynchronousGreedy",
         ),
     ],
+)
+
+
+AcquisitionRuleFunction = (
+    Callable[[], AcquisitionRule[TensorType, SearchSpace, TrainableProbabilisticModel]]
+    | Callable[
+        [],
+        AcquisitionRule[
+            State[TensorType, AsynchronousRuleState | BatchTrustRegionBox.State],
+            Box,
+            TrainableProbabilisticModel,
+        ],
+    ]
 )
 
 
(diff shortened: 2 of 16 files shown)


trieste/acquisition/rule.py
```diff
@@ -148,6 +148,32 @@ def acquire_single(
...
+        :return: The updated datasets.
+        """
+        # Account for the case where there may be an initial dataset that is not tagged
+        # per region. In this case, only the global dataset will exist in datasets. We
+        # want to copy this initial dataset to all the regions.
+        #
+        # If a tag from tagged_output does not exist in datasets, then add it to
+        # datasets by copying the dataset from datasets with the same tag-prefix.
+        # Otherwise keep the existing dataset from datasets.
+        # TODO: this could mean that when we have a global model, the global dataset
```
vpicheny: why TODO?

khurram-ghani: I hadn't tidied that yet, that was part of the "double-check". It is done now. Thanks.

tests/unit/test_ask_tell_optimization.py
```diff
@@ -171,7 +180,7 @@ def test_ask_tell_optimizer_copies_state(
     ask_tell.tell(new_data)
     state_end: Record[None] = ask_tell.to_record(copy=copy)
 
-    assert_datasets_allclose(state_start.dataset, init_dataset if copy else init_dataset + new_data)
+    assert_datasets_allclose(state_start.dataset, init_dataset)
```
khurram-ghani: Note: this is needed due to the following change in `ask_tell_optimization.py`. The previous version updated the dataset in place, which is not the case anymore.

```
-        for tag in self._datasets:
-            self._datasets[tag] += new_data[tag]

+        updated_datasets = {}
+        for tag, new_dataset in new_data.items():
+            _, old_dataset = get_value_for_tag(
+                self._datasets, [tag, LocalizedTag.from_tag(tag).global_tag]
+            )
+            assert old_dataset is not None
+            updated_datasets[tag] = old_dataset + new_dataset
+        self._datasets = updated_datasets
```

trieste/acquisition/rule.py
```diff
@@ -148,6 +162,36 @@ def acquire_single(
             datasets=None if dataset is None else {OBJECTIVE: dataset},
         )
 
+    def update_datasets(
```
vpicheny: Do we have to have this as a method of `AcquisitionRule`? 

In general an acquisition rule should not update any dataset, it should just receive a dataset and model and make a query point recommendation based on this. So this could confuse users.

trieste/acquisition/rule.py
```diff
@@ -972,6 +1019,107 @@ def update(
         """
         ...
 
+    def _get_tags(self, tags: Set[Tag]) -> Tuple[Set[Tag], Set[Tag]]:
```
vpicheny, line 1010: Would this be better placed in utils/misc?

trieste/acquisition/rule.py
```diff
@@ -944,6 +988,9 @@ def acquire(
 class UpdatableTrustRegion(SearchSpace):
     """A search space that can be updated."""
 
+    def __init__(self) -> None:
+        self.index: Optional[int] = None
```
vpicheny: explain what this index is in a docstring? might also be beneficial to give it a more explicit name?

khurram-ghani: I have added an explanation (and a new arg), plus renamed it.

trieste/acquisition/rule.py
khurram-ghani: I think it is quite specific to the current use case, i.e. it returns two list containing only global parts of the tags. I propose we leave it here for now, but move it in the future if we find another use case for it.

trieste/acquisition/rule.py
khurram-ghani: The reason is that the behaviour of datasets-update changes depending on which rule we have. For most rules the datasets are simply concatenated; but for `BatchTrustRegion` rule there is quite a bit of code to filter-out points outside the subspaces; and perhaps we will have other ways in the future. So it seemed sensible to encapsulate that functionality as part of the rule.

We could add it in the `BayesianOptimizer`/`AskTellOptimizer` instead (which is where the previous dataset update was). However, that code would then likely need to use dynamic type detection, and have different behaviour depending on the rule. That seemed more messy. We can certainly go down that route though.

trieste/objectives/utils.py
```diff
@@ -57,3 +61,54 @@ def mk_multi_observer(**kwargs: Callable[[TensorType], TensorType]) -> MultiObse
...
+            # Convert to a mapping with a default key.
+            obs_or_dataset = {default_key: obs_or_dataset}
+
+        datasets = {}
+        for key, dataset in obs_or_dataset.items():
+            # Include overall dataset and per batch dataset.
+            obs = dataset.observations
+            qps = tf.reshape(qps, [-1, batch_size, qps.shape[-1]])
+            obs = tf.reshape(obs, [-1, batch_size, obs.shape[-1]])
+            _datasets = {
```
uri-granta: not sure this is clearer than just looping:

```python
datasets[key] = dataset
for i in range(batch_size):
    datasets[LocalTag(key, i)] = Dataset(qps[:, i], obs[:, i])
```

trieste/objectives/utils.py
```diff
@@ -57,3 +61,54 @@ def mk_multi_observer(**kwargs: Callable[[TensorType], TensorType]) -> MultiObse
...
+
+    @check_shapes("qps: [n_points, batch_size, n_dims]")
+    # Note that the return type is not correct, but that is what mypy is happy with. It should be
+    # Mapping[Tag, Dataset] if key is not None, otherwise Dataset.
+    # One solution is to create two separate functions, but that will result in some duplicate code.
+    def _observer(qps: TensorType) -> Mapping[Tag, Dataset]:
+        # Call objective with rank 2 query points by flattening batch dimension.
+        # Some objectives might only expect rank 2 query points, so this is safer.
+        batch_size = qps.shape[1]
+        qps = tf.reshape(qps, [-1, qps.shape[-1]])
```
uri-granta: I would use a different variable name for clarity, e.g. `flat_qps`.

trieste/acquisition/rule.py
khurram-ghani: I have changed this so that dataset updates are back in `BayesianOptimizer`/`AskTellOptimizer`. The rule now contains only filtering of the datasets.

The reason for this change (other than this discussion) was that the previous version threw away points as soon as they were outside a region. If a region ever came back over the same space, those points were lost. The new version keeps the full datasets around. Hence a region can use points irrespective of when they were collected.

trieste/utils/misc.py
```diff
@@ -220,21 +233,84 @@ def map_values(f: Callable[[U], V], mapping: Mapping[K, U]) -> Mapping[K, V]:
...
-    :param tag: A tag.
-    :return: The value of the tag in the mapping, or None if the mapping is None.
-    :raises ValueError: If the tag is not in the mapping and the mapping is not None.
+    :param tags: A tag or a sequence of tags. Sequence is searched in order.
+    :return: The chosen tag and value of the tag in the mapping, or None for each if the mapping is
+        None.
+    :raises ValueError: If none of the tags are in the mapping and the mapping is not None.
     """
 
+    if isinstance(tags, Tag):
```
uri-granta: This is very fragile, as `Tag = Hashable`, so passing in a tuple for example would be interpreted as a single tag rather than a sequence of them. My suggestion would be to force the user to be explicit as to whether they're passing in a single tag or a sequence of them. Easiest way is probably to use *args:

```python
def get_value_for_tag(
    mapping: Optional[Mapping[Tag, T]], *tags: Tag)
) -> Tuple[Optional[Tag], Optional[T]]:
    if not tags:
        tags = [OBJECTIVE]
```

And either use `get_value_for_tag(mapping, tag)` or `get_value_for_tag(mapping)` as before, or `get_value_for_tag(mapping, *tags)`.

trieste/utils/misc.py
```diff
@@ -220,21 +233,84 @@ def map_values(f: Callable[[U], V], mapping: Mapping[K, U]) -> Mapping[K, V]:
...
+        tags = [tags]
+
     if mapping is None:
-        return None
-    elif tag in mapping.keys():
-        return mapping[tag]
+        return None, None
     else:
-        raise ValueError(f"tag '{tag}' not found in mapping")
+        matched_tags = sorted(set(tags) & set(mapping.keys()), key=tags.index)
```
uri-granta: Clearer and more efficient?

```suggestion
        matched_tag = next(tag for tag in tags if tag in mapping.keys(), None)
        if matched_tag is None:
            raise ValueError(...)
        return matched_tag, mapping[matched_tag]
```

trieste/utils/misc.py
```diff
@@ -220,21 +233,84 @@ def map_values(f: Callable[[U], V], mapping: Mapping[K, U]) -> Mapping[K, V]:
...
+            return matched_tags[0], mapping[matched_tags[0]]
+        else:
+            raise ValueError(f"none of the tags '{tags}' found in mapping")
+
+
+@dataclass(frozen=True)
+class LocalTag:
+    """Manage a tag for a local model or dataset."""
+
+    global_tag: Tag
```
uri-granta, line 253: Maybe add docstrings (or at least comments) for these?

Also consider the class name. The fact that `LocalTag(OBJECTIVE, None).is_local is False` is certainly counterintuitive! Maybe `LocalizedTag`?

trieste/utils/misc.py
```diff
@@ -220,21 +233,84 @@ def map_values(f: Callable[[U], V], mapping: Mapping[K, U]) -> Mapping[K, V]:
...
+    @property
+    def is_local(self) -> bool:
+        """Return True if the tag is a local tag."""
+        return self.local_index is not None
+
+    @property
+    def tag(self) -> Tag:
+        """The local tag."""
+        if self.is_local:
+            return f"{self.global_tag}__{self.local_index}"
```
uri-granta: Maybe make this more robust, to avoid a clash with some other user generated tag scheme. Perhaps:

```suggestion
            return f"{self.global_tag}__LocalTag{self.local_index}"
```

**Update**: actually wouldn't it be better to not stringify the global tag at all? In fact, can't we just use the `LocalTag` object itself as the tag, as it's frozen and hence hashable?

trieste/utils/misc.py
```diff
@@ -220,21 +233,84 @@ def map_values(f: Callable[[U], V], mapping: Mapping[K, U]) -> Mapping[K, V]:
...
+        else:
+            return self.global_tag
+
+    def __repr__(self) -> str:
```
uri-granta: If you use `@dataclass(frozen=True, repr=True, eq=True)` then repr, hash and eq will be defined for you. And arguably you shouldn't stringify the Tag at all (see previous comment).

trieste/utils/misc.py
```diff
@@ -220,21 +233,84 @@ def map_values(f: Callable[[U], V], mapping: Mapping[K, U]) -> Mapping[K, V]:
...
+        """Return True if the local tag is equal to the other object."""
+        return hash(self) == hash(other)
+
+    @staticmethod
+    def from_tag(tag: Union[Tag, LocalTag]) -> LocalTag:
+        """Return a LocalTag from a given tag."""
+        if isinstance(tag, LocalTag):
+            return tag
+        else:
+            tag = str(tag)
```
uri-granta: This will become much simpler if we use LocalTags as tags unchanged! (Though if that doesn't work out then I might have some other comments...)

uri-granta (commented):
Some initial comments on the first couple of files. Will hopefully have more time to continue looking at this.

trieste/utils/misc.py
khurram-ghani: `repr=True, eq=True` is the default anyway, so as you say, I didn't need those implementations. Removed.

trieste/utils/misc.py
khurram-ghani: In a few places the tag was used directly already, but in other places it was used as a string. Removing the string version completely works quite nicely.

trieste/utils/misc.py
```diff
@@ -220,21 +221,56 @@ def map_values(f: Callable[[U], V], mapping: Mapping[K, U]) -> Mapping[K, V]:
...
+def get_value_for_tag(
+    mapping: Optional[Mapping[Tag, T]], *tags: Tag
+) -> Tuple[Optional[Tag], Optional[T]]:
+    """Return the value from a mapping for the first tag found from a sequence of tags.
 
     :param mapping: A mapping from tags to values.
-    :param tag: A tag.
-    :return: The value of the tag in the mapping, or None if the mapping is None.
-    :raises ValueError: If the tag is not in the mapping and the mapping is not None.
+    :param tags: A tag or a sequence of tags. Sequence is searched in order.
```
uri-granta: (mention default tag if no tags are given explicitly?)

trieste/acquisition/utils.py
```diff
@@ -138,6 +142,22 @@ def get_local_dataset(local_space: SearchSpaceType, dataset: Dataset) -> Dataset
     return local_dataset
 
 
+def copy_to_local_models(
```
uri-granta, line 146: (add basic unit test?)

trieste/bayesian_optimizer.py
```diff
@@ -96,18 +98,26 @@ class Record(Generic[StateType]):
     @property
     def dataset(self) -> Dataset:
         """The dataset when there is just one dataset."""
-        if len(self.datasets) == 1:
-            return next(iter(self.datasets.values()))
+        # Ignore local datasets.
+        datasets: Mapping[Tag, Dataset] = dict(
```
uri-granta: More Pythonic to use comprehension?

```suggestion
        datasets: Mapping[Tag, Dataset] = {
            k : v for k, v in self.datasets.items()
            if not LocalizedTag.from_tag(k).is_local
        }
```

uri-granta: Update: since you use this repeatedly, it might even be worth defining a helper function in acquisition.utils

```python
def ignoring_local_models(model_map: Mapping[Tag, T]) -> Mapping[Tag, T]:
    return { ... }
```

trieste/bayesian_optimizer.py
```diff
@@ -627,19 +645,24 @@ def optimize(
...
         # reassure the type checker that everything is tagged
         datasets = cast(Dict[Tag, Dataset], datasets)
         models = cast(Dict[Tag, TrainableProbabilisticModelType], models)
 
         if num_steps < 0:
             raise ValueError(f"num_steps must be at least 0, got {num_steps}")
 
-        if datasets.keys() != models.keys():
+        # Get set of dataset and model keys, ignoring any local tag index.
+        datasets_keys = {LocalizedTag.from_tag(tag).global_tag for tag in datasets.keys()}
```
uri-granta, line 657: (simpler?)

```suggestion
        datasets_keys = ignoring_local_models(datasets).keys()
        models_keys = ignoring_local_models(models).keys()
```

trieste/bayesian_optimizer.py
```diff
@@ -731,26 +757,42 @@ def optimize(
...
                         if isinstance(observer_output, Mapping)
                         else {OBJECTIVE: observer_output}
                     )
 
-                    datasets = {tag: datasets[tag] + tagged_output[tag] for tag in tagged_output}
+                    # See explanation in ask_tell_optimization.tell().
+                    updated_datasets = {}
+                    for tag, new_dataset in tagged_output.items():
+                        _, old_dataset = get_value_for_tag(
+                            datasets, *[tag, LocalizedTag.from_tag(tag).global_tag]
```
uri-granta:
```suggestion
                            datasets, tag, LocalizedTag.from_tag(tag).global_tag
```

trieste/ask_tell_optimization.py
```diff
@@ -188,19 +189,23 @@ def __init__(
...
+        if not isinstance(models, Mapping):
+            models = {OBJECTIVE: models}
 
         # reassure the type checker that everything is tagged
-        datasets = cast(Dict[Tag, Dataset], datasets)
         models = cast(Dict[Tag, TrainableProbabilisticModelType], models)
 
-        if datasets.keys() != models.keys():
+        # Get set of dataset and model keys, ignoring any local tag index.
+        datasets_keys = {LocalizedTag.from_tag(tag).global_tag for tag in datasets.keys()}
```
uri-granta, line 208: (same comments for this file as for bayesian_optimizer.py)

trieste/acquisition/rule.py
```diff
@@ -148,6 +162,16 @@ def acquire_single(
             datasets=None if dataset is None else {OBJECTIVE: dataset},
         )
 
+    def filter_datasets(self, datasets: Mapping[Tag, Dataset]) -> Mapping[Tag, Dataset]:
+        """
+        Filter the datasets.
```
uri-granta: Maybe explain why you might want to override this?

trieste/acquisition/rule.py
```diff
@@ -972,6 +1004,111 @@ def update(
...
+        global_tags = set()
+        for tag in tags:
+            ltag = LocalizedTag.from_tag(tag)
+            if not ltag.is_local:
+                global_tags.add(tag)
+            elif ltag.local_index == self.region_index:
+                local_gtags.add(ltag.global_tag)
+
+        # Only keep global tags that don't have a matching local tag.
+        global_tags = global_tags.difference(local_gtags)
```
uri-granta: (suggestion)

```suggestion
        global_tags -= local_gtags
```

trieste/acquisition/rule.py
```diff
@@ -972,6 +1004,111 @@ def update(
...
+        return local_gtags, global_tags
+
+    def select_models(
+        self, models: Optional[Mapping[Tag, ProbabilisticModelType]]
+    ) -> Optional[Mapping[Tag, ProbabilisticModelType]]:
+        """
+        Select models belonging to this region for acquisition.
+
+        :param models: The model for each tag.
+        :return: The models belonging to this region.
```
```suggestion
        :return: The models belonging to this region (or `None` if there aren't any).
```

trieste/acquisition/rule.py
```diff
@@ -972,6 +1004,111 @@ def update(
...
+            _models = {}
+            for tag in local_gtags:
+                ltag = LocalizedTag(tag, self.region_index)
+                _models[ltag] = models[ltag]
+            for tag in global_tags:
+                _models[tag] = models[tag]
+
+        return _models if _models else None
+
+    def select_datasets(
```
uri-granta: This is basically identical to select_models. Maybe make it generic?

```
def select_in_region(self, mapping: Optional[Mapping[Tag, T]]) -> Optional[Mapping[Tag, T]]:
    ....
```

and then you can pass in either `models` or `datasets` as required.

trieste/acquisition/rule.py
```diff
@@ -1055,10 +1197,38 @@ def acquire(
             points from the previous acquisition state.
         """
 
+        # Subspaces should be set by the time we call `acquire`.
+        assert self._tags is not None
+        assert self._init_subspaces is not None
+
+        num_local_models: Dict[Tag, int] = defaultdict(int)
```
uri-granta: Simpler to use collections.Counter:

```python
num_local_models_vals = Counter(
    LocalizedTag.from_tag(tag).global_tag for tag in models
    if LocalizedTag.from_tag(tag).is_local
).values()
```

trieste/acquisition/rule.py
```diff
@@ -1055,10 +1197,38 @@ def acquire(
...
+        for tag in models:
+            ltag = LocalizedTag.from_tag(tag)
+            if ltag.is_local:
+                num_local_models[ltag.global_tag] += 1
+        num_local_models_vals = set(num_local_models.values())
+        assert (
+            len(num_local_models_vals) <= 1
+        ), f"The number of local models should be the same for all tags, got {num_local_models}"
+        _num_local_models = 0 if len(num_local_models_vals) == 0 else num_local_models_vals.pop()
```
uri-granta: Pithier?

```suggestion
        _num_local_models = sum(num_local_models_vals)
```

trieste/acquisition/rule.py
```diff
@@ -1055,10 +1197,38 @@ def acquire(
...
+
+        num_subspaces = len(self._tags)
+        assert _num_local_models in [0, num_subspaces], (
+            f"When using local models, the number of subspaces {num_subspaces} should be equal to "
+            f"the number of local models {_num_local_models}"
+        )
+
+        # If we have local models, run the (deepcopied) base rule sequentially for each subspace.
+        # Otherwise, run the base rule as is, once with all models and datasets.
+        # Note: this should only trigger on the first call to `acquire`, as after that we will
```
uri-granta, line 1195: I'm confused as to what prevents this triggering on subsequent calls to acquire? Did you mean to include `and self._rules is not None`?

trieste/acquisition/rule.py
```diff
@@ -1024,9 +1161,14 @@ def __init__(
             if not isinstance(init_subspaces, Sequence):
                 init_subspaces = [init_subspaces]
             self._init_subspaces = tuple(init_subspaces)
+            for index, subspace in enumerate(self._init_subspaces):
+                subspace.region_index = index  # Override the index.
             self._tags = tuple([str(index) for index in range(len(init_subspaces))])
 
         self._rule = rule
+        self._rules: Optional[
```
uri-granta, line 1141: Maybe add a comment explaining what this is?

trieste/acquisition/rule.py
```diff
@@ -1092,9 +1262,32 @@ def state_func(
...
+                        _datasets = {
+                            LocalizedTag.from_tag(tag).global_tag: dataset
+                            for tag, dataset in _datasets.items()
+                        }
+                    _points.append(rule.acquire(subspace, _models, _datasets))
+                points = tf.stack(_points, axis=1)
+            else:
+                points = self._rule.acquire(acquisition_space, models, datasets)
+
+            return state_, tf.reshape(points, [-1, len(subspaces), points.shape[-1]])
```
uri-granta, line 1262: So are we always returning rank 3 now, even if there are no subspaces? (or am I misunderstanding?)

uri-granta (commented):
Still need to look at the tests but looking good so far! Various small comments, suggestions and questions.

trieste/ask_tell_optimization.py
```diff
@@ -421,18 +437,44 @@ def tell(self, new_data: Mapping[Tag, Dataset] | Dataset) -> None:
...
+        #
+        # Note: this replication of initial data can potentially cause an issue when a global model
+        # is being used with local datasets, as the points may be repeated. This will only be an
+        # issue if two regions overlap and both contain that initial data-point -- as filtering
+        # (in BatchTrustRegion) would otherwise remove duplicates. The main way to avoid the issue
+        # in this scenario is to provide local initial datasets, instead of a global initial
+        # dataset.
+        updated_datasets = {}
+        for tag, new_dataset in new_data.items():
+            _, old_dataset = get_value_for_tag(
```
uri-granta: Would it be better (clearer but also more efficient) to reuse rather than copy the dataset if it is already correctly tagged?

```python
for tag, new_dataset in new_data.items():
    if tag in self._datasets:
        self._datasets[tag] += new_dataset
    else:
        global_tag = LocalizedTag.from_tag(tag).global_tag
        assert global_tag in self._datasets
        self._datasets[tag] = self._datasets[global_tag] + new_data
```

tests/unit/test_ask_tell_optimization.py
uri-granta: (see comment about that change; it might be nice to avoid copying the dataset when we don't need to)

tests/unit/test_ask_tell_optimization.py
```diff
@@ -427,3 +436,74 @@ def __deepcopy__(self, memo: dict[int, object]) -> _UncopyableModel:
...
+            for i in range(batch_size)
+        }
+        init_data[OBJECTIVE] = mk_dataset([[0.5], [1.5]], [[0.25], [0.35]])
+
+    query_points = tf.reshape(
+        tf.constant(range(batch_size * num_query_points_per_batch), tf.float64),
+        (num_query_points_per_batch, batch_size, 1),
+    )
+
+    class DatasetChecker(QuadraticMeanAndRBFKernel, PseudoTrainableProbModel):
```
uri-granta: Can you define this at the module level and import it from test_baysian_optimizer?

tests/unit/objectives/test_utils.py
```diff
@@ -49,3 +56,51 @@ def test_mk_multi_observer() -> None:
...
+    # Check keys.
+    exp_keys: Set[Union[Tag, LocalizedTag]] = set()
+    for key in keys:
+        exp_keys.update({LocalizedTag(key, i) for i in range(batch_size)})
+        exp_keys.add(key)
+    assert ys.keys() == exp_keys
+
+    # Check datasets.
+    for key in keys:
+        # Different observers (in parameterize above) return different observation values.
```
uri-granta: (might be cleaner to move this to the parametrize? maybe replace "keys" with a mapping of keys to lambda?))

tests/integration/test_ask_tell_optimization.py
```diff
@@ -47,26 +49,58 @@
...
     [
         pytest.param(
-            20, False, lambda: EfficientGlobalOptimization(), id="EfficientGlobalOptimization"
+            20, False, lambda: EfficientGlobalOptimization(), 1, id="EfficientGlobalOptimization"
         ),
         pytest.param(
             20,
             True,
             lambda: EfficientGlobalOptimization(),
+            1,
```
uri-granta: It's annoying that we have to put these everywhere for the non-local tests. Is there any way to infer it from the acquisition rule?

docs/notebooks/trust_region.pct.py
```diff
@@ -245,6 +245,44 @@ def plot_history(result: trieste.bayesian_optimizer.OptimizationResult) -> None:
 # %%
 plot_history(result)
 
+# %% [markdown]
+# ## TEST
```
uri-granta: (?)

uri-granta (commented):
(a few more comments)

trieste/bayesian_optimizer.py
khurram-ghani: That is not quite the same functionality. Existing code returns the global-tag portion from all tags, including local ones. Whereas your suggestion would ignore the local tags.

trieste/bayesian_optimizer.py
khurram-ghani: For consistency sake, I am actually going to change it to look like your suggestion in ask_tell_optimization.py.

docs/notebooks/trust_region.pct.py
khurram-ghani: This is just a section to show how you can use local models. I mentioned in the PR description that I was going to delete it before merging. However, since the TURBO PR now shows how to use this as well, I'll delete it now.

tests/unit/test_ask_tell_optimization.py
khurram-ghani: Change reverted due to suggested change.

trieste/acquisition/rule.py
khurram-ghani: Yes that is correct. Which does mean that the caller has to use the new batched observer with this. `BayesianOptimizer` class has rank detection in it now. For ask-tell optimiser, the user has to use the correct observer, e.g. see `integration/test_ask_tell_optimization.py` for similar rank detection.

tests/integration/test_ask_tell_optimization.py
khurram-ghani: I don't think there is a neat way to infer this from the rule. For example, the tests `BatchTrustRegionBox` and `BatchTrustRegionBox/LocalModels` create almost identical rules, except one is using local models. We could choose an artificial difference (e.g. num_query_points) but that doesn't seem good. I tried to find out if there is a way to have default values with `pytest.parametrize`, but couldn't find a way. Do you have any other ideas?

khurram-ghani: I agree BTW that it is annoying.

trieste/acquisition/rule.py
khurram-ghani: Good spot. Yes an expression is missing. It got lost in translation from a previous implementation.

trieste/acquisition/rule.py
khurram-ghani: Yes, with some small modifications.

tests/integration/test_ask_tell_optimization.py
uri-granta: One possibility is to make acuqisition_rule_fn optionally a tuple:

```python
AcquisitionRuleFunction = Callable[
    [], AcquisitionRule[TensorType, SearchSpace, TrainableProbabilisticModel]
]
...
acquisition_rule_fn: AcquisitionRuleFunction | tuple[AcquisitionRuleFunction, int]
```

trieste/acquisition/rule.py
uri-granta: Is this a breaking change then? Our next release is probably going to include some anyway, so not necessarily an issue if it is, though we'll want to make sure we document it properly in the release notes.

khurram-ghani: Updated the PR description.

trieste/ask_tell_optimization.py
```diff
@@ -186,18 +187,28 @@ def __init__(
...
         # reassure the type checker that everything is tagged
         datasets = cast(Dict[Tag, Dataset], datasets)
         models = cast(Dict[Tag, TrainableProbabilisticModelType], models)
 
-        if datasets.keys() != models.keys():
+        # Get set of dataset and model keys, ignoring any local tag index. That is, only the
+        # global tag part is considered.
+        datasets_keys = {LocalizedTag.from_tag(tag).global_tag for tag in datasets.keys()}
+        models_keys = {LocalizedTag.from_tag(tag).global_tag for tag in models.keys()}
+        if datasets_keys != models_keys:
```
hstojic, line 210: perhaps use set operator? in case order of keys is different?

trieste/objectives/utils.py
```diff
@@ -57,3 +61,49 @@ def mk_multi_observer(**kwargs: Callable[[TensorType], TensorType]) -> MultiObse
...
+    :return: A multi-observer across the batch dimension of query points, returning the data from
+        ``objective_or_observer``.
+    """
+
+    @check_shapes("qps: [n_points, batch_size, n_dims]")
+    def _observer(qps: TensorType) -> Mapping[Tag, Dataset]:
+        # Call objective with rank 2 query points by flattening batch dimension.
+        # Some objectives might only expect rank 2 query points, so this is safer.
+        batch_size = qps.shape[1]
+        flat_qps = tf.reshape(qps, [-1, qps.shape[-1]])
```
hstojic: we have `flatten_lead_dims` function that I think does this for you and returns unflatten function so you can revert it

trieste/acquisition/rule.py
```diff
@@ -972,6 +1007,78 @@ def update(
         """
         ...
 
+    def _get_tags(self, tags: Set[Tag]) -> Tuple[Set[Tag], Set[Tag]]:
+        # Separate tags into local (matching index) and global tags (without matching
+        # local tag).
+        local_gtags = set()
```
hstojic: gtags is intentional?

trieste/acquisition/rule.py
```diff
@@ -1055,10 +1169,38 @@ def acquire(
...
+        ), f"The number of local models should be the same for all tags, got {num_local_models}"
+        _num_local_models = sum(num_local_models_vals)
+
+        num_subspaces = len(self._tags)
+        assert _num_local_models in [0, num_subspaces], (
+            f"When using local models, the number of subspaces {num_subspaces} should be equal to "
+            f"the number of local models {_num_local_models}"
+        )
+
+        # If we have local models, run the (deepcopied) base rule sequentially for each subspace.
```
hstojic, line 1193: why do we need to copy the rules? is there an issue with iterating through the local models and sending them to the acquire call of a same acquisition rule?

hstojic (commented):
I haven't gone through it in details, there is a lot going on here...

my biggest comment is on adding  additional complexity to ask-tell and bayesian_optimizer - do we really need to take care of the local models there?
very few acq rules will make use of local models, why not taking care of datasets and model training in the acquisition rule itself, in those few that make use of it?

hstojic (approved):
ok after a discussion

trieste/acquisition/rule.py
khurram-ghani: Added a comment to explain better.

trieste/acquisition/rule.py
khurram-ghani: We only copy rules when we either have local models, or not using EGO. As the rules compile the acquisition functions, the model trainable variables will be built into the graph, so we need to build separate graphs for local models. Similarly, non-EGO rules may have some state that needs to be maintained between calls for each region, so it is safer to use separate instances.

PR #790, opened 2023-10-13
Fix trajectory sampling to handle active dims

khurram-ghani:

## Summary
Fixes issue where trajectory sampling was ignoring active-dims in the kernel

## PR checklist


hstojic (commented):
this is great, thanks!
but we should also add a test for active_dims, to make sure we would catch smth like this

hstojic (approved):

PR #791, opened 2023-10-17
Batch trust regions implementation of TURBO

khurram-ghani:

## Summary

This PR adds a TURBO implementation based on the new batch-trust-region classes and replaces the original `TURBO` implementation.

The entry point `filter_datasets` in acquisition rules from the earlier local-models PR is called after acquisition and observation of new query-points, but before model training. Trust region algorithms do not update their regions in `acquire` anymore, instead the update is now in this new method. This enables the behaviour of the new `TURBOBox` class to match the previous `TURBO` implementation. 

Existing notebook, unit tests and integration tests have been updated to use the new classes. The `trust_region` notebook shows how local models can be used with `TURBOBox`.

We have agreed that we don't need backwards compatibility.

Previously, a `TURBO` rule was created like this:
``` python
rule = TURBO(search_space)
```

This will need to be replaced with the following:
``` python
rule = BatchTrustRegionBox(TURBOBox(search_space))
```

## PR checklist


trieste/acquisition/rule.py
```diff
@@ -1647,233 +1690,183 @@ def get_dataset_min(
...
-        y_min: TensorType
-        """ The minimum observed value. """
-
-        def __deepcopy__(self, memo: dict[int, object]) -> TURBO.State:
-            box_copy = copy.deepcopy(self.acquisition_space, memo)
-            return TURBO.State(
-                box_copy, self.L, self.failure_counter, self.success_counter, self.y_min
-            )
-
     def __init__(
```
uri-granta, line 1706: What's your opinion of defining a `TURBO` helper function that returns `BatchTrustRegionBox(TURBOBox(search_space, ...), ...)` for backwards compatibility?

trieste/acquisition/utils.py
```diff
@@ -159,6 +159,29 @@ def copy_to_local_models(
     return {LocalizedTag(key, i): copy.deepcopy(global_model) for i in range(num_local_models)}
 
 
+def add_local_datasets(
```
uri-granta: The name suggests that you're updating the input in place (though the type hint says otherwise). Maybe rename?
 
```suggestion
def with_local_datasets(
```

trieste/ask_tell_optimization.py
```diff
@@ -231,20 +228,33 @@ def __init__(
...
+        # Only applies to a subset of acquisition rules, i.e. ones that have subspaces and
+        # hence use local datasets.
+        #
+        # Note: this replication of initial data can potentially cause an issue when a global
+        # model is being used with local datasets, as the points may be repeated. This will only
+        # be an issue if two regions overlap and both contain that initial data-point -- as
+        # filtering below would otherwise remove duplicates. The main way to avoid the issue in
+        # this scenario is to provide local initial datasets, instead of a global initial
+        # dataset.
+        if hasattr(self._acquisition_rule, "num_subspaces"):
```
uri-granta: I think we can do better: if we need to know whether an acquisition rule uses subspaces we can either

1. define `class LocalDatasetAcquisitionRule(AcquisitionRule)` and use that for all such rules
2. or define a `has_subspaces` property on all rules that defaults to False and override it when appropriate

Option 1 sounds simpler to me (and we do a similar thing with VectorizedAcquisitionFunctionBuilder).

trieste/ask_tell_optimization.py
```diff
@@ -231,20 +228,33 @@ def __init__(
         else:
             self._acquisition_rule = acquisition_rule
 
-        if (fit_model) and isinstance(acquisition_rule, TURBO):
-            warnings.warn(
-                """
```
uri-granta, line 238: Is this warning still relevant for `BatchTrustRegionBox`?

trieste/acquisition/utils.py
```diff
@@ -159,6 +159,29 @@ def copy_to_local_models(
...
+    datasets: Mapping[Tag, Dataset],
+    num_local_datasets: int,
+) -> Mapping[Tag, Dataset]:
```
uri-granta: since you rely on this being a dict (or at least a mutable mapping) elsewhere:

```suggestion
) -> Dict[Tag, Dataset]:
```

trieste/ask_tell_optimization.py
```diff
@@ -231,20 +228,33 @@ def __init__(
...
+        #
+        # Note: this replication of initial data can potentially cause an issue when a global
+        # model is being used with local datasets, as the points may be repeated. This will only
+        # be an issue if two regions overlap and both contain that initial data-point -- as
+        # filtering below would otherwise remove duplicates. The main way to avoid the issue in
+        # this scenario is to provide local initial datasets, instead of a global initial
+        # dataset.
+        if hasattr(self._acquisition_rule, "num_subspaces"):
+            datasets = add_local_datasets(self._datasets, self._acquisition_rule.num_subspaces)
+            self._datasets = cast(Dict[Tag, Dataset], datasets)
```
uri-granta: Shouldn't be necessary if you correctly type add_local_datasets (see other comment)
```suggestion
```

uri-granta (commented):
(still looking but a few initial comments as I might not get a chance to get back to this before tomorrow)

trieste/acquisition/rule.py
khurram-ghani: We discussed this for `TREGO` and decided we don't need to define helper classes/methods for that. E.g. https://github.com/secondmind-labs/trieste/pull/778#pullrequestreview-1593955989

trieste/ask_tell_optimization.py
khurram-ghani: I considered 2, but was really trying not to add extra items to the base `AcquisitionRule` class. So 1 is a good option.

trieste/ask_tell_optimization.py
khurram-ghani: No, as we now always use the main `models` mapping, unlike `TURBO` which had separate internal state for local models.

It is possible that the users can have too many models in the mapping and we end up training them, but that is a general misuse issue not specific to trust-regions; and if they use the recommended `copy_to_local_models`, that shouldn't be the case.

trieste/acquisition/utils.py
khurram-ghani: Agreed.. did consider it, but wasn't sure.

trieste/bayesian_optimizer.py
```diff
@@ -713,12 +710,11 @@ def optimize(
...
                             datasets_copy = copy.deepcopy(datasets)
                             models_copy = copy.deepcopy(models)
-                            acquisition_state_copy = copy.deepcopy(acquisition_state)
-                            record = Record(datasets_copy, models_copy, acquisition_state_copy)
+                            record = Record(datasets_copy, models_copy, tracked_acquisition_state)
                             history.append(record)
                         else:
                             track_path = Path(track_path)
-                            record = Record(datasets, models, acquisition_state)
+                            record = Record(datasets, models, tracked_acquisition_state)
```
uri-granta: This is copying it twice, since saving a record by necessity creates a copy.

More importantly, we are now copying acquisition state even when track_state is False, which is a problem as it may not be copyable.

trieste/bayesian_optimizer.py
```diff
@@ -760,6 +764,10 @@ def optimize(
                         else:
                             query_points = points_or_stateful
 
+                    # Save acquisition state for tracking immediately after acquisition. Some rules,
+                    # such as BatchTrustRegion, may mutate the state in `update_and_filter` below.
+                    tracked_acquisition_state = copy.deepcopy(acquisition_state)
```
uri-granta: Why are we doing this? What's not right with returning the updated state in the Records (both the tracked state and the final computation result)? (Also see comment above about state not necessarily being copyable.)

trieste/experimental/plotting/plotting.py
```diff
@@ -590,21 +590,28 @@ def plot_trust_region_history_2d(
     else:
         spaces = [acquisition_space]
 
+    num_spaces = len(spaces)  # Assum  # Assume one query point per subspace.
```
uri-granta:
```suggestion
    num_spaces = len(spaces)
```

trieste/acquisition/rule.py
```diff
@@ -1121,30 +1136,32 @@ def __init__(
         :param rule: The acquisition rule that defines how to search for a new query point in each
             subspace. Defaults to :class:`EfficientGlobalOptimization` with default arguments.
```
uri-granta: (do we want to fully document the default rule heuristic?)

trieste/acquisition/rule.py
```diff
@@ -1356,7 +1385,36 @@ def filter_datasets(self, datasets: Mapping[Tag, Dataset]) -> Mapping[Tag, Datas
...
+        global_search_space: SearchSpace,
+        region_index: Optional[int] = None,
+    ):
+        """
+        :param global_search_space: The global search space this search space lives in.
+        :param region_index: The index of the region in a multi-region search space. This is used to
+            identify the local models and datasets to use for acquisition. If `None`, the
+            global models and datasets are used.
+        """
+        super().__init__(global_search_space.lower, global_search_space.upper)
```
uri-granta: I think this would be much clearer without super:

```python
Box.__init__(self, global_search_space.lower, global_search_space.upper)
UpdatableTrustRegion.__init__(self, region_index)
```

trieste/acquisition/rule.py
```diff
@@ -1356,7 +1385,36 @@ def filter_datasets(self, datasets: Mapping[Tag, Dataset]) -> Mapping[Tag, Datas
...
+        super(Box, self).__init__(region_index)
+        self._global_search_space = global_search_space
+        # Random initial location in the global search space.
+        self.location = tf.squeeze(global_search_space.sample(1), axis=0)
```
uri-granta, line 1399: Tempted to suggest putting this in a property too (with a setter) to make it clear that it's part of the interface.

uri-granta (commented):
(A few more comments...)

trieste/bayesian_optimizer.py
khurram-ghani: One of the issues is with plotting; knowing what region was used to find a query point. For `BatchTrustRegion` rules, the acquistion for a query point (which is added to the dataset below) is from a certain region, which is subsequently updated in `update_and_filter` below. Without this copy, the relationship between tracked query_point and the associated region is lost. We copy here to keep a record of the region used to compute the query point.

On the issue of state not being copyable, I can make the use of copy condional on `track_state` being set; which is how I wrote it originally, but then "simplified" it :-)

khurram-ghani: Here is the change to make the copy conditional: [4e5aad6](https://github.com/secondmind-labs/trieste/pull/791/commits/4e5aad673c8de63c2e991ea78dc20194979a227b)

trieste/bayesian_optimizer.py
khurram-ghani: Reverted, as discussed. Copy is now inside the rule.

trieste/bayesian_optimizer.py
khurram-ghani: Reverted, as discussed. Copy is now inside the rule.

uri-granta (approved):

PR #792, opened 2023-10-27
Avoid use of python bool in qmc_normal_samples

ChrisMorter:
Avoid use of python `if` statement in `qmc_normal_samples` which prevents it, or any code which calls it, from being saved as a `tf.module`.

## Summary

`qmc_normal_samples` contains a precondition check to handle the edge cases where the number of samples is zero, or the sample dim is zero. This was implemented using a python `if` statement, but this means that the function can't be saved as a `tf.function`. 

The error raised if you try this is:

```
tensorflow.python.framework.errors_impl.OperatorNotAllowedInGraphError: Using a symbolic `tf.Tensor` as a Python `bool` is not allowed: AutoGraph is disabled in this function. Try decorating it directly with @tf.function.
``` 

As the error message suggests, this can be worked around by ensuring that any usages of `qmc_normal_samples` are wrapped in `tf.function`, but this seems like an unnecessary burden to place on calling code when we can just replace the problematic `if` statement with `tf.cond` instead.

...

## PR checklist

### DIFF
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -942,7 +942,7 @@ def test_qmc_samples_shapes(num_samples: int, n_sample_dim: int) -> None:
 
 
 def test_qmc_samples__save_as_tf_function(tmp_path: Path) -> None:
-    def get_samples():
+    def get_samples() -> tf.Tensor:
         return qmc_normal_samples(
             num_samples=tf.constant(5),
             n_sample_dim=tf.constant(2),
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -941,12 +941,16 @@ def test_qmc_samples_shapes(num_samples: int, n_sample_dim: int) -> None:
     assert samples.shape == expected_samples_shape
 
 
-def test_qmc_samples__save_as_tf_function(tmp_path: Path) -> None:
+@pytest.mark.parametrize("decorate_with_tf_function", [True, False])
+def test_qmc_samples__save_as_tf_function(tmp_path: Path, decorate_with_tf_function: bool) -> None:
     def get_samples() -> tf.Tensor:
         return qmc_normal_samples(
             num_samples=tf.constant(5),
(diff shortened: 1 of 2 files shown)


khurram-ghani:
I would like to understand a bit more about the context in which the error occurs please. Are we not using auto-graph when saving the function? If we are, I am a bit surprised that it didn't convert the code to effectively what you have done manually. There are plenty of other places in trieste where we use python `if` expressions and rely on auto-graph to convert them. So it would be good to understand what is different here. It is concerning if we had to make this kind of change repeatedly.

ChrisMorter:
> I would like to understand a bit more about the context in which the error occurs please. Are we not using auto-graph when saving the function? If we are, I am a bit surprised that it didn't convert the code to effectively what you have done manually. There are plenty of other places in trieste where we use python `if` expressions and rely on auto-graph to convert them. So it would be good to understand what is different here. It is concerning if we had to make this kind of change repeatedly.

So the issue occurs when code outside of Trieste attempts to save models which call into Trieste with `autograph=False`. I'm slightly surprised we haven't hit this issue sooner, because it seems like it would happen whenever the calling code encounters an `if` expression. Maybe we've just been lucky so far due to mainly calling prediction methods which don't tend to have much branching logic.

In this case, I'm not sure why the calling code doesn't use autograph. Perhaps the resolution should be to leave Trieste as is it and to update the callers to use autograph. I'll look into how feasible that is... I agree with your point that if we wish for Trieste functions to generally support saving without using autograph then we'd need to update many places, which would be a pain.

hstojic:
@ChrisMorter and @khurram-ghani do we have a resolution on this topic?

khurram-ghani:
@ChrisMorter has updated the calling code to always use autograph and that has resolved the original issue; so I believe this trieste change is not needed anymore. @ChrisMorter are you okay with closing this PR?

ChrisMorter:
> @ChrisMorter has updated the calling code to always use autograph and that has resolved the original issue; so I believe this trieste change is not needed anymore. @ChrisMorter are you okay with closing this PR?

@khurram-ghani apologies for the slow reply, I must have missed the notification for your message. Yes, that's correct, I'm happy to close this PR.

PR #794, opened 2023-11-17
ScipyOptimizer

### LINKED ISSUE #793: `randomize_hyperparameters` ignores `prior_on`
The `randomize_hyperparameters` helper does not check parameters' `prior_on` attributes [here](https://github.com/secondmind-labs/trieste/blob/aa94c861c360c4ef4710eb94e837f598a81dcbae/trieste/models/gpflow/utils.py#L79). Hopefully a quick fix:
```
if param.prior_on is PriorOn.UNCONSTRAINED:
    param.unconstrained_variable.assign(sample)
else:
    param.assign(sample)
```

j-wilson:
**Related issue(s)/PRs:** 

Resolves #793 

## Summary

This PR introduces a `ScipyOptimizer` class for GPFlow Modules that automatically extracts parameter bounds and passes them to `gpflow.optimizers.Scipy`. This fixes an issue where `unconstrained_variables` that are actually constrained under the prior cause model fitting to fail. 

Additionally, this PR fixes a bug in `resample_hyperparameters` where `prior_on` was ignored.

There are edge-cases which may cause this code to fail. For example, `Scipy` may choose to filter out parameters that it deems inactive and, in so doing, cause the provided bounds to be misspecified (since it does not filter out the corresponding bounds).

## PR checklist

### DIFF
--- a/trieste/models/gpflow/interface.py
+++ b/trieste/models/gpflow/interface.py
@@ -41,6 +41,7 @@ from ..utils import (
     write_summary_kernel_parameters,
     write_summary_likelihood_parameters,
 )
+from .optimizer import ScipyOptimizer
 from .sampler import BatchReparametrizationSampler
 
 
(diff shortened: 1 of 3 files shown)


j-wilson:
@hstojic Here is a quick PR related to #793. I've opted to upload this here to better illustrate the goal of these changes. Additional tests and docstrings are needed. Ultimately, we may want to separate out the fix for #793, since this can likely be merged as is.

PR #796, opened 2023-11-24
Handle unconstrained priors in randomize_hyperparameters

uri-granta:
**Related issue(s)/PRs:** #793

## Summary

Fix randomize_hyperparameters to handle unconstrained priors.

Also change the logic for handling Sigmoid constraints: we now prioritise sampling from the prior if there is one, and sample from the constrained domain only if there isn't.

New Sigmoid behaviour is more logical and flexible, and old behaviour can be easily obtained by removing any prior.

## PR checklist

### DIFF
--- a/trieste/models/gpflow/utils.py
+++ b/trieste/models/gpflow/utils.py
@@ -53,7 +53,8 @@ def assert_data_is_compatible(new_data: Dataset, existing_data: Dataset) -> None
 def randomize_hyperparameters(object: gpflow.Module) -> None:
     """
     Sets hyperparameters to random samples from their prior distributions or (for Sigmoid
-    constraints with no priors) their constrained domains.
+    constraints with no priors) their constrained domains. Note that it is up to the caller
+    to ensure that the prior, if defined, is compatible with the transform.
 
     :param object: Any gpflow Module.
     """
--- a/trieste/models/gpflow/utils.py
+++ b/trieste/models/gpflow/utils.py
@@ -14,11 +14,12 @@
 
 from __future__ import annotations
 
-from typing import Tuple, Union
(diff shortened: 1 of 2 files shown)


trieste/models/gpflow/utils.py
```diff
@@ -76,6 +68,18 @@ def randomize_hyperparameters(object: gpflow.Module) -> None:
                 sample = param.prior.sample(tf.shape(param))
             else:
                 sample = param.prior.sample()
+            if param.prior_on is gpflow.base.PriorOn.UNCONSTRAINED:
+                param.unconstrained_variable.assign(sample)
+            else:
+                param.assign(sample)
```
frgsimpson, line 75: There's a risk here that we hit a commonly encountered error where we can't assign a value incompatible with the transform. This could be eg if we have a Gaussian prior on a sigmoid transform-ed parameter. But that is the user's responsibility to take so I think it's fine here.

frgsimpson (approved):
Looks good!

trieste/models/gpflow/utils.py
hstojic: we could add a note in the docstring about it

hstojic:
@uri-granta did you check @j-wilson PR where he also addressed the same issue: https://github.com/secondmind-labs/trieste/pull/794?

uri-granta:
> @uri-granta did you check @j-wilson PR where he also addressed the same issue: #794?

AFAIU that's doing more than just fixing the bug in randomize_hyperparameters. Will look at that as soon as this is merged.

trieste/models/gpflow/utils.py
uri-granta: Added

PR #805, opened 2024-01-10
Filter-out local datasets when calling base-rule

khurram-ghani:

## Summary
Filter out local datasets when calling the EGO base rule. This will reduce the burden of filtering on most acquisition rules/functions that we currently have. If in the future we have acquisition rules/functions that need to see the local datasets, we can expand the functionality then.

Also took the opportunity to fix a very minor typing issue in `mk_batch_observer`.

## PR checklist

### DIFF
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -1816,7 +1816,7 @@ def test_multi_trust_region_box_acquire_filters() -> None:
 
     # Create a BatchTrustRegionBox instance with the mock base_rule.
     subspaces = [SingleObjectiveTrustRegionBox(search_space) for _ in range(2)]
-    rule = BatchTrustRegionBox(subspaces, mock_base_rule)  # type: ignore[var-annotated]
+    rule: BatchTrustRegionBox[ProbabilisticModel] = BatchTrustRegionBox(subspaces, mock_base_rule)
 
     rule.acquire(search_space, models, datasets)(None)
 
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import copy
 from collections.abc import Mapping
 from typing import Callable, Optional
+from unittest.mock import ANY, MagicMock
 
 import gpflow
 import numpy as np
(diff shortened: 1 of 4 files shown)


hstojic (approved):
looks good I think, but perhaps would be good that @uri-granta takes a look as well

trieste/acquisition/rule.py
```diff
@@ -1282,7 +1282,19 @@ def state_func(
                     _points.append(rule.acquire(subspace, _models, _datasets))
                 points = tf.stack(_points, axis=1)
             else:
-                points = self._rule.acquire(acquisition_space, models, datasets)
+                # Filter out local datasets as this is the EGO rule with normal acquisition
```
uri-granta: The fact that `self._rules` implies (EGO + no local models) isn't obvious here. Maybe rename `._rules` to 
`._subspace_rules` and change the initialisation (above) to:

```python
        if self._rules is None and not (
            _num_local_models == 0 and isinstance(self._rule, EfficientGlobalOptimization)
        ):
```

?

tests/unit/acquisition/test_rule.py
```diff
@@ -1798,6 +1799,33 @@ def test_multi_trust_region_box_updated_datasets_are_in_regions(
...
+        "global_tag": MagicMock(),
+    }
+
+    search_space = Box([0.0], [1.0])
+    mock_base_rule = MagicMock(spec=EfficientGlobalOptimization)
+    mock_base_rule.acquire.return_value = tf.constant([[[0.0], [0.0]]], dtype=tf.float64)
+
+    # Create a BatchTrustRegionBox instance with the mock base_rule.
+    subspaces = [SingleObjectiveTrustRegionBox(search_space) for _ in range(2)]
+    rule = BatchTrustRegionBox(subspaces, mock_base_rule)  # type: ignore[var-annotated]
```
uri-granta: (what's the mypy error here?)

uri-granta (approved):
Looks good. Couple of minor comments.

tests/unit/acquisition/test_rule.py
khurram-ghani: The error is:
`tests/unit/acquisition/test_rule.py:1819: error: Need type annotation for "rule"  [var-annotated]
`

It wants the type of the rule spelled out, as it seems it can't figure that out iteself (via the `base_rule`). This is a standard error in most instantiations of TR rules as they have a base rule (I think). This can be avoided by using the following (for this instance). However, for unit tests I chose to be not that verbose; we instance the rules in many places.

`    rule: BatchTrustRegionBox[ProbabilisticModel] = BatchTrustRegionBox(subspaces, mock_base_rule)
`

trieste/acquisition/rule.py
khurram-ghani: I am not sure about renaming to `_subspace_rules`, but am happy to apply De Morgan's law to the condition expression.

Also, I'll change the comment slightly to be not so definitive about being EGO. That is the case for now, but may support more rules in the future.

PR #807, opened 2024-01-13
Various linting suggestions

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Investigate pylint and apply some of its suggestions. Also try running ruff.

The main thing holding us back from adopting pylint in CI is speed: running it on the codebase takes over a minute, compared to 3 seconds for flake8. If this PR exposes any concerning gaps then we could investigate whether there are any flake8 plugins that would cover them.

Ruff meanwhile is very fast but seems to catch less than flake8, so there is no incentive to switch.

## PR checklist


trieste/models/gpflow/sampler.py
```diff
@@ -839,12 +835,12 @@ def __init__(
             kernel_K(self._inducing_points, x)
         )
 
-    def call(self, x: TensorType) -> TensorType:  # [N, D] -> [N, F + M] or [L, N, F + M]
+    def call(self, inputs: TensorType) -> TensorType:  # [N, D] -> [N, F + M] or [L, N, F + M]
```
khurram-ghani, line 838: What does the linter complain about here?

khurram-ghani (approved):
All are good suggestions; just a question on one of them below.

trieste/models/gpflow/sampler.py
uri-granta: This was overriding an ABC method (in gpflux) where the parameters are named differently.

PR #808, opened 2024-01-16
Initial point samplers

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Add support for custom samplers for generating the optimization initial point candidates. This solves two problems:

1. It allows including pre-computed points among the candidates
2. It allows batching the random sampling to avoid running out of memory for high dimensional problems.

Given how often `generate_continuous_optimizer` is used (and often with keyword arguments) I think we should make an extra effort to maintain backwards compatibility here. The current approach is simply to expand `num_initial_samples` so it can take _either_ an `int` or a sampler. An alternate approach would be to add an additional optional `initial_point_sampler` parameter, but the downside of that is that it wouldn't be possible to catch cases where people accidentally pass in both a `num_initial_samples` and a `initial_sampler`.

## PR checklist


hstojic:
this can work as a short term solution, but we would want to make this more flexible - we could pass generators of initial samples to the optimizer, and provide a nice function/class for random sample generators - but this would then allow us to pass other types of generators when needed, e.g. qp generated by some other optimization process which would provide better starting points

uri-granta:
> this can work as a short term solution, but we would want to make this more flexible - we could pass generators of initial samples to the optimizer, and provide a nice function/class for random sample generators - but this would then allow us to pass other types of generators when needed, e.g. qp generated by some other optimization process which would provide better starting points

Is generators definitely the way to go here? Space.sample doesn't currently support this, and I worry that a generator giving 10,000 samples one at a time would be much less efficient than a single call to sample(10_000).

Another approach would be to provide an additional optional argument `initial_sampler: Callable[[int], TensorType]` that lets you specify a function that returns a given number of initial sample points. The default behaviour would be equivalent to `initial_sampler = space.sample` but we could provide a helper function so people could write something like `initial_sampler = select_from_samples(generator)` if they want to. This would work nicely with `split_initial_samples`, and (more importantly) not break the current `generate_continuous_optimizer` API.

hstojic:
> > this can work as a short term solution, but we would want to make this more flexible - we could pass generators of initial samples to the optimizer, and provide a nice function/class for random sample generators - but this would then allow us to pass other types of generators when needed, e.g. qp generated by some other optimization process which would provide better starting points
> 
> Is generators definitely the way to go here? Space.sample doesn't currently support this, and I worry that a generator giving 10,000 samples one at a time would be much less efficient than a single call to sample(10_000).
> 
> Another approach would be to provide an additional optional argument `initial_sampler: Callable[[int], TensorType]` that lets you specify a function that returns a given number of initial sample points. The default behaviour would be equivalent to `initial_sampler = space.sample` but we could provide a helper function so people could write something like `initial_sampler = select_from_samples(generator)` if they want to. This would work nicely with `split_initial_samples`, and (more importantly) not break the current `generate_continuous_optimizer` API.

it doesn't have to be generators, that's just what seemed like what could allow us to pass a variety of initial samples, but
`initial_sampler` could do the trick -  though, it would perhaps need to be a list of samplers, e.g. we would want to have some random samples and some pre-optimised initial points

@vpicheny any thoughts?

uri-granta:
I've added something that I think is flexible enough but still easy to use. Opinions?

trieste/acquisition/optimizer.py
```diff
@@ -168,11 +169,63 @@ def optimize_discrete(
...
+    """
+    Initial point sampler that returns points from a prebuilt sequence (e.g. a list or Tensor).
+    Raises a :exc:`ValueError` if there are insufficiently many points to return.
+    """
+
+    def _sampler(space: SearchSpace, num_samples: int, offset: int) -> TensorType:
+        if len(sequence) < offset + num_samples:
+            raise ValueError(
+                f"Insufficient samples ({offset+num_samples} required, {len(sequence)} available"
+            )
```
hstojic: should we also do a basic check that dimensionality is correct?

trieste/acquisition/optimizer.py
```diff
@@ -168,11 +169,63 @@ def optimize_discrete(
...
 
+InitialPointSampler = Callable[[SearchSpace, int, int], TensorType]
+"""
+Type alias for a function that generates initial samples for an optimizer.
+Takes a search space, the number of samples to generate, and an offset, and returns
+that many samples (at the offset).
+"""
+
+
+def sample_from_space(space: SearchSpace, num_samples: int, offset: int) -> TensorType:
```
hstojic: perhaps move all of these helper functions to utils.py file?

trieste/acquisition/optimizer.py
```diff
@@ -168,11 +169,63 @@ def optimize_discrete(
...
+
+    return _sampler
+
+
 def generate_continuous_optimizer(
     num_initial_samples: int = NUM_SAMPLES_MIN,
     num_optimization_runs: int = 10,
     num_recovery_runs: int = 10,
     optimizer_args: Optional[dict[str, Any]] = None,
+    split_initial_samples: Optional[int] = 100_000,
```
hstojic: I'm thinking that with initial samplers we can simplify arguments here, i.e. we could subsume `split_initial_samples` and `num_initial_samples` within the samplers, then each call of the sampler gives a splitted sample and we call them until they are exhausted, what do you think?
it would be a breaking change, but we are releasing 3.0 anyway :)

trieste/acquisition/optimizer.py
```diff
@@ -202,11 +255,22 @@ def generate_continuous_optimizer(
     :param optimizer_args: The keyword arguments to pass to the Scipy L-BFGS-B optimizer.
         Check `minimize` method  of :class:`~scipy.optimize` for details of which arguments
         can be passed. Note that method, jac and bounds cannot/should not be changed.
+    :param split_initial_samples: Maximum number of initial samples to process at a time.
+        Decreasing this can reduce memory usage at the start, at the slight cost of performance.
```
hstojic: lets make this a bit more concrete and add a use-case here, its beneficial for high-dimensional problems with a lot of initial points

trieste/acquisition/optimizer.py
```diff
@@ -202,11 +255,22 @@ def generate_continuous_optimizer(
...
+    :param initial_sampler: Function for generating initial samples. This should accept
+        a search space, the number of samples to generate and an offset. It may be called multiple
```
hstojic: here as well, lets add a sentence with concrete use case for this, one could add pre-optimised starting points to random starting points etc

trieste/acquisition/optimizer.py
```diff
@@ -202,11 +255,22 @@ def generate_continuous_optimizer(
...
+        times (with different offsets) if `split_initial_samples` is specified. By default,
+        samples are generated by calling `space.sample(num_samples)`.
     :return: The acquisition optimizer.
```
hstojic, line 348: should we add raise bit here in the docs?

trieste/acquisition/optimizer.py
```diff
@@ -249,63 +313,87 @@ def optimize_continuous(
...
-            tiled_candidates = tf.tile(candidates, [1, multiple, 1])  # [num_initial_samples, V, D]
-        else:
-            tf.debugging.assert_rank(
-                candidates,
-                2,
+        samples_left = num_initial_samples
+        top_fun_values: Optional[TensorType] = None  # [V, num_optimization_runs]
+        top_candidates: Optional[TensorType] = None  # [V, num_optimization_runs, D]
+
+        while samples_left > 0:
```
hstojic: should we extract this as a standalone function so we can properly test it?

trieste/acquisition/optimizer.py
```diff
@@ -249,63 +313,87 @@ def optimize_continuous(
...
+            _, top_k_indices = tf.math.top_k(
+                top_fun_values, k=num_optimization_runs
+            )  # [V, num_optimization_runs]
+
+            top_candidates = tf.gather(
+                top_candidates, top_k_indices, batch_dims=1
+            )  # [V, num_optimization_runs, D]
+            top_fun_values = tf.gather(
+                top_fun_values, top_k_indices, batch_dims=1
+            )  # [V, num_optimization_runs]
```
hstojic: neat

hstojic (commented):
left a few comments for improvement, but overall looks much better, I think this would fit the bill - with a sampler I can create a custom one where I mix random samples with points obtained in other ways to provide a better start to the optimiser

trieste/acquisition/optimizer.py
uri-granta: That is a much better idea! I've pushed a first attempt at this but will see what breaks and write some tests before asking for a review.

As an aside, it's actually possible to do this without breaking the interface by still letting people specify an int for `num_initial_samples` if they want to. There's an argument that we should rename the argument from `num_initial_samples` to `initial_samples`, but I'm worried that this would break a _lot_ of code (and not just ours). Another option is to add a separate `initial_samples` argument and make people specify just one of this and `num_initial_samples`.

tests/unit/acquisition/test_optimizer.py
```diff
@@ -824,3 +826,64 @@ def side_effect(*args: Any, **kwargs: Any) -> spo.OptimizeResult:
     else:
         received_constraints = None
     assert received_constraints == expected_constraints
+
+
+@pytest.mark.parametrize("num_initial_points", [0, 1, 2, 3, 4])
+def test_generate_initial_points(num_initial_points: int) -> None:
+    def sampler(space: SearchSpace) -> Iterable[TensorType]:
+        assert space == Box([-1], [2])
```
hstojic, line 834: you are using this space in multiple tests, perhaps create a constant at the top and reuse it? or a fixture?

tests/unit/acquisition/test_optimizer.py
```diff
@@ -824,3 +826,64 @@ def side_effect(*args: Any, **kwargs: Any) -> spo.OptimizeResult:
     else:
         received_constraints = None
     assert received_constraints == expected_constraints
+
+
+@pytest.mark.parametrize("num_initial_points", [0, 1, 2, 3, 4])
```
hstojic, line 831: what about covering the vectorization case, > 1?

tests/unit/acquisition/test_optimizer.py
```diff
@@ -824,3 +826,64 @@ def side_effect(*args: Any, **kwargs: Any) -> spo.OptimizeResult:
...
+def test_generate_initial_points(num_initial_points: int) -> None:
+    def sampler(space: SearchSpace) -> Iterable[TensorType]:
+        assert space == Box([-1], [2])
+        yield tf.range(-1, 2, 0.1)[:, None]
+
+    best_four_samples = tf.constant([1.0, 0.9, 1.1, 0.8])
+    points = generate_initial_points(
+        num_initial_points, sampler, Box([-1], [2]), _quadratic_sum([1.0])
+    )
+    npt.assert_allclose(points, best_four_samples[:num_initial_points, None, None], atol=1e-6)
```
hstojic, line 842: can you also check that the shape is as required?

tests/unit/acquisition/test_optimizer.py
```diff
@@ -824,3 +826,64 @@ def side_effect(*args: Any, **kwargs: Any) -> spo.OptimizeResult:
...
+    npt.assert_allclose(points, best_samples[:num_initial_points, None, None], atol=1e-6)
+
+
+@pytest.mark.parametrize("num_samples,batch_size", [(1, None), (5, None), (5, 2), (5, 5), (5, 10)])
+def test_sample_from_space(num_samples: int, batch_size: Optional[int]) -> None:
+    batches = list(sample_from_space(num_samples, batch_size)(Box([0], [1])))
+    assert len(batches) == ceil(num_samples / (batch_size or num_samples))
+    assert sum(len(batch) for batch in batches) == num_samples
+    assert all(0 <= x <= 1 for batch in batches for x in batch)
+    assert len(set(float(x) for batch in batches for x in batch)) == num_samples
```
hstojic, line 901: can you also check that the shape is as required?

hstojic (approved):
great work! left a few minor comments, but I think this is it :)

it would still be good to pass it by @khurram-ghani as well before merging

p.s. this will make it way easier to test alternative `generate_initial_points`, e.g. with clustering to achieve better diversity of starting points between the runs

trieste/acquisition/optimizer.py
```diff
@@ -168,8 +168,144 @@ def optimize_discrete(
...
+        raise ValueError(f"num_samples must be positive, got {num_samples}")
+
+    if isinstance(batch_size, int) and batch_size <= 0:
+        raise ValueError(f"batch_size must be positive, got {batch_size}")
+
+    batch_size_int = batch_size or num_samples
+
+    def sampler(space: SearchSpace) -> Iterable[TensorType]:
+        for offset in range(0, num_samples, batch_size_int):
+            yield space.sample(min(num_samples, offset + batch_size_int) - offset)
```
khurram-ghani:
```suggestion
            yield space.sample(min(num_samples - offset, batch_size_int))
```

trieste/acquisition/optimizer.py
```diff
@@ -168,8 +168,144 @@ def optimize_discrete(
...
+    Return the best starting points for an optimization from those generated by a given sampler.
+
+    :param num_initial_points: Number of best starting points to return.
+    :param initial_sampler: Initial point sampler.
+    :param space: Search space.
+    :param target_func: Target function being optimized.
+    :param vectorization: Vectorization of the target function.
+    """
+    top_fun_values: Optional[TensorType] = None  # [V, num_optimization_runs]
+    top_candidates: Optional[TensorType] = None  # [V, num_optimization_runs, D]
```
khurram-ghani, line 231: Is it worth initialising these directly here, instead of starting off as `None` and being initialised in the loop? With the current code it is possible for `top_candiates` to be `None` I think. See comment below.

trieste/acquisition/optimizer.py
```diff
@@ -168,8 +168,144 @@ def optimize_discrete(
...
+        )  # [V, num_initial_points]
+
+        top_candidates = tf.gather(
+            top_candidates, top_k_indices, batch_dims=1
+        )  # [V, num_initial_points, D]
+        top_fun_values = tf.gather(
+            top_fun_values, top_k_indices, batch_dims=1
+        )  # [V, num_initial_points]
+
+    initial_points = tf.transpose(top_candidates, [1, 0, 2])  # [num_initial_points,V,D]
```
khurram-ghani, line 307: `top_candidates` can be `None` if the generator is empty, right? So I don't understand how this passes type checking.

trieste/acquisition/optimizer.py
```diff
@@ -168,8 +168,144 @@ def optimize_discrete(
...
+                """
+            ),
+        )
+
+        if top_candidates is None:
+            top_candidates = tf.zeros(
+                [vectorization, 0, tf.shape(candidates)[-1]], dtype=candidates.dtype
+            )
+        if top_fun_values is None:
+            top_fun_values = tf.zeros([vectorization, 0], dtype=target_func_values.dtype)
```
khurram-ghani, line 284: This assumes `top_fun_values` can only be positive. Should be initialised to `-inf` I think.

tests/unit/acquisition/test_optimizer.py
```diff
@@ -824,3 +826,87 @@ def side_effect(*args: Any, **kwargs: Any) -> spo.OptimizeResult:
...
+
+@pytest.mark.parametrize("num_initial_points", [0, 1, 2, 3, 4])
+def test_generate_initial_points(num_initial_points: int) -> None:
+    def sampler(space: SearchSpace) -> Iterable[TensorType]:
+        assert space == Box([-1], [2])
+        yield tf.range(-1, 2, 0.1)[:, None]
+
+    best_four_samples = tf.constant([1.0, 0.9, 1.1, 0.8])
+    points = generate_initial_points(
+        num_initial_points, sampler, Box([-1], [2]), _quadratic_sum([1.0])
```
khurram-ghani, line 839: Could include a test where all the points (including the best) have negative function values. See comment in `optimizer.py`.

khurram-ghani (commented):
One important comment I think (on `-inf`), and some minor ones.

trieste/acquisition/optimizer.py
uri-granta: This is actually an **empty** value of shape [v, 0] so it doesn't actually matter what we initialize it to! But I can add a comment.

trieste/acquisition/optimizer.py
uri-granta: Problem is that we don't know D until we've seen some candidates. Also we don't know the dtypes of either of them. That's why we only initialize them inside the loop.

trieste/acquisition/optimizer.py
uri-granta: Good spot. I'll take a look why it's not complaining (and probably raise an exception if the initial point sampler is empty).

uri-granta: It's not complaining because tensorflow doesn't have type stubs and TensorType is treated as Any. I've fixed it and added a test.

trieste/acquisition/optimizer.py
khurram-ghani: Ah.. didn't spot that; so it is fine. A comment might be useful but not needed.

khurram-ghani (approved):
It is a really good and flexible implementation!

Issue #781, opened 2023-08-24
Investigate improvements to the get-unique-points algorithm


khurram-ghani:
**Describe the feature you'd like**
Investigate whether [this algorithm](https://github.com/secondmind-labs/trieste/blob/ef62eebcf68a52052a25a8eb73efce10361a15f8/trieste/acquisition/utils.py#L141-L173) can be parallelised. There is a previous parallel but incomplete implementation and the current sequential one. See [this PR](https://github.com/secondmind-labs/trieste/pull/773#discussion_r1300181968) for context.

j-wilson:
This can be done in `O(N + M^2)` time and space by using a bloom filter, where `N` is the number of points and `M` is the number of duplicate points. Roughly:

1. Quantized each point to the desired precision.
2. Convert each quantized point to a string type.
3. Hash each stringified+quantized point.
4. Test for equality within each bin where a collision occured.

Here is a naive implementation with `O(N^2)` time and space complexity that computes all pairwise distances and returns the indices of each unique row
```
@tf.function
def get_unique_rows_dense(matrix: tf.Tensor, precision: Optional[int] = None) -> tf.Tensor:
    matrix = (
        matrix
        if precision is None
        else tf.math.round(10 ** precision * matrix)
    )
    sq_norms = tf.reduce_sum(tf.square(matrix), axis=-1, keepdims=True)
    sq_dists = (
        sq_norms 
        + tf.transpose(sq_norms) 
        - tf.matmul(2 * matrix, matrix, transpose_b=True)
    )
    argmin = tf.argmin(sq_dists, axis=-1) 
    unique = tf.where(argmin == tf.range(tf.shape(matrix)[0], dtype=argmin.dtype))
    return tf.squeeze(unique, axis=-1)
```
And, here is a fancy implementation using the approach suggested above
```
@tf.function
def get_unique_rows(
    matrix: tf.Tensor, precision: Optional[int] = None, **kwargs: Any,
) -> tf.Tensor:
    matrix = (
        matrix
        if precision is None
        else tf.math.round(10 ** precision * matrix)
    )
    strings = tf.strings.reduce_join(tf.strings.as_string(matrix), axis=-1)
    bin_ids = tf.strings.to_hash_bucket_fast(strings, 2 ** 63 - 1)
    unique_ids, membership = tf.unique(bin_ids, out_idx=tf.int64)

    nrows = tf.shape(matrix)[0]
[...]

PR #809, opened 2024-01-16
Make get_unique_points more efficient

uri-granta:
**Related issue(s)/PRs:** #781 

## Summary

Make get_unique_points more efficient and document its precise behaviour.

**Fully backwards compatible:** ish

The behaviour changes as we now use rounding to select points rather than a distance based loop. But the old behaviour was never fully documented, and the rounding approach should work fine for BatchTrustRegionBox.get_initialize_subspaces_mask which is the only place we currently use it.

## PR checklist

### DIFF
--- a/tests/unit/acquisition/test_utils.py
+++ b/tests/unit/acquisition/test_utils.py
@@ -182,12 +182,12 @@ def test_with_local_datasets(
         (
             tf.constant([[1.0], [2.0], [1.0], [3.0], [1.71], [1.699999], [3.29], [3.300001]]),
             0.3,
-            tf.constant([True, True, False, True, False, True, False, True]),
+            tf.constant([True, True, False, True, False, False, False, True]),
         ),
         (
             tf.constant([[1.0], [2.0], [1.0], [3.0], [1.699999], [1.71], [3.300001], [3.29]]),
             0.3,
-            tf.constant([True, True, False, True, True, False, True, False]),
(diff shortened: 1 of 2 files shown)


tests/unit/acquisition/test_utils.py
```diff
@@ -180,12 +180,12 @@ def test_with_local_datasets(
...
         ),
         (
-            tf.constant([[1.0], [2.0], [1.0], [3.0], [1.71], [1.699999], [3.29], [3.300001]]),
+            tf.constant([[1.0], [2.0], [1.0], [3.0], [2.1], [2.10001], [3.29], [3.300001]]),
             0.3,
             tf.constant([True, True, False, True, False, True, False, True]),
         ),
         (
-            tf.constant([[1.0], [2.0], [1.0], [3.0], [1.699999], [1.71], [3.300001], [3.29]]),
+            tf.constant([[1.0], [2.0], [1.0], [3.0], [2.10001], [2.1], [3.300001], [3.29]]),
```
khurram-ghani, line 188: This seems problematic. The `2.10001` point is only `0.10001` away and well within the tolerance of `0.3` but is still marked as unique. I hadn't quite realised that the rounding boundary had such a big error effect.

uri-granta:
Closing PR following offline discussion. Will create a separate PR for parallelising the existing greedy set cover algorithm.

PR #813, opened 2024-01-23
Make get_unique_points_mask more efficient

uri-granta:
**Related issue(s)/PRs:** #781 

## Summary

Make get_unique_points_mask run in parallel. Since this approach detects connected graph components, it can trip up on examples like [1,2,3,4] with tolerance 1, which should return two points not one, even though all the points are connected. However, we can fix this using what will almost certainly be a small number of iterations.

## PR checklist


khurram-ghani (approved):
Looks really good. Glad we could come up with a nice parallel algorithm :-)

PR #815, opened 2024-01-24
Make AskTellOptimizer generic in how model updates are handled

### LINKED ISSUE #814: AskTellOptimizer needs an instances of TrainableProbabilisticModel
The AskTellOptimizer needs instances of `TrainableProbabilisticModel` even if no training is being performed by the optimizer.

In other words, if only the `ask` method is called, and not the `tell` method, then having a _trainable_ model seems unnecessary, but this is required by the type declaration.

In our case we train models outside of Trieste, so it would be nicer if we could specify just a `ProbabilisticModel` if we only intend to use the `ask` method, and not the `tell` method.

uri-granta:
**Related issue(s)/PRs:** resolves #814

## Summary

Generalize `AskTellOptimizer` so that we can use it with non-trainable models.

1. `AskTellOptimizer` works as before (and still requires trainable models)
2. `AskTellOptimizerNoTraining` does no training at tell time and during initialization (and supports all models).
3. New `AskTellOptimizerABC` subclasses can be defined to incorporate other model handling into the tell stage.

Runtime behaviour unchanged, but Record, FrozenRecord and OptimizationResult are now also generic in the model type, which could affect type checking type-annotated code. The fix is to add model type annotations were appropriate.

## PR checklist

### DIFF
--- a/trieste/ask_tell_optimization.py
+++ b/trieste/ask_tell_optimization.py
@@ -511,7 +511,7 @@ class AskTellOptimizer(AskTellOptimizerABC[SearchSpaceType, TrainableProbabilist
 class AskTellOptimizerNoTraining(AskTellOptimizerABC[SearchSpaceType, ProbabilisticModelType]):
     """
     This class provides Ask/Tell optimization interface with no model training performed
-    during the tell stage.
+    during the Tell stage or at initialization.
     """
 
     def update_model(self, model: ProbabilisticModelType, dataset: Dataset) -> None:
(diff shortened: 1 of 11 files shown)


trieste/ask_tell_optimization.py
```diff
@@ -483,3 +488,30 @@ def tell(self, new_data: Mapping[Tag, Dataset] | Dataset) -> None:
...
+
+    def update_model(self, model: TrainableProbabilisticModelType, dataset: Dataset) -> None:
+        model.update(dataset)
+        optimize_model_and_save_result(model, dataset)
+
+
+class AskTellOptimizerNoTraining(AskTellOptimizerABC[SearchSpaceType, ProbabilisticModelType]):
+    """
+    This class provides Ask/Tell optimization interface with no model training performed
+    during the tell stage.
```
khurram-ghani: Should say no model training during `tell` or `__init__`?

khurram-ghani (approved):
LGTM.

PR #821, opened 2024-02-09
Trust region with mixed search spaces

khurram-ghani:

## Summary
Trust regions with mixed search spaces are implemented by introducing the concept of a "product" trust region (and matching rule). This is analogous to `TaggedProductSearchSpace`.

`TaggedProductSearchSpace` internally contains a list of `SearchSpace` spaces; similarly `UpdatableTrustRegionProduct` contains a list of `UpdatableTrustRegion` regions.

The `mixed_search_spaces` notebook has been updated to show usage, and unit and integration tests have been added.

~~Integration tests do not exist for mixed search spaces (trust region or otherwise), mainly because we do not have a synthetic problem. The notebook creates a problem by discretizing one of the Branin dimensions. We could move this out of the notebook to create integration tests, or alternatively treat the notebook as "integration testing" for now.~~

## PR checklist

### DIFF
--- a/docs/notebooks/mixed_search_spaces.pct.py
+++ b/docs/notebooks/mixed_search_spaces.pct.py
@@ -67,6 +67,8 @@ fig.show()
 # We create our mixed search space by instantiating this class with a list containing the discrete
 # and continuous spaces, without any explicit tags (hence using default tags).
 # This can be easily extended to more than two search spaces by adding more elements to the list.
+#
+# Note: the dtype of all the component search spaces must be the same.
 
 # %%
 from trieste.space import Box, DiscreteSearchSpace, TaggedProductSearchSpace
--- a/docs/notebooks/mixed_search_spaces.pct.py
+++ b/docs/notebooks/mixed_search_spaces.pct.py
@@ -238,8 +238,15 @@ model = GaussianProcessRegression(gpflow_model)
 # acquisition rule as the base rule. The trust region acquisition rule is initialized with a set of
 # trust regions; 5 in this example. Each trust regions is defined as a product of a discrete and a
 # continuous trust sub-region. The base rule is then used to optimize the acquisition function
-# within each trust region. This setup is analogous to the one used in the "Batch trust region rule"
+# within each trust region. This setup is similar to the one used in the "Batch trust region rule"
 # section of the [trust region Bayesian optimization notebook](trust_region.ipynb).
+#
+# Analogous to a `TaggedProductSearchSpace`, each trust region is a product of a discrete and a
+# continuous trust (sub-)region. The discrete sub-region is defined by
+# `FixedPointTrustRegionDiscrete` that selects a random point from the discrete space, which is then
+# fixed for the duration of the optimization. The continuous sub-region is defined by
+# `SingleObjectiveTrustRegionBox`, just as in the trust region notebook, where the region is updated
+# at each step of the optimization.
 
 # %%
 from trieste.acquisition import ParallelContinuousThompsonSampling
--- a/docs/notebooks/mixed_search_spaces.pct.py
+++ b/docs/notebooks/mixed_search_spaces.pct.py
@@ -218,6 +218,161 @@ for point in points:
     )
 
 # %% [markdown]
+# ## Trust region with mixed search spaces
+#
+# In this section, we demonstrate the use of trust region acquisition rules with mixed search
+# spaces. We use the same mixed search space and observer as before, and the same initial data.
+# See [trust region Bayesian optimization notebook](trust_region.ipynb) for an introduction to
+# trust region acquisition rules.
+#
+# First we build a Gaussian process model of the objective function using the initial data.
+
+# %%
+gpflow_model = build_gpr(
+    initial_data, mixed_search_space, likelihood_variance=1e-7
+)
+model = GaussianProcessRegression(gpflow_model)
+
+# %% [markdown]
+# We create a trust region acquisition rule that uses the efficient global optimization (EGO)
+# acquisition rule as the base rule. The trust region acquisition rule is initialized with a set of
+# trust regions; 5 in this example. Each trust regions is defined as a product of a discrete and a
+# continuous trust sub-region. The base rule is then used to optimize the acquisition function
+# within each trust region. This setup is analogous to the one used in the "Batch trust region rule"
+# section of the [trust region Bayesian optimization notebook](trust_region.ipynb).
+
+# %%
+from trieste.acquisition import ParallelContinuousThompsonSampling
+from trieste.acquisition.rule import (
+    BatchTrustRegionProduct,
+    EfficientGlobalOptimization,
+    FixedPointTrustRegionDiscrete,
+    SingleObjectiveTrustRegionBox,
+    UpdatableTrustRegionProduct,
+)
+
+num_query_points = 5
+init_regions = [
+    UpdatableTrustRegionProduct(
+        [
+            FixedPointTrustRegionDiscrete(discrete_space),
+            SingleObjectiveTrustRegionBox(continuous_space),
+        ]
+    )
+    for _ in range(num_query_points)
+]
+base_rule = EfficientGlobalOptimization(  # type: ignore[var-annotated]
+    builder=ParallelContinuousThompsonSampling(),
+    num_query_points=num_query_points,
+)
+tr_acq_rule = BatchTrustRegionProduct(init_regions, base_rule)
+
+# %% [markdown]
+# We run the optimization loop for 15 steps using the trust region acquisition rule.
+
+# %%
+bo = BayesianOptimizer(observer, mixed_search_space)
+
+num_steps = 15
+tr_result = bo.optimize(
+    num_steps, initial_data, model, tr_acq_rule, track_state=True
+)
+dataset = tr_result.try_get_final_dataset()
+
+# %% [markdown]
+# The best point found by the optimizer is obtained as before.
+
+# %%
+query_point, observation, arg_min_idx = tr_result.try_get_optimal_point()
+
+print(f"query point: {query_point}")
+print(f"observation: {observation}")
+
+# %% [markdown]
+# Plot of the optimization loop over the mixed search space, similar to the previous plot.
+
+# %%
+query_points = dataset.query_points.numpy()
+observations = dataset.observations.numpy()
+
+_, ax = plot_function_2d(
+    scaled_branin,
+    ScaledBranin.search_space.lower,
+    ScaledBranin.search_space.upper,
+    contour=True,
+)
+plot_bo_points(query_points, ax[0, 0], num_initial_points, arg_min_idx)
+ax[0, 0].set_xlabel(r"$x_1$")
+ax[0, 0].set_ylabel(r"$x_2$")
+
+for point in points:
+    ax[0, 0].vlines(
+        point,
+        mixed_search_space.lower[1],
+        mixed_search_space.upper[1],
+        colors="b",
+        linestyles="dashed",
+        alpha=0.6,
+    )
+
+# %% [markdown]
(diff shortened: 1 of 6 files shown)


docs/notebooks/mixed_search_spaces.pct.py
```diff
@@ -217,6 +217,161 @@
...
+    SingleObjectiveTrustRegionBox,
+    UpdatableTrustRegionProduct,
+)
+
+num_query_points = 5
+init_regions = [
+    UpdatableTrustRegionProduct(
+        [
+            FixedPointTrustRegionDiscrete(discrete_space),
+            SingleObjectiveTrustRegionBox(continuous_space),
```
hstojic, line 268: this is the key bit, it deserves another small paragraph above explaining how exactly we use mixed search spaces here, what is `FixedPointTrustRegionDiscrete` exactly etc

trieste/acquisition/rule.py
```diff
@@ -1889,6 +1896,304 @@ def get_dataset_min(
...
+    """
+
+    def __init__(
+        self,
+        global_search_space: DiscreteSearchSpace,
+        region_index: Optional[int] = None,
+    ):
+        super().__init__(global_search_space, region_index)
+        # Random initial point from the global search space.
+        self._points = self.global_search_space.sample(1)
```
hstojic, line 2052: so the acq fnc optimizer won't really be able to find the best initial discrete values and fix them there, instead it will get discrete dimensions fixed to a single point, right?
however, we have discussed that we should leave it to the optimizer to fix the discrete dims, no?

trieste/acquisition/rule.py
```diff
@@ -1889,6 +1896,304 @@ def get_dataset_min(
...
+
+    @property
+    def global_search_space(self) -> DiscreteSearchSpace:
+        """The global space this search space lives in."""
+        return self._global_search_space
+
+
+class FixedPointTrustRegionDiscrete(UpdatableTrustRegionDiscrete):
+    """
+    A discrete trust region with a fixed point location that does not change.
```
hstojic: perhaps expand on this, does not change across active learning steps?

trieste/acquisition/rule.py
```diff
@@ -1889,6 +1896,304 @@ def get_dataset_min(
...
+            )
+
+        self._global_search_space = TaggedProductSearchSpace(
+            [region.global_search_space for region in regions]
+        )
+
+        # Find the most general dtype of all the regions.
+        dtypes = [np.array(region.location).dtype for region in regions]
+        most_general_dtype = np.result_type(*dtypes)
+        self.dtype = most_general_dtype
```
hstojic: what is this all about?

trieste/acquisition/rule.py
```diff
@@ -1889,6 +1896,304 @@ def get_dataset_min(
...
+
+UpdatableTrustRegionWithGlobalSearchSpace = Union[
+    UpdatableTrustRegionBox, UpdatableTrustRegionDiscrete
+]
+"""A type alias for updatable trust regions with a global search space."""
+
+
+class UpdatableTrustRegionProduct(TaggedProductSearchSpace, UpdatableTrustRegion):
+    """
+    An updatable mixed search space that is the product of multiple updatable trust sub-regions.
```
hstojic, line 2084: perhaps a sentence about what its useful for - mixed search spaces

trieste/acquisition/rule.py
hstojic: also why np operations?

trieste/acquisition/rule.py
```diff
@@ -1889,6 +1896,304 @@ def get_dataset_min(
...
+        return _regions
+
+    @property
+    def location(self) -> TensorType:
+        """
+        The location of the product trust region, concatenated from the locations of the
+        sub-regions.
+        """
+        return tf.concat(
+            [tf.cast(region.location, self.dtype) for region in self.regions.values()], axis=-1
```
hstojic: i see, so we have to convert everything to floats

hstojic: i wonder if we have a failure mode here and in `Dataset` object (or more generally with mixed search spaces), could models require integers to properly function? i.e. dataset should be treating different types differently? probably a very low likelihood that models do need that, but we should probably document this assumption in few places (e.g. Dataset object, mixed search space notebook etc)

tests/unit/acquisition/test_rule.py
```diff
@@ -1860,6 +1870,288 @@ def test_multi_trust_region_box_state_deepcopy() -> None:
...
+def test_updatable_tr_product_raises_on_missing_index(
+    discrete_search_space: DiscreteSearchSpace, continuous_search_space: Box
+) -> None:
+    region1 = FixedPointTrustRegionDiscrete(discrete_search_space, region_index=0)
+    region2 = SingleObjectiveTrustRegionBox(continuous_search_space, region_index=1)
+    with pytest.raises(AssertionError, match="regions can only have a region_index"):
+        UpdatableTrustRegionProduct([region1, region2])
+
+
+def test_updatable_tr_product_raises_on_mimatch_index(
```
hstojic:
```suggestion
def test_updatable_tr_product_raises_on_mismatch_index(
```

tests/unit/acquisition/test_rule.py
```diff
@@ -1860,6 +1870,288 @@ def test_multi_trust_region_box_state_deepcopy() -> None:
...
+                    OBJECTIVE: Dataset(
+                        tf.constant([[0.5], [0.0], [1.0]], dtype=tf.float64),
+                        tf.constant([[0.5], [0.0], [1.0]], dtype=tf.float64),
+                    )
+                },
+            ],
+        ),
+    ],
+)
+def test_updatable_tr_product_method_calls_subregions(
```
hstojic, line 2083: not sure whats going on here - what was the goal here?

hstojic (approved):
I have only a few minor comments and one major - we don't have an integration test for mixed search spaces - we should definitely add it, and we could use the branin problem from the notebook for it - this could be a separate PR though (hence I'm approving this one)

hstojic:
ok, now I have seen you have mentioned the integration tests in the description... I think we should definitely have them, I wouldn't rely on the notebook

tests/unit/acquisition/test_rule.py
khurram-ghani: This is ensuring when any of the 3 methods in the product region are called, the same methods in the sub-regions are called with the correct arguments.

trieste/acquisition/rule.py
khurram-ghani: As we discussed earlier today, this is a simple initial version. Yes, it fixes points at random at init and then doesn't change them or use the acq function.

To use the acq function, we can use model based intitialisation, similar to what we are currently implementing elsewhere. This fits in nicely with that structure.

trieste/acquisition/rule.py
khurram-ghani: As you noted, this is so we can concatenate different types of locations into a single tensor in `self.location`. I am determining the least general common subtype (and not just using `float64` by default).

On using `np`, there is only an `experimental` version of this available in tensorflow, so I didn't use it. But thinking about it now, I guess that is fine, so I can switch to using that (but see other comment about removing this altogether).

trieste/acquisition/rule.py
khurram-ghani: Not necessarily floats, whatever is the least general common type, but yes likely floats in most cases.

We already have this requirement in many places in existing code; for instance, wherever we use `TaggedProductSearchSpace`. For example, when we call `sample` in the optimiser -- it combines the samples by calling `tf.concat`. I think we force the user to create all search spaces with a common type, i.e. float64.

Here I am just being extra nice by doing the casting for the user if they didn't create the spaces with the same type. But I agree this is of limited value, since other places in our code will fallover anyway. So I can remove this for consistency sake with other code :-)

hstojic: no need to remove it then :)

but note about putting a warning in the docs is still valid

khurram-ghani: It's fine, I have removed it. It convinced me that the added complexity was not worth it.

I have added a note about dtypes in 3 obvious places, `TaggedProductSearchSpace`, `UpdatableTrustRegionProduct` and `mixed_search_spaces.pct.py`. I wasn't sure about adding it to `Dataset`, since that doesn't directly reference mixed search spaces.

khurram-ghani:
Integration test added.

trieste/acquisition/rule.py
```diff
@@ -1050,7 +1112,7 @@ def select_in_region(self, mapping: Optional[Mapping[Tag, T]]) -> Optional[Mappi
         elif self.region_index is None:
             # If no index, then return the global items.
             _mapping = {
-                tag: item
+                tag: self.with_input_active_dims(item)
```
uri-granta, line 1139: slightly surprised this type checks, given T isn't guaranteed to satisfy the input requirements for with_input_active_dims. Can you bind T to `Union[TensorType, Dataset, ProbabilisticModel]`?

trieste/space.py
```diff
@@ -999,6 +999,8 @@ class TaggedProductSearchSpace(CollectionSearchSpace):
         decision_space = Box([-1, -2], [2, 3])
         mixed_space = TaggedProductSearchSpace(spaces=[context_space, decision_space])
 
+    Note: the dtype of all the component search spaces must be the same.
```
uri-granta, line 1002: do we check for this and raise an Exception?

uri-granta (commented):
Overall approach of using UpdatableTrustRegionProduct and input_active_dims looks good to me. Happy to review in more detail if you like, but also happy for you to check in with the current tick.

trieste/acquisition/rule.py
khurram-ghani: Good spot. Not sure why typing works. I have changed it.

trieste/space.py
khurram-ghani: Tensorflow will do that anytime we use `lower`, `sample` etc.

trieste/acquisition/rule.py
```diff
@@ -994,13 +994,20 @@ def acquire(
...
+        self,
+        region_index: Optional[int] = None,
+        input_active_dims: Optional[Union[slice, Sequence[int]]] = None,
+    ) -> None:
         """
         :param region_index: The index of the region in a multi-region search space. This is used to
             identify the local models and datasets to use for acquisition. If `None`, the
             global models and datasets are used.
+        :param input_active_dims: The active dimensions of the input space, either a slice or list
+            of indices into the columns of the space. If `None`, all dimensions are active.
```
hstojic, line 1007: perhaps add a sentence on is this used for?

trieste/acquisition/rule.py
```diff
@@ -2005,18 +2118,26 @@ def __init__(
...
+        # Set active dimensions for each sub-region.
+        dim_ix = 0
+        for region, dims in zip(regions, self.subspace_dimension):
+            # Check the region's input active dims are not already set.
+            assert region.input_active_dims is None, (
+                f"input_active_dims ({region.input_active_dims}) should not be set for sub-regions "
+                f" ({region}) of a product region"
+            )
+            region.input_active_dims = slice(dim_ix, dim_ix + dims)
+            dim_ix += dims
```
hstojic, line 2144: shouldn't this be informed by the global search space, which dimensions exactly are discrete there?

hstojic (commented):
few final comments

trieste/acquisition/rule.py
khurram-ghani: It is already informed by the global search space, effectively. `self.subspace_dimension` returns the list of dimensions of all the subspaces/regions. The same subspace/regions are used a few lines above to create `self._global_search_space`.

Also, here we are creating the active dims for all types of search spaces, not just discrete.

PR #825, opened 2024-02-28
Updatable discrete trust region

khurram-ghani:

## Summary
This PR adds a new updatable discrete trust region. This builds on the the fixed-point discrete TR added in a previous PR. The new class `SingleObjectiveTrustRegionDiscrete` is similar in behaviour to `SingleObjectiveTrustRegionBox`, but for a discrete search space.

The mixed search spaces notebook and integ tests now use this class, instead of `FixedPointTrustRegionDiscrete` from the previous PR.

## PR checklist

### DIFF
--- a/docs/notebooks/mixed_search_spaces.pct.py
+++ b/docs/notebooks/mixed_search_spaces.pct.py
@@ -236,27 +236,29 @@ gpflow_model = build_gpr(
 model = GaussianProcessRegression(gpflow_model)
 
 # %% [markdown]
-# We create a trust region acquisition rule that uses the efficient global optimization (EGO)
-# acquisition rule as the base rule. The trust region acquisition rule is initialized with a set of
+# We create a trust region meta acquisition rule that uses efficient global optimization (EGO)
+# as the base rule. The trust region rule is initialized with a set of
 # trust regions; 5 in this example. Each trust regions is defined as a product of a discrete and a
-# continuous trust sub-region. The base rule is then used to optimize the acquisition function
-# within each trust region. This setup is similar to the one used in the "Batch trust region rule"
-# section of the [trust region Bayesian optimization notebook](trust_region.ipynb).
-#
-# Analogous to a `TaggedProductSearchSpace`, each trust region is a product of a discrete and a
-# continuous trust (sub-)region. The discrete sub-region is defined by
-# `FixedPointTrustRegionDiscrete` that selects a random point from the discrete space, which is then
-# fixed for the duration of the optimization. The continuous sub-region is defined by
-# `SingleObjectiveTrustRegionBox`, just as in the trust region notebook, where the region is updated
-# at each step of the optimization.
+# continuous trust sub-region, analogous to a `TaggedProductSearchSpace`. The base rule is then
+# called to optimize the acquisition function within each region.
+
+# This setup is similar to the one used in the "Batch trust region rule"
+# section of the [trust region Bayesian optimization notebook](trust_region.ipynb). That notebook
+# creates trust regions of type `SingleObjectiveTrustRegionBox`. Here, we create trust regions that
+# are a product of a discrete and a continuous trust sub-region with
+# `UpdatableTrustRegionProduct`. The continuous part `SingleObjectiveTrustRegionBox` is the same as
+# in the trust region notebook, but now as a sub-region. The discrete sub-region is implemented by
+# `SingleObjectiveTrustRegionDiscrete`, which follows a very similar algorithm to the continuous
+# one, but with a region defined by a set of neighboring points. Both the continuous and
+# discrete sub-regions are updated at each step of the optimization.
 
 # %%
 from trieste.acquisition import ParallelContinuousThompsonSampling
 from trieste.acquisition.rule import (
     BatchTrustRegionProduct,
     EfficientGlobalOptimization,
-    FixedPointTrustRegionDiscrete,
     SingleObjectiveTrustRegionBox,
+    SingleObjectiveTrustRegionDiscrete,
     UpdatableTrustRegionProduct,
 )
 
@@ -264,7 +266,7 @@ num_query_points = 5
 init_regions = [
     UpdatableTrustRegionProduct(
         [
-            FixedPointTrustRegionDiscrete(discrete_space),
+            SingleObjectiveTrustRegionDiscrete(discrete_space),
             SingleObjectiveTrustRegionBox(continuous_space),
         ]
     )
@@ -329,8 +331,9 @@ for point in points:
 # step. The trust regions are shown as translucent boxes, with each box in a different color. The
 # new query point for earch region is plotted in matching color.
 #
-# Note that since the discrete dimension is on the x-axis, the trust regions appear as vertical
-# lines with zero width.
+# Note that since the discrete dimension is on the x-axis, the trust regions sometimes appear as
+# vertical lines with zero width. This occurs when a region contains only a single point along the
+# discrete dimension.
 
 # %%
 import base64
@@ -362,7 +365,7 @@ def plot_history(
             hist,
             num_query_points=num_query_points,
             num_init=num_initial_points,
-            alpha=1.0,
+            alpha=0.6,
         )
 
         if fig is not None:
--- a/docs/notebooks/mixed_search_spaces.pct.py
+++ b/docs/notebooks/mixed_search_spaces.pct.py
@@ -251,6 +251,10 @@ model = GaussianProcessRegression(gpflow_model)
 # `SingleObjectiveTrustRegionDiscrete`, which follows a very similar algorithm to the continuous
 # one, but with a region defined by a set of neighboring points. Both the continuous and
 # discrete sub-regions are updated at each step of the optimization.
+#
+# Note that `SingleObjectiveTrustRegionDiscrete` is designed for discrete numerical
+# variables only, which we have in this example. It is not suitable for qualitative (categorical,
+# ordinal and binary) variables.
 
 # %%
 from trieste.acquisition import ParallelContinuousThompsonSampling
--- a/docs/notebooks/mixed_search_spaces.pct.py
+++ b/docs/notebooks/mixed_search_spaces.pct.py
@@ -241,7 +241,7 @@ model = GaussianProcessRegression(gpflow_model)
 # trust regions; 5 in this example. Each trust regions is defined as a product of a discrete and a
 # continuous trust sub-region, analogous to a `TaggedProductSearchSpace`. The base rule is then
 # called to optimize the acquisition function within each region.
-
+#
 # This setup is similar to the one used in the "Batch trust region rule"
 # section of the [trust region Bayesian optimization notebook](trust_region.ipynb). That notebook
 # creates trust regions of type `SingleObjectiveTrustRegionBox`. Here, we create trust regions that
--- a/tests/integration/test_mixed_space_bayesian_optimization.py
+++ b/tests/integration/test_mixed_space_bayesian_optimization.py
@@ -31,6 +31,7 @@ from trieste.acquisition.rule import (
     AcquisitionRule,
     BatchTrustRegionProduct,
     EfficientGlobalOptimization,
+    FixedPointTrustRegionDiscrete,
     SingleObjectiveTrustRegionBox,
     SingleObjectiveTrustRegionDiscrete,
     UpdatableTrustRegionProduct,
@@ -95,6 +96,35 @@ mixed_search_space = _get_mixed_search_space()
                 num_query_points=3,
             ),
             id="LocalPenalization",
+        ),
+        pytest.param(
+            8,
+            BatchTrustRegionProduct(
+                [
+                    UpdatableTrustRegionProduct(
+                        [
+                            FixedPointTrustRegionDiscrete(
+                                cast(
+                                    DiscreteSearchSpace, mixed_search_space.get_subspace("discrete")
+                                )
+                            ),
+                            SingleObjectiveTrustRegionBox(
+                                mixed_search_space.get_subspace("continuous")
+                            ),
+                        ],
+                        tags=mixed_search_space.subspace_tags,
+                    )
+                    for _ in range(10)
+                ],
+                EfficientGlobalOptimization(
+                    ParallelContinuousThompsonSampling(),
+                    # Use a large batch to ensure discrete init finds a good point.
+                    # We are using a fixed point trust region for the discrete space, so
+                    # the init point is randomly chosen and then never updated.
+                    num_query_points=10,
+                ),
+            ),
+            id="TrustRegionSingleObjectiveFixed",
         ),
         pytest.param(
             8,
--- a/tests/integration/test_mixed_space_bayesian_optimization.py
+++ b/tests/integration/test_mixed_space_bayesian_optimization.py
@@ -31,8 +31,8 @@ from trieste.acquisition.rule import (
     AcquisitionRule,
     BatchTrustRegionProduct,
     EfficientGlobalOptimization,
-    FixedPointTrustRegionDiscrete,
     SingleObjectiveTrustRegionBox,
+    SingleObjectiveTrustRegionDiscrete,
     UpdatableTrustRegionProduct,
 )
 from trieste.bayesian_optimizer import BayesianOptimizer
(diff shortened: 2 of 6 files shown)


docs/notebooks/mixed_search_spaces.pct.py
```diff
@@ -236,35 +236,37 @@
...
-# at each step of the optimization.
+# This setup is similar to the one used in the "Batch trust region rule"
+# section of the [trust region Bayesian optimization notebook](trust_region.ipynb). That notebook
+# creates trust regions of type `SingleObjectiveTrustRegionBox`. Here, we create trust regions that
+# are a product of a discrete and a continuous trust sub-region with
+# `UpdatableTrustRegionProduct`. The continuous part `SingleObjectiveTrustRegionBox` is the same as
+# in the trust region notebook, but now as a sub-region. The discrete sub-region is implemented by
+# `SingleObjectiveTrustRegionDiscrete`, which follows a very similar algorithm to the continuous
+# one, but with a region defined by a set of neighboring points. Both the continuous and
+# discrete sub-regions are updated at each step of the optimization.
```
hstojic, line 253: we need to point out that this makes sense only for discrete numerical variables, not qualitative variables (ordinal, categorical, binary)

tests/integration/test_mixed_space_bayesian_optimization.py
```diff
@@ -102,7 +102,7 @@ def _get_mixed_search_space() -> TaggedProductSearchSpace:
                 [
                     UpdatableTrustRegionProduct(
                         [
-                            FixedPointTrustRegionDiscrete(
```
hstojic: we should still test this one, no?

trieste/acquisition/optimizer.py
```diff
@@ -603,25 +604,35 @@ def _objective_value_and_gradient(x: TensorType) -> Tuple[TensorType, TensorType
...
                 f"""
-                The vectorization of the target function {V} must be a multiple of the length
-                of the bounds sequence {len(bounds)}.
+                The number of optimization runs {num_optimization_runs} must be a multiple of the
+                length of the bounds sequence {len(bounds)}.
                 """
             ),
         )
-        multiple = V // len(bounds)
-        bounds = bounds * multiple * num_optimization_runs_per_function
```
hstojic: what happened here, was this a bug?

trieste/acquisition/optimizer.py
```diff
@@ -603,25 +604,35 @@ def _objective_value_and_gradient(x: TensorType) -> Tuple[TensorType, TensorType
...
-            for lower, upper in zip(space.subspace_lower, space.subspace_upper)
+            [
+                get_bounds_of_box_relaxation_around_point(ss, starting_points[i : i + 1, j])
+                if isinstance(ss, TaggedProductSearchSpace)
+                else spo.Bounds(ss.lower, ss.upper)
+                for j, ss in enumerate(subspaces)
+            ]
+            for i in tf.range(num_optimization_runs_per_function)
         ]
+        bounds = list(chain.from_iterable(bounds))
```
hstojic: this is important to get right and perhaps a bit complex to test as a part of the whole optimizer? should we think about extracting creating bounds as a function and then more comprehensively test it?

trieste/acquisition/rule.py
```diff
@@ -2073,6 +2073,183 @@ def update(
...
+                    1,
+                    global_search_space.points.shape[0],
+                )
+            ),
+            1,
+        )[0]
+
+        # Pairwise distances for each axis in the global search space.
+        points = global_search_space.points
+        self._global_distances = tf.abs(tf.expand_dims(points, -2) - tf.expand_dims(points, -3))
```
hstojic: we should probably create a (private) method for computing this, in case somebody wants to be creative and do it differently, easier to subclass and change - and perhaps the same for determining the initial point above (and you also repeat that bit of the code below in `initialize`)

trieste/acquisition/rule.py
```diff
@@ -2073,6 +2073,183 @@ def update(
         pass
 
 
+class SingleObjectiveTrustRegionDiscrete(UpdatableTrustRegionDiscrete):
+    """
+    An updatable discrete trust region that maintains a set of neighboring points around a
+    single location point. The region is updated based on the best point found in the region.
```
hstojic: we should add some notes on:
- type of discrete variables that should be used here, i.e. discrete numerical
- scaling when it comes to the number of value combinations, i.e. describe the use case

trieste/acquisition/rule.py
```diff
@@ -2073,6 +2073,183 @@ def update(
...
+
+    @property
+    def location(self) -> TensorType:
+        """Center point of the region."""
+        return tf.reshape(tf.gather(self.global_search_space.points, self._location_ix), (-1,))
+
+    def _init_eps(self) -> None:
+        global_lower = self.global_search_space.lower
+        global_upper = self.global_search_space.upper
+        self.eps = 0.5 * (global_upper - global_lower) / (5.0 ** (1.0 / global_lower.shape[-1]))
```
hstojic, line 2150: we should parametrize the multiplier and update the eps rule, this one is outdated,  we overlooked this previously....

hstojic: can you raise a ticket to come back to this in a separate PR?

trieste/acquisition/rule.py
```diff
@@ -2073,6 +2073,183 @@ def update(
         pass
 
 
+class SingleObjectiveTrustRegionDiscrete(UpdatableTrustRegionDiscrete):
```
hstojic, line 2076: @vpicheny you should take a look as well!

hstojic (commented):
I did a quick round now, I think I want to step through the code for some of the stuff, but don't have time right now - I had some comments though that can be addressed already

tests/integration/test_mixed_space_bayesian_optimization.py
khurram-ghani: I was trying to limit the tests, but yes sure. I'll add that back in.

trieste/acquisition/optimizer.py
khurram-ghani: No not a bug. It is simply a change in behaviour in the the bounds calculation above. Previously the `bounds` array length was equal to `num-subspaces`, and the final line replicated that to be the full desired length. The new code creates an array of length `num-subspaces * num_optimization_runs_per_function` instead. So the checks and final replication is slightly different.

trieste/acquisition/rule.py
khurram-ghani: Ah yes, I forgot to mention the scaling limitations. I'll add that and the other point. Thanks.

trieste/acquisition/rule.py
khurram-ghani: I can add a parameter as part of this PR, e.g. `zeta` can default to `0.5`, but what about the divisor?

trieste/acquisition/optimizer.py
khurram-ghani: Yes, that makes sense. I'll look into that.

docs/notebooks/mixed_search_spaces.pct.py
khurram-ghani: I have added something. Please let me know if that isn't clear.

trieste/acquisition/rule.py
khurram-ghani: Done

trieste/acquisition/rule.py
khurram-ghani: I have added some notes. Is that clear?

trieste/acquisition/optimizer.py
khurram-ghani: Done

trieste/acquisition/rule.py
khurram-ghani: I guess we also want to change `SingleObjectiveTrustRegionBox`, so perhaps it makes sense to do it in a separate PR like you suggest.

tests/unit/acquisition/test_optimizer.py
```diff
@@ -567,6 +568,121 @@ def test_get_bounds_of_box_relaxation_around_point(
...
+                ([-3, 27], [0, 27]),
+                ([-1, 31], [2, 31]),
+                ([-3, 33], [0, 33]),
+                ([-1, 35], [2, 35]),
+                ([-3, 37], [0, 37]),
+            ],
+        ),
+    ],
+)
+def test_get_bounds_of_optimization(
```
hstojic, line 667: much better coverage, thats great!

tests/unit/acquisition/test_rule.py
```diff
@@ -1997,6 +2001,182 @@ def test_fixed_trust_region_discrete_update(
...
+) -> None:
+    tr = SingleObjectiveTrustRegionDiscrete(discrete_search_space)
+    with pytest.raises(ValueError, match="a single OBJECTIVE dataset must be provided"):
+        tr.get_dataset_min(datasets)
+
+
+def test_trust_region_discrete_get_dataset_min(discrete_search_space: DiscreteSearchSpace) -> None:
+    # Check get_dataset_min picks the minimum x and y values from the dataset.
+    dataset = Dataset(
+        tf.constant([[1, 1], [4, 4], [3, 4], [8, 8], [4, 4]], dtype=tf.float64),
```
hstojic, line 2023: seeing dtype hardcoded reminded me about our single precision stuff :)
perhaps we should add a test where dtype is correctly handled by the search space?

trieste/acquisition/rule.py
```diff
@@ -2073,6 +2073,191 @@ def update(
...
+        global_lower = self.global_search_space.lower
+        global_upper = self.global_search_space.upper
+        self.eps = 0.5 * (global_upper - global_lower) / (5.0 ** (1.0 / global_lower.shape[-1]))
+
+    def _compute_global_distances(self) -> None:
+        # Pairwise distances along each axis in the global search space.
+        points = self.global_search_space.points
+        self._global_distances = tf.abs(
+            tf.expand_dims(points, -2) - tf.expand_dims(points, -3)
+        )  # [num_points, num_points, D]
```
hstojic, line 2157: this is symmetric matrix, right?
in which case this is unnecessary wasteful in terms of memory, we only need lower/upper triangular matrix, not the full one

trieste/acquisition/rule.py
```diff
@@ -2073,6 +2073,191 @@ def update(
...
+        ``1 / beta``. Conversely, if it was unsuccessful, the size is reduced by the factor
+        ``beta``.
+
+        :param models: The model for each tag.
+        :param datasets: The dataset for each tag.
+        """
+
+        if not self._initialized or tf.reduce_any(self.eps < self._min_eps):
+            self.initialize(models, datasets)
+            return
```
hstojic, line 2213: how is this coordinated with other search space types?
if min eps is hit here the whole product TR should be initialised as well, we have `test_trust_region_discrete_update_initialize` test but do we have a test that it propagates?

trieste/acquisition/rule.py
```diff
@@ -2073,6 +2073,191 @@ def update(
...
+
+        datasets = self.select_in_region(datasets)
+        x_min, y_min = self.get_dataset_min(datasets)
+        self._location_ix = tf.where(
+            tf.reduce_all(self.global_search_space.points == x_min, axis=-1)
+        )
+        self._location_ix = tf.squeeze(self._location_ix, axis=-1)
```
hstojic, line 2220: this seems like the same code as `location` property?
perhaps a private method that is used both here and in the property?

hstojic (commented):
stepped through the code in detail and have few more comments, see below

tests/unit/acquisition/test_rule.py
khurram-ghani: That is a pretty common pattern we use for the tests in this file. In any case, I'll update one test that calls all the TR methods to tests both precisions. Probably not worth doing that for all tests, and hence doubling the number of them. That would at least test the basics.

trieste/acquisition/rule.py
khurram-ghani: This is the reverse operation to the one in the property. The property is using the location-index to select a point from the global set. Whereas here we are doing a search in the global set to find an index.

trieste/acquisition/rule.py
khurram-ghani: Yes it is symmetric. I did think about this, and decided it isn't worth the complexity of keeping only part of the matrix. You could in principle use a `tf.sparse.SparseTensor`, but that seems unnecessary since we expect the number of points to be small. Also, that might have a performance impact due to the operations being less likely to be parallelizable on a GPU. Note: in gpflow we also have symmetric covariance matrices, but we choose to always keep the full matrices.

trieste/acquisition/rule.py
khurram-ghani: Currently this isn't coordianated at all in the base implementation. Each sub-region only reacts and sees itself. So the whole product search space will not be initialised, only a part of it. This would need to be explicitly handled in a custom product class. However, it does seem like an important operation to support by default. I can look at how to propagate the intialisation up to the product... not sure how to do that yet. Do you think that is important to support here?

khurram-ghani: @hstojic I have an idea, but it requires a new property and hence touching multiple classes. So I think it is best if I resolve this in a separate PR.

khurram-ghani: Open PR #827 resolves this issue.

hstojic: ok, thats fair enough

trieste/acquisition/rule.py
hstojic: ok, if the benefits are not clear, lets stick to this simpler implementation

tests/unit/acquisition/test_rule.py
hstojic: I know its a pretty common, but we should start with explicitly testing handling of dtypes with anything new we implement - doesnt have to be dispersed throughout, can be a single test that just focuses on this aspect

hstojic (approved):
looks good

docs/notebooks/mixed_search_spaces.pct.py
```diff
@@ -236,35 +236,41 @@
...
-# continuous trust (sub-)region. The discrete sub-region is defined by
-# `FixedPointTrustRegionDiscrete` that selects a random point from the discrete space, which is then
-# fixed for the duration of the optimization. The continuous sub-region is defined by
-# `SingleObjectiveTrustRegionBox`, just as in the trust region notebook, where the region is updated
-# at each step of the optimization.
+# This setup is similar to the one used in the "Batch trust region rule"
+# section of the [trust region Bayesian optimization notebook](trust_region.ipynb). That notebook
+# creates trust regions of type `SingleObjectiveTrustRegionBox`. Here, we create trust regions that
+# are a product of a discrete and a continuous trust sub-region with
+# `UpdatableTrustRegionProduct`. The continuous part `SingleObjectiveTrustRegionBox` is the same as
```
vpicheny, line 249: The text is good, but I am wondering (prob not for this PR) if it would be better to do this automatically? As in, it is probably a bit of a headache for the use to know that if they have a mixed input space then they must use a TrustRegionProduct. Ideally they would use SingleObjectiveTR and we dispatch it automatically to the right kind?

trieste/acquisition/optimizer.py
```diff
@@ -766,6 +740,70 @@ def get_bounds_of_box_relaxation_around_point(
     return spo.Bounds(space_with_fixed_discrete.lower, space_with_fixed_discrete.upper)
 
 
+def get_bounds_of_optimization(space: SearchSpace, starting_points: TensorType) -> List[spo.Bounds]:
```
vpicheny, line 747: I would improve the docstring here if possible. 

I find it very difficult to understand what this does. Especially with the notion of "relaxed bounds".

I don't know if it is feasible at all but maybe even show what this does for a very simple example?

vpicheny (approved):
looks good overall! I was wondering if we could have abstracted things a bit more since this new class has a very bespoke TR scheme (e.g. init_eps is hard-coded, etc.), but this might not be possible.

docs/notebooks/mixed_search_spaces.pct.py
khurram-ghani: I'll think about this for a separate PR.

trieste/acquisition/optimizer.py
khurram-ghani: I have improved the docstring and also added an example.

trieste/acquisition/rule.py
khurram-ghani: PR #832 addresses this.

PR #826, opened 2024-02-29
Test DeepEnsemble with float32

uri-granta:
**Related issue(s)/PRs:** 

## Summary

...

## PR checklist

### DIFF
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -637,7 +637,7 @@ def _test_optimizer_finds_minimum(
     original_tf_cast = tf.cast
 
     def patched_tf_cast(x: TensorType, dtype: tf.DType) -> TensorType:
-        if x.dtype in (tf.float32, tf.float64) and x.dtype != dtype:
+        if isinstance(x, tf.Tensor) and x.dtype in (tf.float32, tf.float64) and x.dtype != dtype:
             raise ValueError(f"unexpected cast: {x} to {dtype}")
         return original_tf_cast(x, dtype)
 
--- a/tests/integration/test_bayesian_optimization.py
+++ b/tests/integration/test_bayesian_optimization.py
@@ -633,12 +633,21 @@ def _test_optimizer_finds_minimum(
             tf.cast(search_space.upper, dtype=tf.float32),
         )
 
-    # ensure there are no unnecessary casts from float64 to float32 or vice versa
     original_tf_cast = tf.cast
 
     def patched_tf_cast(x: TensorType, dtype: tf.DType) -> TensorType:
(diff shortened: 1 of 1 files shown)


tests/integration/test_bayesian_optimization.py
```diff
@@ -603,12 +622,21 @@ def _test_optimizer_finds_minimum(
...
+            tf.cast(search_space.upper, dtype=tf.float32),
+        )
+
     if model_type in [SparseVariational, DeepEnsemble]:
         num_initial_query_points = 20
     elif model_type in [DeepGaussianProcess]:
         num_initial_query_points = 25
 
     initial_query_points = search_space.sample(num_initial_query_points)
+    assert initial_query_points.dtype is tf.float32 if single_precision else tf.float64
```
hstojic: maybe also test whether generated query points are of correct dtype?

tests/integration/test_bayesian_optimization.py
```diff
@@ -16,7 +16,7 @@
 import tempfile
 from functools import partial
 from pathlib import Path
-from typing import Any, List, Mapping, Optional, Tuple, Type, Union, cast
+from typing import Any, Callable, List, Mapping, Optional, Tuple, Type, Union, cast
```
hstojic, line 21: we should do the same with ask-tell, no?

hstojic (commented):
a step in the right direction :)
left some comments - this checks if all runs, which is necessary, but certain parts of the pipeline might nevertheless cast back and forth which we don't test here - I'm not sure whats the best way to do that, just saying this is not sufficient - we'll probably need to  test each individual component, or perhaps each individual object should have a dtype property which could facilitate that...

cannot recall now, but do we have dtype checks in unit tests of the DE model - e.g. checking prediction, sample methods, and trajectory sampler?

tests/integration/test_bayesian_optimization.py
uri-granta: Yes, though I'll postpone that to a future PR as at the moment we don't have any DeepEnsemble tests with ask-tell. This PR was basically an initial nothing-blows-up check. Next step will be to (1) figure out how to check whether any conversions are happening under the hood (2) write lots more tests.

uri-granta:
By patching tf.cast I've verified that there are no unexpected casts happening in the test. The only casts are in `sample_model_index` (which casts deep ensemble size to tf.int32) and `write_summary_data_based_metrics` (which casts  dataset observations to the model prediction dtype).

hstojic (approved):
the diff became a bit mangled, but I think its good now

uri-granta:
> the diff became a bit mangled, but I think its good now

FYI you can add ?w=1 to the diff url to get it to ignore whitespace, which makes it much more readable!

PR #827, opened 2024-03-08
Add `requires_initialization` property to trust regions

khurram-ghani:
**Related issue(s)/PRs:** resolves re-init issue discussed in #825 

## Summary
This PR adds a property to trust-regions, allowing a rule to determine if a region needs to be re-initialized. The rule can then call the `initialize` method directly on the region. Previously this decision was made internally by the regions. This resulted in an issue for product spaces where the re-init of sub-regions wasn't coordinated. This PR resolves that issue.

There is a minor incompatibility that the regions are now required to set `self._initialized` to `True` in their `initialize` method. This instance variable starts off as `False` and should be set to `True` in any derived `initialize` methods.

## PR checklist

### DIFF
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -1410,8 +1410,8 @@ def test_trust_region_box_initialize() -> None:
     npt.assert_array_equal(trb._y_min, tf.constant([np.inf], dtype=tf.float64))
 
 
-# Update call initializes the box if eps is smaller than min_eps.
-def test_trust_region_box_update_initialize() -> None:
+# Box requires initialization if eps is smaller than min_eps.
+def test_trust_region_box_needs_initialize() -> None:
     search_space = Box([0.0, 0.0], [1.0, 1.0])
     datasets = {
         OBJECTIVE: Dataset(  # Points outside the search space should be ignored.
@@ -1423,15 +1423,17 @@ def test_trust_region_box_update_initialize() -> None:
     trb.initialize(datasets=datasets)
     location = trb.location
 
-    trb.update(datasets=datasets)
+    assert trb.needs_init
+    trb.initialize(datasets=datasets)
     npt.assert_array_compare(np.not_equal, location, trb.location)
     location = trb.location
 
-    trb.update(datasets=datasets)
+    assert trb.needs_init
+    trb.initialize(datasets=datasets)
     npt.assert_array_compare(np.not_equal, location, trb.location)
 
 
-# Update call does not initialize the box if eps is larger than min_eps.
+# Box does not need initialization if eps is larger than min_eps.
 def test_trust_region_box_update_no_initialize() -> None:
     search_space = Box([0.0, 0.0], [1.0, 1.0])
     datasets = {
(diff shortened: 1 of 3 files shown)


tests/unit/acquisition/test_rule.py
```diff
@@ -1410,8 +1410,8 @@ def test_trust_region_box_initialize() -> None:
     npt.assert_array_equal(trb._y_min, tf.constant([np.inf], dtype=tf.float64))
 
 
-# Update call initializes the box if eps is smaller than min_eps.
-def test_trust_region_box_update_initialize() -> None:
+# Box requires initialization if eps is smaller than min_eps.
```
hstojic: move the comments to the docstrings of the test functions?

hstojic (approved):
it looks good to me, this should resolve the issue with product search spaces!
but lets wait for @uri-granta 's approval as well :)

trieste/acquisition/rule.py
```diff
@@ -1013,6 +1013,11 @@ def __init__(
         self.region_index = region_index
         self.input_active_dims = input_active_dims
 
+    @property
+    def needs_init(self) -> bool:
```
uri-granta: (suggestion)

```suggestion
    def requires_initialization(self) -> bool:
```

trieste/acquisition/rule.py
```diff
@@ -1013,6 +1013,11 @@ def __init__(
...
+        """Return `True` if the search space needs to be initialized, and `False` otherwise."""
```
uri-granta: Maybe expand here a bit?

```suggestion
        """Return `True` if the search space needs to be re-initialized with the latest datasets, and `False` if it can be just updated."""
```

trieste/acquisition/rule.py
```diff
@@ -1013,6 +1013,11 @@ def __init__(
...
+        return False
```
uri-granta: Do we want to move `_initialized` into the UpdatableTrustRegion base class rather than implementing it each time? We could then default to returning `self._initialized`. (If you do this, then you can also remove the "initialization" in `FixedPointTrustRegionDiscrete.__init__`.)

uri-granta (approved):
Looks good. A couple of suggestions/questions.

trieste/acquisition/rule.py
khurram-ghani: Yes, makes sense. I thought about this before, but didn't go ahead with it. Now that you mention it ...

I have made the change, though I kept the "initialization" in `FixedPointTrustRegionDiscrete.__init__`. Safer to set the points to something (rather than `None`), and random init is the most obvious.

khurram-ghani: @uri-granta note though that this change adds a minor incompatibility, explained in the description:

> There is a minor incompatibility that the regions are now required to set `self._initialized` to `True` in their `initialize` method. This instance variable starts off as `False` and should be set to `True` in any derived `initialize` methods.

Thoughts?

uri-granta: I think it's fine, though I would document this in the `initialize` abstractmethod docstring. Alternatively, you could make `initialize` not abstract and default to just setting `self._initialized` to `True`, and call it via `super()` everywhere. Though that adds more code for no obvious benifit.

khurram-ghani: Thanks. I have added a comment in the docstring.

I agree that making `initialize` not abstract would be of limited benefit, plus it would not fix the backwards incompatibility, since existing classes would still need to be updated to call `super()...`.

PR #829, opened 2024-03-11
Remove some unnecessary tf.casts

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Remove some unnecessary casting, in particular in the various samplers.

## PR checklist

### DIFF
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -26,6 +26,7 @@ import tensorflow_probability as tfp
 from check_shapes.exceptions import ShapeMismatchError
 from scipy import stats
 
+from tests.unit.models.gpflow.test_interface import _QuadraticPredictor
 from tests.util.misc import TF_DEBUGGING_ERROR_TYPES, ShapeLike, quadratic, random_seed
 from tests.util.models.gpflow.models import (
     GaussianProcess,
@@ -353,9 +354,12 @@ def test_batch_reparametrization_sampler_samples_are_repeatable(qmc: bool, qmc_s
 
 @pytest.mark.parametrize("qmc", [True, False])
 @pytest.mark.parametrize("qmc_skip", [True, False])
-def test_batch_reparametrization_sampler_doesnt_cast(qmc: bool, qmc_skip: bool) -> None:
-    sampler = BatchReparametrizationSampler(100, _dim_two_gp(), qmc=qmc, qmc_skip=qmc_skip)
-    xs = tf.random.uniform([3, 5, 7, 2], dtype=tf.float64)
+@pytest.mark.parametrize("dtype", [tf.float32, tf.float64])
+def test_batch_reparametrization_sampler_doesnt_cast(
+    qmc: bool, qmc_skip: bool, dtype: tf.DType
+) -> None:
+    sampler = BatchReparametrizationSampler(100, _QuadraticPredictor(), qmc=qmc, qmc_skip=qmc_skip)
+    xs = tf.random.uniform([3, 1, 7, 7], dtype=dtype)
 
     original_tf_cast = tf.cast
 
--- a/tests/unit/models/gpflow/test_sampler.py
+++ b/tests/unit/models/gpflow/test_sampler.py
@@ -370,7 +370,9 @@ def test_batch_reparametrization_sampler_doesnt_cast(
         return original_tf_cast(x, dtype)
 
     with patch("tensorflow.cast", side_effect=patched_tf_cast):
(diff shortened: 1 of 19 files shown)


trieste/acquisition/function/active_learning.py
```diff
@@ -222,7 +222,7 @@ def acquisition(x: TensorType) -> TensorType:
         t = (threshold - mean) / stdev
         t_plus = t + alpha
         t_minus = t - alpha
-        normal = tfp.distributions.Normal(tf.cast(0, x.dtype), tf.cast(1, x.dtype))
+        normal = tfp.distributions.Normal(tf.constant(0, x.dtype), tf.constant(1, x.dtype))
```
hstojic, line 225: why is using constant better than casting?

uri-granta: tf.cast should be used for Tensors, tf.constant for static, non-Tensor types, and tf.convert_to_tensor for any mix of Tensor-like objects. Also, it reads better here, and makes testing for unnecessary casts easier.

tests/unit/models/gpflux/test_sampler.py
```diff
@@ -151,7 +152,17 @@ def test_dgp_reparam_sampler_sample_is_repeatable(
 
     sampler = DeepGaussianProcessReparamSampler(100, model)
     xs = tf.random.uniform([100, 2], minval=-10.0, maxval=10.0, dtype=tf.float64)[:, None, :]
-    npt.assert_allclose(sampler.sample(xs), sampler.sample(xs))
+
+    # also check there are no unnecessary casts from float64 to float32 or vice versa
```
uri-granta, line 156: (this should be a separate test but the sampler tests are currently very slow due to training; I'll try to get them to use a session-scoped fixture in a separate PR)

trieste/acquisition/function/active_learning.py
hstojic: thanks for the education :)

tests/unit/models/gpflux/test_sampler.py
hstojic: yes, that sounds good

tests/unit/models/gpflow/test_sampler.py
```diff
@@ -350,6 +352,29 @@ def test_batch_reparametrization_sampler_samples_are_repeatable(qmc: bool, qmc_s
...
+    def patched_tf_cast(x: TensorType, dtype: tf.DType) -> TensorType:
+        # ensure there are no unnecessary casts from float64 to float32 or vice versa
+        if isinstance(x, tf.Tensor) and x.dtype in (tf.float32, tf.float64) and x.dtype != dtype:
+            raise ValueError(f"unexpected cast: {x} to {dtype}")
+        return original_tf_cast(x, dtype)
+
+    with patch("tensorflow.cast", side_effect=patched_tf_cast):
+        samples = sampler.sample(xs)
+        assert samples.dtype is dtype
+        npt.assert_allclose(samples, sampler.sample(xs))
```
hstojic, line 375: neat test

tests/unit/models/keras/test_sampler.py
```diff
@@ -74,7 +75,16 @@ def test_ensemble_trajectory_sampler_returns_trajectory_function_with_correctly_
...
+    original_tf_cast = tf.cast
+
+    def patched_tf_cast(x: TensorType, dtype: tf.DType) -> TensorType:
+        # ensure there are no unnecessary casts from float64 to float32 or vice versa
+        if isinstance(x, tf.Tensor) and x.dtype in (tf.float32, tf.float64) and x.dtype != dtype:
+            raise ValueError(f"unexpected cast: {x} to {dtype}")
+        return original_tf_cast(x, dtype)
+
+    with patch("tensorflow.cast", side_effect=patched_tf_cast):
+        assert trajectory(test_data).shape == (num_evals, batch_size, num_outputs)
```
hstojic: why not testing the dtype of the output?
for DeepGPs perhaps not necessary as we should always use 64 there, but here we could conceivably use both...

hstojic (commented):
some additional comments, see below...

we should probably also test that output of certain methods in models is consistent, i.e. dtype is as expected - but this can be done in a separate PR 

an additional question, acq fnc optimizer is not doing any conversion/hardcoding?

tests/unit/models/keras/test_sampler.py
uri-granta: So it turns out that the hidden layers in DeepEnsemble meant that we were always using float32 for the predictions. I've tentatively changed it so it uses the dtypes in the KerasEnsemble networks. Does it look ok?

hstojic (approved):
looks good

PR #831, opened 2024-03-20
Fix slow tests

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Remove slow deep ensemble test with float32 as randomness has this currently failing (plus it's fairly slow). We still have the other faster tests ensuring that float32 isn't broken.

Perplexedly replace one tf.constant with a tf.cast as the latter was resulting in a result.save error in the expected improvements notebook on line 298 (but not when run in quick mode)!

## PR checklist


khurram-ghani (approved):
Looks good.

PR #834, opened 2024-03-26
Investigate AskTellOptimizer that doesn't track datasets

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Support an AskTellOptimizer that doesn't track datasets: instead you use tell to tell it that the datasets have been updated (passing in the complete updated datasets, not just the new points).

Also see: https://github.com/Prowler-io/automotive/pull/5598

## PR checklist


trieste/ask_tell_optimization.py
```diff
@@ -71,6 +72,21 @@
...
 
 
+@dataclass(frozen=True)
+class AskTellOptimizerState(Generic[StateType, ProbabilisticModelType]):
+    """
+    Internal state for an Ask/Tell optimizer. This can be obtained using the optimizer's
+    `state` property, and can be used to initialise a new instance of the optimizer.
+    """
+
+    record: Record[StateType, ProbabilisticModelType]
```
avullo, line 82: Will this then contain the min on each trust region? This is being added in this PR too?

trieste/ask_tell_optimization.py
```diff
@@ -447,30 +499,77 @@ def ask(self) -> TensorType:
 
         return query_points
 
-    def tell(self, new_data: Mapping[Tag, Dataset] | Dataset) -> None:
+    def tell(
+        self,
+        new_data: Mapping[Tag, Dataset] | Dataset,
+        new_data_ixs: Optional[Sequence[TensorType]] = None,
```
avullo, line 534: Wouldn't it be cleaner, easier to aggregate these two into a more general concept of `new_data`, i.e. data and local indices?

trieste/ask_tell_optimization.py
uri-granta: Once the acquisition rule in automotive has been updated, these will be accessible in `record.acquisition_state`.

trieste/ask_tell_optimization.py
uri-granta: In the mainline use case we don't bother providing indices, and instead have the AskTellOptimizer infer them from the change of the data size and number of regions. Also, this ways preserves backwards compatibility in a widely used interface.

trieste/bayesian_optimizer.py
```diff
@@ -999,6 +999,9 @@ def write_summary_observations(
 ) -> None:
     """Write TensorBoard summary for the current step observations."""
     for tag in models:
+        if tag not in tagged_output:
```
khurram-ghani, line 1002: Very minor point... feel free to ignore. It is probably more intuitive to check against `datasets`, since those concepts exist in multiple places. `datasets` and `tagged_output` keys should match anyway.

```suggestion
        if tag not in datasets:
```

trieste/ask_tell_optimization.py
```diff
@@ -69,6 +72,21 @@
 AskTellOptimizerType = TypeVar("AskTellOptimizerType")
 
 
+@dataclass(frozen=True)
+class AskTellOptimizerState(Generic[StateType, ProbabilisticModelType]):
+    """
+    Internal state for an Ask/Tell optimizer. This can be obtained using the optimizer's
+    `state` property, and can be used to initialise a new instance of the optimizer.
```
khurram-ghani: There is no `state` property, is there? Is this meant to say the `to_state` method?

trieste/ask_tell_optimization.py
```diff
@@ -410,6 +475,14 @@ def to_result(self, copy: bool = True) -> OptimizationResult[StateType, Probabil
         record: Record[StateType, ProbabilisticModelType] = self.to_record(copy=copy)
         return OptimizationResult(Ok(record), [])
 
+    def to_state(self) -> AskTellOptimizerState[StateType, ProbabilisticModelType]:
```
khurram-ghani: Should we add the `copy` argument like `to_result`, instead of hard-coding it to `False` internally?

trieste/ask_tell_optimization.py
```diff
@@ -447,30 +520,77 @@ def ask(self) -> TensorType:
...
-        if self._datasets.keys() != new_data.keys() and datasets_indices != {None}:
+        # the existing datasets are all global and the new data contains local datasets too.
+        if all(LocalizedTag.from_tag(tag).local_index is None for tag in self._datasets.keys()):
+            global_old = {LocalizedTag.from_tag(tag).global_tag for tag in self._datasets.keys()}
+            global_new = {LocalizedTag.from_tag(tag).global_tag for tag in new_data.keys()}
+            if global_new != global_old:
+                raise ValueError(
+                    f"new_data global keys {global_new} doesn't "
+                    f"match dataset global keys {global_old}"
+                )
```
khurram-ghani, line 556: I think we should only allow this exception when `new_data_ixs is None`. Otherwise we would have both existing local datasets in `new_data` and indices for them. In which case the `with_local_datasets` call below would ignore those indices (it only uses them if the local tags do not exist). This could be confusing for the user.

trieste/ask_tell_optimization.py
```diff
@@ -483,7 +603,7 @@ def tell(self, new_data: Mapping[Tag, Dataset] | Dataset) -> None:
         if summary_writer:
             with summary_writer.as_default(step=logging.get_step_number()):
                 write_summary_observations(
-                    self._datasets,
+                    datasets,
```
khurram-ghani, line 613: Similar to this change, `write_summary_initial_model_fit` call in `__init__` above should probably use `datasets` instead of `self._datasets`, as datasets are updated for rules with local dataset.

Also, the `write_summary_query_points` call in `ask` should probably also track `_dataset_ixs` for completeness. Otherwise it is missing some key information when `track_data==False`.

trieste/ask_tell_optimization.py
```diff
@@ -447,30 +520,77 @@ def ask(self) -> TensorType:
...
+                self._datasets[tag] += new_dataset
+            datasets: Mapping[Tag, Dataset] = self._datasets
+        elif not isinstance(self._acquisition_rule, LocalDatasetsAcquisitionRule):
+            datasets = new_data
+        else:
+            if new_data_ixs is None:
+                # infer dataset indices from change in dataset sizes
+                new_dataset_len = self.dataset_len(new_data)
+                num_new_points = new_dataset_len - self._dataset_len
+                if num_new_points < 0 or num_new_points % len(self._dataset_ixs) != 0:
```
khurram-ghani: Maybe create a local variable `num_local_datasets = len(self._dataset_ixs)` and use it in the various places in this `else` section (8?) for clarity? Even though `self._dataset_ixs` gets updated part way through, its length wouldn't change.

tests/unit/acquisition/test_utils.py
```diff
@@ -166,6 +166,50 @@ def test_with_local_datasets(
...
+) -> None:
+    original_datasets = dict(datasets).copy()
+    global_tags = {t for t in original_datasets if not LocalizedTag.from_tag(t).is_local}
+    num_global_datasets = len(global_tags)
+
+    num_local_datasets = len(indices)
+    datasets = with_local_datasets(datasets, num_local_datasets, indices)
+    for d in datasets.items():
+        print(d)
+        print("\n\n")
```
khurram-ghani: Prints left in by accident?

khurram-ghani (approved):
Looks good. Left a few comments/suggestions.

trieste/ask_tell_optimization.py
avullo: What do we need to do to the `FeasibleSetTrustRegionAcquisitionRule` in order for the record to contain the correct acquisition state?

PR #835, opened 2024-04-03
Refactor `SingleObjectiveTrustRegion*` classes to make them more generic

khurram-ghani:

## Summary
This PR refactors `SingleObjectiveTrustRegionBox` and `SingleObjectiveTrustRegionDiscrete` classes to reduce code duplication, plus it introduces more flexibility for algorithms (e.g. concept of `location_candidate`). There is a new common class called `HypercubeTrustRegion`.

## PR checklist


trieste/acquisition/rule.py
```diff
@@ -1484,122 +1484,113 @@ def filter_datasets(
...
-        Box.__init__(self, global_search_space.lower, global_search_space.upper)
-        UpdatableTrustRegion.__init__(self, region_index, input_active_dims)
-        self._global_search_space = global_search_space
-        # Random initial location in the global search space.
-        self.location = tf.squeeze(global_search_space.sample(1), axis=0)
+        models: Optional[Mapping[Tag, ProbabilisticModelType]] = None,
+        datasets: Optional[Mapping[Tag, Dataset]] = None,
+        location_candidate: Optional[TensorType] = None,
+    ) -> None:
+        # Initialize the location of the region.
```
uri-granta: Given that this is now called in subclasses, it should probably have a docstring rather than a comment.

trieste/acquisition/rule.py
```diff
@@ -1484,122 +1484,113 @@ def filter_datasets(
...
+        # If no candidate is provided, take a random location from the global search space.
+        if location_candidate is not None:
+            self.location = location_candidate
+        else:
+            self.location = tf.squeeze(self.global_search_space.sample(1), axis=0)
```
uri-granta: Definitely looks nicer with a setter :-)

trieste/acquisition/rule.py
```diff
@@ -1484,122 +1484,113 @@ def filter_datasets(
...
+    @abstractmethod
     def global_search_space(self) -> SearchSpace:
-        """The global search space this search space lives in."""
-        return self._global_search_space
+        """The global search space this region lives in."""
 
 
-class SingleObjectiveTrustRegionBox(UpdatableTrustRegionBox):
-    """An updatable box search space for use with trust region acquisition rules."""
+# TODO: what is a good name for this class?
```
uri-granta: Can you explain the type hierarchy here? AFAICT there are the following (and probably some more I've missed):

1. `UpdatableTrustRegion(SearchSpace)`
    1. `UpdatableTrustRegionWithLocationInGlobalSearchSpace(UpdatableTrustRegion)` — are there any updatable trust regions that _don't_ have a location in global search space? or can we merge this with the parent class?
        1. `UpdatableTrustRegionBox(Box, UpdatableTrustRegionWithLocationInGlobalSearchSpace)`
        1. `UpdatableTrustRegionDiscrete(DiscreteSearchSpace, UpdatableTrustRegionWithLocationInGlobalSearchSpace)`
        1. `GenericUpdatableTrustRegion(UpdatableTrustRegionWithLocationInGlobalSearchSpace)` — is the point of this to add the mechanism for defining single objective updatable trust regions for use with trust region acquisition rules? If so, maybe call it `SingleObjectiveTrustRegion`?
            1. `SingleObjectiveTrustRegionBox(UpdatableTrustRegionBox, GenericUpdatableTrustRegion)`
            1. `SingleObjectiveTrustRegionDiscrete(UpdatableTrustRegionDiscrete, GenericUpdatableTrustRegion)`
    1. `UpdatableTrustRegionProduct(TaggedProductSearchSpace, UpdatableTrustRegion)`

uri-granta (commented):
(still reviewing but thought I'd post the question re class hierarchy now so you can respond!)

trieste/acquisition/rule.py
khurram-ghani: That is the correct hierarchy. The main concept is that `GenericUpdatableTrustRegion` implements the core algorithm, and depends on location/global-space but doesn't have an implementation for them. Classes `UpdatableTrustRegionBox` and `UpdatableTrustRegionDiscrete` provide this implementation of location/global-space for different types of regions. These ingredients are brought together in the final concrete classes `SingleObjectiveTrustRegion*`.

> ... can we merge this with the parent class?

The intention was for `UpdatableTrustRegion` to be more generic. We don't have any trust-regions without location etc., but @hstojic uses these classes for completely different use. So after discussion with @hstojic, the right answer probably is to rename `UpdatableTrustRegion` as `UpdatableSearchSpace` and move it to `space.py`; and then drop `WithLocationInGlobalSearchSpace` from the other class (plus some other related minor refactor).

> ... If so, maybe call it SingleObjectiveTrustRegion?

The `GenericUpdatableTrustRegion` now support algorithms with multiple datasets, so not just single objective trust regions. So that is not a suitable name.

khurram-ghani: Actually, I don't think I will move `UpdatableSearchSpace` to `space.py`. It depends on `Dataset` and `Tag` etc., and that would add extra dependencies to that file.

trieste/acquisition/rule.py
```diff
@@ -991,7 +991,7 @@ def acquire(
         return thompson_samples
 
 
-class UpdatableTrustRegion(SearchSpace):
+class UpdatableSearchSpace(SearchSpace):
```
hstojic, line 994: perhaps I made a mistake - there is a bunch of methods here that are obviously trust region specific, I forgot about that... so perhaps we need a class one level up that doesnt have all the TR stuff, or we can move the TR methods below to `UpdatableTrustRegion`?

trieste/acquisition/rule.py
```diff
@@ -1484,122 +1525,81 @@ def filter_datasets(
         return filtered_datasets
 
 
-class UpdatableTrustRegionBox(Box, UpdatableTrustRegion):
-    """
-    A simple updatable box search space with a centre location and an associated global search
-    space.
-    """
+# TODO: what is a good name for this class?
+class GenericUpdatableTrustRegion(UpdatableTrustRegion):
```
hstojic: why exactly cant we combine this one with `UpdatableTrustRegion`? why do we need separate classes?

hstojic (commented):
few quick questions

trieste/acquisition/rule.py
hstojic: ok, re-read the thread above, so I think now understand it a bit better

trieste/acquisition/rule.py
```diff
@@ -1484,122 +1525,81 @@ def filter_datasets(
...
+    """A generic updatable search space for use with trust region acquisition rules."""
```
hstojic: add further explanation, how is this one different from the other classes, perhaps adding concrete use cases - and same goes for docstrings of other classes

trieste/acquisition/rule.py
khurram-ghani: I moved one method (`get_datasets_filter_mask`) over to `UpdatableTrustRegion` already. The 3 main remaining ones are `initialize`, `update` and `select_in_region` (and some related ones it depends on). So it is just `select_in_region` that we could move, but I thought that it could be more generally useful.

trieste/acquisition/rule.py
khurram-ghani: This is a specific algorithm for intialize/update. For example, TURBO is different and doesn't use this class.

hstojic (approved):
there are some minor comments for improving the docstrings but approved it
though, honestly, I'm a bit lost with the whole hierarchy of classes, so there is a bit of trust on this one that this refactoring is the right way forward

khurram-ghani:
@hstojic thanks for the review. Did you have a name suggestion for `GenericUpdatableTrustRegion`? I think it could have "Box" somewhere in the name because that is how it defines local neighbours, for both continuous and discrete variables. It then grows and shrinks these boxes. However, we can't use "Box" because Trieste uses that to mean continuous spaces, and that would be confusing.

hstojic:
Not a strong opinion at the end as I don't have a clear enough view on how classes are used and the differences between them

Perhaps if its specific in a way that its using hyper rectangles rather than other types such as ellipsoid's, we could put that in the name?Hypercube actually?

trieste/acquisition/rule.py
khurram-ghani: Thinking about it some more, I think I will move `select_in_region` to `UpdatableTrustRegion`. Especially since it has "region" in the name. I will also move the related code, e.g. `input_active_dims`.

khurram-ghani:
@hstojic on the hierarchy of classes, I had discussions with @uri-granta about that and he is happy with it :-)

PR #841, opened 2024-05-10
Add state handling to filter_datasets

uri-granta:
**Related issue(s)/PRs:** 

## Summary

...

## PR checklist

### DIFF
--- a/tests/unit/acquisition/test_rule.py
+++ b/tests/unit/acquisition/test_rule.py
@@ -681,9 +681,7 @@ def test_trego_successful_global_to_global_trust_region_unchanged(
 
     assert current_state is not None
     current_subspace = current_state.acquisition_space.get_subspace("0")
-    assert isinstance(current_subspace, TREGOBox)
-    npt.assert_array_almost_equal(current_subspace.lower, lower_bound)
-    npt.assert_array_almost_equal(current_subspace.upper, upper_bound)
+    assert current_subspace is subspace
 
     npt.assert_array_almost_equal(subspace._eps, eps)
     assert subspace._is_global
@@ -728,9 +726,7 @@ def test_trego_for_unsuccessful_global_to_local_trust_region_unchanged(
 
     assert current_state is not None
     current_subspace = current_state.acquisition_space.get_subspace("0")
-    assert isinstance(current_subspace, TREGOBox)
-    npt.assert_array_almost_equal(current_subspace.lower, acquisition_space.lower)
-    npt.assert_array_almost_equal(current_subspace.upper, acquisition_space.upper)
+    assert current_subspace is subspace
 
     npt.assert_array_almost_equal(subspace._eps, eps)
     assert not subspace._is_global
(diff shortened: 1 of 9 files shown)


tests/unit/acquisition/test_rule.py
```diff
@@ -610,19 +612,15 @@ def test_trego_for_default_state(
...
-    tr.filter_datasets({OBJECTIVE: model}, {OBJECTIVE: dataset})
+    state, _ = tr.filter_datasets({OBJECTIVE: model}, {OBJECTIVE: dataset})(state)
 
     assert state is not None
     ret_subspace = state.acquisition_space.get_subspace("0")
     assert isinstance(ret_subspace, TREGOBox)
     npt.assert_array_almost_equal(ret_subspace.lower, lower_bound)
     npt.assert_array_almost_equal(ret_subspace.upper, upper_bound)
-
-    npt.assert_array_almost_equal(query_point, [expected_query_point], 5)
```
khurram-ghani, line 621: Did you mean to remove the query-point assertion as well?

trieste/acquisition/rule.py
```diff
@@ -1434,11 +1448,11 @@ def state_func(
...
                 else:
                     _datasets = None
+                acquisition_space = TaggedMultiSearchSpace(subspaces, self._tags)
                 points = self._rule.acquire(acquisition_space, models, _datasets)
 
-            # We may modify the regions in filter_datasets later, so return a copy.
-            state_ = BatchTrustRegion.State(copy.deepcopy(acquisition_space))
-            return state_, tf.reshape(points, [-1, len(self._subspaces), points.shape[-1]])
+            state_ = BatchTrustRegionState(subspaces, self._tags)
+            return state_, tf.reshape(points, [-1, len(self._init_subspaces), points.shape[-1]])
```
khurram-ghani: Nitpick: probably slightly clearer to use `subspaces` as this is what the surrounding code is referring to.
 
```suggestion
            return state_, tf.reshape(points, [-1, len(subspaces), points.shape[-1]])
```

trieste/acquisition/rule.py
```diff
@@ -1488,46 +1502,70 @@ def get_initialize_subspaces_mask(
...
-                    used_masks[tag] = tf.logical_or(used_masks[tag], in_region)
-
-        filtered_datasets = {}
-        for tag, used_mask in used_masks.items():
-            filtered_datasets[tag] = Dataset(
-                tf.boolean_mask(datasets[tag].query_points, used_mask),
-                tf.boolean_mask(datasets[tag].observations, used_mask),
-            )
+                assert self._init_subspaces is not None, "the subspaces have not been initialized"
+                assert self._tags is not None
```
khurram-ghani, line 1529: I think there is an existing issue here that has been highlighted by this.

Contrary to what we discussed previously, `filter_datasets` does need to handle the case when `state` is `None`, as it can be called first by bayesian/ask_tell optimizers; which is what you are doing here. However, in this case we also want to support the case where `self._init_subspaces` is None. The derived classes (e.g. `BatchTrustRegionBox`) handle this in `acquire` by creating some default subspaces; but they don't do this if `filter_datasets` is called first.

So I think we need a way (in the derived classes) to also create default subspaces in `filter_datasets`.

uri-granta: One issue is that acquire is passed the search space while filter_datasets isn't.

Can you describe the workflow that is currently broken? AFAICT filter_datasets is *always* called first, right?

trieste/ask_tell_optimization.py
```diff
@@ -288,7 +288,17 @@ def __init__(
                 datasets = with_local_datasets(
                     self._datasets, num_local_datasets, self._dataset_ixs
                 )
-        self._filtered_datasets = self._acquisition_rule.filter_datasets(self._models, datasets)
+            self._acquisition_rule.initialize_subspaces(search_space)
```
khurram-ghani, line 291: We should also do this for `bayesian_optimizer.py` under `step == 1`, and add a unit test (like you did for ask_tell).

uri-granta: Good spot. Done.

khurram-ghani (approved):

PR #843, opened 2024-05-16
Don't save TR `ymin` from init dataset

khurram-ghani:

## Summary
Tweaked TR initialization algorithm to not take the `y_min` value from the initial dataset, and instead leave it at `inf`. The previous version had the inconsistency that at initialisation the saved `y_min` value was not at the region center/location. This change means that the first update/step will always be successful.

Discussed and agreed this change with @vpicheny.

## PR checklist


uri-granta (approved):
LGTM

PR #849, opened 2024-05-31
Support initialising AskTellOptimizer with extended datasets

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Another use case for `track_data=False`: being able to initialise a new AskTellOptimizer using the state saved after ask was called but before the datasets were updated. The acquisition state is already updated correctly due to filter_dataset. This fix adds the dataset length to the AskTellOptimiserState so that the local_data_ixs can also be updated automatically.

## PR checklist


avullo (approved):
lgtm

PR #853, opened 2024-06-06
Fix ask_tell global `dataset_len` detection

khurram-ghani:

## Summary
Quick fix for `dataset_len` detection. `tf.unique` returns a list of indices of the same length as the input, so we should use `unique_lens` to check the length instead.

## PR checklist


avullo (approved):
lgtm

PR #858, opened 2024-06-22
Support TF 2.16

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Add support for Tensorflow 2.16.1. Since this uses Keras 3 as the default keras, I've had to switch all uses of `tf.keras` (including hidden ones like `tf.optimizer`) to gpflow's `tf_keras` version-agnostic shortcut.

I've also put a maximum supported TF version, to prevent people trying to install trieste with unsupported versions in the future (since TF minor versions are offten not backwards compatible).

And I've removed support for Python 3.7 (which is already end-of-life) and 3.8 (which ends of life in less than a month) and switched the minimum supported version to 3.9.

## PR checklist


khurram-ghani (approved):
LGTM. (Had a quick chat offline).

PR #863, opened 2024-07-31
Categorical search spaces

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Add support for Categorical search spaces.

Modelling support will come in a future PR (see #862).

## PR checklist

### DIFF
--- a/trieste/space.py
+++ b/trieste/space.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import operator
 from abc import ABC, abstractmethod
 from functools import reduce
+from itertools import chain
 from typing import Callable, Optional, Sequence, Tuple, TypeVar, Union, overload
 
 import numpy as np
@@ -24,6 +25,7 @@ import scipy.optimize as spo
 import tensorflow as tf
 import tensorflow_probability as tfp
 from check_shapes import check_shapes
+from typing_extensions import Protocol, runtime_checkable
 
 from .types import TensorType
 
@@ -32,6 +34,9 @@ SearchSpaceType = TypeVar("SearchSpaceType", bound="SearchSpace")
 
 DEFAULT_DTYPE: tf.DType = tf.float64
 """ Default dtype to use when none is provided. """
+
+EncoderFunction = Callable[[TensorType], TensorType]
+""" Type alias for point encoders. These transform points from one search space to another. """
 
 
 class SampleTimeoutError(Exception):
@@ -263,6 +268,11 @@ class SearchSpace(ABC):
 
     @property
     @abstractmethod
+    def has_bounds(self) -> bool:
+        """Whether the search space has meaningful numerical bounds."""
+
+    @property
+    @abstractmethod
     def lower(self) -> TensorType:
         """The lowest value taken by each search space dimension."""
 
@@ -369,22 +379,15 @@ class SearchSpace(ABC):
     @property
     def has_constraints(self) -> bool:
         """Returns `True` if this search space has any explicit constraints specified."""
-        # By default assume there are no constraints; can be overridden by a subclass.
+        # By default, assume there are no constraints; can be overridden by a subclass.
         return False
 
 
-class DiscreteSearchSpace(SearchSpace):
-    r"""
-    A discrete :class:`SearchSpace` representing a finite set of :math:`D`-dimensional points in
-    :math:`\mathbb{R}^D`.
-
-    For example:
-
-        >>> points = tf.constant([[-1.0, 0.4], [-1.0, 0.6], [0.0, 0.4]])
-        >>> search_space = DiscreteSearchSpace(points)
-        >>> assert tf.constant([0.0, 0.4]) in search_space
-        >>> assert tf.constant([1.0, 0.5]) not in search_space
-
+class GeneralDiscreteSearchSpace(SearchSpace):
+    """
+    An ABC representing different types of discrete search spaces (not just numerical).
+    This contains a default implementation using explicitly provided points which subclasses
+    may ignore.
     """
 
     def __init__(self, points: TensorType):
@@ -397,20 +400,6 @@ class DiscreteSearchSpace(SearchSpace):
         self._points = points
         self._dimension = tf.shape(self._points)[-1]
 
-    def __repr__(self) -> str:
-        """"""
-        return f"DiscreteSearchSpace({self._points!r})"
-
-    @property
-    def lower(self) -> TensorType:
-        """The lowest value taken across all points by each search space dimension."""
-        return tf.reduce_min(self.points, -2)
-
-    @property
-    def upper(self) -> TensorType:
-        """The highest value taken across all points by each search space dimension."""
-        return tf.reduce_max(self.points, -2)
-
     @property
     def points(self) -> TensorType:
         """All the points in this space."""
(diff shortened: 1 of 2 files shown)


trieste/space.py
```diff
@@ -480,6 +502,175 @@ def __eq__(self, other: object) -> bool:
...
+        if not isinstance(categories, Sequence) or not (
+            all(
+                isinstance(x, Sequence)
+                and not isinstance(x, str)
+                and all(isinstance(y, str) for y in x)
+                for x in categories
+            )
+            or all(isinstance(x, int) for x in categories)
+        ):
+            raise TypeError("Invalid category description: expected either numbers or names.")
```
khurram-ghani: Include `categories` in the error message?

trieste/space.py
```diff
@@ -480,6 +502,175 @@ def __eq__(self, other: object) -> bool:
...
+    such sets.
+
+    For example:
+
+        >>> CategoricalSearchSpace(5)
+        CategoricalSearchSpace([('0', '1', '2', '3', '4')])
+        >>> CategoricalSearchSpace(["Red", "Green", "Blue"])
+        CategoricalSearchSpace([('Red', 'Green', 'Blue')])
+        >>> CategoricalSearchSpace([2,3])
+        CategoricalSearchSpace([('0', '1'), ('0', '1', '2')])
```
khurram-ghani, line 534: Perhaps also include an example of multidimensional string case.

trieste/space.py
```diff
@@ -480,6 +502,175 @@ def __eq__(self, other: object) -> bool:
...
+
+    @property
+    def one_hot_encoder(self) -> EncoderFunction:
+        """A one-hot encoder for the numerical indices."""
+
+        def encoder(x: TensorType) -> TensorType:
+            if tf.rank(x) != 2:
+                raise ValueError(
+                    f"Invalid input for one-hot encoding: expected rank 2, got {tf.rank(x)}"
+                )
```
khurram-ghani: Should we consider supporting an arbitrary number of batch dimensions, and not just 1? I think this will require flattening/unflattening them as `tf.keras.layers.CategoryEncoding` only supports 1D (for the one hot case).

trieste/space.py
```diff
@@ -480,6 +502,175 @@ def __eq__(self, other: object) -> bool:
...
+        category tags.
+
+        :param indices: A tensor of integer indices.
+        :return: A tensor of string tags.
+        """
+
+        def extract_tags(row: TensorType) -> TensorType:
+            return tf.stack(
+                [tf.gather(tf.constant(self._tags[i]), row[i]) for i in range(len(row))]
+            )
```
khurram-ghani, line 658: I _think_ you can probably use `tf.gather_nd` instead of `tf.gather` and a for loop. Not sure, but you might be able to also get rid of `tf.map_fn`.

trieste/space.py
```diff
@@ -1136,6 +1336,23 @@ def product(self, other: TaggedProductSearchSpace) -> TaggedProductSearchSpace:
...
+
+        def encoder(x: TensorType) -> TensorType:
+            components = []
+            for tag in self.subspace_tags:
+                component = self.get_subspace_component(tag, x)
+                space = self.get_subspace(tag)
+                if isinstance(space, HasOneHotEncoder):
+                    component = space.one_hot_encoder(component)
+                components.append(component)
+            return tf.concat(components, axis=1)
```
khurram-ghani: Nitpick: suggest using `axis=-1`, allowing an arbitrary number of batch dimensions.

khurram-ghani (commented):
Some comments on the main changes. I haven't looked at the unit testing yet.

trieste/space.py
```diff
@@ -443,6 +432,39 @@ def sample(self, num_samples: int, seed: Optional[int] = None) -> TensorType:
...
+        >>> points = tf.constant([[-1.0, 0.4], [-1.0, 0.6], [0.0, 0.4]])
+        >>> search_space = DiscreteSearchSpace(points)
+        >>> assert tf.constant([0.0, 0.4]) in search_space
+        >>> assert tf.constant([1.0, 0.5]) not in search_space
+
+    """
+
+    def __repr__(self) -> str:
+        """"""
+        return f"DiscreteSearchSpace({self._points!r})"
```
hstojic, line 453: `self.__class__.__name__` instead of a string?

hstojic (commented):
done a quick pass, looks ok to me, no major comments
I'll let @khurram-ghani approve it

trieste/space.py
uri-granta: Not sure that's possible, as self._tags is a non-rectangular Sequence[Sequence[str], not a tensor.

trieste/space.py
uri-granta: Good suggestion!

trieste/space.py
khurram-ghani: Okay :+1: .

There might be a way ([see example](https://stackoverflow.com/questions/67123407/tensorflow-how-to-use-a-ragged-tensor-as-an-index-into-a-normal-tensor)) to use ragged tensors with `batch_dims` arg to `tf.gather`, but that might not be more efficient than a `for` loop. Happy for you to ignore this :-)

khurram-ghani (approved):
Looks good.

However, my personal preference is to not put the `one_hot_encoder` in the categorical and product search space classes. This could be a separate class that takes a search space as an `__init__` argument.

I realise that these are just default implementations and users can always write separate implementations. As we know, there would be cases where we want to encode in different ways, especially for the product space (perhaps the categorical-only case is fine). I just think it is better to have the implied flexibility and not burn a default implementation in the spaces. I guess it just depends on our opinion on what is the most likely use case.

trieste/space.py
```diff
@@ -1136,6 +1348,23 @@ def product(self, other: TaggedProductSearchSpace) -> TaggedProductSearchSpace:
         """
         return TaggedProductSearchSpace(spaces=[self, other])
 
+    @property
+    def one_hot_encoder(self) -> EncoderFunction:
```
qiq208, line 1352: @khurram-ghani @uri-granta am i right in thinking we will need an equivalent of this for the embedding layers?

trieste/space.py
```diff
@@ -480,6 +503,186 @@ def __eq__(self, other: object) -> bool:
...
+                    f"expected {len(self.tags)} tags, got {flat_x.shape[-1]}"
+                )
+            columns = tf.split(flat_x, flat_x.shape[-1], axis=1)
+            encoders = [
+                tf.keras.layers.CategoryEncoding(num_tokens=len(ts), output_mode="one_hot")
+                for ts in self.tags
+            ]
+            encoded = tf.concat(
+                [encoder(column) for encoder, column in zip(encoders, columns)], axis=1
+            )
```
khurram-ghani, line 637: Given `x` is now a float type, do we need to do some checking/casting similar to `to_tags`? I'm not sure if `tf.keras.layers.CategoryEncoding` allows non-integer inputs.

uri-granta: It's only the underling self.points in GeneralDiscreteSearchSpace that are now floats. The one hot encoding uses self.tags directly so isn't affected. (And this is all tested.)

khurram-ghani: The `x` passed in by the user are the underlying indices, which can be float. We then call the encoder (after flattening) with these. From what I can see, the testing uses integer indices only, but I could have easily missed that. But in any case, it doesn't matter too much as I just tried out `tf.keras.layers.CategoryEncoding` with float types and it silently converts them into integers (in a weird way, see below). But perhaps we want to be explicit with our checking/assertions and not rely on tensorflow.

```
> e = tf.keras.layers.CategoryEncoding(4, output_mode="one_hot")

> e(tf.constant([3.0, 2.999999], dtype=tf.float32))
<tf.Tensor: shape=(2, 4), dtype=float32, numpy=
array([[0., 0., 0., 1.],
       [0., 0., 1., 0.]], dtype=float32)>

> e(tf.constant([3.0, 2.9999999], dtype=tf.float32))
<tf.Tensor: shape=(2, 4), dtype=float32, numpy=
array([[0., 0., 0., 1.],
       [0., 0., 0., 1.]], dtype=float32)>
```

PR #864, opened 2024-08-02
Encoded models

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Support for encoded models. Follows on from #863.

## PR checklist


trieste/models/interfaces.py
```diff
@@ -742,3 +743,159 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
         :return: The covariance with the top fidelity for the `query_points`, of shape [N, P]
         """
         raise NotImplementedError
+
+
+class EncodedProbabilisticModel(ProbabilisticModel):
```
pio-neil, line 753: Since this is not abstract (not an interface) I guess this means that none of our interfaces in automotive can extend from it. But I'm not sure if this is a problem, since I don't fully understand the intended usage at this point.

trieste/models/interfaces.py
```diff
@@ -742,3 +743,159 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+        """Encode points or a Dataset using the query point encoder."""
+        if self.encoder is None:
+            return points
+        elif isinstance(points, Dataset):
+            return Dataset(self.encoder(points.query_points), points.observations)
+        else:
+            return self.encoder(points)
+
+    @abstractmethod
+    def predict_encoded(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
```
pio-neil, line 785: I guess I'm missing some context, but I'm wondering when we need to predict on encoded query points (`predict_encoded`) vs raw query points (`predict`).

trieste/models/interfaces.py
pio-neil: I guess the main question is "do automotive models (like [DeepEnsembleBOModel](https://github.com/Prowler-io/automotive/blob/1ce61f36e17c105b740bc2734807842286300301/engine-tuning/sm_calibration/model/trieste_models/deep_ensemble.py#L98)) need to implement `predict_encoded` and `predict_encoded`, or is it sufficient for them to implement `predict` and `sample` as they do now?

In other words, is it sufficient for only the wrapped models (like Trieste's `DeepEnsemble`) to implement these methods?

trieste/models/interfaces.py
```diff
@@ -742,3 +743,159 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+
+    Classes that inherit from this (or the other associated mixins below) should implement the
+    relevant _encoded methods (e.g. predict_encoded instead of predict), to which the public
+    methods delegate after encoding their input. Take care to use the correct methods internally
+    to avoid encoding twice accidentally.
+    """
+
+    @property
+    @abstractmethod
+    def encoder(self) -> EncoderFunction | None:
```
pio-neil, line 764: It seems a little strange to me that `encoder` and `encode` are public methods. But maybe this is necessary.

pio-neil (commented):
Added a few comments ... just discussion points for our chat at 1.30

trieste/models/interfaces.py
uri-granta: Yes, the various TriesteBOModels can stay exactly as they are — as they simply delegate predict, etc to the internal trieste model (which will now do the encoding for them).

trieste/models/interfaces.py
uri-granta: I'll explain it better in our chat, but essentially external actors will always call `predict` (and typically won't even know that this is an encoded model). Internally, we may need to call `predict_encoded` e.g. to generate samples, as by that point encoding will already have happened.

trieste/models/interfaces.py
uri-granta: Encoder is an abstract property to ensure that any encoded model has an actual encoder it can use.

Encode is a helper method for models that implement additional methods that need encoding (e.g. `DeepEnsemble.sample_ensemble`).

tests/integration/test_mixed_space_bayesian_optimization.py
```diff
@@ -190,3 +199,86 @@ def test_optimizer_finds_minima_of_the_scaled_branin_function(
...
+    problem = categorical_scaled_branin(tf.random.shuffle(points))
+    initial_query_points = problem.search_space.sample(5)
+    observer = mk_observer(problem.objective)
+    initial_data = observer(initial_query_points)
+
+    # model uses one-hot encoding for the categorical inputs
+    encoder = one_hot_encoder(problem.search_space)
+    encoded_space = one_hot_encoded_space(problem.search_space)
+    model = GaussianProcessRegression(
+        build_gpr(encode_dataset(initial_data, encoder), encoded_space, likelihood_variance=1e-8),
```
hstojic: perhaps `encode_dataset` call should happen in the builder itself automatically if a mixed search space is used?

tests/integration/test_mixed_space_bayesian_optimization.py
```diff
@@ -190,3 +199,86 @@ def test_optimizer_finds_minima_of_the_scaled_branin_function(
...
+def test_optimizer_finds_minima_of_the_categorical_scaled_branin_function(
+    num_steps: int,
+    acquisition_rule: AcquisitionRule[
+        TensorType, TaggedProductSearchSpace, TrainableProbabilisticModel
+    ],
+) -> None:
+    # 6 categories mapping to 3 random points plus the 3 minimizer points
+    points = tf.concat(
+        [tf.random.uniform([3], dtype=tf.float64), ScaledBranin.minimizers[..., 0]], 0
+    )
```
hstojic, line 253: I'm not sure I get the logic behind this problem...

tests/integration/test_mixed_space_bayesian_optimization.py
```diff
@@ -190,3 +199,86 @@ def test_optimizer_finds_minima_of_the_scaled_branin_function(
         acquisition_function = acquisition_rule._acquisition_function
         if isinstance(acquisition_function, AcquisitionFunctionClass):
             assert acquisition_function.__call__._get_tracing_count() <= 4  # type: ignore
+
+
+def categorical_scaled_branin(
```
hstojic, line 202: can you clarify what are you doing here?

tests/unit/test_space.py
```diff
@@ -1756,51 +1756,51 @@ def test_categorical_search_space__to_tags_raises_for_non_integers() -> None:
     [
         (
             CategoricalSearchSpace(["V"]),
-            tf.constant([[0], [0]]),
-            tf.constant([[1], [1]], dtype=tf.float32),
+            tf.constant([[0], [0]], dtype=tf.float64),
+            tf.constant([[1], [1]], dtype=tf.float64),
```
hstojic, line 1760: why testing only on 64?

trieste/models/interfaces.py
```diff
@@ -742,3 +743,164 @@ def covariance_with_top_fidelity(self, query_points: TensorType) -> TensorType:
...
+            return encode_dataset(points, self.encoder)
+        else:
+            return self.encoder(points)
+
+    @abstractmethod
+    def predict_encoded(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
+        """Implementation of predict on encoded query points."""
+
+    @abstractmethod
+    def sample_encoded(self, query_points: TensorType, num_samples: int) -> TensorType:
```
khurram-ghani, line 789: Minor: perhaps add shape checks for various `*_encoded` methods?

khurram-ghani (approved):
Looks good.

One question though, what is the reason for not updating the multi-fidelity models? Is that going to be in a separate PR? I think for gpflux models we already discussed and said this would be done in a later PR.

hstojic (commented):
I'm really confused with changes in the `models` package, to me it seems we are unnecessarily appending `_encoded` to all of our methods - why is that needed? we don't seem to keep the old methods without it - why not simply keep the old names, we define a method `maybe_encode` that encodes the data / query points if encoder is passed through as an argument?

uri-granta:
> I'm really confused with changes in the `models` package, to me it seems we are unnecessarily appending `_encoded` to all of our methods - why is that needed? we don't seem to keep the old methods without it - why not simply keep the old names, we define a method `maybe_encode` that encodes the data / query points if encoder is passed through as an argument?

Keeping the old names is not possible for multiple reasons. One is that we need both the external (encoding) method and internal (already encoded) method as method implementations often call each other: e.g. sample might call predict and we don't want it to encode twice. Another is that automatically rewriting class methods using a mixin requires either a metaclass (which doesn't work with ABCs) or a class decorator (which confuses the IDE). A third is that we want to ensure people subclassing an encoded class with a new implementation of predict don't accidentally break encoding. Also worth noting that encoding affects a _lot_ of methods (including some ad hoc ones only used by specific model types) and that the datasets and query points are bare tensors so have no way of remembering whether they've already been encoded.

tests/integration/test_mixed_space_bayesian_optimization.py
uri-granta: The first dimension is a categorical space with 6 points, three of which are derived from the minimizers (to ensure the problem minimum corresponds to the global scaled branin minimum) and the other three randomly selected. We then shuffle these, as the order of the categories shouldn't affect how we treat them.

tests/unit/test_space.py
uri-granta: (changed a couple to float32, though dtypes are also tested elsewhere)

tests/integration/test_mixed_space_bayesian_optimization.py
uri-granta: Good suggestion. I'll also add optional arguments in case people want to use a different encoding to one_hot.

trieste/models/interfaces.py
uri-granta: I wasn't sure about that, as this is essentially an internal interface, so if someone wants to mess about with the batch dimensions as part of the encoding it's kinda fine as long as the external interface (which is shape checked) is still satisfied.

PR #865, opened 2024-08-07
Categorical trust regions

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Follows on from #864.

## PR checklist


trieste/acquisition/rule.py
```diff
@@ -1624,7 +1626,13 @@ def __init__(
         self._y_min = tf.constant(np.inf, dtype=self.location.dtype)
 
     def _init_eps(self) -> None:
-        self.eps = self._zeta * (self.global_search_space.upper - self.global_search_space.lower)
+        if not isinstance(self.global_search_space, HasOneHotEncoder):
```
khurram-ghani, line 1629: Using an `isinstance` check in this file to detect categorical spaces makes me a bit uncomfortable, as it forces developers to have to inherit from `HasOneHotEncoder`, and it feels too much of a special case. I think our original design decision was to use a property like `is_categorical` to do this. I don't have a strong objection, but just wanted to highlight that. I think we already discussed this before, but can't remeber the conclusion.

trieste/acquisition/rule.py
```diff
@@ -2256,27 +2267,36 @@ def location(self, location: TensorType) -> None:
...
-            tf.expand_dims(points, -2) - tf.expand_dims(points, -3)
-        )  # [num_points, num_points, D]
+        if isinstance(self.global_search_space, HasOneHotEncoder):
+            # use Hamming distance for categorical spaces
+            return tf.math.reduce_sum(
+                tf.where(tf.expand_dims(points, -2) == tf.expand_dims(points, -3), 0, 1),
+                axis=-1,
+                keepdims=True,  # (keep last dim for distance calculation below)
+            )  # [num_points, num_points, 1]
+        else:
```
khurram-ghani: I think maybe we should add more of an explanation here, as the categorical and numerical cases are slightly inconsistent.

The size of the last dimension for the numerical case is `D`, i.e. the distance in each dimension is calculated separately. Each dimesion is then separately tested against `distance` in `_get_points_within_distance` and it selects neighbors if all dimensions are within distance (i.e. `reduce_all` below).

For the categorical case the last dimension is `1` as we do a `reduce_sum`. I can see why that is, as we want to effectively do a `reduce_any` in `_get_points_within_distance`, i.e. the neighbors are selected if they are within `distance` in any dimension.

So we can add an explanation, or alternatively do the selection below and explicitly add a `reduce_any`.

tests/unit/acquisition/test_rule.py
```diff
@@ -2269,28 +2317,33 @@ def test_trust_region_discrete_update_size(
...
-        npt.assert_allclose(eps * tr._beta, tr.eps)
+        # Check that the region is smaller by beta (except for categorical)
+        npt.assert_allclose(1 if categorical else eps * tr._beta, tr.eps)
 
     # Check the new set of neighbors.
-    neighbors_mask = tf.abs(discrete_search_space.points - tr.location) <= tr.eps
-    neighbors_mask = tf.reduce_all(neighbors_mask, axis=-1)
-    neighbors = tf.boolean_mask(discrete_search_space.points, neighbors_mask)
+    if categorical:
+        # Hammond distance
```
khurram-ghani:
```suggestion
        # Hamming distance
```

tests/integration/test_mixed_space_bayesian_optimization.py
```diff
@@ -234,11 +234,50 @@ def objective(x: TensorType) -> TensorType:
...
+
+cat_problem = _get_categorical_problem()
+
+
 @random_seed
 @pytest.mark.parametrize(
     "num_steps, acquisition_rule",
     [
         pytest.param(25, EfficientGlobalOptimization(), id="EfficientGlobalOptimization"),
+        pytest.param(
```
khurram-ghani, line 273: Is it worth adding a `FixedPointTrustRegionDiscrete` test as well, like the discrete-numerical case?

Also, is it possible to combine `test_optimizer_finds_minima_of_the_categorical_scaled_branin_function` with `test_optimizer_finds_minima_of_the_scaled_branin_function`, so we don't have to maintain them separately?

trieste/acquisition/rule.py
uri-granta: Remind me: are there any categorical spaces (where we would wish to use Hamming distances) that are also numerically bounded? Because looking at it now, it seems much more natural to write

```suggestion
        if self.global_search_space.has_bounds:
```

Certainly we shouldn't calculate eps for unbounded spaces.

tests/integration/test_mixed_space_bayesian_optimization.py
uri-granta: I've merged them now using a helper function. The tests are reasonably slow so my preference is not to add a categorical `FixedPointTrustRegionDiscrete` test unless it's testing an important combination.

trieste/acquisition/rule.py
khurram-ghani: I don't think so. We won't have categorical spaces that have bounds. However, we could potentially have the reverse, i.e. spaces that are not bounded and are also not categorical.

uri-granta: Well we can't use eps as written for unbounded spaces, even if they're not categorical, as it uses the bounds. Is there any reason not to default to Hamming distance for those cases, at least for now?

khurram-ghani: That seems fine to me. @vpicheny what do you think?

khurram-ghani (approved):

PR #866, opened 2024-08-13
Handle the case where either query points or observations have unspec…

avullo:
…ified leading dimension.

**Related issue(s)/PRs:** 

## Summary

There might be situations where not only the query points, but also the observations might have an unspecified leading dim. This PR is addressing this case. It is also adding previously missing tests for the existing case, and also in combination for the one considered here.

## PR checklist


uri-granta (approved):
LGTM

Issue #304, opened 2021-07-19
`tf.debugging` & tests for `tf.function`


hstojic:
This is a follow up on discussion in PR #299 - at the moment we use `tf.function` for optimizing the model, but users might pass any other piece of code through it e.g. whole BO loop through it. Hence, it would be good to have some tests that (all of) our code can be used within `tf.function` without errors. Most likely we would have errors, since there is a lot of python code can cause errors within `tf.function`. This is not a high priority probably, but it could get out of hand with time so perhaps at least we could start testing any new piece of code that we want to merge into `develop`?

Related to this, we should use `tf.debugging` everywhere in the code instead of if-else statements with some `raise` statement that we dominantly use at the moment.

joelberkeley:
I checked and, contrary to what I said before, I can't find any evidence that standard error handling is a problem in `tf.function`. In particular, this test passes
```python
def test_raise_incompatible_with_tf_function() -> None:
    def f(a: tf.Tensor) -> tf.Tensor:
        if a < 0:
            raise ValueError

        return a

    f_compiled = tf.function(f)

    f_compiled(1)

    with pytest.raises(ValueError):
        f_compiled(-1)
```

joelberkeley:
the situation is actually much worse than I realised re `raise` and `tf.function`. The following test passes
```python
def test_raise_is_incompatible_with_tf_function() -> None:
    @tf.function
    def f(a: tf.Tensor) -> tf.Tensor:
        if a <= tf.constant(0):
            raise ValueError

        return a

    with pytest.raises(ValueError):
        f(tf.constant(1))
```
which suggests that the `ValueError` is raised even though the value shouldn't execute the error branch (perhaps due to tracing covering every branch). This suggests that any function that has any standard python exceptions in it cannot work for any arguments with `tf.function`.

hstojic:
thanks for making this concrete, this seems quite indicative about exceptions

uri-granta:
~~(connected` to #722)~~

uri-granta:
Bumping this ticket in light of https://github.com/secondmind-labs/trieste/pull/880.

PR #880, opened 2024-10-11
Make one hot encoder serialisable

uri-granta:
**Related issue(s)/PRs:** 

## Summary

Quick fix to allow one hot encoders to be serialised. Further investigation will happen as part of #304.

(Note that second change to tf.debugging in encoder doesn't seem to be necessary for the serialisation to work, but is included for consistency.)

## PR checklist


qiq208 (approved):
awesome, thanks!!

avullo:
Awesome 2. Out of curiosity, why tf debugging work?

uri-granta:
The problem is that raise gets called when you compile the function, not when you run it. Meanwhile `tf.debugging.assert` is similar to `tf.print` in that it's intended for runtime. That said, I still haven't found a clear description of the best practice re exceptions (will investigate as part of #304).

PR #885, opened 2024-11-05
Use different seeds for subspace sampling

uri-granta:
**Related issue(s)/PRs:** 

## Summary

When sampled with a specific seed, product spaces currently pass that seed to each of their subspaces. This means that a space A * A for example would return identical samples from both subspaces. This fixes that by offsetting the seed passed to each subspace.

## PR checklist


hstojic (approved):
looks good

PR #890, opened 2025-01-17
using tf.data for fit method in DeepEnsemble model

hstojic:
this simple change should improve memory handling, should be better optimized for GPUs, and generally gives more control to the user over preparing data for training

### DIFF
--- a/trieste/models/keras/models.py
+++ b/trieste/models/keras/models.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import copy
 import re
-from typing import Any, Dict, Mapping, Optional
+from typing import Any, Dict, Mapping, Optional, Union
 
 import dill
 import tensorflow as tf
@@ -379,24 +379,60 @@ class DeepEnsemble(
         return
 
     def prepare_tf_data(
-        self, x: Dict[str, TensorType], y: Dict[str, TensorType], batch_size: int, num_points: int
-    ) -> tf.data.Dataset:
+        self,
+        x: dict[str, TensorType],
+        y: dict[str, TensorType],
+        batch_size: int,
+        num_points: int,
+        validation_split: float = 0.0,
+    ) -> Union[tf.data.Dataset, tuple[tf.data.Dataset, tf.data.Dataset]]:
         """
         Prepare data for optimization as a `tf.data.Dataset`. This method allows user a more control
         over the data pipeline, e.g. shuffling, batching, prefetching, repeating,etc.
 
-        :param x: The input data.
-        :param y: The output data.
-        :param batch_size: The batch size.
-        :param num_points: The number of points.
-        :return: The prepared data.
+        :param x: Dictionary of input tensors
+        :param y: Dictionary of output tensors
+        :param batch_size: Batch size for the dataset
+        :param num_points: Number of data points
+        :param validation_split: Float between 0 and 1, fraction of data to use for validation
+        :return: If validation_split is 0, returns a single dataset for training.
+                If validation_split > 0, returns a tuple of (training_dataset, validation_dataset)
         """
-        return (
-            tf.data.Dataset.from_tensor_slices((x, y))
-            .prefetch(tf.data.experimental.AUTOTUNE)
-            .shuffle(num_points)
-            .batch(batch_size, drop_remainder=True)
-        )
+        if not 0.0 <= validation_split < 1.0:
+            raise ValueError("validation_split must be between 0 and 1")
+
+        dataset = tf.data.Dataset.from_tensor_slices((x, y))
+
+        if validation_split > 0:
+            # Calculate split sizes
+            val_size = int(num_points * validation_split)
+            train_size = num_points - val_size
+
+            # Shuffle before splitting to ensure randomness
+            dataset = dataset.shuffle(num_points, reshuffle_each_iteration=True)
+
+            # Split into train and validation
+            train_dataset = dataset.take(train_size)
+            val_dataset = dataset.skip(train_size)
+
+            # Prepare training dataset
+            train_dataset = (
+                train_dataset.prefetch(tf.data.AUTOTUNE)
+                .shuffle(train_size, reshuffle_each_iteration=True)
+                .batch(batch_size, drop_remainder=True)
+            )
+
+            # Prepare validation dataset
+            val_dataset = val_dataset.prefetch(tf.data.AUTOTUNE)
+
+            return train_dataset, val_dataset
+        else:
+            # Original behavior when no validation split is requested
+            return (
+                dataset.prefetch(tf.data.AUTOTUNE)
+                .shuffle(train_size, reshuffle_each_iteration=True)
+                .batch(batch_size, drop_remainder=True)
+            )
 
     def optimize_encoded(self, dataset: Dataset) -> tf_keras.callbacks.History:
         """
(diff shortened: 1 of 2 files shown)


trieste/models/keras/models.py
```diff
@@ -377,6 +378,62 @@ def update_encoded(self, dataset: Dataset) -> None:
...
         return
 
+    def prepare_tf_data(
+        self,
+        x: dict[str, TensorType],
+        y: dict[str, TensorType],
+        batch_size: int,
+        num_points: int,
+        validation_split: float = 0.0,
+    ) -> Union[tf.data.Dataset, tuple[tf.data.Dataset, tf.data.Dataset]]:
```
uri-granta, line 388: might be nicer to always return a tuple?

```suggestion
    ) -> tuple[tf.data.Dataset, Optional[tf.data.Dataset]]]:
```

trieste/models/keras/models.py
```diff
@@ -377,6 +378,62 @@ def update_encoded(self, dataset: Dataset) -> None:
...
+        :param x: Dictionary of input tensors
+        :param y: Dictionary of output tensors
+        :param batch_size: Batch size for the dataset
+        :param num_points: Number of data points
+        :param validation_split: Float between 0 and 1, fraction of data to use for validation
+        :return: If validation_split is 0, returns a single dataset for training.
+                If validation_split > 0, returns a tuple of (training_dataset, validation_dataset)
+        """
+        if not 0.0 <= validation_split < 1.0:
+            raise ValueError("validation_split must be between 0 and 1")
```
uri-granta, line 402:
```suggestion
            raise ValueError(f"validation_split must be between 0 and 1: got {validation_split}")
```

trieste/models/keras/models.py
```diff
@@ -377,6 +378,62 @@ def update_encoded(self, dataset: Dataset) -> None:
...
+
+        dataset = tf.data.Dataset.from_tensor_slices((x, y))
+
+        if validation_split > 0:
+            # Calculate split sizes
+            val_size = int(num_points * validation_split)
```
uri-granta, line 408:
```suggestion
            val_size = round(num_points * validation_split)
```

trieste/models/keras/models.py
```diff
@@ -393,20 +450,38 @@ def optimize_encoded(self, dataset: Dataset) -> tf_keras.callbacks.History:
...
+        if "epochs" in fit_args_copy:
+            fit_args_copy["epochs"] = fit_args_copy["epochs"] + self._absolute_epochs
 
         x, y = self.prepare_dataset(dataset)
+
+        validation_split = fit_args_copy.pop("validation_split", 0.0)
+        tf_data = self.prepare_tf_data(
+            x,
+            y,
+            batch_size=fit_args_copy["batch_size"],
```
uri-granta, line 467: "batch_size" isn't guaranteed to exist for a user-supplied fit_args

```suggestion
            batch_size=fit_args_copy.get("batch_size"),
```

trieste/models/keras/models.py
```diff
@@ -393,20 +450,38 @@ def optimize_encoded(self, dataset: Dataset) -> tf_keras.callbacks.History:
...
-
-        if "epochs" in fit_args:
-            fit_args["epochs"] = fit_args["epochs"] + self._absolute_epochs
+        if "epochs" in fit_args_copy:
+            fit_args_copy["epochs"] = fit_args_copy["epochs"] + self._absolute_epochs
 
         x, y = self.prepare_dataset(dataset)
+
+        validation_split = fit_args_copy.pop("validation_split", 0.0)
+        tf_data = self.prepare_tf_data(
```
uri-granta, line 464: (if you change the return type above as suggested)

```suggestion
        train_dataset, val_dataset = self.prepare_tf_data(
```

trieste/models/keras/models.py
```diff
@@ -393,20 +450,38 @@ def optimize_encoded(self, dataset: Dataset) -> tf_keras.callbacks.History:
...
+            y,
+            batch_size=fit_args_copy["batch_size"],
+            num_points=dataset.observations.shape[0],
+            validation_split=validation_split,
+        )
+        fit_args_copy["batch_size"] = None  # batching is done in prepare_tf_data
+
+        if validation_split > 0:
+            train_dataset, val_dataset = tf_data
+            fit_args_copy["validation_data"] = val_dataset
```
uri-granta, line 475: should we maybe raise an exception if "train_dataset, val_dataset" is already present in the fit_args?

trieste/models/keras/models.py
```diff
@@ -393,20 +450,38 @@ def optimize_encoded(self, dataset: Dataset) -> tf_keras.callbacks.History:
...
+            history = self.model.fit(
+                train_dataset, **fit_args_copy, initial_epoch=self._absolute_epochs
+            )
+        else:
+            history = self.model.fit(tf_data, **fit_args_copy, initial_epoch=self._absolute_epochs)
```
uri-granta, line 480:
```suggestion
        history = self.model.fit(tf_data, **fit_args_copy, initial_epoch=self._absolute_epochs)
```

trieste/models/keras/models.py
```diff
@@ -377,6 +378,62 @@ def update_encoded(self, dataset: Dataset) -> None:
...
+
+            # Prepare validation dataset
+            val_dataset = val_dataset.prefetch(tf.data.AUTOTUNE)
+
+            return train_dataset, val_dataset
+        else:
+            # Original behavior when no validation split is requested
+            return (
+                dataset.prefetch(tf.data.AUTOTUNE)
+                .shuffle(train_size, reshuffle_each_iteration=True)
```
uri-granta, line 433: (I think?)

```suggestion
                .shuffle(num_points, reshuffle_each_iteration=True)
```

trieste/models/keras/models.py
```diff
@@ -377,6 +378,62 @@ def update_encoded(self, dataset: Dataset) -> None:
...
+                .batch(batch_size, drop_remainder=True)
+            )
+
+            # Prepare validation dataset
+            val_dataset = val_dataset.prefetch(tf.data.AUTOTUNE)
+
+            return train_dataset, val_dataset
+        else:
+            # Original behavior when no validation split is requested
```
uri-granta, line 430: Q: is this really the same as the original behaviour?

uri-granta (commented):
Various comments/comments. Happy to review again (including the tests) once the tests are passing.

pio-neil:
I do have a small concern about the use of `tf.Dataset.shuffle`. When I did some testing with this before, the shuffle buffer (which is `tf.Dataset`'s internal method of shuffling data) used around 18GB of extra memory. This was with a dataset with 30 million rows, with 15 inputs and one output, and a batch size of 1000. The shuffle buffer also has an impact on speed, but I suspect this is relatively minor compared to the model training time.

This might not be such a problem with smaller datasets. So perhaps it would be a good idea to do some benchmarking? 

However, it's also not clear to me why we're introducing shuffling by default here, when AFAICT it wasn't there before? This seems like a change of behaviour. Do we expect this to improve model accuracy? It may be better to let the user of Trieste control this, rather than making it the default behaviour?

trieste/models/keras/models.py
hstojic: ah, well spotted, I was remembering BatchOptimizer...

hstojic:
> I do have a small concern about the use of `tf.Dataset.shuffle`. When I did some testing with this before, the shuffle buffer (which is `tf.Dataset`'s internal method of shuffling data) used around 18GB of extra memory. This was with a dataset with 30 million rows, with 15 inputs and one output, and a batch size of 1000. The shuffle buffer also has an impact on speed, but I suspect this is relatively minor compared to the model training time.
> 
> This might not be such a problem with smaller datasets. So perhaps it would be a good idea to do some benchmarking?

I have already done some testing, on much smaller data than that, haven't seen any adverse effects.
Users that handle bigger datasets now can handle data preparation themselves by subclassing the model and overwriting the new method, if there are any issues with speed/memory.

> 
> However, it's also not clear to me why we're introducing shuffling by default here, when AFAICT it wasn't there before? This seems like a change of behaviour. Do we expect this to improve model accuracy? It may be better to let the user of Trieste control this, rather than making it the default behaviour?

`shuffle` argument is by default True in Keras' `fit` method, hence why its default here, but its definitely needed - we are typically doing many epochs here and shuffling data at each epoch helps model training - so yes, should improve model accuracy. The user can always overwrite the new method, that was the purpose of extracting this data preparation bit into a method.

pio-neil:
> > However, it's also not clear to me why we're introducing shuffling by default here, when AFAICT it wasn't there before? This seems like a change of behaviour. Do we expect this to improve model accuracy? It may be better to let the user of Trieste control this, rather than making it the default behaviour?
> 
> `shuffle` argument is by default True in Keras' `fit` method, hence why its default here, but its definitely needed - we are typically doing many epochs here and shuffling data at each epoch helps model training - so yes, should improve model accuracy. The user can always overwrite the new method, that was the purpose of extracting this data preparation bit into a method.

Right, yes, I see. I was just wondering why we were doing things differently in terms of shuffling, but we aren't (since `fit` ignores the shuffle argument for datasets):

> shuffle: Boolean, whether to shuffle the training data before each epoch. This argument is ignored when x is a [keras.utils.PyDataset](https://keras.io/api/utils/python_utils#pydataset-class), [tf.data.Dataset](https://www.tensorflow.org/api_docs/python/tf/data/Dataset), torch.utils.data.DataLoader or Python generator function.

hstojic:
after some further thinking, this is currently unnecessary complication of the code, as for smallish datasets that we mainly deal with here current code is good enough

PR #891, opened 2025-01-19
Support copying posterior caches (and thereby float32 predictions from float64 models)

uri-granta:
**Related issue(s)/PRs:** 

## Summary

...

## PR checklist


tests/unit/models/gpflow/test_utils.py
```diff
@@ -330,3 +331,42 @@ def test_check_optimizer_raises_for_invalid_optimizer_wrapper_combination() -> N
...
+        posterior32 = to_float32(model._posterior)
+        model32 = type(model)(gpr32, posterior=posterior32)  # type: ignore[call-arg]
+
+        # check that the predictions match the original float64 model
+        f_mean, f_var = model.predict(x)
+        f_mean32, f_var32 = model32.predict(tf.cast(x, tf.float32))
+        assert f_mean32.dtype is tf.float32
+        npt.assert_allclose(f_mean, f_mean32, rtol=1e-6)
+        assert f_var32.dtype is tf.float32
+        npt.assert_allclose(f_var, f_var32, rtol=1e-6)
```
hstojic, line 372: my assumption is that accuracy will depend a lot on the problem, here we are testing on a simple function and its not surprising we get pretty much the same value - have you tested it on more complex problems of higher dimensionality?

uri-granta: Good call. Can you suggest a problem I can easily test it on?

hstojic: lets start with more complex functions that we have in our synthetic problems  - we have some where we can specify arbitrary dimensionality as well I think

PR #894, opened 2025-01-21
implemented predict_y and predict_noise

hstojic:
this is an important change, it will affect DE model quite a bit, so its important to test downstream consequences
- `predict` method now outputs estimation uncertainty, which is more appropriate
- DE now supports `predict_y` method which outputs combined uncertainty
- there is a new method `predict_noise` that gives mean and variance of the noise (i.e. aleatoric uncertainty), which comes handy in some situations

note that trajectory sampler behaviour will be affected as well, but only in the diversify mode


tests/unit/models/keras/test_models.py
```diff
@@ -370,6 +405,19 @@ def test_deep_ensemble_predict_ensemble_call_shape(
...
+def test_deep_ensemble_uncertainty_separation() -> None:
+    example_data = _get_example_data([100, 2])
+    model, _, _ = trieste_deep_ensemble_model(example_data, _ENSEMBLE_SIZE, False, False)
+
+    qp = example_data.query_points
+    _, epistemic_var = model.predict(qp)
+    aleatoric_var, _ = model.predict_noise(qp)
+    _, total_var = model.predict_y(qp)
+
+    assert tf.reduce_all(total_var >= epistemic_var)
```
avullo:
```suggestion
    assert tf.reduce_all((total_var >= epistemic_var) & (total_var >= aleatoric_var))
```

trieste/models/keras/sampler.py
```diff
@@ -114,7 +114,8 @@ class deep_ensemble_trajectory(TrajectoryFunctionClass):
     Option `diversify` can be used to increase the diversity in case of optimizing very large
     batches of trajectories. We use quantiles from the approximate Gaussian distribution of
     the ensemble as trajectories, with randomly chosen quantiles approximating a trajectory and
-    using a reparametrisation trick to speed up computation. Note that quantiles are not true
+    using a reparametrisation trick to speed up computation. Only epistemic uncertainty is taken
+    into account in sampling. Note that quantiles are not true trajectories, so this will likely
     trajectories, so this will likely have some performance costs.
```
avullo:
```suggestion
    using a reparametrisation trick to speed up computation. Only epistemic uncertainty is taken
    into account in sampling. Note that quantiles are not true trajectories, so this will likely
    have some performance costs.
```

trieste/models/keras/models.py
```diff
@@ -252,7 +258,42 @@ def ensemble_distributions(self, query_points: TensorType) -> tuple[tfd.Distribu
 
     def predict_encoded(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
```
ChrisMorter, line 260: Can this be changed to not drop information about the aleatoric uncertainty? Currently Secondmind's  calibration sampling logic uses `IndependentReparametrizationSampler`, which generates samples based upon predictions made using `predict`, which in turn calls through to `predict_encoded`. 

For this to be suitable for query point generation in active learning, we need to know both the epistemic and aleatoric uncertainty. Hypothetically, we could not use `IndependentReparametrizationSampler`, and instead write a new sampler which uses `predict_ensemble_encoded` rather than `predict`, because that does return information about the aleatoric uncertainty. However, this would be a significant change in the calibration product.

An implementation could look something like:

```
        ensemble_means, ensemble_vars = self.predict_ensemble_encoded(query_points)
        predicted_means = tf.math.reduce_mean(ensemble_means, axis=-3)
        epistemic_variance = tf.math.reduce_variance(ensemble_means, axis=-3)
        aleatoric_variance = tf.math.reduce_mean(ensemble_vars, axis=-3)
        aleatoric_variance_var = tf.math.reduce_variance(ensemble_vars, axis=-3)

        means = tf.concat([predicted_means, aleatoric_variance], axis=-1)
        vars = tf.concat([epistemic_variance, aleatoric_variance_var], axis=-1)
        
        return means, vars
```

ChrisMorter: Don't let this issue block this PR from being merged though, I appreciate that's likely a difficult change to make since it requires updating all usages, and would be a breaking change for anyone using `DeepEnsemble.predict`. We can discuss further how to integrate DeepEnsemble for active learning in a future PR.

ChrisMorter (approved):

trieste/models/keras/models.py
```diff
@@ -277,29 +318,38 @@ def predict_encoded(self, query_points: TensorType) -> tuple[TensorType, TensorT
...
+        """
+        _, ensemble_vars = self.predict_ensemble_encoded(query_points)
+        aleatoric_variance_mean = tf.math.reduce_mean(ensemble_vars, axis=-3)
+        aleatoric_variance_variance = tf.math.reduce_variance(ensemble_vars, axis=-3)
+
+        return aleatoric_variance_mean, aleatoric_variance_variance
 
-        return unflatten(predicted_means), unflatten(predicted_vars)
+    def predict_noise(self, query_points: TensorType) -> tuple[TensorType, TensorType]:
+        return self.predict_noise_encoded(self.encode(query_points))
```
uri-granta, line 360: This should have a docstring too, as it's the external API. And also a check_shapes, probably:

    @check_shapes(
        "query_points: [broadcast batch..., D]",
        "return[0]: [batch..., E...]",
        "return[1]: [batch..., E...]",
    )


Also, does this method generalise to other models beyond `DeepEnsemble`? If so, we could preemptively define `SupportsPredictNoise` and `EncodedSupportsPredictNoise` protocols, just like `[Encoded]SupportsPredictY`.

trieste/models/keras/models.py
```diff
@@ -277,29 +318,38 @@ def predict_encoded(self, query_points: TensorType) -> tuple[TensorType, TensorT
         :return: The predicted mean and variance of the observations at the specified
             ``query_points``.
```
uri-granta, line 320:
```suggestion
            ``query_points``, including noise contributions.
```

uri-granta (approved):
LGTM! Just one comment. Also, as discussed, we should avoid changing predict to include the aleatoric uncertainty for now, though there may be a number of different ways we could do this in the future without breaking existing usages.

trieste/models/keras/models.py
hstojic: I spoke to @uri-granta about this, and its not clear that this usage fits well the interfaces - we decided to proceed as is for now and think through how to deal with this kind of use case

trieste/models/keras/models.py
hstojic: other models in trieste don't have it, so probably not needed at the moment

hstojic: added doc and checkshapes otherwise

avullo (approved):
lgtm!

PR #897, opened 2025-01-23
DeepEnsemble: training networks in parallel

hstojic:
attempt at fixing DE issue with training networks serially

the main observation this bug was based on was reduction in training time when decreasing the size of the ensemble


hstojic:
not needed, we used a wrong test

PR #898, opened 2025-01-23
adding a parallelisation test for DeepEnsemble

hstojic:
we explicitly test that increasing number of networks in the ensemble, once accounted for difference in number of parameters, will lead to similar training time

### DIFF
--- a/tests/unit/models/keras/test_models.py
+++ b/tests/unit/models/keras/test_models.py
@@ -780,6 +780,7 @@ def test_deep_ensemble_log(
     assert mocked_summary_histogram.call_count == num_histogram
 
 
+@pytest.mark.slow
 @random_seed
 def test_deep_ensemble_parallel_training_performance() -> None:
     """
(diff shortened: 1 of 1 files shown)


uri-granta (commented):
Good idea for a test! Couple of questions:

1. how long does it to take to run? If it's slow you could mark it with `@pytest.mark.slow`.
2. why does the smaller ensemble have more units per layer? wouldn't that increase the run time and make the comparison less fair?
3. just to be certain, have manually confirmed that the training is happening in parallel (e.g. by setting a breakpoint? or disabling parallelisation and seeing a change?)

hstojic:
> Good idea for a test! Couple of questions:
> 
> 1. how long does it to take to run? If it's slow you could mark it with `@pytest.mark.slow`.

about a minute on CPU, probably similar on GPU - I'll do that

> 2. why does the smaller ensemble have more units per layer? wouldn't that increase the run time and make the comparison less fair?

that's specifically to account for the number of parameters in each model, so this is the fair comparison

> 3. just to be certain, have manually confirmed that the training is happening in parallel (e.g. by setting a breakpoint? or disabling parallelisation and seeing a change?)

its not possible to do that I think, disabling parllelisation would require a completely different DE architecture

hstojic:
@uri-granta can you review again please?

uri-granta (approved):
:+1:

PR #899, opened 2025-02-05
more flexible prepare_dataset method in DE model

hstojic:
this is to make it easier to create validation data to pass it to the fit method, if one would overwrite the optimize method


trieste/data.py
```diff
@@ -202,3 +202,47 @@ def add_fidelity_column(query_points: TensorType, fidelity: int) -> TensorType:
...
+    if len(dataset) == 0:
+        return dataset, dataset
+
+    # Set random seed if provided
+    if seed is not None:
+        tf.random.set_seed(seed)
+
+    # Generate random permutation of indices
+    num_points = len(dataset)
+    indices = tf.random.shuffle(tf.range(num_points))
```
ChrisMorter: It's probably best to pass the seed to `tf.random.shuffle` directly, rather than setting the global random seed because that could affect other random ops.

ChrisMorter (approved):
LGTM, just one comment regarding setting the seed for splitting the data.

trieste/data.py
```diff
@@ -202,3 +202,46 @@ def add_fidelity_column(query_points: TensorType, fidelity: int) -> TensorType:
...
+        points.
+    :raise ValueError: If proportion is not between 0 and 1
+    """
+    if not 0 <= proportion <= 1:
+        raise ValueError(f"proportion must be between 0 and 1, got {proportion}")
+
+    if len(dataset) == 0:
+        return dataset, dataset
+
+    if seed is not None:  # ensure reproducibility
```
ChrisMorter, line 225: I think this can be removed now that we're passing the seed into `tf.random.shuffle`?

PR #900, opened 2025-02-05
fix jitter bug in Deep ensemble trajectory sampler

hstojic:
there is a subtle bug in the sampler when adding jitter to the variances (only when diversify is true) - because the constant is too large it introduces discontinuities in trajectories


uri-granta (approved):
LGTM!

PR #904, opened 2025-02-25
Liyuan/add logei

liyuan9988:
## Summary
Add log version of EI and AEI aquisition function

Motivated by [Unexpected Improvements to Expected Improvement for Bayesian Optimization](https://arxiv.org/abs/2310.20708), the log version of EI and AEI are added

## PR checklist


tests/util/misc.py
```diff
@@ -47,6 +48,7 @@
 TF_DEBUGGING_ERROR_TYPES: Final[tuple[type[Exception], ...]] = (
     ValueError,
     tf.errors.InvalidArgumentError,
+    StagingError,
```
hstojic: what is this bit about?

trieste/acquisition/function/function.py
```diff
@@ -223,6 +224,167 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    """Accurately computes log(phi(u) + u * Phi(u)) in a differentiable manner for u in
+    [-10^100, 10^100] in double precision, and [-10^20, 10^20] in single precision.
+    The implementation is inspired by the BoTorch implementation
+    (https://github.com/pytorch/botorch/blob/main/botorch/acquisition/analytic.py)
+    """
+    if not (u.dtype == tf.float32 or u.dtype == tf.float64):
+        raise TypeError(
+            f"LogExpectedImprovement only supports torch.float32 and torch.float64 "
+            f"dtypes, but received {u.dtype=}."
+        )
```
hstojic: I think this may cause issues with compiling the __call__ function above

hstojic: torch.float32 and torch.float64 :)

trieste/acquisition/function/function.py
```diff
@@ -103,12 +104,14 @@ class ExpectedImprovement(SingleModelAcquisitionBuilder[ProbabilisticModel]):
     the maximum of the posterior mean at all observed points.
     """
 
-    def __init__(self, search_space: Optional[SearchSpace] = None):
+    def __init__(self, search_space: Optional[SearchSpace] = None, log_transform: bool = False):
```
hstojic: one principle that we are trying to follow in trieste is to have sensible defaults, to not expect that users are experts - here to me it seems the default should be True and if someone really needs the original one they can get it

trieste/acquisition/function/function.py
```diff
@@ -223,12 +232,132 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
 class AugmentedExpectedImprovement(SingleModelAcquisitionBuilder[SupportsGetObservationNoise]):
     """
     Builder for the augmented expected improvement function for optimization single-objective
     optimization problems with high levels of observation noise.
     """
 
+    def __init__(self, log_transform: bool = False):
+        """
+        :param log_transform: If `True`, use logEI + log of augmented term.
```
hstojic: default to True, same as the comment above?

trieste/acquisition/function/function.py
```diff
@@ -325,6 +463,63 @@ def __call__(self, x: TensorType) -> TensorType:
...
+class log_augmented_expected_improvement(AcquisitionFunctionClass):
+    def __init__(self, model: SupportsGetObservationNoise, eta: TensorType):
+        r"""
+        Return the logarithm of Augmented Expected Improvement (AEI) acquisition function
+        for single-objective global optimization under homoscedastic observation noise.
+        Improvement is with respect to the current "best" observation ``eta``, where an
+        improvement moves towards the objective function's minimum and the expectation is
+        calculated with respect to the ``model`` posterior.
+        This shares the same optima as original AEI but expected to be easier to optimize.
+        See
```
hstojic: see?

trieste/acquisition/function/function.py
```diff
@@ -325,6 +463,63 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+        :param model: The model of the objective function.
+        :param eta: The "best" observation.
+        :return: The expected improvement function. This function will raise
+            :exc:`ValueError` or :exc:`~tf.errors.InvalidArgumentError` if used with a batch size
+            greater than one or a model without homoscedastic observation noise.
```
hstojic: homoskedastic?

trieste/acquisition/function/function.py
```diff
@@ -325,6 +463,63 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            [(x, [..., 1, None])],
+            message="This acquisition function only supports batch sizes of one.",
+        )
+        eps = (
+            tf.constant(-1e12, dtype=tf.float64)
+            if x.dtype == tf.float64
+            else tf.constant(-1e6, dtype=tf.float32)
+        )
+        mean, variance = self._model.predict(tf.squeeze(x, -2))
+        sigma = tf.sqrt(tf.clip_by_value(variance, clip_value_min=eps, clip_value_max=x.dtype.max))
```
hstojic: we have `ensure_positive` function now that we have started using elsewhere to make sure variance is positive - can you re-use that?

trieste/acquisition/function/function.py
```diff
@@ -325,6 +463,63 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        u = (self._eta - mean) / sigma
+        logei = _log_ei_helper(u) + tf.math.log(sigma)
+
+        """
+        It can be problematic to take the log of original augmentaiton term in AEI
+        since variance + noise_variance  can be numerically equal to noise_variance.
+        Hence, we use numerically stable version
+        """
```
hstojic: perhaps add a docstring for the __call__ and put this there?

trieste/acquisition/function/function.py
```diff
@@ -325,6 +463,63 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+        log_augmentation = tf.math.log(variance) - tf.math.log(
```
hstojic: wouldn't you need to use a variance where you made sure all values are positive here as well?

trieste/acquisition/function/function.py
```diff
@@ -325,6 +463,63 @@ def __call__(self, x: TensorType) -> TensorType:
...
+            tf.math.sqrt(variance + self._noise_variance)
+        )
+        log_augmentation = log_augmentation - tf.math.log(
+            tf.math.sqrt(variance + self._noise_variance) + tf.math.sqrt(self._noise_variance)
```
hstojic: you could perhaps eliminate duplication with the lines above: `tf.math.sqrt(variance + self._noise_variance)`

trieste/acquisition/function/function.py
```diff
@@ -223,12 +232,132 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        "return: [N..., L]",
+    )
+    def __call__(self, x: TensorType) -> TensorType:
+        eps = (
+            tf.constant(-1e12, dtype=tf.float64)
+            if x.dtype == tf.float64
+            else tf.constant(-1e6, dtype=tf.float32)
+        )
+        mean, variance = self._model.predict(tf.squeeze(x, -2))
+        sigma = tf.sqrt(tf.clip_by_value(variance, clip_value_min=eps, clip_value_max=x.dtype.max))
```
hstojic: see below my comment and perhaps re-using `ensure_positive` used elsehwere for this purpose?

trieste/acquisition/function/function.py
```diff
@@ -223,12 +232,132 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        calculated with respect to the ``model`` posterior. LogEI mitigates the issue of
+        Expected Improvement (EI) that it is numerically zero in many regions.
+        Although LogEI has (approximately) same optima as EI, it is much easier to optimze,
+        and empirically shows significantly improved optimization performance.
+        See :cite: `ament2023unexpected` for details.
+
+        :param model: The model of the objective function.
+        :param eta: The "best" observation.
+        :return: The expected improvement function. This function will raise
+            :exc:`ValueError` or :exc:`~tf.errors.InvalidArgumentError` if used with a batch size
```
hstojic: is check_shapes actually raising this at runtime?
I vaguely recall that it doesn't do that...

trieste/acquisition/function/function.py
```diff
@@ -223,12 +232,132 @@ def __call__(self, x: TensorType) -> TensorType:
         return (self._eta - mean) * normal.cdf(self._eta) + variance * normal.prob(self._eta)
 
 
+class log_expected_improvement(AcquisitionFunctionClass):
```
hstojic: you can inherit from `expected_improvement`, no?
then there is no need for init and update methods...

hstojic: docstring can be moved to class docstring

trieste/acquisition/function/function.py
```diff
@@ -223,12 +232,132 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    # To this end, a second branch is necessary, depending on whether or not u is
+    # smaller than approximately the negative inverse square root of the machine
+    # precision. Below this point, numerical issues in computing log(1 - exp(w)) occur
+    # as w approaches zero from below, even though the relative contribution to log_ei
+    # vanishes in machine precision at that point.
+    neg_inv_sqrt_eps = (
+        tf.constant(-1e6, dtype=tf.float64)
+        if u.dtype == tf.float64
+        else tf.constant(-1e3, dtype=tf.float32)
+    )
```
hstojic, line 311: these constants seem a bit too large?

trieste/acquisition/function/function.py
```diff
@@ -325,6 +463,63 @@ def __call__(self, x: TensorType) -> TensorType:
         return ei * augmentation
 
 
+class log_augmented_expected_improvement(AcquisitionFunctionClass):
```
hstojic: here you can inherit from `augmented_expected_improvement`...

hstojic (commented):
looks fine, I only have 2 bigger comments:
- add these new acquisition rules in the integration test in `tests/integration/test_bayesian_optimization.py`
- there are some follow up PRs that we would need: constrained EI, MCEI variants, hypervolume improvement

also, add something about this in one of the notebooks (e.g. `expected_improvement`?

trieste/acquisition/function/function.py
liyuan9988: Fixed in https://github.com/secondmind-labs/trieste/pull/904/commits/2f45eb8e8035fbfa7105f6f2acd9a470edb32906

trieste/acquisition/function/function.py
liyuan9988: Fixed in https://github.com/secondmind-labs/trieste/pull/904/commits/2f45eb8e8035fbfa7105f6f2acd9a470edb32906

trieste/acquisition/function/function.py
liyuan9988: No, it raises `StageError`, so I removed it in https://github.com/secondmind-labs/trieste/pull/904/commits/0cec4201066bbdf89577739a0723d9c21816304b and do shape check by `tf.debugging.assert_shapes`

tests/util/misc.py
liyuan9988: That was the artifact of using `check_shapes`. I have removed it in https://github.com/secondmind-labs/trieste/pull/904/commits/0cec4201066bbdf89577739a0723d9c21816304b

trieste/acquisition/function/function.py
liyuan9988: Fixed in https://github.com/secondmind-labs/trieste/pull/904/commits/dee5e41e7fbeae04efe9116bae586d309f8772a0

trieste/acquisition/function/function.py
liyuan9988: Fixed in https://github.com/secondmind-labs/trieste/pull/904/commits/dee5e41e7fbeae04efe9116bae586d309f8772a0

trieste/acquisition/function/function.py
liyuan9988: Agreed, but it is consistent with original implementation: https://github.com/pytorch/botorch/blob/01aa45e2c1287b53b9ce15f683e27be8a4ad78c5/botorch/acquisition/analytic.py#L1000

trieste/acquisition/function/function.py
liyuan9988: "c" seems to be more popular?

https://en.wikipedia.org/wiki/Homoscedasticity_and_heteroscedasticity

trieste/acquisition/function/function.py
liyuan9988: Fixed!

trieste/acquisition/function/function.py
liyuan9988: Fixed in https://github.com/secondmind-labs/trieste/pull/904/commits/2f45eb8e8035fbfa7105f6f2acd9a470edb32906

trieste/acquisition/function/function.py
liyuan9988: Fixed in https://github.com/secondmind-labs/trieste/pull/904/commits/2f45eb8e8035fbfa7105f6f2acd9a470edb32906

trieste/acquisition/function/function.py
liyuan9988: Fixed in https://github.com/secondmind-labs/trieste/pull/904/commits/0f219f2d5177c96c82d2efe5d5ac784a52e26fbe

trieste/acquisition/function/function.py
liyuan9988: Fixed in https://github.com/secondmind-labs/trieste/pull/904/commits/0f219f2d5177c96c82d2efe5d5ac784a52e26fbe to create logEI class

trieste/acquisition/function/function.py
liyuan9988: Same above

liyuan9988:
@hstojic Thank you for the reivew, I addressed the all comments and add logEI, logAugEI to the integration test.
Could you review them again?

I will make seperate pull-req for other aquisition functions and notebook, once this is merged.

trieste/acquisition/function/function.py
```diff
@@ -104,14 +105,14 @@ class ExpectedImprovement(SingleModelAcquisitionBuilder[ProbabilisticModel]):
     the maximum of the posterior mean at all observed points.
     """
 
-    def __init__(self, search_space: Optional[SearchSpace] = None, log_transform: bool = False):
+    def __init__(self, search_space: Optional[SearchSpace] = None):
         """
         :param search_space: The global search space over which the optimisation is defined. This is
             only used to determine explicit constraints.
         :param log_transform: If `True`,  use logEI proposed by :cite:`ament2023unexpected`.
```
hstojic: leftover arg

trieste/acquisition/function/function.py
```diff
@@ -232,8 +227,28 @@ def __call__(self, x: TensorType) -> TensorType:
...
+        """
+        :param search_space: The global search space over which the optimisation is defined. This is
+            only used to determine explicit constraints.
+        """
+        self._search_space = search_space
+        self._acq_function_cls = log_expected_improvement
+
+    def __repr__(self) -> str:
+        """"""
+        return f"LogExpectedImprovement({self._search_space!r})"
```
hstojic: you can modify the repr to return `self.__class__.__name__` (I think its that, perhaps smth similar) in `ExpectedImprovement` and then you don't need to overwrite it this way

trieste/acquisition/function/function.py
```diff
@@ -104,14 +105,14 @@ class ExpectedImprovement(SingleModelAcquisitionBuilder[ProbabilisticModel]):
...
-    def __init__(self, search_space: Optional[SearchSpace] = None, log_transform: bool = False):
+    def __init__(self, search_space: Optional[SearchSpace] = None):
         """
         :param search_space: The global search space over which the optimisation is defined. This is
             only used to determine explicit constraints.
         :param log_transform: If `True`,  use logEI proposed by :cite:`ament2023unexpected`.
         """
         self._search_space = search_space
-        self._log_transform = log_transform
+        self._acq_function_cls = expected_improvement
```
hstojic: I think a typical way to do this would be to create a private method, smth like this:
```
def _prepare_acq_function(model, eta):
    return expected_improvement(model, eta)
```

which you would then overwrite, overwriting init is a bit unusual way to do this  I think...

trieste/acquisition/function/function.py
```diff
@@ -232,8 +227,28 @@ def __call__(self, x: TensorType) -> TensorType:
...
+
+    def __repr__(self) -> str:
+        """"""
+        return f"LogExpectedImprovement({self._search_space!r})"
+
+
+class log_expected_improvement(expected_improvement):
+
+    @tf.function
+    def __call__(self, x: TensorType) -> TensorType:
```
hstojic, line 269: docsting below in the call should go to the class docstring I think, thats the actual builder this method is just evaluating it

hstojic: actually, it should probably go to LogExpectedImprovement, that's the one users will actually use and therefore read that doc

trieste/acquisition/function/function.py
```diff
@@ -232,8 +227,28 @@ def __call__(self, x: TensorType) -> TensorType:
         return (self._eta - mean) * normal.cdf(self._eta) + variance * normal.prob(self._eta)
 
 
-class log_expected_improvement(AcquisitionFunctionClass):
-    def __init__(self, model: ProbabilisticModel, eta: TensorType):
+class LogExpectedImprovement(ExpectedImprovement):
+    """
+    Builder for the log expected improvement function.
```
hstojic: needs some more explanation what is this, reference, perhaps also why to use instead of the original, as a guidence to users

trieste/acquisition/function/function.py
```diff
@@ -463,59 +449,53 @@ def __call__(self, x: TensorType) -> TensorType:
...
+    Builder for the augmented expected improvement function for optimization single-objective
+    optimization problems with high levels of observation noise.
+    """
+
+    def __init__(self) -> None:
+        self._acq_function_cls = log_augmented_expected_improvement
+
+    def __repr__(self) -> str:
+        """"""
+        return "LogAugmentedExpectedImprovement()"
```
hstojic: same comments as above

trieste/acquisition/function/function.py
```diff
@@ -463,59 +449,53 @@ def __call__(self, x: TensorType) -> TensorType:
...
-        eps = (
-            tf.constant(-1e12, dtype=tf.float64)
-            if x.dtype == tf.float64
-            else tf.constant(-1e6, dtype=tf.float32)
-        )
+
         mean, variance = self._model.predict(tf.squeeze(x, -2))
-        sigma = tf.sqrt(tf.clip_by_value(variance, clip_value_min=eps, clip_value_max=x.dtype.max))
+        variance = ensure_positive(variance)
+        sigma = tf.sqrt(ensure_positive(variance))
```
hstojic: why do you need another pass through `ensure_positive`?

hstojic (commented):
left some minor comments to adress

trieste/acquisition/function/function.py
hstojic: ok I guess

hstojic (approved):
looks good to go now!

PR #905, opened 2025-02-26
Repeat dataset when steps_per_epoch is used with DeepEnsemble

ChrisMorter:
**Related issue(s)/PRs:** 

## Summary

...

This PR repeats the dataset if necessary if `steps_per_epoch` is used with DeepEnsemble. Previously, if the dataset didn't contain at least `epochs` * `steps_per_epoch` many rows, then training would finish early after every row was seen, albeit with a warning message that could be easily missed. By setting `steps_per_epoch` and `epochs`, users likely intend to execute that precise number of steps and epochs, so it's preferable to repeat the dataset as needed to fulfil this expectation.

The current behaviour is particularly problematic for small datasets, as the model cannot train to convergence by seeing each row only once.

## PR checklist

### DIFF
--- a/tests/unit/models/keras/test_models.py
+++ b/tests/unit/models/keras/test_models.py
@@ -25,6 +25,7 @@ import numpy.testing as npt
 import pytest
 import tensorflow as tf
 import tensorflow_probability as tfp
+from _pytest.logging import LogCaptureFixture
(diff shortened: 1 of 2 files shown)


hstojic:
@ChrisMorter we can close this one now, right?

PR #907, opened 2025-03-05
Pass TF dataset to DeepEnsemble model.fit

ChrisMorter:
**Related issue(s)/PRs:** 

[using tf.data for fit method in DeepEnsemble model](https://github.com/secondmind-labs/trieste/pull/890/files)

## Summary

This PR enables passing a `tf.data.Dataset` to Keras's `model.fit` in `DeepEnsemble`. This allows us to support training for a specified number of steps (rather than a number of epochs) for small datasets, which might require data repetition to meet the desired step count.

A limitation with Keras's `model.fit` is its incompatibility with simultaneously accepting `tf.data.Dataset` inputs and a `validation_split` parameter. This means, that if we want to pass `tf.data.Dataset`s to `model.fit`, then we need to form the validation split in Trieste, before passing the training split to `model.fit`. It's possible to do this in Trieste because we have access to the raw `x`, and `y` tensors before creating the `tf.data.Dataset`.

This PR essentially does the same thing as this [PR](https://github.com/secondmind-labs/trieste/pull/890) from @hrvoje but limiting the amount of additional complexity introduced to Trieste due to having to split the dataset ourselves. This is achieved by using the same method that Keras's `model.fit` uses internally to form the validation split.

...

## PR checklist

### DIFF
--- a/tests/unit/models/keras/test_models.py
+++ b/tests/unit/models/keras/test_models.py
@@ -18,6 +18,7 @@ import operator
 import tempfile
 import unittest.mock
 from typing import Any, Optional
+from unittest.mock import MagicMock
 
 import numpy as np
 import numpy.testing as npt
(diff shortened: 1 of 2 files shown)


trieste/models/keras/models.py
```diff
@@ -480,9 +481,18 @@ def optimize_encoded(self, dataset: Dataset) -> tf_keras.callbacks.History:
             fit_args["epochs"] = fit_args["epochs"] + self._absolute_epochs
 
         x, y = self.prepare_dataset(dataset)
+
+        if "validation_split" in fit_args:
+            (x, y), (x_val, y_val) = train_validation_split(
```
ChrisMorter: `train_validation_split` is what keras uses internally

hstojic: the issue with validation data using output of `prepare_dataset` is that its typically bootstrapped since its needed for uncertainty quantification - thats not great for creating validation data, ideally we would separate out validation set and then create x,y that are bootstrapped for training - I address this in #899, i.e. modify methods and provided some utils to make it easier

ChrisMorter: I've rebased this onto [#899](https://github.com/secondmind-labs/trieste/pull/899) and have updated it to form the validation split using `split_dataset_randomly` before potentially bootstrapping the data with `prepare_dataset`.

tests/unit/models/keras/test_models.py
```diff
@@ -542,6 +542,30 @@ def test_deep_ensemble_predict_ensemble() -> None:
...
+        "steps_per_epoch": 20,
+        "validation_split": 0.5,
+    }
+    optimizer_wrapper = KerasOptimizer(optimizer, fit_args)
+
+    model = DeepEnsemble(keras_ensemble, optimizer_wrapper)
+    model.optimize(example_data)
+
+    losses = model.model.history.history["loss"]
+    assert len(losses) == fit_args["epochs"]
```
hstojic: perhaps check underlying keras model `iterations` to verify number of steps?

hstojic (approved):
looks good now (after #899 is merged in)

PR #910, opened 2025-05-18
Float32 GPR predictions

uri-granta:
**Related issue(s)/PRs:** 

## Summary

...

## PR checklist


hstojic (commented):
so, can you please leave out `freeze_as_float32` methods from the models?

providing utilities, expanding arguments and making it easy is fine, but these methods will be a very rare (public) use case of trieste so in my opinion this is unnecessary bloat here

PR #911, opened 2025-05-20
Support pregenerated posterior cache

uri-granta:
**Related issue(s)/PRs:** 

## Summary

...

## PR checklist


trieste/models/gpflow/models.py
```diff
@@ -129,7 +133,10 @@ def __init__(
         self._num_rff_features = num_rff_features
         self._use_decoupled_sampler = use_decoupled_sampler
         self._ensure_variable_model_data()
-        self.create_posterior_cache()
+        if posterior is None:
+            self.create_posterior_cache()
+        else:
+            self._posterior = posterior
```
hstojic, line 139: perhaps we need to perform some basic checks of the posterior that has been passed? dimensionality?, not sure what the object contains and what could be used as some sensible check that something wrong isnt passed

hstojic (approved):
looks good, but left a small comment, not sure whether its sensible to do that though...

trieste/models/gpflow/models.py
uri-granta: Will have a think of what we can do, thanks - though will merge for now to unblock the automotive work.

PR #921, opened 2025-09-02
Unpin dill (for real this time)

uri-granta:
**Related issue(s)/PRs:** 

## Summary

...

## PR checklist


trieste/utils/misc.py
```diff
@@ -468,3 +469,16 @@ def ensure_positive(x: TensorType) -> TensorType:
     """Ensure that all the elements in `x` are strictly positive (using a dtype-dependent
     capping threshold)."""
     return tf.math.maximum(x, 1e-15 if x.dtype == tf.float32 else 1e-30)
+
+
+def restore_tfp_symbols_to_modules() -> None:
```
hstojic, line 474: is there any way to test this?

hstojic (approved):
looks good I think, but is there any way to test that?

trieste/utils/misc.py
uri-granta: The _purpose_ of the code (to make serialisation of tfp kernels possible with the latest dill) is already tested, both with the latest version of dill and with an older supported version: without this fix, a number of unit tests fail when trying to use latest dill. Not sure if there's any point to writing an explicit unit test for the function, as we don't really care about its actual behaviour (exposing additional symbols that were previously removed), only the fact that it fixes serialisation.

PR #929, opened 2026-05-21
Faster Deep Ensembles

avullo:
**Related issue(s)/PRs:** 

## Summary

...

## PR checklist


trieste/models/keras/architectures.py
```diff
@@ -114,20 +213,91 @@ def ensemble_size(self) -> int:
...
+        input_dim = int(np.prod(network.input_tensor_spec.shape))
+
+        # Single stacked input: [batch, E * input_dim]
+        stacked_input = keras.Input(shape=(E * input_dim,), dtype=dtype, name="ensemble_input")
+
+        # Reshape to [E, batch, input_dim] for vectorized layers
+        h = keras.layers.Lambda(
+            lambda x: tf.transpose(tf.reshape(x, [-1, E, input_dim]), [1, 0, 2]),
+            name="ensemble_reshape",
+        )(stacked_input)
```
hstojic, line 264: you forgot the dtype here, so the whole network drops to float32 in case data is float64 before the first matmul

this is perhaps why you had to increase tolerance on some of the tests?

trieste/models/keras/architectures.py
```diff
@@ -114,20 +213,91 @@ def ensemble_size(self) -> int:
...
+                name=f"vec_dense_{j}",
+                dtype=dtype,
+            )(h)
+
+        # Output layer: 2 parameters per member (mean + raw scale)
+        params = vectorized_dense_layer(E, 2, None, name="vec_params", dtype=dtype)(h)
+        # [E, batch, 2] → [batch, E, 2]
+        params_T = keras.layers.Lambda(
+            lambda x: tf.transpose(x, [1, 0, 2]), name="ensemble_transpose"
+        )(params)
```
hstojic, line 281: dtype not set here as well

tests/integration/test_bayesian_optimization.py
```diff
@@ -559,7 +559,8 @@ def test_bayesian_optimizer_with_deep_ensemble_finds_minima_of_scaled_branin(
     "num_steps, acquisition_rule",
     [
         pytest.param(5, lambda: EfficientGlobalOptimization(), id="EfficientGlobalOptimization"),
-        pytest.param(10, lambda: DiscreteThompsonSampling(1000, 1), id="DiscreteThompsonSampling"),
+        # 12 steps: vectorized ensemble + some CI CPUs need margin over 10 for DTS float64.
+        pytest.param(12, lambda: DiscreteThompsonSampling(1000, 1), id="DiscreteThompsonSampling"),
```
hstojic, line 563: this may be now needed bcs of the dtype bug with lambda layers, check whether you still need this after fixing the bug

tests/unit/models/keras/test_sampler.py
```diff
@@ -423,7 +424,13 @@ def test_ensemble_trajectory_sampler_trajectory_on_subsets_same_as_set(diversify
     eval_2 = trajectory(test_data[100:200, :])
     eval_3 = trajectory(test_data[200:300, :])
 
-    npt.assert_allclose(eval_all, tf.concat([eval_1, eval_2, eval_3], axis=0), rtol=1e-5)
+    # Vectorized batched inference can differ slightly by total batch size (float32 order).
+    npt.assert_allclose(
+        eval_all,
+        tf.concat([eval_1, eval_2, eval_3], axis=0),
+        rtol=1e-5,
+        atol=1e-6,
```
hstojic, line 432: this may be now needed bcs of the dtype bug with lambda layers, check whether you still need this after fixing the bug

trieste/models/keras/architectures.py
```diff
@@ -114,20 +213,91 @@ def ensemble_size(self) -> int:
...
+        """Vectorized ensemble with a single stacked input and a single batched output.
+
+        A single input ``"ensemble_input"`` of shape ``[E * D]`` is reshaped to
+        ``[E, batch, D]`` inside the model and fed through one shared set of
+        :class:`VectorizedEnsembleDenseLayer` layers (one batched matmul per layer).
+        The ``[E, batch, 2]`` parameter tensor is transposed to ``[batch, E, 2]`` and
+        wrapped in a single :class:`~tfp.layers.DistributionLambda`, yielding a
+        ``Normal`` with ``batch_shape = [batch, E, 1]``.
+
+        On prod pins (standalone ``tf_keras`` with TFP 0.22), the functional model is built
```
hstojic, line 240: what does prod refers to here?

trieste/models/keras/architectures.py
```diff
@@ -114,20 +213,91 @@ def ensemble_size(self) -> int:
...
+        handling overhead per training step in the ``tf.data`` pipeline and Keras metrics.
+        """
+        assert all(isinstance(n, GaussianNetwork) for n in self._networks)
+
+        keras = _keras_for_vectorized_ensemble_build()
+        self._vectorized_uses_tensorflow_keras = keras is tf.keras
+        vectorized_dense_layer = _vectorized_ensemble_dense_layer_class(keras)
+
+        E = len(self._networks)
+        network = self._networks[0]
```
hstojic, line 253: this is a bug I think, only the first architecture is taken and whole ensemble built from it, while in the builder  intentionally has more flexibility, we should be able to specify different architectures across members - this way that would be ignored, right?

hstojic: I think we don't have a test right now that would catch this, in fixing this lets add that test

trieste/models/keras/architectures.py
hstojic: do we have a simple test that would catch this?
e.g. perturbing a query point by something below float32 detection, say 1e-10, should change the prediction, while if model drops to low precision we would get zero.

trieste/models/keras/architectures.py
```diff
@@ -114,20 +213,91 @@ def ensemble_size(self) -> int:
...
-        to construct individual networks.
+
+        For :class:`GaussianNetwork` ensembles with a single output dimension, a vectorized
+        model is constructed using batched matmuls across all ensemble members in one pass.
+        For other network types, falls back to the original per-member functional model.
 
         :return: The Keras model.
         """
+        if all(
+            isinstance(n, GaussianNetwork) and n.flattened_output_shape == 1 for n in self._networks
```
hstojic, line 224: isinstance(n, GaussianNetwork) bypasses subclasses, says AI :)

trieste/models/keras/architectures.py
```diff
@@ -114,20 +213,91 @@ def ensemble_size(self) -> int:
...
+            lambda x: tf.transpose(tf.reshape(x, [-1, E, input_dim]), [1, 0, 2]),
+            name="ensemble_reshape",
+        )(stacked_input)
+
+        # Vectorized hidden layers
+        for j, layer_args in enumerate(network._hidden_layer_args):  # type: ignore
+            h = vectorized_dense_layer(
+                E,
+                layer_args["units"],
+                layer_args.get("activation"),
```
hstojic, line 271: what about other Dense kwargs?
everything except units and activation is dropped?
builder was intentionally flexible there and any dense kwarg could be used

trieste/models/keras/models.py
```diff
@@ -157,14 +162,29 @@ def __init__(
         if self.optimizer.loss is None:
             self.optimizer.loss = negative_log_likelihood
 
-        if self.optimizer.metrics is None:
-            self.optimizer.metrics = ["mse"]
```
hstojic, line 161: was this intentional?
I don't particularly mind if we drop it but then the init docstring needs to be amended

trieste/models/keras/models.py
```diff
@@ -157,14 +162,29 @@ def __init__(
...
+        base_loss = (
+            self.optimizer.loss if self.optimizer.loss is not None else negative_log_likelihood
+        )
+        n_outputs = len(model.model.outputs)
+        if n_outputs == 1:
+            compile_loss = aggregate_member_losses(base_loss)
+        else:
+            compile_loss = [base_loss] * n_outputs  # type: ignore
+        self._compile_args = dict(compile_args)
+        compile_metrics = compile_metrics_for_ensemble(
```
hstojic, line 176: the new helper  mishandles the multi-output (non-vectorized) path, because it returns a flat list of per-member metrics and Keras broadcasts a flat list to every output

hstojic: try ["mse"] and ["mse", "mae"] on 3-output ensemble

hstojic: check also `log` method for vectorized models, history key is different and not caught by log method, probably `test_deep_ensemble_log` needs to be udpated

trieste/models/keras/utils.py
```diff
@@ -134,3 +135,141 @@ def negative_log_likelihood(
...
+
+    def aggregated(y_true: TensorType, y_pred: Any) -> TensorType:
+        values = loss_fn(y_true, y_pred)
+        if len(values.shape) == 0:
+            return values
+        if len(values.shape) >= 2:
+            return tf.reduce_sum(tf.reduce_mean(values, axis=0))
+        return tf.reduce_mean(values, axis=0)
+
+    return aggregated
```
hstojic, line 163: this doesn't cover tf_keras.losses.Loss instance, which already reduces scalar form already, so it results in 1/E scaling that this aims to prevent... computing such losses with reduction="none" before aggregating would fix it

trieste/models/keras/models.py
```diff
@@ -221,19 +241,61 @@ def prepare_dataset(
...
-        outputs = {}
-        for index in range(self.ensemble_size):
+        E = self.ensemble_size
+        single_input = len(self.model.input_names) == 1
+
+        if single_input:
+            # Single stacked input/output model: pack all E members' data into one tensor.
+            n_rows = dataset.observations.shape[0]
             if self._bootstrap and not do_not_bootstrap:
-                resampled_data = sample_with_replacement(dataset)
```
hstojic, line 228: this function had empty dataset guard, but in new version is gone, you need to intorduce it again I think

hstojic: probably we need a test for this

hstojic (commented):
see comments

PR #935, opened 2026-06-05
feat(space): add HierarchicalSearchSpace core

mawbray:
Add the HierarchicalSearchSpace class with construction and validation, equality/repr, the indicator/dimension/bounds properties, and sampling, containment and subspace-component extraction. Indicators must be a BooleanSearchSpace or a dimension-1 discrete subspace; the hierarchy is described by a sequence of gpflow HierarchyNode objects built via hierarchy_node_from_tags. Adds construction, property, sampling, containment, validation and equality/repr tests.

**Related issue(s)/PRs:** 

## Summary

...

## PR checklist

### DIFF
--- a/tests/unit/test_hierarchical_search_space.py
+++ b/tests/unit/test_hierarchical_search_space.py
@@ -14,15 +14,19 @@
 """Tests for the hierarchical-search-space API."""
 from __future__ import annotations
 
+import gpflow.kernels
 import numpy.testing as npt
 import pytest
 import tensorflow as tf
 
 from trieste.space import (
+    ActivityCondition,
     BooleanSearchSpace,
     Box,
     CategoricalSearchSpace,
     DiscreteSearchSpace,
+    HierarchicalSearchSpace,
+    HierarchyNode,
     SearchSpace,
     hierarchy_node_from_tags,
 )
@@ -43,6 +47,40 @@ def _worked_example_subspaces() -> dict[str, SearchSpace]:
         "x4": Box([-2.0], [2.0]),
         "x3": Box([-1.0], [1.0]),
     }
+
+
+def _worked_example_hierarchy(subspaces: dict[str, SearchSpace]) -> list[HierarchyNode]:
+    return [
+        hierarchy_node_from_tags(
+            "shared",
+            subspace_tags=["x1"],
+            subspaces=subspaces,
+            indicator_tags=["y1"],
+        ),
+        hierarchy_node_from_tags(
+            "branch_A",
+            subspace_tags=["x2", "x4"],
+            activity_condition_tags={"y1": 1},
+            subspaces=subspaces,
+            indicator_tags=["y1"],
+        ),
+        hierarchy_node_from_tags(
+            "branch_B",
+            subspace_tags=["x3"],
+            activity_condition_tags={"y1": 0},
+            subspaces=subspaces,
+            indicator_tags=["y1"],
+        ),
+    ]
+
+
+def _make_worked_example_hss() -> HierarchicalSearchSpace:
+    subspaces = _worked_example_subspaces()
+    return HierarchicalSearchSpace(
+        subspaces,
+        _worked_example_hierarchy(subspaces),
+        indicator_tags=["y1"],
+    )
 
 
 # ===== hierarchy_node_from_tags =====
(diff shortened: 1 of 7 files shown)


trieste/space.py
```diff
@@ -1559,6 +1559,408 @@ def hierarchy_node_from_tags(
...
+                subspaces=subspaces, indicator_tags=["y1"],
+            ),
+            hierarchy_node_from_tags(
+                "branch_B", subspace_tags=["x3"],
+                activity_condition_tags={"y1": 0},
+                subspaces=subspaces, indicator_tags=["y1"],
+            ),
+        ]
+        space = HierarchicalSearchSpace(
+            subspaces, hierarchy, indicator_tags=["y1"])
```
pakhaykwok, line 1803: C: Same as my comment in the other PR, I think `indicator_tags` is redundant.

pakhaykwok, line 1803: C: The current design is a little clunky in the sense that `subspaces` need to be passed to so many places. Ideally it should only be passed once in `HierarchicalSearchSpace`. Perhaps we can have a trieste version of `HierarchyNode` which doesn't need `subspaces` and only resolve it to the gpflow version in `HierarchicalSearchSpace`?

trieste/space.py
```diff
@@ -1559,6 +1559,408 @@ def hierarchy_node_from_tags(
...
+        self._column_to_tag: Dict[int, str] = {}
+        cursor = 0
+        for tag in self._tags:
+            size = int(self._subspace_sizes_by_tag[tag])
+            cols = list(range(cursor, cursor + size))
+            self._tag_to_columns[tag] = cols
+            for c in cols:
+                self._column_to_tag[c] = tag
+            cursor += size
+        self._total_columns = cursor
```
pakhaykwok: SH: Consider making this a util, as it's also being used in the other PR.

trieste/space.py
```diff
@@ -1559,6 +1559,408 @@ def hierarchy_node_from_tags(
...
+
+        self._validate()
+
+    def _validate(self) -> None:
```
pakhaykwok, line 1894: SH: Also consider taking out the common validations.

trieste/space.py
mawbray: You're right that it can be inferred from the hierarchy, but I think it is also not a bad thing to have as sanity check, cross validation. I will plan to make it optional. Eg BooleanSearchSpaces have bounds so can become a feature if not correctly included in the hierarchy. Making the indicators explicit is a way of protecting against this.

mawbray: Seems sensible I will look into this.

pakhaykwok: I'm more about that it should always be inferred from subspaces.

mawbray: Okay, I guess we will have to assert strict requirements on the feature subspaces types? E.g. we can never have overlapping types between indicators and features.

mawbray: I will hold off on this, subject to @uri-granta opinion

mawbray: Also holding off on this, subject to @uri-granta

trieste/space.py
```diff
@@ -1610,6 +1610,416 @@ def hierarchy_node_from_tags(
     )
 
 
+class HierarchicalSearchSpace(CollectionSearchSpace):
```
uri-granta, line 1738: **Update**: just noticed that you define product in #937. Haven't reviewed that yet, but I think the changes below still make sense for this PR. (You can also then replace the ValueError in #937 with NotImplementedError, which would allow creating a product of two HierarchicalSearchSpace with overlapping tags via a TaggedProductSearchSpace.)

ES: AFAICT this is missing the `product` abstract method, so none of the tests will currently run. Since we don't want to care about implementing a proper product, you can just write:

```
    def product(self, other: "HierarchicalSearchSpace") -> "HierarchicalSearchSpace":
        raise NotImplementedError
```

However, since this is the first time we've skipped implementing product, you should also update the SearchSpace base methods to handle this: (update the docstring in product, and update `__mul__` to default to using a tagged product space:

```
    @abstractmethod
    def product(self: SearchSpaceType, other: SearchSpaceType) -> SearchSpaceType:
        """
        :param other: A search space of the same type as this search space.
        :return: The Cartesian product of this search space with the ``other``.
            Subclasses that do not support same-type products should raise
            :exc:`NotImplementedError`; :meth:`__mul__` will fall back to a
            :class:`TaggedProductSearchSpace` in that case.
        """


    def __mul__(self, other: SearchSpace) -> SearchSpace:
        """
        :param other: A search space.
        :return: The Cartesian product of this search space with the ``other``.
            If both spaces are of the same type then this calls the :meth:`product` method.
            Otherwise, it generates a :class:`TaggedProductSearchSpace`.
        """
        # If the search space has any constraints, always return a tagged product search space.
        if not self.has_constraints and not other.has_constraints and isinstance(other, type(self)):
            try:
                return self.product(other)
            except NotImplementedError:
                pass
        return TaggedProductSearchSpace((self, other))
```

tests/unit/test_hierarchical_search_space.py
```diff
@@ -274,3 +577,98 @@ def test_helper_accepts_discrete_non_indicator_subspace() -> None:
...
+            indicator_tags=["y1"],
+        ),
+    ]
+    other = HierarchicalSearchSpace(subspaces, renamed_hierarchy, indicator_tags=["y1"])
+    assert _make_worked_example_hss() != other
+
+
+def test_hss_eq_false_for_non_hierarchical_object() -> None:
+    assert _make_worked_example_hss() != 5
+    assert _make_worked_example_hss() != _make_categorical_hss()
```
uri-granta: `_make_categorical_hss` isn't defined (in this PR at least)!

tests/unit/test_hierarchical_search_space.py
```diff
@@ -216,6 +226,299 @@ def test_helper_rejects_activity_condition_key_not_a_subspace() -> None:
...
+
+
+def test_hss_infers_indicator_tags_when_omitted() -> None:
+    # Omitting indicator_tags must reproduce the explicit result on the worked example.
+    subspaces = _worked_example_subspaces()
+    inferred = HierarchicalSearchSpace(subspaces, _worked_example_hierarchy(subspaces))
+    explicit = _make_worked_example_hss()
+    assert inferred.indicator_tags == explicit.indicator_tags == ("y1",)
+    assert list(inferred.indicator_dims) == list(explicit.indicator_dims) == [1]
+    assert inferred.enumerate_tasks() == explicit.enumerate_tasks()
```
uri-granta: `enumerate_tasks` is not defined (in this PR at least)!

trieste/space.py
```diff
@@ -1610,6 +1610,416 @@ def hierarchy_node_from_tags(
...
+    def __eq__(self, other: object) -> bool:
+        """
+        :param other: A search space.
+        :return: Whether the search space is identical to this one, including its
+            hierarchy and indicator tags (the base-class comparison only covers the
+            subspaces and their tags).
+        """
+        if not isinstance(other, HierarchicalSearchSpace):
+            return NotImplemented
+        if super().__eq__(other) is not True:
```
```suggestion
        if not super().__eq__(other):
```

trieste/space.py
```diff
@@ -1610,6 +1610,416 @@ def hierarchy_node_from_tags(
...
+            self._hierarchy, other._hierarchy
+        )
+
+    def __repr__(self) -> str:
+        return f"""{self.__class__.__name__}(spaces =
+                {[self.get_subspace(tag) for tag in self.subspace_tags]},
+                tags = {self.subspace_tags},
+                hierarchy = {list(self._hierarchy)},
+                indicator_tags = {list(self._indicator_tags)})
+                """
```
uri-granta: This currently has whitespace and blanklines in the repr.

```suggestion
        return (
            f"{self.__class__.__name__}("
            f"spaces={[self.get_subspace(tag) for tag in self.subspace_tags]}, "
            f"tags={list(self.subspace_tags)}, "
            f"hierarchy={list(self._hierarchy)}, "
            f"indicator_tags={list(self._indicator_tags)})"
        )
```

tests/unit/test_hierarchical_search_space.py
```diff
@@ -216,6 +226,299 @@ def test_helper_rejects_activity_condition_key_not_a_subspace() -> None:
...
+            feature_dims=[0],
+            feature_bounds=tf.constant([[0.0, 1.0]], dtype=tf.float64),
+            activity_condition=ActivityCondition({0: 1}),
+        )
+    ]
+    with pytest.raises(ValueError, match="BooleanSearchSpace"):
+        HierarchicalSearchSpace(subspaces, hierarchy, indicator_tags=["y1"])
+
+
+def test_hss_raises_if_feature_dim_points_at_indicator_column() -> None:
```
uri-granta: test_hss_raises_if_feature_dim_points_at_indicator_column, test_hss_raises_if_feature_dim_out_of_range,
  test_hss_raises_if_feature_bounds_disagree_with_subspace, and
  test_hss_raises_if_activity_condition_key_not_an_indicator_column can be merged into one parametrised test (I think)

tests/unit/test_hierarchical_search_space.py
```diff
@@ -91,6 +92,15 @@ def _worked_example_hierarchy(subspaces: dict[str, SearchSpace]) -> list[Hierarc
     ]
 
 
+def _make_worked_example_hss() -> HierarchicalSearchSpace:
```
uri-granta, line 82: Should we also test multi-indicator hierarchies?

trieste/space.py
```diff
@@ -1610,6 +1610,416 @@ def hierarchy_node_from_tags(
...
+    of ``subspaces`` (a ``dict`` preserves insertion order in Python 3.7+), the
+    same convention as :class:`TaggedProductSearchSpace`.
+
+    .. note::
+        ``sample`` and ``contains`` operate on the full flat vector and do **not**
+        enforce "active-branch" semantics: every subspace always contributes its
+        columns regardless of the indicator values, and inactive-branch columns are
+        neither masked nor ignored. This matches the flat-vector convention expected
+        by the hierarchical GP kernel, which reads the indicator columns itself to
+        decide which features are active. Use :meth:`active_subspace_tags` /
```
uri-granta: These methods don't exist AFAIK. Are they in later PRs?

trieste/space.py
uri-granta: Sounds like a good idea to me, though it might be easier to do it in a subsequent PR. In general, the less repetition and plumbing the better.

uri-granta: Happy to chat about this

trieste/space.py
```diff
@@ -1610,6 +1610,416 @@ def hierarchy_node_from_tags(
...
+        """The hierarchy specification."""
+        return self._hierarchy
+
+    @property
+    def indicator_tags(self) -> tuple[str, ...]:
+        """The declared indicator tags (Boolean or categorical)."""
+        return self._indicator_tags
+
+    @property
+    def indicator_dims(self) -> List[int]:
```
uri-granta: NP: the new code uses a mix of new-style list[...], tuple[...] and old-style List[...], Tuple[...]. Unless there's a reason, we should stick to the former.

uri-granta (commented):
Looking good...

